PR #334, opened 2026-03-12
fix: calorie computation bugs affecting training load accuracy

fiddur:
## Summary

Fixes four calorie computation bugs that caused incorrect training load (ATL/CTL/TSB) curves:

- **Formula gives total, not active calories**: Subtract resting baseline (formula output at resting_hr × 1.2) so sleep/rest → 0 kcal. Multiplier validated against Oura sleep HR data.
- **Midnight aggregate spike**: `health_connect_aggregate` dumps full day's calories at 00:00, inflating overnight fatigue. Now `calories_active` uses only per-minute `aurboda` source.
- **Tooltip double-division**: `cal.avg / bucketMinutes` divided an already-per-minute value by minutes again, showing values ~15x too small.
- **Full recompute**: `recalculate_calories` MCP tool now supports omitting start/end to recompute all historical data.

## Expected values after fixes
- 36-min strength session (avg HR ~111): 356 → **247 kcal active**
- Sleep (HR ~60): 128 kcal/hr → **0 kcal/hr**
- Light walking (HR 80): → **~127 kcal/hr**

## Test coverage
- 24 unit tests covering baseline subtraction, sleep → 0, edge cases
- All 918 backend unit tests pass
- `pnpm check` and `pnpm fix` pass clean

### DIFF
--- a/apps/backend/coverage-baseline.json
+++ b/apps/backend/coverage-baseline.json
@@ -1,5 +1,5 @@
 {
-  "statements": 60.83,
+  "statements": 60.74,
   "branches": 86.63,
-  "functions": 79.19
+  "functions": 78.67
 }
--- a/apps/backend/src/db/cumulative-query.ts
+++ b/apps/backend/src/db/cumulative-query.ts
@@ -1,19 +1,31 @@
 /**
  * Helper for splitting metric queries by cumulative/non-cumulative type.
  *
- * Cumulative metrics (steps, distance, etc.) use only the aggregate source
+ * Cumulative metrics (steps, distance, etc.) use only trusted sources
  * to avoid mixing raw readings with deduplicated daily totals.
+ *
+ * Some cumulative metrics (like calories_active) are "aurboda-only":
+ * when aurboda computes per-minute values, the health_connect_aggregate
+ * daily totals must be excluded to avoid nonsense averages.
  */
-import { cumulativeMetrics, type MetricType } from '@aurboda/api-spec'
+import { aurbodaOnlyMetrics, cumulativeMetrics, type MetricType } from '@aurboda/api-spec'
 import type { QueryResultRow } from 'pg'
 
 /**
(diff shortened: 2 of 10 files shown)


fiddur:
## Ideas for recovering Oura non-exercise calorie data

The current fix excludes `health_connect_aggregate` entirely for `calories_active`, which loses Oura's movement-based calorie estimates for periods without HR monitoring. Here are some approaches to get that back without reintroducing the midnight spike:

### 1. Residual distribution (recommended)
After computing aurboda per-minute calories from HR data, compare the daily total against the Oura/HC aggregate. If the aggregate is higher, distribute the **residual** (aggregate − aurboda sum) evenly across hours where no HR data exists. This captures Oura's wrist-based movement detection for non-exercise periods.

**Pros:** Preserves Oura's total daily estimate, spreads it realistically, no midnight spike.
**Cons:** Assumes Oura's total is trustworthy; non-HR hours get a flat rate rather than variable.

### 2. Time-proportional spreading
When ingesting the HC aggregate at midnight, instead of storing it as a single point, spread it across all 1440 minutes of the day proportionally to known activity. If no activity data exists for a minute, use a flat baseline.

**Pros:** Simple to implement at ingestion time.
**Cons:** Less accurate than approach 1 since it doesn't have the HR-based computation to compare against.

### 3. Gap-filling only
Only use the Oura/HC value for time ranges where we have **no** HR data at all. For those ranges, compute kcal/min = (daily_aggregate - aurboda_sum) / gap_minutes and fill the gaps.

**Pros:** Most accurate — HR-based where available, Oura-estimated elsewhere.
**Cons:** Slightly more complex; needs to identify HR data gaps.

---

Approach 1 or 3 would work well as a post-processing step in `computeAndStoreCalories` — after storing aurboda points, check if an HC aggregate exists for the same day and distribute the difference. The key invariant: never store a single large value at midnight.

PR #716, opened 2026-05-06
chore: upgrade to Node 26 with native Temporal API

fiddur:
## Summary

- Bump Node 25 → 26 (Dockerfile build + prod stages, root `package.json` engines, `.nvmrc`, CI + Android workflows)
- Drop `@js-temporal/polyfill` from runtime — Node 26 ships Temporal natively (V8)
- Keep the polyfill as a `devDependencies`-only source of TypeScript types (TS 5.9 still ships no Temporal types)
- Add `apps/backend/src/temporal-global.d.ts` exposing `Temporal` as a global namespace + const, mirroring the polyfill's shape
- Drop value-side imports in the four backend files that used Temporal (`tz-utils`, `location-stays`, `metrics`, `training-load/aggregation`)

## Test plan

- [x] `pnpm check` passes (tsgo + oxlint, 0 errors)
- [x] All 2053 backend tests pass on Node 26 (`nvm use 26 && pnpm --filter aurboda-backend test`)
- [x] All 324 web tests pass on Node 26
- [x] Verified `globalThis.Temporal` exists on Node 26 (`Temporal.Now.instant().toString()` works)
- [ ] CI green on Node 26 runner (`actions/setup-node` pulls from nodejs.org, should be fine)
- [ ] Docker image build green (**blocked — see below**)
- [ ] Deploy to https://aurboda.net via Docker image build

## Pending upstream

- **`node:26-alpine` not yet on Docker Hub.** Node 26.0.0 was released so recently that the `docker-library/node` images haven't been published. `docker pull node:26-alpine` currently returns `manifest unknown`; the latest tag on Docker Hub is `25.9.0-trixie`. Hold this PR as Draft until the image lands (typically a few days after a Node release), then merge to trigger deploy. CI itself is unaffected — `actions/setup-node` downloads directly from nodejs.org.
- **TypeScript doesn't ship Temporal types yet.** As of TS 5.9 there is no `Temporal` in any `lib.*.d.ts`. We borrow types from `@js-temporal/polyfill` (kept as devDep) via `temporal-global.d.ts`. Once TypeScript ships Temporal types in stdlib, we can remove `@js-temporal/polyfill` and `temporal-global.d.ts` entirely.


fiddur (commented):
Clean and well-scoped. Read through the whole diff and the d.ts approach is sound:

- The 8 types exported in `temporal-global.d.ts` cover every `Temporal.*` symbol the backend currently uses (`grep -r 'Temporal\.[A-Z]'` in `apps/backend/src` shows only `Instant`, `PlainDate`, `PlainTime`). `Temporal.Now` isn't used anywhere yet, and the `const Temporal: typeof TemporalImpl` line gives full value-side access regardless. So no missing exports today; just a heads-up if anyone starts using `Temporal.TimeZone` / `Temporal.Calendar` they'll need to add them to the namespace block.
- `apps/backend/tsconfig.json` includes `src`, so `temporal-global.d.ts` is picked up automatically — no `types` array bump needed.
- Keeping `@js-temporal/polyfill` as devDep purely for types is the standard trick until TS stdlib ships `Temporal`. Worth a short note in `temporal-global.d.ts` to delete the file + the devDep when that day comes (you already capture this in the PR description, but a TODO at the top of the .d.ts would help the next person who finds it).

A few small thoughts:

1. **Root `tsconfig.json` still extends `@tsconfig/node24`.** Currently it works because lib levels are forward-compatible, but `@tsconfig/node26` likely exists now. Not blocking — just inconsistent.
2. **`engines.node: "~25"` was strict; now it's `~26`.** Means anyone on Node 25 LTS gets an `EBADENGINE` warn. Fine for development, but if the production server isn't running Node 26 by the time this merges, pnpm's `engine-strict=true` (if set anywhere) would block install. Worth a sanity check on the deploy host.
3. **No runtime assertion that `globalThis.Temporal` exists.** If, for some reason, a future Alpine image regression strips it (vanishingly unlikely), the failure mode is `ReferenceError: Temporal is not defined` deep in some query. A one-liner `if (typeof Temporal === 'undefined') throw new Error('Temporal not available — Node 26+ required')` at backend startup would fail loud and early. Optional, very paranoid.

Pending the `node:26-alpine` image hitting Docker Hub — once that lands and the manual deploy smoke runs, this looks ready. LGTM.

PR #733, opened 2026-05-08
fix(entity-detail): navigate to override row after saving a synced activity

fiddur:
## Summary

After #715 / #732, editing a synced activity (Garmin, Strava, Health Connect, Oura) on the detail page creates a new aurboda **override row with a different id**. The frontend's `saveMutation` then invalidated the query keyed on the URL's source id, so it just re-fetched the source — which still had the old values. The user's edits appeared to vanish.

The user reported: *"I edited the activity on the detail page, the changes weren't visible at all. Going to the timeline showed my override, and clicking that got correct."* — i.e. the override was created server-side, but the detail page never showed it.

## Fix

- `updateActivity` (state/api) now reads the PATCH response's `data.id` and returns it. PATCH on a synced source returns the override's id; PATCH on an aurboda or already-overridden row returns the same id back.
- `EntityDetail`'s `saveMutation.onSuccess` compares the returned id against the URL's `rawEntityId`. If they differ, it routes to `/detail/activity/<override_id>` via `useLocation().route()`. The page then displays the override the user just saved.
- Same-id edits fall through to the existing invalidate-and-redraw path.

Route format matches the source-record links elsewhere in the page (line 57: `/detail/activity/<id>`).

## Test plan

- [x] Type-check + lint clean
- [x] Existing test suite passes (367/367 web)
- [ ] Manual: edit a Garmin-sourced activity (e.g. change `meditation` → `pipe_ceremony`); save. Page should navigate to the new override and show the new type, not silently sit on the source.
- [ ] Manual: edit the override again; save. Page should stay on the same URL (same id returned).
- [ ] Manual: edit an aurboda-sourced activity; save. Page should stay on the same URL.

### DIFF
--- a/apps/web/src/pages/EntityDetail/index.tsx
+++ b/apps/web/src/pages/EntityDetail/index.tsx
@@ -38,6 +38,7 @@ import { MetricContent } from './MetricContent'
 import { MusicPlaylist } from './MusicPlaylist'
 import { NotesSection } from './NotesSection'
 import { ProductivityDetail } from './ProductivityDetail'
+import { activityRouteAfterSave } from './saveNavigation'
 import { SchemaDataFields } from './SchemaDataFields'
 import {
   computeSleepMinutesFromStages,
@@ -530,7 +531,7 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
     onSuccess: () => invalidate(),
   })
 
-  const saveMutation = useMutation<{ id: string } | null, Error, void>({
+  const saveMutation = useMutation<{ id: string | undefined } | null, Error, void>({
     mutationFn: async () => {
       if (!activity) return null
       const body: {
@@ -559,19 +560,16 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
         body.data = draft.data
       }
       if (Object.keys(body).length === 0) return null
-      return await updateActivity(rawEntityId, body)
+      return updateActivity(rawEntityId, body)
     },
     onSuccess: (result) => {
       setIsEditing(false)
       invalidate()
-      // Editing a synced activity (e.g. Garmin) for the first time creates a
-      // new aurboda override row with a different id. Re-fetching the source
-      // id would just show the unedited source — the user's edits would
-      // appear to vanish. Navigate to the override so the page reflects
-      // what they just saved.
-      if (result && result.id !== rawEntityId) {
-        route(`/detail/activity/${result.id}`)
-      }
+      const target = activityRouteAfterSave(result?.id, rawEntityId)
+      // `replace=true` so back-button doesn't return to the source-id detail
+      // (which would now show the unedited source — exactly the confusing
+      // state this fix avoids).
+      if (target) route(target, true)
     },
   })
 
--- a/apps/web/src/pages/EntityDetail/index.tsx
+++ b/apps/web/src/pages/EntityDetail/index.tsx
@@ -8,7 +8,7 @@
 import { exerciseTypeNames, getExerciseTypeName as getExerciseTypeNameFromValue } from '@aurboda/api-spec'
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { format } from 'date-fns'
-import { useRoute } from 'preact-iso'
+import { useLocation, useRoute } from 'preact-iso'
 import { useCallback, useEffect, useState } from 'preact/hooks'
 
 import type { Activity, ActivityTypeDefinition, ExerciseTypeName, SourceRecord } from '../../state/api'
@@ -456,6 +456,7 @@ const ResyncDetailButton = ({
 
 const ActivityContent = ({ entityId }: { entityId: string }) => {
   const queryClient = useQueryClient()
+  const { route } = useLocation()
   const isMerged = entityId.startsWith('merged:')
   const rawEntityId = isMerged ? entityId.slice('merged:'.length) : entityId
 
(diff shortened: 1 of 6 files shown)


apps/web/src/state/api/activities.ts
```diff
@@ -71,11 +71,19 @@ export const softDeleteActivity = async (id: string): Promise<void> => {
...
+ * holding the user's edits.
+ */
+export const updateActivity = async (id: string, body: UpdateActivityBody): Promise<{ id: string }> => {
   const { token } = auth.value
-  await axios.patch(`${API_URL}/activities/${id}`, body, {
+  const response = await axios.patch<{ data: { id: string } }>(`${API_URL}/activities/${id}`, body, {
     headers: { Authorization: `Bearer ${token}` },
   })
+  return { id: response.data.data.id }
 }
```
fiddur, line 92: **P2:** Use the existing `UpdateActivityResponse` type from `@aurboda/api-spec` instead of an inline cast. The schema (`packages/api-spec/src/schemas/activities.ts`) is what the backend handler types against, and the rest of this file imports response types from the spec (e.g. `AddActivityResponse`). Worth noting the schema marks `data` as optional — switching to the spec type makes that explicit.

```ts
import type { UpdateActivityBody, UpdateActivityResponse } from @aurboda/api-spec

const response = await axios.patch<UpdateActivityResponse>(`${API_URL}/activities/${id}`, body, { ... })
return { id: response.data.data!.id }
```

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -557,12 +558,20 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
...
+    onSuccess: (result) => {
       setIsEditing(false)
       invalidate()
+      // Editing a synced activity (e.g. Garmin) for the first time creates a
+      // new aurboda override row with a different id. Re-fetching the source
+      // id would just show the unedited source — the user's edits would
+      // appear to vanish. Navigate to the override so the page reflects
+      // what they just saved.
+      if (result && result.id !== rawEntityId) {
+        route(`/detail/activity/${result.id}`)
```
fiddur: **P3:** Consider `route(`/detail/activity/${result.id}`, true)` to replace the current history entry rather than push a new one. Otherwise the back-button takes the user back to the source-id detail, which still shows the unedited source — exactly the confusing state this PR is fixing.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -557,12 +558,20 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
       if (draft.data && JSON.stringify(draft.data) !== JSON.stringify(orig.data)) {
         body.data = draft.data
       }
-      if (Object.keys(body).length === 0) return Promise.resolve()
-      return updateActivity(rawEntityId, body)
+      if (Object.keys(body).length === 0) return null
+      return await updateActivity(rawEntityId, body)
```
fiddur: **P3:** `return await` here is equivalent to a bare `return updateActivity(...)` (no surrounding try/catch). Some lint configs flag this via `@typescript-eslint/return-await`. Trivial.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -557,12 +558,20 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
...
-    onSuccess: () => {
+    onSuccess: (result) => {
       setIsEditing(false)
       invalidate()
+      // Editing a synced activity (e.g. Garmin) for the first time creates a
+      // new aurboda override row with a different id. Re-fetching the source
+      // id would just show the unedited source — the user's edits would
+      // appear to vanish. Navigate to the override so the page reflects
+      // what they just saved.
+      if (result && result.id !== rawEntityId) {
```
fiddur: **P2 (testing):** Worth a small unit test here — mock `updateActivity` to return `{ id: override-id }`, fire the mutation with `rawEntityId = source-id`, assert `route` was called with `/detail/activity/override-id`. And the inverse: same id back → `route` not called. This is exactly the branch that regressed; protecting it is cheap.

fiddur (commented):
## Summary

Fix is well-scoped and the PR description traces the bug clearly. Two P2s and a couple P3s — nothing blocking.

## P2

- **Use `UpdateActivityResponse` from `@aurboda/api-spec`** instead of the inline `<{ data: { id: string } }>` cast in `updateActivity`. The schema already exists and is what the backend handler types against (`apps/backend/src/routes/activities-router.ts`). The rest of this file uses spec types for response shapes (e.g. `addActivity` → `AddActivityResponse`), so the inline shape is inconsistent. Note also that the schema marks `data` as optional — using the proper type forces you to acknowledge that.
- **No new tests.** The diff changes (a) the return type of an API-client function and (b) a navigation side-effect in `saveMutation.onSuccess`. Both are testable: a unit test that PATCH returning a different id triggers `route()`, and that same id does not, would cover the regression that motivated this PR. The PR notes the existing 367 tests still pass, but none of them exercise the new branch.

## P3

- `apps/web/src/pages/EntityDetail/index.tsx:573` — consider `route(url, true)` (replace) so the back-button doesn’t return the user to the source-id detail page, which would now show the unedited source — exactly the confusing state this PR is trying to fix.
- `apps/web/src/pages/EntityDetail/index.tsx:562` — bare `return updateActivity(...)` is equivalent to `return await ...` here (no try/catch), and avoids `@typescript-eslint/return-await` flagging it. Trivial.
- The three manual-test items in the PR description are still unchecked. Worth verifying before flipping out of Draft.

 / Claude

PR #734, opened 2026-05-10
feat(meals): multi-day overview tab with NNR2023 nutrient ranges

fiddur:
## Summary

- Adds an **Overview** tab to the Meals page with averaged nutrient intake over a selectable window (7 / 14 / 30 / 90 days), each rendered against a recommended min/max range using the existing `ReferenceRangeBar` component.
- Curated NNR2023 defaults seeded into a new `shared_nutrient_recommendations` table in the central DB; per-user overrides live in `user_nutrient_recommendations` (override row wins; explicit NULL/NULL suppresses a range).
- Energy section compares avg **eaten** vs. averaged daily **calories_total** burned, with a graceful fallback when the metric isn't recorded.
- Full REST + MCP parity for the new endpoints; reuses `aggregateNutrients`-style logic from the meals service so nutrient field set stays in sync via `NUTRIENT_FIELD_NAMES`.

## Test plan

- [ ] `pnpm --filter aurboda-backend test` -- 2147+ tests pass (incl. new unit + integration tests for merge logic and DB roundtrip)
- [ ] `pnpm --filter @aurboda/web test` -- 367 tests pass
- [ ] `pnpm --filter aurboda-backend exec tsgo --noEmit` -- clean
- [ ] `pnpm --filter @aurboda/web exec tsgo --noEmit` -- clean
- [ ] Manual: navigate to `/meals?view=overview`, switch windows, confirm `ReferenceRangeBar` renders for nutrients with NNR defaults; confirm energy section shows / hides burn comparison correctly
- [ ] MCP smoke: `get_nutrient_recommendations`, `set_nutrient_recommendation { nutrient_name: "protein", recommended_low: 80, recommended_high: 200 }`, re-fetch -> source: 'user'; `clear_nutrient_recommendation` reverts to central
- [ ] Sanity check that the Day tab is unchanged at `/meals`


fiddur (commented):
## Review summary

Solid feature — clean separation of central / user / effective layers, decent test coverage, good docs. Two real bugs and a handful of validation gaps below.

**P1 (blocking):**
- REST `PUT /nutrient-recommendations/:nutrient_name` returns **500** for legitimate suppression payloads (`{recommended_low: null, recommended_high: null}`) even though the DB write succeeds — see comment on `routes/nutrient-recommendations-router.ts`.
- `days_with_data` is per-nutrient, not per-day-with-meals — overstates nutrient adequacy on intermittently-eaten nutrients (fiber, vitamin K, etc.). The PR description and code disagree on the semantics — see comment on `services/queries/meal-period-summary.ts`.

**P2 (should fix):**
- `nutrient_name` URL/MCP arg has no validation against the known nutrient set.
- `tz` query param has no IANA validation; bad values crash `Intl.DateTimeFormat` and bubble up as 500.
- A few small redundancies: `setUserNutrientRecommendation` discards the upsert RETURNING row, `clearUserNutrientRecommendation` does an unneeded user-side SELECT after DELETE, and `unitFor` references its lookup table declared after it.

**P3 (nitpicks):** see inline comments on `ReferenceRangeBar`, the seed loader, the DELETE response shape, and the flag-color semantics in the Overview.

 / Claude

apps/backend/src/routes/nutrient-recommendations-router.ts
```diff
@@ -0,0 +1,61 @@
...
+  )
+
+  router.put<{ nutrient_name: string }, NutrientRecommendationResponse, UpsertNutrientRecommendationBody>(
+    '/:nutrient_name',
+    authMiddleware,
+    validateBody(upsertNutrientRecommendationBodySchema),
+    async (req, res) => {
+      const effective = await setUserNutrientRecommendation(req.user!, req.params.nutrient_name, req.body)
+      if (!effective) {
+        return res.status(500).json({ error: 'Failed to resolve effective recommendation', success: false })
```
fiddur: **P1: 500 on legitimate suppression**

The body validator allows `{recommended_low: null, recommended_high: null}` (just needs at least one key present), and a fresh upsert with only `{recommended_high: null}` defaults the other column to NULL. Both write successfully and the merge correctly returns `null` (suppression) — but this branch then surfaces it as a 500 even though nothing failed.

The MCP variant handles this fine (returns `data: null, success: true`). Drop the `if (!effective)` branch and just return `res.json({ data: effective ?? undefined, success: true })`, mirroring DELETE.

The existing test `'500 if the service can’t resolve effective state'` enshrines this bug — replace with a "200 with no data on full suppression" test.

apps/backend/src/routes/nutrient-recommendations-router.ts
```diff
@@ -0,0 +1,61 @@
...
+    '/',
+    authMiddleware,
+    async (req, res) => {
+      const recommendations = await getEffectiveRecommendations(req.user!)
+      res.json({ recommendations, success: true })
+    },
+  )
+
+  router.put<{ nutrient_name: string }, NutrientRecommendationResponse, UpsertNutrientRecommendationBody>(
+    '/:nutrient_name',
```
fiddur, line 54: **P2: no validation on `nutrient_name`**

`req.params.nutrient_name` is taken straight from the URL and persisted as TEXT with no length cap or allowlist. A user can `PUT /nutrient-recommendations/<arbitrary-string>` and accumulate junk rows in their per-user table.

Self-DoS only (the table is per-user), but worth a `z.enum(NUTRIENT_FIELD_NAMES)` validator or a membership check against the known nutrient set before the upsert. Same applies to the MCP path.

apps/backend/src/routes/nutrient-recommendations-router.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+      res.json({ data: effective, success: true })
+    },
+  )
+
+  router.delete<{ nutrient_name: string }, NutrientRecommendationResponse>(
+    '/:nutrient_name',
+    authMiddleware,
+    async (req, res) => {
+      const { effective } = await clearUserNutrientRecommendation(req.user!, req.params.nutrient_name)
```
fiddur: **P3:** the service returns `{ cleared, effective }` but the response only forwards `effective`. A client that wants to distinguish "I removed an existing override" from "there was no override and there's no central default either" can't — both shapes look like `data: undefined`. Consider adding `cleared` to `NutrientRecommendationResponse` or a dedicated DELETE response type.

apps/backend/src/services/queries/meal-period-summary.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/** Convert a UTC Date to its local YYYY-MM-DD key for the given tz. */
+const localDateKey = (d: Date, tz: string): string => getDateKeyFormatter(tz).format(d)
+
+const aggregateNutrientsFromLinks = (links: MealFoodItemLink[]): Map<string, number> => {
+  const totals = new Map<string, number>()
+  for (const link of links) {
+    for (const field of NUTRIENT_FIELD_NAMES) {
+      const v = link[field]
+      if (typeof v === 'number' && v > 0) {
```
fiddur, line 72: **P1: per-nutrient `days_with_data` overstates adequacy**

The MCP tool description and PR text both say:

> "avg = mean across days that have any meal (so sparse logs are not dragged toward zero)"

But this filter (`v > 0`) makes the day count per-nutrient: `perNutrientDayCount` ends up counting days where *this specific nutrient* was non-zero, not days where the user logged any meal.

Concrete impact: a user who logs meals all 30 days but only eats fiber-containing food on 10 shows fiber avg = `total / 10` instead of `total / 30`. The Reference Range bar then reports them in-range when they're actually at a third of the recommendation. For a UX whose entire job is "are you hitting your targets?", this is systematically misleading.

Two reasonable fixes — pick one and document it:
- (a) Track a single per-day "had any meal logged" set (size = denominator for *all* nutrients) — matches the description.
- (b) Keep current semantics but rename the field to `days_with_value` and update the description.

(a) is closer to what users want; (b) is defensible for nutrients legitimately null in the food DB (e.g. manual entries with only calories) but is the more surprising default.

The existing test `'averages over days that have data, not the full range'` only uses calories so doesn't exercise this — add a multi-nutrient case with a 0-fiber day to lock the chosen behavior in.

packages/api-spec/src/schemas/nutrient-recommendations.ts
```diff
@@ -0,0 +1,181 @@
...
+  const a = new Date(`${start}T00:00:00Z`).getTime()
+  const b = new Date(`${end}T00:00:00Z`).getTime()
+  return Math.floor((b - a) / 86_400_000) + 1
+}
+
+export const nutrientPeriodSummaryQuerySchema = z
+  .object({
+    start: dateOnlySchema.meta({ description: 'Inclusive start date (YYYY-MM-DD)' }),
+    end: dateOnlySchema.meta({ description: 'Inclusive end date (YYYY-MM-DD)' }),
+    tz: z
```
fiddur, line 171: **P2: `tz` not validated as a real IANA zone**

`?tz=Foo/Bar` will throw `RangeError: Invalid time zone specified` from `new Intl.DateTimeFormat('sv-SE', { timeZone: 'Foo/Bar' })` inside `getDateKeyFormatter`, which Express turns into an unhandled 500.

Options:
- Refine in Zod using `Intl.supportedValuesOf('timeZone').includes(v)` (Node 18+ / modern browsers), or
- try/catch the formatter construction in `meal-period-summary.ts` and fall back to UTC.

apps/backend/src/services/nutrient-recommendations.ts
```diff
@@ -0,0 +1,130 @@
...
+  return mergeOne(nutrientName, centralRow ?? undefined, userRow ?? undefined)
+}
+
+export const setUserNutrientRecommendation = async (
+  user: string,
+  nutrientName: string,
+  input: UserNutrientRecommendationInput,
+): Promise<NutrientRecommendation | null> => {
+  await dbUpsertUserNutrientRecommendation(user, nutrientName, input)
+  return getEffectiveRecommendation(user, nutrientName)
```
fiddur: **P2:** the upsert already returns the row via RETURNING, but it's discarded here and `getEffectiveRecommendation` then fires two more queries to refetch it. Pass the upsert's return value into a synchronous merge with a single `getSharedNutrientRecommendation` call instead. Cheap, but trivially avoidable.

apps/backend/src/services/nutrient-recommendations.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export const clearUserNutrientRecommendation = async (
+  user: string,
+  nutrientName: string,
+): Promise<{ cleared: boolean; effective: NutrientRecommendation | null }> => {
+  const cleared = await dbClearUserNutrientRecommendation(user, nutrientName)
+  const effective = await getEffectiveRecommendation(user, nutrientName)
```
fiddur: **P2:** after `dbClearUserNutrientRecommendation` succeeds there can't be a user row, so `getEffectiveRecommendation` does a wasted `dbGetUserNutrientRecommendation`. Inline a direct `getCentralDb().getSharedNutrientRecommendation(...)` lookup.

apps/backend/src/services/nutrient-recommendations.ts
```diff
@@ -0,0 +1,130 @@
...
+  row.source_version ? `${row.source} ${row.source_version}` : row.source
+
+/**
+ * Resolve the unit for a nutrient name. Prefer the central row's stored unit,
+ * fall back to the api-spec NUTRIENT_FIELDS table so a user override on a
+ * nutrient that has no central default still surfaces the right unit instead
+ * of an empty string.
+ */
+const unitFor = (name: string, centralRow: CentralRow | undefined): string =>
+  centralRow?.unit ?? NUTRIENT_UNIT_BY_NAME.get(name) ?? ''
```
fiddur, line 50: **P2 (style):** `unitFor` reads `NUTRIENT_UNIT_BY_NAME` declared on the next line. Works at runtime (the call site is reached after module init), but reads backward — define the lookup table first, then the helper that consumes it.

apps/backend/src/mcp/nutrient-recommendation-tools.ts
```diff
@@ -0,0 +1,81 @@
...
+  )
+
+  server.tool(
+    'set_nutrient_recommendation',
+    [
+      'Upsert a per-user override of a nutrient recommendation range.',
+      'Pass `null` for either bound to suppress the central default for that bound; pass both as null to suppress the recommendation entirely (the nutrient will still be measured but no range is shown).',
+    ].join(' '),
+    {
+      nutrient_name: z
```
fiddur: **P2: same `nutrient_name` validation gap as the REST route.**

`z.string()` with no allowlist means a model can call `set_nutrient_recommendation` with any string and pollute the per-user table. Constrain to the known field set (`z.enum(NUTRIENT_FIELD_NAMES)` or similar).

apps/web/src/components/ReferenceRangeBar.tsx
```diff
@@ -17,29 +17,91 @@ const FLAG_COLORS: Record<string, string> = {
...
+    }
   }
+  if (high !== undefined) {
+    const displayMin = Math.min(0, value)
+    const displayMax = Math.max(high * 1.3, value * 1.1)
+    const displaySpan = displayMax - displayMin || 1
+    return {
+      displayMin,
+      displayMax,
+      normalStartPct: ((displayMin - displayMin) / displaySpan) * 100, // 0
```
fiddur: **P3:** `((displayMin - displayMin) / displaySpan) * 100` is a verbose `0` — the trailing comment says so. Just write `0`.

apps/web/src/components/ReferenceRangeBar.tsx
```diff
@@ -17,29 +17,91 @@ const FLAG_COLORS: Record<string, string> = {
...
+      normalEndPct: ((high - displayMin) / displaySpan) * 100,
+    }
+  }
+  // low !== undefined && high === undefined
+  const lo = low!
```
fiddur: **P3:** `low!` works but a `low as number` (or relying on the conditional narrowing from the two `if` branches above) reads better than the non-null assertion.

apps/backend/src/services/central-nutrient-recommendations.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Wrapped in a transaction with `pg_advisory_xact_lock` so multiple backend
+ * instances starting in parallel serialize through one writer instead of
+ * each issuing the same ~30 INSERTs.
+ */
+export const seedSharedNutrientRecommendations = async (client: pg.Client): Promise<void> => {
+  await client.query('BEGIN')
+  try {
+    await client.query('SELECT pg_advisory_xact_lock($1)', [NNR2023_SEED_ADVISORY_LOCK_KEY.toString()])
+    for (const entry of nnr2023Seed) {
+      if (entry.recommended_low === null && entry.recommended_high === null) continue
```
fiddur: **P3:** silently skipping seed entries where both bounds are null hides bugs in `nnr2023Seed.ts`. The table CHECK constraint already forbids that state, so a malformed entry would only fail via `continue` — never via a clear error. Throw or `console.warn` so a typo is caught at startup.

Also: this loop is upsert-only — a nutrient *removed* from `nnr2023Seed` will linger in `shared_nutrient_recommendations` forever. Worth either a delete pass for unknown rows or a comment noting that drift is acceptable.

apps/web/src/pages/Meals/MealsOverview.tsx
```diff
@@ -0,0 +1,298 @@
...
+ * that maps to a future tier on the override schema.
+ */
+const flagFor = (
+  value: number,
+  low: number | null | undefined,
+  high: number | null | undefined,
+): ReportFlag => {
+  if (low !== null && low !== undefined && value < low) return 'low'
+  if (high !== null && high !== undefined && value > high) return 'high'
+  return 'normal'
```
fiddur, line 76: **P3:** `FLAG_COLORS` was designed for lab reports (HRV, body fat) where low = red and high = yellow. That semantics doesn't fit dietary contexts symmetrically — exceeding the salt cap or the saturated-fat cap is *more* concerning than missing the floor on protein. Acceptable for v1 but worth a follow-up to either invert per-nutrient or introduce a separate `dietary` flag palette.

apps/backend/src/services/central-nutrient-recommendations.ts
fiddur: Fixed in 52bdecd: now throws on a null/null seed entry instead of silent continue. Added a docstring noting upsert-only / removed-nutrient drift behavior.

apps/web/src/components/ReferenceRangeBar.tsx
fiddur: Fixed in 52bdecd: bound to a typed local `const lo = low as number` so the narrowing carries through the local arithmetic without the non-null assertion.

apps/web/src/pages/Meals/MealsOverview.tsx
fiddur: Acknowledged in 52bdecd with a comment in MealsOverview explaining the v1 symmetric-palette simplification and the per-nutrient-invert / dedicated dietary-palette options for follow-up. Tracking as a P3 follow-up rather than blocking this PR.

apps/backend/src/services/nutrient-recommendations.ts
fiddur: Fixed in 52bdecd: setUserNutrientRecommendation now passes the upsert RETURNING row + a single `getSharedNutrientRecommendation` into mergeOne. One DB write + one read instead of three reads.

apps/backend/src/services/nutrient-recommendations.ts
fiddur: Fixed in 52bdecd: clearUserNutrientRecommendation only fetches the central row after DELETE — the user side is known empty.

apps/backend/src/mcp/nutrient-recommendation-tools.ts
fiddur: Fixed in 52bdecd: shared `nutrientNameMcpSchema = z.enum(NUTRIENT_FIELD_NAMES)` on both set_nutrient_recommendation and clear_nutrient_recommendation MCP tools.

apps/backend/src/services/nutrient-recommendations.ts
fiddur: Fixed in 52bdecd: NUTRIENT_UNIT_BY_NAME declared before unitFor.

apps/web/src/components/ReferenceRangeBar.tsx
fiddur: Fixed in 52bdecd: replaced with literal `0`.

apps/backend/src/routes/nutrient-recommendations-router.ts
fiddur: Fixed in 52bdecd: dropped the 500 branch, PUT now returns 200 with no data on full suppression — mirrors DELETE. Test replaced with 'returns 200 with no data when fully suppressed and no central default remains'.

apps/backend/src/routes/nutrient-recommendations-router.ts
fiddur: Fixed in 52bdecd: REST adds requireKnownNutrient gate (400 if name not in NUTRIENT_FIELD_NAMES) before any service call. Covered by new 400 tests for both PUT and DELETE.

apps/backend/src/services/queries/meal-period-summary.ts
fiddur: Fixed in 52bdecd: top-level `days_with_meals` is now the single avg denominator for every nutrient. The per-nutrient field is renamed to `days_with_value` and kept as a diagnostic. Added a multi-nutrient test ('intermittent nutrients average over days_with_meals, not days_with_value') with a fiber-on-only-1-of-3-days case to lock the new behavior in. Frontend EnergySection updated to read days_with_meals.

apps/backend/src/routes/nutrient-recommendations-router.ts
fiddur: Fixed in 52bdecd: added ClearNutrientRecommendationResponse with a `cleared` boolean. DELETE now returns `{ cleared, data?, success }`; tests cover both cleared=true and cleared=false branches.

packages/api-spec/src/schemas/nutrient-recommendations.ts
fiddur: Fixed in 52bdecd: tz is refined in the Zod schema via `Intl.supportedValuesOf('timeZone').includes(v)` with a DateTimeFormat try/catch fallback for older runtimes. Invalid tz now lands as 400 instead of bubbling a 500.

PR #735, opened 2026-05-10
feat(activities): proper multi-target overrides + enable edit on merged-activity view

fiddur:
## Summary

Replaces the single-target `overrides_id` FK with a proper many-to-many `activity_override_targets` join table. One aurboda override row can now claim a cross-source merge group as a whole (Garmin + Strava + Health Connect → one override), instead of targeting one source and relying on overlap-based merging to drag the rest along.

Per user feedback on the earlier single-target hack:
> I think it would be better to have a proper multi-activity override, not limiting the override to a single activity, considering the multi-sync is a common pattern.

## What changes

**Schema** — new `activity_override_targets (override_id, target_id)` join table. UNIQUE on `target_id` (each target overridden by at most one override). Both FKs cascade-delete; an AFTER DELETE trigger removes the override row when its LAST target is gone, restoring the old `ON DELETE CASCADE` semantics. Migration backfills from the existing `overrides_id` column then drops it.

**Activity type** — `Activity.overrides_id?: string` → `Activity.override_target_ids?: string[]`. Single-target overrides become 1-element arrays. SELECTs in `queries.ts`/`supersession.ts` populate the array via correlated subquery against the join table.

**Service layer** — `UpdateActivityInput` and `updateActivityBodySchema` gain optional `override_target_ids: string[]`. `dbInsertOverride(user, targetIds, ...)` inserts the activity row + N join-table rows inside a transaction. Single-target callers still work (default to `[targetedId]`).

**Merge logic** — `isOverrideLinked`, `pickOverrideWinner`, `indexOverrides`, `pairAnchor` rewritten to walk the array. The pairing step unions the override with EVERY target it claims regardless of the 120-second inner-loop window.

**Frontend** — `EntityDetail.saveMutation` collects all non-aurboda source ids when saving from a merged view and sends them in the body. Combined with the `forceMergedSpanForOverride` helper, the new override spans the whole merged group.

**Edit button on merged view** (carried from the original PR scope) — `mergedEditAction` routes to an existing aurboda source when one exists; otherwise enables in-place edit that creates the multi-target override on save.

## Why this design

- **One override per target** (UNIQUE on `target_id`): prevents ambiguous "which override wins for this source" at merge time. If you want to change which sources are overridden, edit the existing override; you can't have two overrides claiming the same source.
- **Override survives partial target loss**: deleting Garmin's row leaves the override claiming Strava + HC; the override row is only auto-deleted when its LAST target goes (matches the spirit of the old single-FK CASCADE).
- **Source-priority hierarchy still works**: when an override's targets span multiple sources (Garmin + Strava + HC), the override still picks the merge-group winner via the existing `pickOverrideWinner` walking the array.

## Test plan

- [x] Backend suite 2139/2139
- [x] Web suite 379/379
- [x] New merge.test.ts case: 4-row group (3 sources + multi-target override) collapses to one with the override winning
- [x] Cascade test: deleting the only target removes the override via the trigger
- [x] `pnpm fix` + `pnpm check` clean
- [ ] Manual on staging:
  - Open the linked merged URL `/detail/activity/merged:1778bcdc-…`; click edit; verify in-place edit. Save → navigates to the new override; the override has all source ids in `override_target_ids`.
  - Re-sync Garmin → override unchanged; merged view still shows override values.
  - Delete one source row in DB → override survives; remaining targets still claimed.
  - Delete last target → override row gone (trigger).

## Migration risk

The `overrides_id` column is dropped after backfill. A rollback to a previous version would error on the missing column. Standard schema migration risk; for this single-developer project it's acceptable.


apps/backend/src/services/mutations.ts
```diff
@@ -566,8 +573,16 @@ export async function updateActivity(
     // carrying the synced row's snapshot overlaid with the user's input.
     // Mirror in-place semantics: input.notes / input.title === undefined keeps
     // the snapshot value; explicit '' (or null) clears it.
+    //
+    // Multi-target (#735): the caller may pass `override_target_ids` from a
+    // merged-view edit so the new override claims all the merged source rows
+    // at once. Always include `target.id` in the list (defaulting to just
+    // that single source when the caller didn't pass anything).
+    const targetIds = input.override_target_ids?.length
+      ? Array.from(new Set([target.id!, ...input.override_target_ids]))
```
fiddur, line 583: **P2:** Backend trusts `input.override_target_ids` shape only — frontend filters `r.source !== 'aurboda'` (`EntityDetail/index.tsx:597`) but the backend doesn't enforce it. A buggy or non-web client could pass an aurboda id as a target, creating an "override of an override" the merge logic isn't designed for.

**Suggestion:** add a defensive check before `dbInsertOverride`:
```ts
const targetSources = await query(
  user,
  `SELECT id, source FROM activities WHERE id = ANY($1)`,
  [targetIds],
)
if (targetSources.rows.some(r => r.source === 'aurboda')) {
  throw new Error('override targets must be synced rows')
}
```

apps/backend/src/db/activities/queries.ts
```diff
@@ -9,6 +9,20 @@ import { query } from '../connection.ts'
...
+ * `override_target_ids` is built from the `activity_override_targets` join
+ * table via a correlated subquery; NULL when the row has no targets and is
+ * normalised to `undefined` by the row mapper.
+ *
+ * Two variants because some queries use a table alias (`a`) and some don't.
+ */
+const ACTIVITY_COLUMNS_BARE = `id, source, external_id, activity_type, start_time, end_time, title, notes, data, deleted_at, superseded_by,
+  (SELECT array_agg(target_id) FROM activity_override_targets WHERE override_id = activities.id) AS override_target_ids`
+
+const ACTIVITY_COLUMNS_ALIAS = `a.id, a.source, a.external_id, a.activity_type, a.start_time, a.end_time, a.title, a.notes, a.data, a.deleted_at, a.superseded_by,
```
fiddur: **P3:** `ACTIVITY_COLUMNS_BARE` and `ACTIVITY_COLUMNS_ALIAS` are kept manually in sync — easy to drift if a column gets added or the subquery shape changes. Consider a single helper:
```ts
const activityColumns = (alias = 'activities') => `
  ${alias}.id, ${alias}.source, ${alias}.external_id, ${alias}.activity_type,
  ${alias}.start_time, ${alias}.end_time, ${alias}.title, ${alias}.notes,
  ${alias}.data, ${alias}.deleted_at, ${alias}.superseded_by,
  (SELECT array_agg(target_id) FROM activity_override_targets
   WHERE override_id = ${alias}.id) AS override_target_ids
`
```
Then call sites pass `'a'` or omit it. Mild — feel free to ignore if you prefer the explicit pair.

apps/backend/src/schema/activities.ts
```diff
@@ -28,8 +25,56 @@ export const activitiesTables: Record<string, string> = {
...
+    END;
+    $$ LANGUAGE plpgsql;
+
+    DO $$ BEGIN
+      IF NOT EXISTS (
+        SELECT 1 FROM information_schema.triggers
+         WHERE event_object_table = 'activity_override_targets'
+           AND trigger_name = 'aot_delete_orphan_override'
+      ) THEN
+        CREATE TRIGGER aot_delete_orphan_override
```
fiddur, line 73: **P2:** The PR description's test plan checks "Cascade test: deleting the only target removes the override via the trigger", but I don't see a corresponding test in the diff (neither in `merge.test.ts` nor in `activities.integration.test.ts`). The trigger is a sharp edge — worth at least one integration test that:

1. Inserts an override targeting two synced rows; deletes one target → asserts override row still exists with the remaining target in `override_target_ids`.
2. Deletes the last target → asserts the override row is gone.

Without a test, a future refactor of the trigger function (or accidental drop of the AFTER DELETE binding) would silently break the cascade semantics.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -559,6 +573,30 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
...
+      // guarantees `body` is non-empty, so the PATCH always reaches the
+      // backend (which then creates the override row).
+      const forced = forceMergedSpanForOverride(
+        isMergedActivity,
+        Boolean(aurbodaSource),
+        activity.merged_start_time ?? activity.start_time,
+        activity.merged_end_time ?? activity.end_time,
+      )
+      if (forced?.start_time && body.start_time === undefined) body.start_time = forced.start_time
+      if (forced?.end_time && body.end_time === undefined) body.end_time = forced.end_time
```
fiddur, line 594: **P3:** The `body.start_time === undefined` / `body.end_time === undefined` guards mean an explicit user edit to one side wins, but the merged-span value still patches the *other* side. That's probably the intent — but it's subtle: if the user shrinks the end_time on a merged-view edit, we silently force-write the merged-group start_time too. Worth a one-line comment at the call site noting this asymmetry, so a future reader doesn't "simplify" by removing the guards.

fiddur (commented):
## Summary

Reviewed the multi-target override migration. No blockers found — the per-user `Client` (in `dbByUser`) keeps `BEGIN`/`COMMIT` on one connection, and the per-user database (`aurboda_${user}`) means cross-tenant target ids would FK-fail anyway, so the new code is safe on both fronts.

A handful of P2/P3 follow-ups inline. One bonus observation that isn't inline-able:

**P3 — `connection.ts` migration dance is wasteful on fresh DBs.** Lines 678 / 690 / 696 (`ALTER TABLE ADD COLUMN overrides_id`, the CHECK constraint, the partial unique index) are kept as pre-existing migration steps, then immediately dropped at lines 824-826. On a brand-new database we add a column + constraint + index purely to drop them seconds later. Cleanest fix: wrap the three legacy create-steps in the same `if (hasOverridesIdColumn.rows[0]?.exists)` guard you already have for the drop steps — fresh DBs never see `overrides_id`, only existing DBs run the backfill+drop dance.

 / Claude

PR #736, opened 2026-05-10
fix(meals): resolve sensitivity flags live at meal read time

fiddur:
## Summary

- Drops the `meal_food_items.sensitivities` snapshot column. Flags are now resolved live from `food_item_sensitivities` at meal read time, mirroring the existing live name/icon resolution.
- `meals.sensitivities` keeps holding *direct* meal-level flags only; the API response exposes the **deduped union of direct + item-derived flags**, so toggling a flag on a food item propagates immediately to every past meal that references it (no resnapshot needed).
- Snapshots still exist for nutrients (frozen historicals); names, icons, and now flags do not.

## Why

From the meals UI, marking e.g. *Gluten* on a food item chip:
1. POST `/food-items/:id/sensitivities` → 200 ✓ (junction updated)
2. The chip dot didn't appear — the meals query refetched the **old snapshot** off `meal_food_items.sensitivities`.
3. Reload appeared to lose the flag for the same reason.

The fix removes the snapshot vs. truth divergence rather than papering over it.

## Test plan

- [x] `pnpm check` — 0 errors.
- [x] Backend unit tests — 1757 passed.
- [x] Backend meal integration tests — 13 passed (incl. 3 new: post-flagging propagates to past meals; un-flagging removes from past meals; direct + item-derived dedupe into union).
- [ ] Manual smoke on aurboda.net after deploy: flag a food item from a meal chip → dot appears immediately, persists through reload.


fiddur (commented):
## Summary

Nice fix — replacing the snapshot with live resolution removes the stale-flag class of bug entirely, and the trade-off (live `food_item_sensitivities` per read) mirrors what you already do for name/icon. The three integration tests cover the behavior shifts (propagate, un-propagate, direct + item-derived union) clearly.

A few things worth a look before merging:

- **P2**: Hard-deleted food items now silently lose their sensitivity tags on historical meals (no `legacy_*` fallback equivalent). See inline.
- **P2**: `DROP COLUMN` is destructive + runs in `migrateSchema` at startup. If there's ever any rolling/overlapping deploy of backend instances, the old instance still has `sensitivities` in `JUNCTION_COLUMNS` and will 500 on SELECTs. For single-instance deploys this is a non-issue.
- **P3**: A couple of minor wording / test-assertion nits inline.

No P1s. / Claude

apps/backend/src/services/meals.ts
```diff
@@ -162,7 +181,7 @@ const linksToFoodItems = (
       carbs: link.carbs as number | undefined,
       fat: link.fat as number | undefined,
       fiber: link.fiber as number | undefined,
-      sensitivities: (link as { sensitivities?: string[] }).sensitivities,
+      sensitivities: sensitivities.length > 0 ? sensitivities : undefined,
```
fiddur, line 191: **P2: No fallback for sensitivities when the canonical food item is gone.**

The doc-comment a few lines up calls out the hard-deleted-without-merge-repointer case for name/icon, and `name: display?.name ?? link.legacy_food_item_name ?? ''` (line 175) preserves the last-known label. Sensitivities have no analogous fallback — `sensitivityMap.get(link.food_item_id) ?? []` just returns `[]` when the canonical is gone, so historical meals whose food item was hard-deleted will silently lose every sensitivity tag they used to display.

If hard-delete-without-repointer is a real path (the existing legacy-name fallback suggests it is), this is a small behavioral regression for users who relied on those tags for filtering / awareness. Options:

1. Accept it (sensitivities aren't safety-critical) and call it out in the docstring so the asymmetry vs. name/icon is intentional.
2. Keep a `legacy_sensitivities TEXT[]` column on the junction (analogous to `legacy_food_item_name`) populated only when the canonical disappears, so the old snapshot survives on orphaned rows.

Worth at least confirming this is the trade-off you want.

apps/backend/src/db/connection.ts
```diff
@@ -1031,9 +1031,10 @@ export const migrateSchema = async (user: string) => {
     for (const field of NUTRIENT_FIELD_NAMES) {
       await query(db, `ALTER TABLE meal_food_items ADD COLUMN IF NOT EXISTS ${field} DOUBLE PRECISION`)
     }
-    // Snapshot a food item's sensitivity flag names at meal-add time so
-    // meal totals carry the inheritance even after the food's flags change.
-    await query(db, `ALTER TABLE meal_food_items ADD COLUMN IF NOT EXISTS sensitivities TEXT[]`)
+    // Sensitivity flags used to be snapshotted here at meal-add time, but
+    // that froze stale flag state across food-item edits. They're now
+    // resolved live from food_item_sensitivities at meal read time.
+    await query(db, `ALTER TABLE meal_food_items DROP COLUMN IF EXISTS sensitivities`)
```
fiddur, line 1107: **P2: One-way destructive migration; flag the deploy plan.**

`DROP COLUMN IF EXISTS sensitivities` runs as part of `migrateSchema` on startup. Two things to be aware of:

1. **Rolling deploy.** If V1 (with `sensitivities` in `JUNCTION_COLUMNS`) is still serving while V2 starts and runs the migration, V1's SELECTs against `meal_food_items` will fail with `column "sensitivities" does not exist`. If aurboda always restarts as a single instance you can ignore this; if not, consider running the column drop as a separate post-deploy migration after V1 is fully drained.
2. **Rollback.** Once dropped, the snapshot data is gone. If you need to roll back to a code version that still references the column, you'd have to re-add it as an empty `TEXT[]`. Worth knowing before deploy.

apps/backend/src/services/meals.ts
```diff
@@ -117,6 +117,22 @@ interface MealsResult {
 
 type EnrichedMeal = Meal & { nutrients?: Record<string, number>; nutrient_data_incomplete?: boolean }
 
+/**
+ * Combine the meal's directly-set flags (stored on the meal row) with the
+ * flags currently set on each of its food items. Flags from items are NOT
+ * snapshotted — they're resolved live, so renames/edits on a food item
+ * propagate automatically to every meal that references it.
+ */
+const computeMealSensitivities = (meal: EnrichedMeal): string[] | undefined => {
```
fiddur, line 126: **P3 (nit):** The trailing sentence — *"renames/edits on a food item propagate automatically to every meal that references it"* — is true of the broader live-resolution design, but this function only merges flag arrays. Renames/edits propagate via `displayMap` in `linksToFoodItems`, not here. Consider trimming to something flag-specific, e.g. *"flag changes on a food item propagate automatically to every meal that references it."*

apps/backend/src/services/meals.integration.test.ts
```diff
@@ -343,6 +338,96 @@ describe('Meals service integration tests', () => {
...
+
+      const meal = await addMeal(user, {
+        food_items: [{ food_item_id: milk.id, name: 'Milk', quantity: 200, unit: 'ml' }],
+        meal_type: 'breakfast',
+        time: '2026-04-26T08:00:00Z',
+      })
+      const mealId = meal.data!.id
+
+      const before = await getMeal(user, mealId)
+      expect(before.data!.food_items?.[0].sensitivities).toEqual(['Dairy'])
```
fiddur, line 390: **P3:** While you're here, an extra assertion would lock down the union behavior for the "only fromItems, no direct" branch:

```ts
expect(before.data!.sensitivities).toEqual(['Dairy'])
```

As written, the food-item-level flag is asserted but the meal-level union (the very thing `computeMealSensitivities` exists for, when direct is empty) is not. Symmetrical to the assertion on line 397 for the post-unflag state.

PR #763, opened 2026-05-13
feat(daily-summary): surface activity notes and structured data

fiddur:
## Summary

- Daily summary now passes through each activity's inline `notes` and `data` blob. The yoga note typed on an exercise activity (and structured fields like `partner: Sara`) finally reach the AI consumer.
- Retires the redundant `exerciseType` / `exerciseTypeName` keys we used to mirror inside `data`. `activity_type` already carries the specific exercise type (`yoga`, `running`, `weightlifting`, …) since the unified-activities migration, so duplicating it in `data` only polluted the new daily-summary `data` passthrough.

## What changed

**api-spec**
- `ActivitySummary` gains `notes` and `data`, loses `exercise_type`.
- `AddActivityBody` / `UpdateActivityBody` lose `exercise_type` — callers pass `activity_type` directly.
- `SourceRecord` loses `exercise_type_name`, gains `activity_type`.

**Backend**
- `getDailySummary` surfaces `notes` + `data`; drops the `data.exerciseType → exercise_type` derivation.
- REST `POST/PATCH /activities` and MCP `add_activity` / `update_activity` no longer accept `exercise_type` or write `exerciseType`/`exerciseTypeName` into `data`.
- FIT upload sets the specific `activity_type` directly (`running_treadmill` instead of generic `exercise` + side-channel).
- Health Connect ingest still resolves the HC integer to a specific `activity_type`, but strips both keys before persisting to `activities.data`. Raw HC payloads remain intact in `raw_records`.
- `mergeOverlappingActivities`' generic-exercise absorption no longer needs the HC integer fallback — `activity_type === 'exercise'` is now the source of truth.
- `buildMergedResponse` exposes `activity_type` per source record instead of the HC name.
- Startup migration strips `exerciseType` / `exerciseTypeName` from existing `activities.data` rows (idempotent JSONB `-` operator).

**Web**
- `Timeline/getExerciseTypeName` and `EntityDetail/resolveExerciseType` resolve from `activity_type` instead of the dropped `data` keys.

## Test plan

- [x] `pnpm --filter aurboda-backend check` — 0 errors
- [x] `pnpm --filter aurboda-web check` — 0 errors
- [x] `pnpm --filter @aurboda/api-spec check` — 0 errors
- [x] Backend unit tests: 1769 / 1769 passing
- [x] Backend integration tests for activities: 36 / 36 passing
- [x] Web tests: 397 / 397 passing
- [x] New tests in `daily-summary.test.ts` verify `notes` + `data` passthrough and the "no notes/data" case
- [ ] Manual smoke after deploy: open a daily summary in the UI, confirm a yoga activity surfaces its `notes` text and any `data.partner` etc.

## Follow-up (PR #2)

Migrate the `activities.notes` column into the `notes` table so activities have a single text channel (currently both `notes` column and attached `comments[]`). Tracked separately because it touches Health Connect imports and Android.

### DIFF
--- a/apps/backend/src/db/activities/merge.ts
+++ b/apps/backend/src/db/activities/merge.ts
@@ -385,13 +385,7 @@ export const mergeOverlappingActivities = (
   return absorbGenericExercises(result)
 }
 
-const GENERIC_EXERCISE_CODES = new Set([0, 2]) // UNKNOWN=0, OTHER_WORKOUT=2
-
-const isGenericExercise = (a: MergedActivity): boolean => {
-  if (a.activity_type !== 'exercise') return false
-  const code = a.data?.exerciseType
-  return typeof code !== 'number' || GENERIC_EXERCISE_CODES.has(code)
-}
+const isGenericExercise = (a: MergedActivity): boolean => a.activity_type === 'exercise'
 
 /** Check if generic's duration overlaps >50% with another activity. */
 const findAbsorbingActivity = (
--- a/apps/backend/src/db/connection.ts
+++ b/apps/backend/src/db/connection.ts
@@ -1199,6 +1199,19 @@ export const migrateSchema = async (user: string) => {
     )
   }
 
+  // Strip redundant exerciseType / exerciseTypeName keys from activities.data.
+  // The specific activity_type (yoga, running, ...) is the authoritative carrier;
+  // keeping both in data duplicates state and leaks into the daily-summary `data`
+  // passthrough. Raw HC payloads still hold the originals in raw_records.
+  if (existingTableNames.has('activities')) {
+    await query(
+      db,
+      `UPDATE activities
+       SET data = data - 'exerciseType' - 'exerciseTypeName'
+       WHERE data ?| ARRAY['exerciseType', 'exerciseTypeName']`,
+    )
+  }
+
   // Step 3: Align Garmin typeKey names with HC exercise type names.
   // Garmin uses modifier_noun (e.g., treadmill_running), HC uses noun_modifier (running_treadmill).
   if (existingTableNames.has('activities')) {
--- a/apps/backend/src/db/health-connect.ts
+++ b/apps/backend/src/db/health-connect.ts
@@ -27,6 +27,18 @@ const resolveExerciseActivityType = (data: Record<string, unknown>): string => {
   return getExerciseTypeName(exerciseType) ?? 'exercise'
 }
 
+/**
+ * Strip the HC exerciseType / exerciseTypeName keys before persisting to
+ * activities.data — activity_type now carries that information, so keeping
+ * them duplicates state and pollutes the daily-summary `data` passthrough.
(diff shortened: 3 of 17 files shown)


fiddur (commented):
Took a thorough read — overall this is a nice cleanup. Single source of truth for the exercise subtype, `data` passthrough finally surfaces what users typed (notes, partner, reps, etc.), and the migration query is correctly gated by a `?|` so it's idempotent. The simplified `isGenericExercise = a.activity_type === 'exercise'` lines up with the new ingest behavior. Tests cover both the new passthrough and the absent-data path.

A few notes:

1. **Redundant casts** (CLAUDE.md says avoid casts). `apps/backend/src/routes/activities-router.ts` casts `bodyData as Record<string, unknown> | undefined` in both the POST (~`:200`) and PATCH (~`:407`) handlers. Since `addActivityBodySchema` / `updateActivityBodySchema` now declare `data: z.record(z.string(), z.unknown()).optional()`, `bodyData` should already be `Record<string, unknown> | undefined` after destructuring. The casts look removable.

2. **AI-consumer ergonomics for `activity_type`.** Removing the MCP `exercise_type` enum drops the friendly *"Invalid exercise_type, valid types include: weightlifting, running, …"* error that was very useful to LLM callers. With `activityTypeSchema` being any snake_case string and the real validation happening at the `activity_type_definitions` runtime check, an AI passing `activity_type: \"running_session\"` (made up) presumably fails at the DB. Worth checking that the runtime error surfaces a similarly enumerated hint — otherwise consider keeping a short examples list in the MCP tool description (looks like you already do — *\"exercise type like yoga/running/weightlifting, meditation, nap, rest, …\"* — but it's only 4 examples; maybe sample more of the common ones).

3. **Daily summary `data` blob can grow.** `daily-summary.ts` now passes `s.data` through verbatim for non-sleep activities. For aurboda-side activities it's tiny — but Garmin/FIT and Health Connect `ExerciseSessionRecord` data can carry `metadata`, `dataOrigin`, sub-objects, and source-specific fields. The migration strips only `exerciseType`/`exerciseTypeName`. Since `get_daily_summary` is the AI-facing surface, a busy day's payload could balloon. Two ideas:
   - allow-list a known set of small structured keys (`partner`, `reps`, `weight`, `dataOrigin`, …) and drop everything else, or
   - add a per-activity byte/key cap and surface a `data_truncated: true` flag instead.

   Not a blocker — but worth thinking about before this PR lands, because once consumers start relying on arbitrary `data` keys it's hard to scope back.

4. **Nit.** `stripExerciseTypeFromData` uses `_et` / `_etn` rename-to-discard. Works fine. If you find yourself adding similar strip helpers, a tiny `omitKeys(obj, ['exerciseType', 'exerciseTypeName'])` utility would read better than the destructure-with-underscore pattern. Two keys, one site — totally fine to leave as is.

LGTM modulo (1) and a thought on (3) before merge.

PR #764, opened 2026-05-13
refactor(activities): move notes from column to notes table

fiddur:
## Summary

PR #2 of the unified-notes refactor (PR #1: #763). Drops the `activities.notes` column; user-typed text now lives in the notes table alongside comments, giving each activity a single text channel.

## What changes

**DB schema**
- `activities.notes` column dropped after a one-shot startup migration that backfills non-empty values into the `notes` table.
  - User/manual rows land as `source IS NULL` (the "user-authored" set).
  - HC/sync rows land with `source = activity.source` so the outbound-sync filter doesn't echo them back.
- `notes` table is now the single source of truth.

**Writers**
- REST POST/PATCH `/activities` and MCP `add_activity` / `update_activity` route their `notes` body field through new helper `replaceUserNotes` — delete-then-insert semantics on `source IS NULL` rows, keyed by activity id. UI behavior is 1:1 with the old single-textbox.
- Health Connect ingest (`processHealthConnectData` / `processHealthConnectBatch`) persists each record's `notes` via `upsertSyncedNote(source='health_connect')`. Idempotent on re-sync.
- `mergeActivities` re-anchors notes from source activities to the merged row (`reanchorNotes`) and applies optional override via `replaceUserNotes`. The DB-level merge (`mergeOverlappingActivities`, `buildMergedActivityData`) no longer touches notes.

**Outbound HC sync**
- Payload `notes` is now `getUserNotesJoined(activity_id)` — chronological `\n`-join of source-NULL notes only. Synced notes are skipped to prevent telephone-game amplification when the same record round-trips.

**Responses**
- `Activity`, `AddedActivity`, daily-summary `ActivitySummary` — `notes` field removed. Consumers read from `comments[]`.
- `CommentSummary` / api-spec `commentSchema` gain a `source` field so callers can distinguish user-authored from synced comments.

**Frontend**
- `EntityDetail` derives the inline textarea content from `comments.filter(c => !c.source)`; saves via PATCH body `notes` (which routes server-side as above).
- `Timeline` activity tooltips switch to the same filter.

## Test plan

- [x] `pnpm check` — 0 errors across backend, web, api-spec
- [x] Backend unit tests: 1764 / 1764 passing (5 obsolete notes-on-column merge tests removed with explanatory comments)
- [x] Backend integration tests for activities: 36 / 36 passing
- [x] Web tests: 397 / 397 passing
- [x] New backend tests verify `replaceUserNotes` invocation from add_activity/update_activity and notes-on-comments passthrough from daily-summary
- [ ] Manual smoke after deploy:
  - [ ] Add an activity with a note via UI → confirm note appears as a comment, not as `activity.notes`
  - [ ] Edit the note → it overwrites the existing user-source comment without touching any HC-source comments on the same activity
  - [ ] HC sync of an exercise with a note → comment appears with `source: 'health_connect'`
  - [ ] Outbound HC sync of an aurboda activity → the HC record's `notes` field carries only user-source content

## Migration safety

- The startup migration guards on column existence (`information_schema.columns`) and idempotent INSERT (`NOT EXISTS`), so a second deploy is a no-op.
- The column drop happens immediately after backfill. If you need to roll back, the data survives in the notes table — but the column itself would have to be re-added manually before the older code could read it.


fiddur (commented):
Read this end-to-end. Nice refactor — the `notes` table as single source of truth matches the existing `comments` shape, the `source IS NULL` vs `source='health_connect'` partition cleanly prevents HC's outbound-sync telephone-game, and the `reanchorNotes` semantics in the explicit `mergeActivities` flow are exactly right. The 5 removed merge tests with explanatory placeholders make the intent clear.

A few things worth a look before this merges:

1. **(Behavior) Cross-source non-override merges silently lose sibling notes.**
   `mergeOverlappingActivities` used to concatenate `notes` across the merged group (`apps/backend/src/db/activities/merge.ts` line 156 in the old version, now removed). For override winners that concat was already skipped (correct — user intent dominates), but for the *non-override* cross-source pairing path (e.g. Garmin + Polar of the same run with `aurboda` override absent), the old code did:
   ```ts
   if (next.notes) current.notes = current.notes ? `${current.notes}\n${next.notes}` : next.notes
   ```
   Post-PR, the merged DB read doesn't carry per-source notes at all — they live in the notes table keyed by each source's `id`. `getNonSleepActivitiesMerged` only attaches notes to the winner row, so HC- or Garmin-sourced auto-notes on the *sibling* rows disappear from the daily-summary `comments[]` and the merged view (until you click through to the source record). Two ways to keep parity:
   - `getCommentsMap(allActivities)` in `daily-summary.ts:489` could include `source_ids` in the lookup, then dedupe by content; or
   - `getNonSleepActivitiesMerged` itself could augment each merged result with `comments` joined across `(winner.id, ...source_ids)`.

   Worth confirming this is intentional — if it is, a one-line note in the PR body would help (and probably in the daily-summary `comments` field doc).

2. **(CLAUDE.md says avoid casts)** `apps/backend/src/db/activities/mutations.ts:~280` and `:~320` each do `id: row.id as string` when building the `InsertedActivityKey` array. The rest of the row-mapping in the project goes through `mapActivityRow` without casts (`apps/backend/src/db/row-mappers.ts:94`). A small `mapInsertedKey(row): InsertedActivityKey` helper (or extending the existing row-mappers module) would match that pattern and remove the casts.

3. **(Edge case) Soft-deleted upsert target won't get notes attached.**
   `INSERT…ON CONFLICT DO UPDATE WHERE activities.deleted_at IS NULL RETURNING…` only returns rows whose UPDATE actually fired. If an HC re-sync targets a soft-deleted row, the update is skipped and the row isn't in the RETURNING set — so the note-attachment loop in `health-connect.ts:235-265` silently drops the note. Probably the right call (you don't want a sync to revive deleted rows), but worth a comment near that loop so the next reader doesn't think it's a bug.

4. **(Docs)** `docs/health-connect.md` describes how HC ingest persists data but doesn't mention that `notes` now go to the separate `notes` table with `source = 'health_connect'`. Per CLAUDE.md ("keep documentation in `docs/` up to date") a one-paragraph addition under "Data Flow" would help anyone debugging where a missing note went.

5. **(Tiny nit)** `replaceUserNotes` in `apps/backend/src/db/notes.ts` is delete-then-insert without an explicit transaction. Today aurboda is single-user-per-instance so concurrent edits to the same activity are vanishingly rare, but a `BEGIN…COMMIT` wrapper would be cheap insurance against a future where the API gets hit by two clients at once.

Sign-off: this is a careful PR with good test coverage. The migration is idempotent (NOT EXISTS guard + DROP COLUMN IF EXISTS + the colCheck wrap), the source-NULL invariant flows through every writer, and the test suite was updated thoughtfully. Just (1) is meaningfully a behavior question worth answering before merge.

PR #765, opened 2026-05-13
feat(meals): keep quick-log strip available after a meal exists

fiddur:
## Summary

- Removes the gate that hid the frequent-food quick-log strip once a meal was logged in a slot. The strip now stays visible and splits into two rows: **Included:** (chips for items already on the meal) and **Add:** (the existing frequent suggestions, filtered to exclude what's already there by `food_item_id`).
- Tapping a suggested chip when a meal already exists appends the food item to it via `updateMealApi` (with an optimistic merge) instead of creating a duplicate meal.
- The sensitivity-flag popover on an included chip now also has a 🗑 **Remove** control with a confirm step, so individual items can be dropped without going to the meal detail page.
- While the slot is mid-save, `meal-calories` grays out with a small spinner — signals that totals are stale and being recomputed server-side.

## Test plan

- [ ] On `/meals`, log a frequent item into an empty Lunch slot — same behavior as before (single chip click creates the meal).
- [ ] With Lunch now populated, the strip still renders. "Included:" shows the just-added chip; "Add:" shows the remaining frequents.
- [ ] Tap another chip in "Add:" → item appends to the existing Lunch meal (no new meal row appears). Calories briefly gray + spinner, then settle to the new total.
- [ ] Tap an "Included:" chip → sensitivity popover opens as before. The 🗑 Remove button at the bottom asks to confirm, then removes the item from the meal.
- [ ] An item already on the meal does **not** appear in "Add:" suggestions (dedupe by `food_item_id`).
- [ ] The Day Totals panel and overview tab refresh correctly after append/remove.
- [ ] Dark mode renders the new row labels and the popover delete divider sensibly.

> Note: type-check, lint and the web vitest suite all pass; the UI itself was not exercised in a browser by Claude — please sanity-check the flow.


fiddur (commented):
Nice — strip-stays-visible UX is a clear win, and the type cleanup on `updateMutation` (`UpdateMealBody & { id: string }` in `apps/web/src/pages/Meals/index.tsx:836`) is good. No `as` casts, snake_case fields preserved, no schema duplication.

Two non-blocking notes:

1. **Index-based removal race** (`apps/web/src/pages/Meals/index.tsx:932`). `handleRemoveFoodItem(meal, index)` and the matching `onDelete={() => onDeleteFoodItem(i)}` in the "Included:" row both target by array position. If another mutation lands between render and click (e.g. an append that resolves while the popover is open), the index now points at the wrong item. In practice React Query's optimistic updates make this unlikely, but matching by `food_item_id` (with fallback to index for ad-hoc items without an id) would be safer and self-documenting. Same applies to append: if the user double-taps two different "Add:" chips faster than the parent re-renders, the second `mapFoodItemsToInput(meal.food_items ?? [])` might read pre-optimistic state.

2. **Self-acknowledged browser check** — the PR body already flags it, just confirming: the calories-stale spinner timing and the popover delete divider in dark mode are the two spots most worth eyeballing live since they're new visuals not covered by the existing vitest suite.

Otherwise this looks good. The `MealCaloriesLine` extraction is a nice incidental cleanup.

PR #766, opened 2026-05-13
fix(meals): id-based food-item matching + icons on included chips

fiddur:
## Summary

Follow-up to #765.

- **Race fix**: `handleAppendFoodItem` and `handleRemoveFoodItem` now read the latest meal from the React Query cache before computing the next `food_items` payload, so two clicks that land between renders no longer both base their update on stale state. `handleRemoveFoodItem` also identifies the target by `food_item_id` (with a `name + quantity + unit` fallback for ad-hoc entries that lack an id) instead of array index — robust against concurrent inserts.
- **Icons on included chips**: `FoodItemChip` now accepts an optional `icon` rendered through the same `ChipIcon` used by the suggestion chips. `.food-item-chip` styling is bumped (inline-flex, gap, larger padding, pill radius) to match `.frequent-chip`, so the "Included:" and "Add:" rows line up at the same height.

## Test plan

- [ ] Append two different "Add:" chips in quick succession on the same slot — both items should land in the meal (no second-tap clobbering the first).
- [ ] While an append is mid-flight, open the popover on an unrelated included chip and Remove it — the right item disappears (not a positional neighbor).
- [ ] Included chips display their icon and sit at the same height as the suggestions in "Add:".
- [ ] Items without a `food_item_id` (legacy / cronometer imports) still removable via the fallback matcher.
- [ ] Dark mode pill rendering of `.food-item-chip` still looks OK.

### DIFF
--- a/apps/web/src/pages/Meals/index.tsx
+++ b/apps/web/src/pages/Meals/index.tsx
@@ -63,6 +63,7 @@ const findMealsForSlot = (meals: Meal[], slotName: string): Meal[] =>
  */
 function FoodItemChip({
   name,
+  icon,
   foodItemId,
   initialFlagNames,
   flags,
@@ -70,6 +71,8 @@ function FoodItemChip({
   isDeletePending,
 }: {
   name: string
+  /** Optional icon — emoji or image path. Same renderer as the suggestion chips so heights match. */
+  icon?: string
   foodItemId?: string
   /** Flag names snapshotted on the meal-junction row — used as the initial popover state. */
   initialFlagNames: string[]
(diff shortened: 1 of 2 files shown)


fiddur (commented):
Clean follow-up that directly addresses both notes from #765. 👍

- `readLatestMeal()` in `apps/web/src/pages/Meals/index.tsx:911` is the right fix — pulling from `queryClient.getQueryData<MealsResult>(mealsQueryKey)` means the second of two rapid appends reads the optimistically-updated cache rather than the stale closure-captured `meal`, so the new item stacks instead of clobbering. The `?? meal` fallback is sensible cache-miss insurance.
- `foodItemMatches()` (`index.tsx:988`) is appropriately minimal: prefer `food_item_id`, fall back to `name + quantity + unit` for ad-hoc/legacy entries. First-match-wins matches the visual chip order, and the `if (matchIdx < 0) return` short-circuit covers the "already removed by a prior mutation" case cleanly.
- Icon pass-through in `FoodItemChip` and the pill-styling alignment in `.food-item-chip` are minor and contained.

No `as` casts, no schema touched, snake_case preserved. Type-only import for `FoodItem` keeps the surface tight.

Nothing else worth flagging — ship it.

PR #767, opened 2026-05-13
feat(meals): multi-window overview with 1/7/30/90d columns + date picker

fiddur:
## Summary
- Meals overview now shows **four windows** (1d / 7d / 30d / 90d) side-by-side on wide screens.
- On narrow screens the existing window selector returns, with non-active columns hidden via CSS (data-active on the root).
- New **DateNav** above the windows controls the end-of-range date (defaults to today), letting you scroll back to yesterday etc.
- **ReferenceRangeBar** extended with a `markers` prop. The Range column shows one bar per nutrient with all four window averages overlaid (tiny "1d/7d/30d/90d" labels above each marker, colored by flag). Single-marker callers (Reports) are unchanged.
- 14d window is gone (replaced by 1d).

## Notes for review
- Frontend-only — backend `meals/period-summary` already accepts arbitrary start/end. Four parallel requests via `useQueries`.
- `ReferenceRangeBar` is backward compatible: `value` + `flag` still works.
- I did **not** manually test in the browser (no live preview available in this session). `pnpm fix`, `pnpm check`, and `pnpm --filter aurboda-web build` are all clean.

## Test plan
- [ ] Wide screen: four columns visible, range bar shows four markers, DateNav scrolls all four windows together.
- [ ] Narrow screen: window selector visible, only one column visible, range bar still shows all four markers (this is intentional — the bar always reflects every window).
- [ ] DateNav: today label when on today; arrows walk by day/month; picker click works.
- [ ] Dark mode reads cleanly.
- [ ] Reports page's ReferenceRangeBar is unaffected.

### DIFF
--- a/apps/web/src/components/ReferenceRangeBar.css
+++ b/apps/web/src/components/ReferenceRangeBar.css
@@ -44,6 +44,24 @@
   box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
 }
 
+.ref-range-marker-label {
+  position: absolute;
+  top: -1rem;
+  left: 50%;
+  transform: translateX(-50%);
+  font-size: 0.6rem;
+  line-height: 1;
+  color: #6b7280;
+  font-variant-numeric: tabular-nums;
+  white-space: nowrap;
+  pointer-events: none;
+}
+
+.ref-range-multi {
+  /* Reserve space above the track for the tiny window labels. */
+  padding-top: 0.85rem;
+}
+
 .ref-range-labels {
   display: flex;
   justify-content: space-between;
@@ -75,5 +93,9 @@
 
   .ref-range-labels {
     color: #6b7280;
+  }
+
+  .ref-range-marker-label {
+    color: #9ca3af;
   }
 }
--- a/apps/web/src/components/ReferenceRangeBar.css
+++ b/apps/web/src/components/ReferenceRangeBar.css
@@ -57,9 +57,17 @@
   pointer-events: none;
 }
 
+/* Stagger label vertical position by index so two markers with similar
+ * values don't overlap their labels. Adjacent windows (1d vs 7d, 30d vs
+ * 90d) are most likely to land close on stable nutrients like sodium. */
+.ref-range-multi .ref-range-marker[data-marker-index='1'] .ref-range-marker-label,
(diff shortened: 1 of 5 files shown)


fiddur (commented):
Solid refactor. The `data-active` + CSS-driven column collapse is a nice way to keep a single render tree and let CSS handle the responsive narrow/wide split — avoids prop-drilling visibility and keeps the four `useQueries` results coherent across renders.

Things I checked:

- `computeDisplayRange` (`apps/web/src/components/ReferenceRangeBar.tsx:62`) cleanly generalizes to N values; padding still scales to span, and the `Math.min(displayMin, minVal - padding*0.1)` / `Math.max(displayMax, maxVal + padding*0.1)` widening keeps overlay markers visible even when a window value falls outside the original two-sided display window. Good.
- Backward compat: `value`/`flag` shorthand still works because `markers && markers.length > 0 ? markers : value !== undefined ? [{flag, value}] : []`. The Reports caller (`Reports/index.tsx`) won't break.
- `clampPct` survives multi-marker mode so markers stay inside the track visually even if a value is wildly outside the reference range.
- `WindowSelector` is rendered inside `.overview-window-control`, which is `display: none` on wide screens and only revealed at `max-width: 759px`. So wide users never see the selector but `activeKey` still drives the data-attribute that hides per-window columns on narrow — clean separation.
- 4× `useQueries` with `staleTime: 60_000` and stable query keys (start/end/tz) won't thrash on `endDate` change; same-date revisits hit cache.
- No new `as` casts. `const lo = low as number` in `computeDisplayRange` is pre-existing.

One observation, non-blocking:

- The PR notes acknowledge no browser test. The single most likely visual gotcha is the marker labels (`1d/7d/30d/90d`) stacking when window values are close — `position: absolute; transform: translateX(-50%)` will overlap them. `ref-range-marker-label` has `white-space: nowrap` but no collision avoidance. Probably fine in practice (1d and 30d averages usually diverge enough), but worth eyeballing once on a low-variance nutrient like sodium where multiple windows might land within a few % of each other.

Otherwise this looks good. The Energy grid restructuring (label-cell + per-window cells) and the `data-active` collapse pattern apply consistently across both the energy section and the nutrient tables — well factored.

PR #768, opened 2026-05-14
fix(meals): require explicit confirmation before creating food items from meal search

fiddur:
## Summary

- Typing in the meal food-item search was creating canonical food items on every autosave (one per partial name — `A`, `As`, `Ask`, `Asko`…), because the autosave posted `food_items` rows with the typed name but no `food_item_id`, and the backend's `resolveCanonical` falls through to `findOrCreate`.
- Autosave now drops draft rows without `food_item_id`; quantity/unit edits on resolved rows still flow through.
- `FoodItemAutocomplete` gains an optional `onCreate` hook — when provided, a `+ Create "<typed name>"` row is always appended as the last dropdown option (Linear/Notion style). Click or Enter creates the canonical item via `POST /food-items` and selects it. Wired up in meal logging; merge dialog and ingredient picker stay search-only.
- `editItemsToBody` / `mealItemsToEdit` extracted to `mealItems.ts` with unit tests covering the draft-filter behavior.

## Test plan

- [x] `pnpm --filter aurboda-web check` — typecheck + lint clean
- [x] `vitest run mealItems.test.ts` — 8/8 pass
- [ ] Manually: open a meal, type a non-existent name slowly, confirm no Food Item is created; click `+ Create "X"` and verify the row resolves with the new id and a `Food Items` page entry appears.
- [ ] Manually: existing meal items still autosave quantity/unit edits.
- [ ] Manually: merge dialog and recipe ingredient picker show no `+ Create` row.

### DIFF
--- a/apps/web/src/components/FoodItemAutocomplete.css
+++ b/apps/web/src/components/FoodItemAutocomplete.css
@@ -56,6 +56,15 @@
   font-size: 0.7rem;
 }
 
+.autocomplete-create {
+  border-top: 1px solid #e5e7eb;
+}
+
+.autocomplete-create .ac-name {
+  color: #2563eb;
+  font-weight: 500;
+}
+
 @media (prefers-color-scheme: dark) {
   .food-autocomplete input {
     background: #374151;
@@ -79,5 +88,13 @@
 
   .ac-cal {
     color: #9ca3af;
+  }
+
+  .autocomplete-create {
+    border-top-color: #4b5563;
+  }
+
+  .autocomplete-create .ac-name {
+    color: #60a5fa;
   }
 }
--- a/apps/web/src/components/FoodItemAutocomplete.tsx
+++ b/apps/web/src/components/FoodItemAutocomplete.tsx
@@ -7,6 +7,14 @@ interface FoodItemAutocompleteProps {
   value: string
   onChange: (name: string) => void
   onSelect: (item: FoodItemEntity) => void
+  /**
+   * Optional explicit-create hook. When provided, a "+ Create '<value>'" row
+   * is appended to the dropdown so the user can opt into creating a new
+   * canonical food item from the typed name. Without this, the autocomplete
+   * is search-only — preventing the backend's findOrCreate path from being
+   * triggered implicitly by autosave keystrokes.
+   */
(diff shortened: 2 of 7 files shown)


fiddur (commented):
Solid fix targeting the actual root cause — the autosave/findOrCreate interaction was a real data-pollution bug, and dropping draft rows from `editItemsToBody` is exactly the right boundary.

Things I checked:

- **The filter is safe for persisted rows** (`apps/web/src/pages/Meals/mealItems.ts:71`). Because the backend's old findOrCreate path always assigned a `food_item_id` to ad-hoc items on save, any row loaded from the server has an id — so the filter only drops *in-progress draft* rows. Worth confirming, but the PR body and the test name `drops draft rows without food_item_id` align with this read.
- **`showCreate = canCreate && !exactMatch`** (`FoodItemAutocomplete.tsx:51`) suppresses the Create row when the typed name matches an existing canonical item case-insensitively. Good — avoids accidental near-duplicates and gives the user the existing pick instead.
- **Keyboard handling** correctly clamps to `optionCount = suggestions.length + (showCreate ? 1 : 0)` and routes Enter to `handleCreate()` only when `activeIndex === createIndex`. The `creating` flag prevents re-entrant create on double-Enter.
- **No `as` casts.** All fields snake_case.
- **`addFoodItemApi` consolidation** removes the duplicate `createFoodItemApi` in `FoodItems/index.tsx` — nice cleanup. The inline response-shape workaround in `state/api/meals.ts:142` is explicitly justified by the comment.
- **Tests** (`mealItems.test.ts`) directly cover the regression and the draft-filter contract. The `mealItemsToEdit` and `scaleNutrient` extractions are well-scoped — extraction to a separate module is what makes them testable without component-rendering machinery, matching the project's testability preference.

One non-blocking note:

- Consider whether the explicit-create change deserves a one-liner in `docs/` describing meal item creation now requiring an explicit confirm — it's a small but user-visible behavior change. Could also be deferred to whoever lands the next docs sweep.

Otherwise this looks ready. The bug had real consequences (per-keystroke canonical items polluting the library), and the fix addresses it surgically without breaking the legitimate "user types a new thing they want to create" flow.

PR #769, opened 2026-05-14
feat(meals): align overview columns, marker legend, completed-day filter

fiddur:
## Summary
- **Column alignment**: convert Energy Balance to a `<table class=\"overview-table\">` matching the per-category tables, with a shared `<colgroup>` + `table-layout: fixed`, so 1d/7d/30d/90d numbers line up across the whole page.
- **Marker legend on wide view**: a small tick under each window header (thin→thick for 1d→90d) plus matching marker thickness on the range bars makes it obvious which marker belongs to which window.
- **Marker filter on narrow view**: only the active window's marker shows on the range bar.
- **Completed-day filter** (backend + frontend): new `count_only_completed` query param on `GET /meals/period-summary` (and `query_meals_period_summary` MCP tool). When set, only days marked as log-completed contribute to averages. The response always reports `days_completed`. Frontend checkbox \"Only include completed days\" defaults on; it's never sent for the 1D window, and the checkbox is hidden on narrow when 1D is the active window.
- **Subtitle**: each multi-day window header shows \"avg from N completed\".

## Test plan
- [x] `pnpm fix` clean
- [x] `pnpm check` clean (0 errors)
- [x] Backend unit + integration tests pass (2202 tests)
- [x] `pnpm --filter aurboda-web build` succeeds
- [ ] Verify in browser: columns align, header ticks visible on wide, single marker on narrow, checkbox toggles avg
- [ ] Verify on mobile width: checkbox hides when 1D selected

### DIFF
--- a/apps/backend/src/db/index.ts
+++ b/apps/backend/src/db/index.ts
@@ -285,6 +285,7 @@ export {
   getFrequentMeals,
   getMealById,
   getMealLogCompleted,
+  getMealLogCompletedInRange,
   getMeals,
   insertMeal,
   upsertMeal,
--- a/apps/backend/src/db/meals.integration.test.ts
+++ b/apps/backend/src/db/meals.integration.test.ts
@@ -7,8 +7,10 @@ import {
   getFrequentFoodItems,
   getFrequentMeals,
   getMealById,
+  getMealLogCompletedInRange,
   getMeals,
   insertMeal,
+  setMealLogCompleted,
   updateMeal,
 } from './meals.ts'
 
@@ -483,6 +485,25 @@ describe('Meals Integration Tests', () => {
       // Without the filter both items are returned.
       const all = await getFrequentFoodItems(user, { limit: 10, since_days: 30 })
       expect(all.map((r) => r.food_item_id).sort()).toEqual([oats, soup].sort())
+    })
+  })
+
+  describe('getMealLogCompletedInRange', () => {
+    test('returns only completed dates within [start, end]', async () => {
+      const user = getTestUser()
+      await setMealLogCompleted(user, '2025-04-01')
+      await setMealLogCompleted(user, '2025-04-05')
+      await setMealLogCompleted(user, '2025-04-10')
+      await setMealLogCompleted(user, '2025-05-01')
+
+      const result = await getMealLogCompletedInRange(user, '2025-04-02', '2025-04-30')
+      expect(result.sort()).toEqual(['2025-04-05', '2025-04-10'])
+    })
+
+    test('returns empty when no dates match', async () => {
+      const user = getTestUser()
+      const result = await getMealLogCompletedInRange(user, '2025-01-01', '2025-01-31')
+      expect(result).toEqual([])
     })
   })
 
--- a/apps/backend/src/db/meals.integration.test.ts
+++ b/apps/backend/src/db/meals.integration.test.ts
@@ -7,6 +7,7 @@ import {
   getFrequentFoodItems,
   getFrequentMeals,
   getMealById,
+  getMealLogCompleted,
   getMealLogCompletedInRange,
   getMeals,
   insertMeal,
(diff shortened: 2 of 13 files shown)


fiddur (commented):
## Review

Polish PR with three coordinated changes (column alignment via shared `<colgroup>` + `table-layout: fixed`, marker/legend weight by window, and a `count_only_completed` filter that also threads `days_completed` through the API and UI). The core wiring looks solid: schema in `api-spec` is updated end-to-end, REST router + MCP tool both pass the flag through, and the service has good unit coverage covering the three meaningful cases (no filter, filter, empty completed set). The new `getMealLogCompletedInRange` correctly avoids the TZ trap by formatting with `TO_CHAR` in SQL (rather than `date.toISOString().slice(0,10)`), and has an integration test for both the matching and empty cases.

A couple of things worth flagging:

### Potential mismatch between header text and actual avg denominator

In `meal-period-summary.ts:189` the response sets `days_completed: completedDates.length` — i.e. all completed dates in the range, regardless of whether those days had any meals.

But when `count_only_completed=true`, the avg denominator is `dayTotals.size` (line 180), which only counts days that are *both* completed *and* had meals.

In the UI (`MealsOverview.tsx:155-157`), the header shows:
```
{countOnlyCompleted ? `avg from ${daysCompleted} completed` : `${daysCompleted} completed`}
```

So if a user marks a day complete with no meals logged (e.g. a fasting day, or just an empty-but-acknowledged day), the header will say "avg from N completed" where N is larger than the denominator actually used. The schema description for `days_with_meals` correctly notes the intersection, but `days_completed` doesn't, and the header copy reads as if the two are the same.

Two plausible fixes:
- Report both numbers (`days_completed` total, plus a count of "completed-with-meals"), and use the latter in "avg from X" copy.
- Or change the semantic of `days_completed` returned when `count_only_completed=true` to be the intersection — but that loses the "X days marked complete" hint for the off-toggle case, so the first option seems cleaner.

Probably small in practice, but worth deciding deliberately.

### Pre-existing toISOString bug in sibling function

The comment on `getMealLogCompletedInRange` correctly calls out the TZ issue; note that the older `getMealLogCompleted` (line 338, the IN-list variant) still uses `r.date.toISOString().slice(0, 10)` and has the same latent bug. Out of scope for this PR, but worth a follow-up.

### Minor

- `meals-router.ts:95` — the route no longer needs the explicit destructure of just three fields since you're now spreading four; consider `...req.query` if the schema is the source of truth (or keep explicit if intentional — fine either way).
- CSS dark-mode block at the bottom of `style.css` is new and only covers a subset of the overview surface (energy block bg, h3, energy-num, window-tick). Worth a separate pass for the rest of the overview later; not blocking.

Overall LGTM modulo the days_completed/denominator semantics question.

fiddur (commented):
## Review

Nicely incremental. Tests cover the new `count_only_completed` paths on the backend and the new `getMealLogCompleted*` DB helpers. A couple of things worth a second look:

### 1. `calories_burned` is not filtered when `count_only_completed=true`

`apps/backend/src/services/queries/meal-period-summary.ts` filters `dayTotals` (per-nutrient averages) by the completed-day set, but `calories_burned` is still computed from `getTimeSeriesBucketed` with no such filter:

```ts
if (input.count_only_completed) {
  const completedSet = new Set(completedDates)
  for (const dayKey of dayTotals.keys()) {
    if (!completedSet.has(dayKey)) dayTotals.delete(dayKey)
  }
}
// …
const calories_burned = await computeCaloriesBurned(user, start, end, tz)
```

The web "Balance" row in `EnergyCell` then subtracts `eaten - burnedAvg`, where `eaten` is averaged over completed days but `burnedAvg` is averaged over every day with a `calories_total` metric. With the toggle on, the two terms come from different day sets, so the balance number is no longer a coherent "avg balance per completed day". Either filter `calories_burned` by the same completed-day set (preferred), or document/labelling that the burned/balance figures ignore the toggle.

### 2. Duplicate `@media (prefers-color-scheme: dark)` block

`apps/web/src/pages/Meals/style.css` already has a dark-mode block around L1220+. This PR adds another one at L1194-1212 that re-styles `.overview-energy`, `.overview-energy h3`, and `.energy-num`. The two blocks disagree on the h3 color (`#e5e7eb` in the new block, `#d1d5db` in the existing one — the existing one wins by source order). Worth merging the new rules (`.overview-group h3`, `.window-tick`) into the existing dark-mode block to keep the cascade obvious.

### Minor

- `WindowHeaderCell` shows `"avg from ${denominator} completed"` where `denominator` is `days_with_meals` (post-filter). If a day is marked completed but has no meals it's not included in the denominator, so the label slightly overstates "completed" coverage. Not a bug per se, just a wording nuance — `days_completed` could be a more honest count there.
- `getMealLogCompletedInRange` integration test asserts `result.sort()` matches a sorted array — fine, but since the SQL has no `ORDER BY`, ordering is undefined; keeping the explicit sort in the assertion is correct.

Everything else (the TZ-safe `TO_CHAR` fix in `getMealLogCompleted`, the shared `<colgroup>` for column alignment, the api-spec/MCP/router parity, doc update) looks good.

PR #770, opened 2026-05-14
fix(activities): dedupe insertActivities batch by SQL conflict key

fiddur:
## Summary
- A batch containing two rows that resolve to the same upsert conflict key triggers Postgres 21000 (`ON CONFLICT DO UPDATE command cannot affect row a second time`), aborting the whole batch.
- Health Connect sync surfaced this: a re-sync chunk could include two copies of the same `ExerciseSessionRecord`, or two records lacking `metadata.id` sharing a `start_time`, blocking initial sync indefinitely (Sara's case).
- Dedupe inside `insertActivities` by the SQL conflict key (last write wins, matching the upsert semantics callers already rely on) so every caller — HC, screentime, lastfm, rescuetime, activitywatch, location-visit — is protected.

## Test plan
- [x] New integration test: duplicate `(source, external_id)` in one batch — second row wins.
- [x] New integration test: duplicate `(source, activity_type, start_time)` without `external_id` — second row wins.
- [x] Existing 36 activities integration tests still pass.
- [ ] Sara retries Health Connect initial sync once deployed and it completes.

### DIFF
--- a/apps/backend/src/db/activities.integration.test.ts
+++ b/apps/backend/src/db/activities.integration.test.ts
@@ -11,6 +11,7 @@ import {
   getOverlappingActivities,
   getOverrideForActivity,
   getSleepSessions,
+  insertActivities,
   insertActivity,
   insertOverride,
   markActivityDetailSynced,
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Targeted fix that pre-dedupes batches by the exact `(source, external_id)` and `(source, activity_type, start_time)` keys that the two `ON CONFLICT` targets use. The dedupe order (`Map.set` last-write-wins) matches the existing upsert semantics — and since Postgres would otherwise raise `21000` and abort the whole batch, only one row could ever win at the SQL layer anyway, so this is strictly an availability improvement with no behavior regression.

Verified all `insertActivities` callers: the HC sync (the obvious place where index alignment could matter) looks rows back up via `external_id` and `(activity_type|start_time)` maps — exactly the dedupe keys — so collapsing duplicates in the returned `InsertedActivityKey[]` is safe. Other callers ignore the return value. Integration tests cover both branches (`external_id` and `(activity_type, start_time)`) and assert last-write-wins, which is the right contract to lock in.

PR #771, opened 2026-05-14
feat(sentry): admin-configurable backend error reporting

fiddur:
## Summary
- Adds an admin-managed Sentry DSN in **Admin Settings → Error Reporting**. When set, the backend initialises `@sentry/node` and registers `Sentry.setupExpressErrorHandler` before the central error handler so all uncaught errors land in Sentry.
- DSN is stored in `server_settings` (key `sentry_dsn`); no env-var bootstrap, no client-side reporting. Changes apply after the next backend restart (noted in the UI).
- `docs/sentry.md` covers setup, scope, and verification.

## Follow-up: source maps (manual)
The wizard step is interactive (browser login to Sentry), so it must be run locally by an admin:

```
npx @sentry/wizard@latest -i sourcemaps --saas --org aurboda --project node-express
```

That step is independent of this PR — error reporting works without source maps; stack traces just won't be original-source mapped.

## Test plan
- [ ] `pnpm check` — passes
- [ ] `pnpm --filter aurboda-backend test` — all 2203 tests pass
- [ ] With no DSN set: backend starts normally, no Sentry init log line, errors handled as before.
- [ ] Set DSN via Admin Settings, restart backend → log line `🛡️ Sentry error reporting enabled` appears.
- [ ] Trigger an error from an authenticated endpoint and confirm it shows up in the Sentry project.
- [ ] Clear DSN and restart → Sentry no longer initialised.

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -66,6 +66,7 @@ import { createGeocodeQueue } from './services/geocode-queue.ts'
 import { createInvitationAuth } from './services/invitation.ts'
 import { getPlaceVisits } from './services/locations.ts'
 import { createPgBoss } from './services/pg-boss.ts'
+import { initSentry, Sentry } from './services/sentry.ts'
 import { createStravaQueue, type StravaQueue } from './services/strava-queue.ts'
 import { createSyncProvider } from './services/sync-provider.ts'
 import { createWebAuthnService } from './services/webauthn.ts'
@@ -94,6 +95,10 @@ const main = async () => {
   // Initialize central database (server settings, admins)
   await initializeCentralDb()
   const centralDb = getCentralDb()
+
+  // Initialize Sentry as early as possible after centralDb is available.
+  // DSN is read from server_settings (configured via Admin Settings).
+  await initSentry(centralDb)
 
   const webHost = process.env.WEB_HOST ?? 'http://localhost:5173'
   const apiBaseUrl = process.env.API_BASE_URL ?? 'http://localhost:3000'
@@ -412,6 +417,10 @@ const main = async () => {
     webHost,
     wellKnown,
   })
+
+  // Sentry must be registered AFTER all controllers and BEFORE any other
+  // error middleware. No-op if Sentry was not initialized.
+  Sentry.setupExpressErrorHandler(httpd)
 
   // Centralized error handler
   httpd.use((err: Error, req: Request, res: Response, _next: NextFunction) => {
--- a/apps/backend/src/routes/admin-router.ts
+++ b/apps/backend/src/routes/admin-router.ts
@@ -46,6 +46,7 @@ export const createAdminRouter = (
         stravaClientId,
         stravaClientSecret,
         ouraWebhookAvailable,
+        sentryDsn,
       ] = await Promise.all([
         centralDb.getSignupMode(),
         centralDb.getAdminCount(),
@@ -57,6 +58,7 @@ export const createAdminRouter = (
         centralDb.getServerSetting('strava_client_id'),
         centralDb.getServerSetting('strava_client_secret'),
         ouraWebhookManager ? ouraWebhookManager.canEnable() : Promise.resolve(false),
+        centralDb.getServerSetting('sentry_dsn'),
       ])
       res.json({
         admin_count: adminCount,
@@ -66,6 +68,7 @@ export const createAdminRouter = (
         oura_client_secret_set: !!ouraClientSecret,
         oura_webhook_available: ouraWebhookAvailable,
         oura_webhook_enabled: ouraWebhookEnabled,
+        sentry_dsn: sentryDsn || null,
         signup_mode: signupMode,
         strava_client_id_set: !!stravaClientId,
         strava_client_secret_set: !!stravaClientSecret,
@@ -87,6 +90,7 @@ export const createAdminRouter = (
         oura_client_id,
         oura_client_secret,
         oura_webhook_enabled,
+        sentry_dsn,
         signup_mode,
         strava_client_id,
         strava_client_secret,
@@ -122,6 +126,9 @@ export const createAdminRouter = (
       if (strava_client_secret !== undefined) {
         await centralDb.setServerSetting('strava_client_secret', strava_client_secret ?? '')
       }
+      if (sentry_dsn !== undefined) {
+        await centralDb.setServerSetting('sentry_dsn', sentry_dsn ?? '')
+      }
       const [
         currentMode,
         adminCount,
@@ -133,6 +140,7 @@ export const createAdminRouter = (
         currentStravaClientId,
         currentStravaClientSecret,
         ouraWebhookAvailable,
+        currentSentryDsn,
       ] = await Promise.all([
         centralDb.getSignupMode(),
         centralDb.getAdminCount(),
@@ -144,6 +152,7 @@ export const createAdminRouter = (
         centralDb.getServerSetting('strava_client_id'),
         centralDb.getServerSetting('strava_client_secret'),
         ouraWebhookManager ? ouraWebhookManager.canEnable() : Promise.resolve(false),
+        centralDb.getServerSetting('sentry_dsn'),
       ])
       res.json({
         admin_count: adminCount,
@@ -153,6 +162,7 @@ export const createAdminRouter = (
         oura_client_secret_set: !!currentOuraClientSecret,
         oura_webhook_available: ouraWebhookAvailable,
         oura_webhook_enabled: ouraWebhookEnabledValue,
+        sentry_dsn: currentSentryDsn || null,
         signup_mode: currentMode,
         strava_client_id_set: !!currentStravaClientId,
         strava_client_secret_set: !!currentStravaClientSecret,
--- a/apps/backend/src/routes/admin-router.ts
+++ b/apps/backend/src/routes/admin-router.ts
@@ -68,6 +68,8 @@ export const createAdminRouter = (
         oura_client_secret_set: !!ouraClientSecret,
         oura_webhook_available: ouraWebhookAvailable,
         oura_webhook_enabled: ouraWebhookEnabled,
+        // DSN is returned in full (not `_set` boolean like other secrets):
+        // Sentry DSNs are public values designed to be embedded in clients.
         sentry_dsn: sentryDsn || null,
         signup_mode: signupMode,
         strava_client_id_set: !!stravaClientId,
--- a/apps/backend/src/services/sentry.ts
+++ b/apps/backend/src/services/sentry.ts
@@ -2,22 +2,25 @@
  * Sentry initialization. DSN is read from server_settings at startup —
  * admins configure it via Admin Settings. Changes take effect on the next
  * backend restart (Sentry SDK is initialized once per process).
+ *
+ * Scope is intentionally errors-only: because `Sentry.init` runs inside
+ * `main()` after all module imports, OpenTelemetry-based auto-instrumentation
+ * (HTTP/express/db tracing, automatic breadcrumbs) will not patch. Only the
+ * explicit `setupExpressErrorHandler` path in `api.ts` captures errors.
+ * Enabling tracing later would require `node --import ./instrument.ts` with
+ * an env-bootstrapped DSN. See docs/sentry.md.
  */
 import * as Sentry from '@sentry/node'
 
 import type { CentralDb } from './central-db.ts'
 
-let initialized = false
-
 export const initSentry = async (centralDb: CentralDb): Promise<boolean> => {
-  if (initialized) return true
   const dsn = await centralDb.getServerSetting('sentry_dsn')
   if (!dsn) return false
   Sentry.init({
     dsn,
     sendDefaultPii: true,
   })
-  initialized = true
   console.info('🛡️ Sentry error reporting enabled')
   return true
 }
--- a/apps/backend/src/services/sentry.ts
+++ b/apps/backend/src/services/sentry.ts
@@ -1,0 +1,25 @@
+/**
+ * Sentry initialization. DSN is read from server_settings at startup —
+ * admins configure it via Admin Settings. Changes take effect on the next
+ * backend restart (Sentry SDK is initialized once per process).
+ */
+import * as Sentry from '@sentry/node'
+
+import type { CentralDb } from './central-db.ts'
+
+let initialized = false
+
+export const initSentry = async (centralDb: CentralDb): Promise<boolean> => {
+  if (initialized) return true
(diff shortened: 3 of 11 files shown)


apps/backend/src/services/sentry.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Sentry initialization. DSN is read from server_settings at startup —
+ * admins configure it via Admin Settings. Changes take effect on the next
+ * backend restart (Sentry SDK is initialized once per process).
+ */
+import * as Sentry from '@sentry/node'
+
+import type { CentralDb } from './central-db.ts'
+
+let initialized = false
```
fiddur: Module-level mutable state violates the project rule "Avoid module level variables" in CLAUDE.md. The `Sentry` SDK itself is a process-global singleton and `Sentry.init` is safe to call repeatedly, so this guard is largely cosmetic — consider dropping it (or returning the DSN-set flag from `getServerSetting` directly) since `initSentry` is only ever called once from `main()`.

apps/backend/src/api.ts
```diff
@@ -95,6 +96,10 @@ const main = async () => {
   await initializeCentralDb()
   const centralDb = getCentralDb()
 
+  // Initialize Sentry as early as possible after centralDb is available.
+  // DSN is read from server_settings (configured via Admin Settings).
+  await initSentry(centralDb)
```
fiddur, line 101: For Sentry v8+/v10, `Sentry.init` should run *before* express and other instrumented modules are imported for auto-instrumentation (HTTP breadcrumbs, performance) to take effect. Calling it inside `main()` after all imports means only the explicit `setupExpressErrorHandler` will work — which is fine for the stated goal, but probably worth calling out in `docs/sentry.md` so a future contributor doesn't expect tracing to work.

apps/backend/src/routes/admin-router.ts
```diff
@@ -66,6 +68,7 @@ export const createAdminRouter = (
         oura_client_secret_set: !!ouraClientSecret,
         oura_webhook_available: ouraWebhookAvailable,
         oura_webhook_enabled: ouraWebhookEnabled,
+        sentry_dsn: sentryDsn || null,
```
fiddur, line 73: Inconsistent with neighbouring secrets: `oura_client_secret` and `strava_client_secret` are exposed only as `_set` booleans, while `sentry_dsn` is returned in full. Sentry DSNs are considered public enough to embed in browser bundles, so this is defensible, but a one-line comment (here or in the schema) explaining *why* the pattern differs would save future readers a double-take.

apps/backend/src/services/sentry.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const initSentry = async (centralDb: CentralDb): Promise<boolean> => {
+  if (initialized) return true
+  const dsn = await centralDb.getServerSetting('sentry_dsn')
+  if (!dsn) return false
+  Sentry.init({
+    dsn,
+    sendDefaultPii: true,
```
fiddur, line 22: `sendDefaultPii: true` is hard-coded — IP and user context will be forwarded to Sentry whenever a DSN is configured. It's documented in `docs/sentry.md`, but if PII handling ever becomes a concern, exposing a second `sentry_send_pii` boolean admin setting would be more flexible than redeploying with a code change.

fiddur (commented):
✅ Approved

Small, focused PR that wires admin-configurable Sentry error reporting into the backend. The schema/router/service/UI layers are all consistent and the docs page is helpful. No blocking issues, but a few notes worth considering:

- **Init ordering for Sentry v8+/v10:** `Sentry.init` is called inside `main()` *after* express and most other modules have already been imported. That works fine for the explicit `setupExpressErrorHandler` (which is all the docs page promises), but Sentry's OpenTelemetry-based auto-instrumentation (HTTP breadcrumbs, performance, etc.) only takes effect when `Sentry.init` runs before instrumented packages are imported. If you ever want tracing/breadcrumbs, you'll need to move init into a separate `--import` entry, or read the DSN from env in a top-level instrument file. Worth a sentence in `docs/sentry.md`.
- **Module-level `initialized` flag** in `services/sentry.ts` is a small violation of the project rule "Avoid module level variables". It's tolerable because the underlying Sentry SDK is itself a process-global singleton, but the guard is also redundant — `Sentry.init` being called twice with the same DSN is harmless, and the function is only invoked once from `main()` anyway. Could drop the flag and just return whether DSN was set.
- **DSN exposure pattern is inconsistent** with other secrets in the same endpoint. `oura_client_secret` / `strava_client_secret` are exposed only as `_set` booleans, while `sentry_dsn` is returned in full. Sentry DSNs are technically considered public, so this is defensible, but worth a note in the schema description so future readers don't think it was an oversight.
- **No backend tests** are added for the new admin settings field or the `initSentry` helper. The repo standard is "Backend should be well covered with tests" — though `admin-router.ts` has no existing tests, so this PR is at least consistent with the pre-existing state. A small unit test that `initSentry` returns false when DSN is empty and calls `Sentry.init` once when set would be worthwhile.
- **`sendDefaultPii: true` is hard-coded.** This sends IP and user context to Sentry by default; it's documented in `docs/sentry.md`, but might be worth a follow-up to make it an admin toggle if PII handling becomes a concern.

None of the above blocks merge.

apps/backend/src/services/sentry.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import type { CentralDb } from './central-db.ts'
+
+let initialized = false
+
+export const initSentry = async (centralDb: CentralDb): Promise<boolean> => {
+  if (initialized) return true
+  const dsn = await centralDb.getServerSetting('sentry_dsn')
+  if (!dsn) return false
+  Sentry.init({
```
fiddur, line 20: `@sentry/node` v8/v10 expects `Sentry.init` to be called before any other module imports so its OpenTelemetry-based auto-instrumentation can patch `http`, `express`, DB drivers, etc. Here init runs inside `main()` long after those modules are imported, so tracing/auto-breadcrumbs will mostly be inert — only the explicit `setupExpressErrorHandler` path will catch errors. If that's the intended scope, consider noting it in `docs/sentry.md`; otherwise the standard pattern is `node --import ./instrument.ts` (or a top-of-entry side-effect import that reads DSN from env).

packages/api-spec/src/schemas/admin.ts
```diff
@@ -188,6 +192,10 @@ export const updateAdminSettingsBodySchema = z
       .boolean()
       .optional()
       .meta({ description: 'Enable or disable Oura webhook push sync' }),
+    sentry_dsn: z.string().nullable().optional().meta({
```
fiddur: Consider stricter validation here — e.g. `z.string().url()` (DSNs are URLs) or a DSN-shaped regex. A typoed DSN currently saves silently and only fails (with a Sentry-side console warning) on the next restart, with no admin-visible signal that reporting is broken.

apps/backend/src/services/sentry.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Sentry initialization. DSN is read from server_settings at startup —
+ * admins configure it via Admin Settings. Changes take effect on the next
+ * backend restart (Sentry SDK is initialized once per process).
+ */
+import * as Sentry from '@sentry/node'
+
+import type { CentralDb } from './central-db.ts'
+
+let initialized = false
+
```
fiddur: Worth a small unit test (`db.integration.test.ts`-style or pure unit): returns `false` when DSN is unset, returns `true` and calls `Sentry.init` when set, and is idempotent on repeat calls. The module is tiny but its contract ("DSN is read once at startup") is load-bearing for the docs.

fiddur (commented):
✅ Approved

Small, focused change that adds an admin-configurable Sentry DSN to the central `server_settings` table, initializes `@sentry/node` at startup when set, and exposes the field through the admin settings API + UI. Schema, router wiring, types, and docs are all consistent with the existing patterns for other admin-managed integrations.

A few non-blocking observations worth noting:

- **Sentry init order.** `@sentry/node` v8/v10 recommends calling `Sentry.init` *before* any other module imports so its OpenTelemetry-based auto-instrumentation (HTTP, Express, DB tracing, breadcrumbs) can patch the relevant modules. Here `initSentry` runs inside `main()` after `express`, route modules, and HTTP integrations have all been imported. The express error handler will still capture thrown errors (the stated goal of this PR), but performance tracing and most automatic breadcrumbs/spans will not work. If only uncaught-error reporting is wanted, the current setup is fine — consider documenting this limitation in `docs/sentry.md`, or moving to an `--import ./instrument.ts` pattern later if richer telemetry is desired.
- **No DSN validation.** `updateAdminSettingsBodySchema` accepts any string. A typoed DSN will silently fail on the next restart (Sentry logs a warning but no admin-visible signal). Consider `z.string().url()` or a regex for DSNs, or a self-test ping.
- **No tests.** `CLAUDE.md` says the backend should be well covered. `initSentry` is straightforward but a small unit test (returns false when DSN unset, initializes when set, idempotent on repeat calls) would lock in the contract.
- **`sendDefaultPii: true`** is enabled unconditionally. The doc mentions this, but admins may want a toggle if their deployment has stricter PII policies. Fine to leave for a follow-up.

Nothing in this PR is incorrect; the items above are improvement suggestions for a future iteration.

apps/backend/src/services/sentry.ts
fiddur: Done — dropped the  flag in 31962f6.  is itself idempotent at the SDK level and the function is only called once from , so the guard was just module-level state for nothing.

apps/backend/src/api.ts
fiddur: Called out in 31962f6:  now has a "Scope: errors only" section explaining that init runs after module imports so OpenTelemetry auto-instrumentation is inert, and how to enable tracing later ( with env-bootstrapped DSN). Same note duplicated in the module docstring.

apps/backend/src/routes/admin-router.ts
fiddur: Done — added an inline comment in 31962f6 right next to the  field in the response explaining that DSNs are public values designed to be embedded in clients (unlike OAuth secrets).

apps/backend/src/services/sentry.ts
fiddur: Noted — leaving  hardcoded for this PR since it matches the SDK's recommended default and is the behavior I want. Happy to expose a  admin setting in a follow-up if PII surfaces become a concern.

apps/backend/src/services/sentry.ts
fiddur: Same fix as in the earlier thread — scope is now explicitly documented in  and the module docstring (31962f6). Errors-only is intentional for the DB-stored DSN model.

packages/api-spec/src/schemas/admin.ts
fiddur: Done —  in the update body schema now goes through  in 31962f6, so a typoed DSN is rejected at save time with a validation error rather than silently failing on the next restart.

apps/backend/src/services/sentry.ts
fiddur: Added in 31962f6 —  covers the three contract cases: unset DSN returns false, empty DSN returns false, configured DSN returns true and calls  with the right args.

apps/backend/src/services/sentry.ts
fiddur: Done — dropped the `initialized` flag in 31962f6. `Sentry.init` is itself idempotent at the SDK level and the function is only called once from `main()`, so the guard was just module-level state for nothing.

apps/backend/src/api.ts
fiddur: Called out in 31962f6: `docs/sentry.md` now has a "Scope: errors only" section explaining that init runs after module imports so OpenTelemetry auto-instrumentation is inert, and how to enable tracing later (`--import ./instrument.ts` with env-bootstrapped DSN). Same note duplicated in the module docstring.

apps/backend/src/routes/admin-router.ts
fiddur: Done — added an inline comment in 31962f6 right next to the `sentry_dsn` field in the response explaining that DSNs are public values designed to be embedded in clients (unlike OAuth secrets).

apps/backend/src/services/sentry.ts
fiddur: Noted — leaving `sendDefaultPii: true` hard-coded for this PR since it matches the SDK's recommended default and is the behavior I want. Happy to expose a `sentry_send_pii` admin setting in a follow-up if PII surfaces become a concern.

apps/backend/src/services/sentry.ts
fiddur: Same fix as in the earlier thread — scope is now explicitly documented in `docs/sentry.md` and the module docstring (31962f6). Errors-only is intentional for the DB-stored DSN model.

packages/api-spec/src/schemas/admin.ts
fiddur: Done — `sentry_dsn` in the update body schema now goes through `z.string().url()` in 31962f6, so a typoed DSN is rejected at save time with a validation error rather than silently failing on the next restart.

apps/backend/src/services/sentry.ts
fiddur: Added in 31962f6 — `services/sentry.test.ts` covers the three contract cases: unset DSN returns false, empty DSN returns false, configured DSN returns true and calls `Sentry.init` with the right args.

fiddur (commented):
✅ Approved

Clean, well-scoped addition: admin-configurable Sentry DSN flowing through `server_settings`, a small `initSentry` service called at startup, and the matching Admin Settings UI. Wiring is correct — `setupExpressErrorHandler` is registered after all routers and before the centralized error middleware as Sentry requires, and the no-op-when-unset behavior is captured by unit tests.

A few minor observations (not blocking):

- The PATCH `/admin/settings` handler keeps growing (10 settings now, eslint-disable for complexity). Worth a follow-up refactor into a generic `setIfDefined` loop, but consistent with the existing pattern.
- `sendDefaultPii: true` will ship request IPs / user context to Sentry. Given the health-data nature of the product, it might be worth mentioning this explicitly in `docs/sentry.md` under "What is captured" (currently only mentions it as a one-liner) so admins make an informed choice. Could also be made admin-toggleable later.
- In the React Save button, `dsn && saveDsn(dsn)` is redundant since the button is already `disabled={!dsn}` — minor.
- The doc mentions `res.sentry` for cross-referencing but the centralized error handler doesn't log/return it, so that line is currently aspirational.

Tests cover the init paths, the schema validates DSNs as URLs, and the clear-by-null pattern is consistent with other settings. Good to ship.

PR #774, opened 2026-05-15
feat(meals): derive missing nutrients from precursors at aggregation

fiddur:
## Summary

Closes #773.

Livsmedelsverket-sourced food items carry precursor fields (`retinol` + `beta_carotene`, `b3_niacin` + `tryptophan`, `salt` or `sodium` alone, `vitamin_d_25oh`, fatty-acid breakdown) but typically leave the combined columns null. As a result, daily/weekly/90d totals understated Vitamin A by 20-40× (e.g. 54 µg avg vs ~640 µg real intake) and missed niacin equivalents, sodium↔salt and 25-OH-D3-only vitamin D entirely.

## Approach

New `apps/backend/src/services/derived-nutrients.ts` exports `withDerivedNutrients(row)`, which fills in a nutrient only when the explicit column is **missing** (null/undefined — a literal `0` stays explicit, so user/data-source "this food has none of this" is preserved).

Derivations:

| Nutrient | Formula (only when explicit field is null) |
|---|---|
| `vitamin_a` (µg RAE) | `retinol + beta_carotene / 12` |
| `niacin_equivalents` (mg NE) | `b3_niacin + tryptophan_g × 1000 / 60` |
| `sodium` ↔ `salt` | cross-fill: `sodium = salt × 400` or `salt = sodium / 400` |
| `vitamin_d` (µg) | adopt `vitamin_d_25oh` (LSV's `VITD_x` is already the combined value) |
| `omega_3` (g) | `ala + epa + dha + dpa` |
| `omega_6` (g) | `la + aa` |
| `net_carbs` (g) | `max(0, carbs − fiber)` |
| `sugars` (g) | `monosaccharides + disaccharides` |

Applied per junction row in both aggregators (`aggregateNutrients` in `meals.ts`, `aggregateNutrientsFromLinks` in `meal-period-summary.ts`) so the fix cascades into daily/weekly/90d/baseline & every dashboard widget without touching canonical food-item rows.

## Test plan

- [x] 21 unit tests covering each derivation, explicit-wins, zero-vs-null, and edge cases (`derived-nutrients.test.ts`)
- [x] Period-summary test exercising mixed explicit + derived rows on a real meal
- [x] `pnpm check` clean
- [x] Full backend test suite (2230 tests) passing
- [ ] After deploy: confirm fiddur's 90d Vitamin A lands in 600-1500 µg/day range

### DIFF
--- a/apps/backend/src/services/derived-nutrients.test.ts
+++ b/apps/backend/src/services/derived-nutrients.test.ts
@@ -1,0 +1,122 @@
+import { describe, expect, test } from 'vitest'
+
+import { withDerivedNutrients } from './derived-nutrients.ts'
+
+describe('withDerivedNutrients', () => {
+  test('passes through a row with no derivable inputs', () => {
+    expect(withDerivedNutrients({ calories: 100, protein: 10 })).toEqual({
+      calories: 100,
+      protein: 10,
+    })
+  })
+
+  test('derives vitamin_a (RAE) from retinol + beta_carotene/12 when vitamin_a is null', () => {
+    // 580 µg retinol + 6720 µg β-carotene → 580 + 560 = 1140 µg RAE
+    const out = withDerivedNutrients({ retinol: 580, beta_carotene: 6720 })
+    expect(out.vitamin_a).toBeCloseTo(1140)
+    expect(out.retinol).toBe(580)
+    expect(out.beta_carotene).toBe(6720)
+  })
+
+  test('vitamin_a explicit value wins (no double-counting against precursors)', () => {
+    const out = withDerivedNutrients({ vitamin_a: 50, retinol: 580, beta_carotene: 6720 })
+    expect(out.vitamin_a).toBe(50)
+  })
+
+  test('vitamin_a derives from a single precursor when the other is missing', () => {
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped change. Each derivation is correct (RAE factor 12 for β-carotene, NE = niacin + tryptophan(g)·1000/60, NaCl ↔ Na ≈ ×400, β-carotene–only contributors via `sumPresent`), and the rules are carefully guarded so a literal `0` is treated as explicit (preventing fabricated zero values from polluting `days_with_value`). The aggregation hook is applied per-link before summation, so the period-summary test correctly demonstrates that a row with explicit `vitamin_a` is not double-counted against its own precursors. Test coverage of the unit (positive paths, explicit-wins, single-precursor, chained, zero handling, non-number fields) is thorough and the period-summary integration-style test pins down the cross-row semantics. Macro recomputation (`recomputeMealMacros`) only consumes the canonical macro keys, none of which are derived, so persisted meal-row macros are unaffected. Two minor non-blocking observations are in inline comments.

_(2 inline comment(s) couldn't be anchored to diff lines and were dropped)_

PR #775, opened 2026-05-20
feat(calories): HR-zone-METs model replaces Keytel formula

fiddur:
## Summary

The omnicalc/Keytel HR-to-kcal formula systematically overshoots at sub-exercise HR — it was tuned against steady-state exercise data and assumes a higher baseline than physiology dictates for sedentary minutes. For a stable-weight user, replaying 14 days of intake vs. various candidate models showed:

| Model | 14-day avg burn | Δ vs intake (2903) |
|---|---:|---:|
| Old: BMR(lab) + Keytel-active(rest×1.2) | 3521 | +21% |
| Variant: BMR-floor + Keytel | 5574 | +92% |
| Variant: HRR-METs linear | 4791 | +65% |
| **New: HR-zone-METs (MET anchors 1/2/4/6/8.5/11 at z1..z5)** | **3189** | **+10%** ✅ |
| HC stored daily aggregate | 2978 | +2.5% |

Picks the right structural model: METs anchored at the user's calibrated zone boundaries, BMR/min as the floor, linear interpolation between zones. Naturally returns 1 MET (= BMR) at rest and capped extrapolation above observed HRmax.

## Changes

- **`calories.ts`**: new `computeCaloriesPerMinuteZoneMets` (+ helpers `computeMetsForHr`, `estimateBmrMifflinStJeor`, `defaultHrZoneThresholds`). Old Keytel exports kept (no consumer left in the orchestrator, but tests still reference them).
- **`calorie-computation.ts`**: orchestrator now reads lab BMR (\`basal_metabolic_rate\` metric, 2-year lookback) → Mifflin-St Jeor fallback. Writes BOTH \`calories_active\` and \`calories_total\` per-minute under source \`'aurboda'\`. Non-HR minutes get BMR/min — daily sum = BMR + active without HC-aggregate gap-fill.
- **`api-spec/common.ts`**: \`aurbodaOnlyMetrics\` extended to include \`calories_total\`. Bucketed cumulative query now reads only aurboda's per-minute series for both calories metrics, avoiding double-count vs HC/Garmin daily aggregates.
- **Removed**: HC-residual gap-fill logic in orchestrator (BMR floor obsoletes it). Gap-fill helpers remain in calories.ts for back-compat.
- **`recalculate-calories`** response: \`vo2_max_source\` → \`bmr_source\` (\`lab\` | \`mifflin_st_jeor\`).

## Migration

After merge, run a full recompute via \`POST /api/metrics/recalculate-calories\` (empty body) or the MCP \`recalculate_calories\` tool. This walks every day with HR data, deletes old aurboda \`calories_active\` / \`aurboda_gap_fill\` / new \`calories_total\` rows, and writes the new per-minute series.

## Test plan

- [x] \`pnpm check\` — 0 errors
- [x] All backend unit tests pass (1812)
- [x] \`time-series.integration.test.ts\` passes (36)
- [x] \`calories.test.ts\` passes (49) — new zone-METs tests cover anchors, interpolation, BMR floor, monotonicity, MAX_HOLD_MINUTES, full-day expectations
- [ ] After merge + recompute: verify meals overview burned-kcal for 2026-05-19 reflects new model (expected ~3100 vs Garmin's stale 1839 from HC)
- [ ] Verify training load impulses recompute (existing watermark logic unchanged)


apps/backend/src/services/calorie-computation.ts
```diff
@@ -409,74 +345,64 @@ export const computeAndStoreCaloriesAll = async (
...
 /**
  * Trigger calorie computation for a time range after HR data ingestion.
- * This is a best-effort operation that never throws.
- * Also runs gap-filling to distribute HC aggregate residual into uncovered minutes.
+ * Best-effort, never throws.
  */
 export const triggerCalorieComputation = async (user: string, start: Date, end: Date): Promise<void> => {
   try {
-    const result = await computeAndStoreCalories(user, start, end)
+    const result = await computeAndStoreCalories(user, start, end, { force: true })
```
fiddur: `force: true` here means every incremental HR-ingest deletes and rewrites aurboda calories for only the HR window. Because `calories_total` is now `aurbodaOnly` (HC aggregate is filtered out at query time), the rest of the day has no BMR floor for `calories_total` unless `computeAndStoreCaloriesAll` has been run for that day. New days post-deploy will under-report daily totals by ~BMR minus the active portion. Either expand `start`/`end` to encompass whole local day(s) here (and arrange to not stomp prior incremental writes for the same day), or hold off on adding `calories_total` to `aurbodaOnlyMetrics` until incremental coverage is whole-day.

fiddur (commented):
🛑 Changes required

The zone-METs model is a clean replacement for Keytel — well-tested, the math reads correctly (resting/zone interpolation, MET caps, BMR-floor invariant), and the BMR resolution priority (lab → Mifflin-St Jeor) is sensible. Tests are thorough and unit-level coverage of the new helpers is solid.

**Blocking concern — partial-day BMR coverage on the incremental path.** `calories_total` joins `aurbodaOnlyMetrics`, so HC aggregates are now excluded from queries; aurboda is the sole authoritative source. But `triggerCalorieComputation` calls `computeAndStoreCalories(user, start, end, { force: true })` with the **raw HR-ingestion window** (e.g. 5 minutes from an Android HR batch). `buildFullDayPoints` then only walks `floor(start/min) … ceil(end/min)` — five minutes of `calories_total`, no BMR floor for the other 1435 minutes of that day. The docstring's promise that "the daily sum naturally equals BMR + active" only holds when the caller is a full-day chunker, which is true for `computeAndStoreCaloriesAll` but not for the live ingestion paths in `sync-router.ts`, `oura/sync.ts`, or `calorie-queue.ts`. Once HC aggregate is filtered out, any day that doesn't get a full-day recompute (new days going forward; any new user after rollout) will report daily `calories_total` as just the small HR-window sums. The previous gap-fill flow against HC aggregate is now gone, so there's no remediation either. Either the trigger path should expand to whole local days (writing BMR floor only for minutes not already populated, to avoid stomping prior incremental writes) or the schema change to `aurbodaOnlyMetrics` should wait until incremental coverage is whole-day.

**Smaller things to consider:**
- `computeAndStoreCaloriesAll` walks chunks via a constant `dayMs = 24h`. After local DST transitions, chunk boundaries drift 1h off from local midnight for the rest of the run. Consider re-anchoring with `getLocalDayStart(chunkEnd, timezone)` each iteration.
- In `computeAndStoreCalories`, the `force` branch deletes existing rows (step 6) before the `hrPoints.length === 0` guard (step 7). If HR samples exist but produce no minute buckets (degenerate edge case), the function returns leaving the day empty.
- `aurbodaOnlySources` still includes `'aurboda_gap_fill'` but nothing writes to that source anymore. The comment calls it back-compat for unrecomputed days; fine but worth a follow-up cleanup once everyone has been recomputed.
- The `getVo2MaxFallback` / `computeCaloriesPerMinute` / `computeGapFillPoints` functions in `calories.ts` remain exported and tested but are no longer used anywhere in production code — candidates for deletion after the rollout is stable.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -126,278 +130,210 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
-  if (newPoints.length > 0) {
-    const rangeStart = newPoints[0].time
-    const rangeEnd = new Date(newPoints[newPoints.length - 1].time.getTime() + 60_000)
-    await deleteTimeSeriesBySource(user, 'calories_active', 'aurboda_gap_fill', rangeStart, rangeEnd)
-  }
+  // 8. Expand to full minutes for the affected range, BMR-floor for non-HR minutes.
+  //    Anchor to whole minutes from start..end (caller controls range; typically a day).
+  const rangeStartMs = Math.floor(start.getTime() / 60_000) * 60_000
+  const rangeEndMs = Math.ceil(end.getTime() / 60_000) * 60_000
+  const { total, active } = buildFullDayPoints(rangeStartMs, rangeEndMs, bmrPerMin, hrPoints)
```
fiddur: The docstring above ("For minutes without HR coverage in a covered day, a BMR/min floor point is written so the daily sum naturally equals BMR + active") only holds when callers pass a whole-day range. `triggerCalorieComputation` doesn't — it passes the HR-ingest window. Either the contract here should require day-aligned ranges (and the trigger path should align them), or the comment should make clear that callers are responsible for day alignment.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -409,74 +345,64 @@ export const computeAndStoreCaloriesAll = async (
...
-  let chunkStart = new Date(range.min)
+  // Walk local-midnight day boundaries from min..max so BMR-floor coverage
+  // aligns to the user's calendar days.
+  let chunkStart = getLocalDayStart(range.min, timezone)
+  const lastDay = getLocalDayStart(range.max, timezone)
 
-  while (chunkStart < range.max) {
-    const chunkEnd = new Date(Math.min(chunkStart.getTime() + dayMs, range.max.getTime() + 60_000))
+  while (chunkStart.getTime() <= lastDay.getTime()) {
+    const chunkEnd = new Date(chunkStart.getTime() + dayMs)
```
fiddur: `chunkStart + dayMs` is a fixed 24h step, but `getLocalDayStart` returns local-midnight UTC, which is 23h or 25h apart across DST transitions. After the first DST event in the user's range, every subsequent chunk is misaligned with local midnight by 1h. Consider `chunkStart = getLocalDayStart(new Date(chunkStart.getTime() + dayMs + 3 * 60 * 60 * 1000), timezone)` or similar.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -126,278 +130,210 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
-  )
-
-  if (newPoints.length === 0) {
-    return {
-      points_computed: caloriePoints.length,
-      points_stored: 0,
-      skipped_reason: 'all minutes already computed',
-      vo2_max_source: vo2MaxSource,
-    }
+  if (hrPoints.length === 0) {
```
fiddur: If `force: true` was set, step 6 has already deleted existing aurboda calorie rows in `[start, end]`. Returning here when `hrPoints.length === 0` leaves the range empty without re-inserting — minor since `hrData.length === 0` is caught earlier, but a degenerate sample timing could still hit this path.

apps/backend/src/services/calories.ts
```diff
@@ -305,3 +305,176 @@ export const computeCaloriesForMinute = (
...
+ */
+export const computeCaloriesPerMinuteZoneMets = (params: ZoneMetsCaloriesParams): ZoneMetsCaloriePoint[] => {
+  const { hr_samples, bmr_kcal_per_day, zone_context } = params
+  if (hr_samples.length === 0) return []
+
+  const bmrPerMin = bmr_kcal_per_day / 1440
+
+  const firstTime = hr_samples[0][0].getTime()
+  const lastTime = hr_samples[hr_samples.length - 1][0].getTime()
+  const endTime = hr_samples.length === 1 ? firstTime : lastTime
```
fiddur: When `hr_samples.length === 1`, `endTime` is forced to `firstTime` so only the bucket containing that single sample is emitted. That's inconsistent with the hold-forward semantics: a single sample at 10:00 covers only 10:00:00–10:00:59 even though the documented behaviour is to hold values up to `MAX_HOLD_MINUTES`. If callers like the incremental trigger ever produce single-sample windows this becomes load-bearing — worth either documenting or letting the normal loop with the `maxHoldMs` guard handle the single-sample case uniformly.

apps/backend/src/services/calories.ts
```diff
@@ -305,3 +305,179 @@ export const computeCaloriesForMinute = (
...
+// resting/light-activity HR. Empirical replay against 14 days of intake data
+// for a stable-weight user showed Keytel+lab-BMR overshooting by ~21% and
+// the BMR/min baseline variant by ~92%, while zone-METs landed within ~10%.
+//
+// MET anchors (1 → 2 → 4 → 6 → 8.5 → 11 → 13) reflect typical physiology at
+// the boundaries of the standard 5-zone model. Below rest: 1 MET (BMR).
+// Above observed max: capped extrapolation.
+
+/** MET anchors at boundaries [resting, z1, z2, z3, z4, z5, observed_max]. */
+export const ZONE_METS_ANCHORS = [1, 2, 4, 6, 8.5, 11, 13] as const
```
fiddur, line 30: Everything above this line (the file-level Keytel docstring, `VO2_MAX_FALLBACK`, `getVo2MaxFallback`, `BASELINE_HR_MULTIPLIER`, `DEFAULT_RESTING_HR`, `CalorieCalcParams`, `CalorieDataPoint`, `GapFillInput`/`GapFillResult`, `computeGapFillPoints`, `computeCaloriesPerMinute`, `computeTotalCaloriesForMinute`, `computeCaloriesForMinute`) is no longer referenced outside this file and its tests. The PR commit message says the zone-METs model *replaces* Keytel — please delete the old code (and the matching test blocks in `calories.test.ts`) rather than leaving it as dead weight.

apps/backend/src/services/calories.ts
```diff
@@ -305,3 +305,179 @@ export const computeCaloriesForMinute = (
...
+    }
+  }
+  return ZONE_METS_ANCHORS[ZONE_METS_ANCHORS.length - 1]
+}
+
+/**
+ * Estimate BMR (kcal/day) using Mifflin-St Jeor when no lab value is known.
+ * Returns null when required inputs are missing (no height etc).
+ */
+export const estimateBmrMifflinStJeor = (
```
fiddur, line 74: Docstring says "Returns null when required inputs are missing (no height etc)" but the signature returns `number`, never null. The null branch is in `resolveBmr`. Either update the docstring or change the function to return `number | null` and move the height check here.

fiddur (commented):
🛑 Changes required

The zone-METs model itself looks well-thought-out: the empirical justification is clear, the helper functions are pure and well unit-tested, the local-day expansion + DST-safe walk in `computeAndStoreCaloriesAll` is correct, and the schema/response changes are consistent. Two blocking items though:

**1. Dead Keytel code left in `apps/backend/src/services/calories.ts`.** The PR's commit message says it *replaces* the Keytel formula, but the entire old implementation is still present: `computeCaloriesPerMinute`, `computeCaloriesForMinute`, `computeTotalCaloriesForMinute`, `computeGapFillPoints`, `getVo2MaxFallback`, plus `VO2_MAX_FALLBACK`, `BASELINE_HR_MULTIPLIER`, `DEFAULT_RESTING_HR`, `GapFillInput`/`GapFillResult`, and a large chunk of the file-level comment describing the Keytel formula. A grep confirms no remaining callers outside the file itself and its own tests. Per the project style guide ("If you are certain that something is unused, you can delete it completely"), all of this — and the corresponding test blocks (`describe('getVo2MaxFallback')`, `describe('computeCaloriesPerMinute')`, `describe('computeCaloriesForMinute')`, `describe('computeTotalCaloriesForMinute')`, `describe('computeGapFillPoints')`) — should go.

**2. `docs/features/calories.md` is now stale.** It still describes the Omnicalculator/Keytel formula (men/women equations), VO2-max as a required input with 730-day lookback, the `resting_hr * 1.2` baseline subtraction, and the HC-aggregate gap-fill mechanism — none of which match the implementation any more. CLAUDE.md explicitly requires docs updates when behavior changes ("When changing APIs, sync behavior, admin/user setup, or data models, update the corresponding docs"). The doc also needs to mention that `calories_total` is now aurboda-authoritative and that Garmin's `calories_total` is now excluded from queries.

**Minor:**
- `estimateBmrMifflinStJeor`'s docstring says "Returns null when required inputs are missing (no height etc)" but the function always returns a number; null-handling lives in `resolveBmr`. Fix the docstring (or actually return null).
- `const restingHr = restingHrMetric ?? 60` in `calorie-computation.ts` uses a magic 60 instead of `DEFAULT_RESTING_HR` from `calories.ts`. (If you delete `DEFAULT_RESTING_HR` along with the rest of the Keytel code, define a new constant locally — the value still matters.)
- `resolveZoneMetsContext` duplicates the `hr_zone_start` priority that `getEffectiveHrZones` already does. You can drop the explicit `if (settings.hr_zone_start)` branch and just use `getEffectiveHrZones(user)` directly.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -126,278 +130,231 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
+  const height = await getLatestMetricValue(user, 'height', expandedEnd, 3650)
+
+  // 4. BMR (lab metric → Mifflin-St Jeor fallback)
+  const bmr = await resolveBmr(user, expandedEnd, { age, height_cm: height, sex, weight_kg: weight })
+  if (bmr === null) return skippedResult('no BMR and no height for fallback')
+  const bmrPerMin = bmr.value / 1440
+
+  // 5. Resting HR + zone-METs context
+  const restingHrMetric = await getLatestMetricValue(user, 'resting_heart_rate', expandedEnd)
+  const restingHr = restingHrMetric ?? 60
```
fiddur: Magic number 60 — use the existing `DEFAULT_RESTING_HR` constant from `./calories.ts` (or, if Keytel cleanup removes it, define a small `DEFAULT_RESTING_HR` near the top of this file). Same value, but named.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -126,278 +130,231 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
+): Promise<ZoneMetsContext> => {
+  const settings = await getSettings(user)
+  const observedMax = settings.training_load?.observed_hr_max ?? 220 - age
+  let zones: HrZoneThresholds
+  if (settings.hr_zone_start) {
+    zones = settings.hr_zone_start
+  } else {
+    const effective = await getEffectiveHrZones(user)
+    zones = effective.zones
+  }
```
fiddur: The `if (settings.hr_zone_start)` branch duplicates what `getEffectiveHrZones` already does (custom → age-based → default). You can collapse this to just `let zones = (await getEffectiveHrZones(user)).zones` and keep only the post-hoc unusable-zones fallback below.

packages/api-spec/src/schemas/common.ts
```diff
@@ -413,19 +413,21 @@ export const cumulativeMetrics: MetricType[] = [
...
- * When aurboda computes per-minute values (e.g., calories from HR data),
- * the daily aggregate from Health Connect is redundant and would produce
- * nonsense if AVG'd with per-minute values in the same bucket.
+ * Aurboda writes per-minute calories_total and calories_active using the
+ * user's calibrated HR zones, lab BMR, and HR data. After a full
+ * recalculate, these cover every minute of every day with HR data — a
+ * complete picture more accurate than any third-party daily total.
  */
-export const aurbodaOnlyMetrics: MetricType[] = ['calories_active']
+export const aurbodaOnlyMetrics: MetricType[] = ['calories_active', 'calories_total']
```
fiddur, line 427: Adding `calories_total` here silently changes query semantics: Garmin's `calories_total` rows (written in `integrations/garmin/process.ts:183`) and any other non-aurboda source will be filtered out of queries. That's the intent per the new docstring, but please update `docs/features/calories.md` and `docs/garmin.md` to call this out so users/operators aren't surprised by the change.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -1,12 +1,17 @@
...
- * as calories_active with source 'aurboda', plus queues outbound sync.
+ * Triggered after HR data is ingested (from Health Connect, Oura, Garmin).
+ * Computes per-minute calories using the zone-METs model (calibrated HR zones
+ * + lab/Mifflin-St Jeor BMR) and stores both `calories_active` and
+ * `calories_total` with source 'aurboda'. Outbound sync is queued so the
+ * mobile app can write `calories_active` back to Health Connect.
+ *
+ * For minutes without HR coverage in a covered day, a BMR/min floor point is
+ * written so the daily sum naturally equals BMR + active.
  */
```
fiddur, line 12: `docs/features/calories.md` still describes the Keytel formula, VO2-max input, the `resting_hr * 1.2` baseline, and the HC-aggregate gap-fill scheme — none of which reflect this new implementation. Per CLAUDE.md "Keep documentation in `docs/` up to date when changing related code", that file needs to be rewritten to describe the zone-METs model, the lab-BMR/Mifflin-St Jeor priority, the BMR/min floor for uncovered minutes, and the fact that `calories_total` is now aurboda-authoritative.

fiddur (commented):
🛑 Changes required

Solid replacement of the Keytel/omnicalculator formula with a zone-METs model anchored on the user's own physiology, and the documentation/data-source story is much cleaner now that aurboda is the authoritative source for `calories_total` as well. The code, tests, and docs are largely well-aligned, and the day-aligned + DST-safe boundary logic is thoughtful.

However there are two issues I'd want resolved before merging:

1. **Outbound sync regression — duplicate-flood.** `computeAndStoreCalories` now always re-emits the full set of HR-covered minutes through `enqueueCalorieSync` every time it runs, because the previous "only new minutes" filter was removed along with the `force`/incremental path. The combination of (a) every HR-ingest call now re-processing the *whole expanded local day*, (b) `enqueueCalorieSync` enqueueing *every* HR-derived minute, and (c) `outbound_sync_queue` having no UNIQUE constraint on `(entity_type, entity_id)` (see `apps/backend/src/schema/system.ts:59-72`, no dedup in `enqueueOutboundSync`), means each subsequent HR sync within the same day appends another full copy of that day's `calories_active|<ts>` rows to the queue. A user with HC background sync firing every ~10 minutes can accumulate dozens of duplicate insert entries per minute of HR coverage per day. Either deduplicate the queue (an `ON CONFLICT (entity_type, entity_id) WHERE status = 'pending' DO NOTHING` and a partial unique index), or restore some form of "only enqueue minutes whose stored aurboda value actually changed" before enqueueing — the change-detection used to live in step 9 of the old code.

2. **`basal_metabolic_rate` lookup is currently unreachable.** `resolveBmr` queries `getLatestMetricValue(user, 'basal_metabolic_rate', …)`, and the docs advertise this as the primary BMR source ("e.g. from an InBody scan"). But `'basal_metabolic_rate'` is not in `validMetrics` in `packages/api-spec/src/schemas/common.ts:12-74`, and no integration writes this metric anywhere in the repo — the lookup will return `[]` for every user, so the function always falls through to Mifflin-St Jeor. If lab BMR really is the intended priority path, add the metric to `validMetrics` and document the ingest route; otherwise drop the lookup and the `lab` branch (and the `bmr_source` enum value) so the API doesn't promise a code path that can't fire.

A couple of smaller things worth fixing while you're in there:

- `computeAndStoreCalories` fetches `getUserSettings(user)` three times per call (line 253, again inside `getDeviceTimezone` at line 263→338, and again inside `resolveZoneMetsContext` via `getSettings`). With this function now running on every HR ingest *and* expanding to a full local day, those extra round-trips add up. Pass the already-loaded settings down (or hand its `device_timezone` to `getLocalDayStart` directly).
- The held-forward minutes from `computeCaloriesPerMinuteZoneMets` can extend up to `MAX_HOLD_MINUTES - 1` past the last HR sample's minute. When the last HR sample of a day is within the last 5 minutes of `expandedEnd`, those tail buckets are present in `hrPoints` but are filtered out by `buildFullDayPoints`' `ms < dayEndMs` loop — yet they still flow into `enqueueCalorieSync`, so the outbound queue ends up with rows that have no DB-side counterpart. Filtering `hrPoints` to `[expandedStart, expandedEnd)` before queueing fixes both the consistency gap and reduces queue churn.
- `computeAndStoreCaloriesAll` walks every local day between `range.min` and `range.max`; for days inside that range with zero HR coverage, you still write 1440 `calories_total` BMR-floor rows. That's the documented intent ("total = BMR even on rest days"), but it's worth confirming you really want that for users who simply didn't wear a device — at scale this is a non-trivial table-bloat decision, and for those minutes a `kcal_total = BMR/min` constant row is information-free.

---

_The following inline notes couldn't be anchored to the diff and are included here instead:_

**`apps/backend/src/services/calorie-computation.ts` line 320:**
> This now enqueues every HR-derived minute on every call, not just newly-computed ones. Because `computeAndStoreCalories` is now invoked on every HR ingestion *and* expands to the full local day, and `outbound_sync_queue` has no UNIQUE constraint on `(entity_type, entity_id)` (see `apps/backend/src/schema/system.ts`), every subsequent HR sync within a day appends another full copy of that day's `calories_active|<ts>` entries to the queue. With typical HC background sync this flood can produce dozens of duplicate `insert` rows per minute per day. Either dedupe at the DB level (partial unique index + `ON CONFLICT DO NOTHING` in `enqueueOutboundSync`) or restore some form of "only enqueue minutes whose value actually changed vs. the previous run" before this call.

**`apps/backend/src/services/calorie-computation.ts` line 168:**
> `'basal_metabolic_rate'` isn't in `validMetrics` (see `packages/api-spec/src/schemas/common.ts`) and nothing in the repo writes this metric, so `getLatestMetricValue` will always return `null` here and `bmr.source === 'lab'` is unreachable. Either add the metric to `validMetrics` and wire up an ingest path (the docs at `docs/features/calories.md` advertise InBody as the source), or drop this branch and the `lab` enum value so the response schema doesn't promise an outcome that can't happen.

**`apps/backend/src/services/calorie-computation.ts` line 263:**
> `getUserSettings(user)` was already awaited on line 253 (where `settings.device_timezone` is available), and `resolveZoneMetsContext` will await `getSettings(user)` a third time on line 284. Three settings round-trips per HR-ingest call is wasteful now that this function runs on every ingestion. Consider threading the already-loaded `settings` (or just its `device_timezone`) through `getLocalDayStart` / `resolveZoneMetsContext`.

**`apps/backend/src/services/calories.ts` line 165:**
> `endMinute` extends `MAX_HOLD_MINUTES - 1` past the last HR sample's minute. When the last sample of a day is within the last 5 minutes of the expanded range, `hrPoints` will contain minutes that fall *outside* `[expandedStart, expandedEnd)`. `buildFullDayPoints` filters these out (its loop is `ms < dayEndMs`), but `enqueueCalorieSync` in `calorie-computation.ts:320` receives the full unfiltered `hrPoints` and queues them — producing outbound-sync entries with no DB-side row. Suggest clamping the loop to `endMinute` or filtering `hrPoints` to the expanded range before queueing.

**`packages/api-spec/src/schemas/common.ts` line 430:**
> Worth a follow-up: now that `calories_total` is in `aurbodaOnlyMetrics`, every `cumulative-query` / daily-aggregate path will *only* see aurboda rows for this metric. Users on the prior build who have HC-aggregate `calories_total` rows but no aurboda recompute yet will see this metric drop to zero until they run `recalculate_calories`. The migration note in `docs/features/calories.md` mentions this, but you may want to surface it more loudly (release note, login-time prompt, or auto-trigger a recompute on next HR sync per-user) so users aren't briefly looking at empty totals.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -124,280 +128,239 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
 
   const sex = settings.sex as BiologicalSex | undefined
   if (!sex) return skippedResult('sex not set')
   if (!settings.birth_date) return skippedResult('birth_date not set')
-
   const age = calculateAge(settings.birth_date)
-
-  // 2. Get latest weight
-  const weight = await getLatestMetricValue(user, 'weight', end)
+  const timezone = (settings.device_timezone as string | undefined) ?? undefined
```
fiddur: `settings.device_timezone` is already typed `string | undefined` on `UserSettings`, so this cast (and the trailing `?? undefined`) is a no-op — `const timezone = settings.device_timezone` does the same thing. The codebase guideline is to avoid casts in favour of real typing/type guards.

apps/backend/src/db/outbound-sync.ts
```diff
@@ -58,6 +63,22 @@ export const enqueueOutboundSync = async (user: string, input: EnqueueOutboundSy
     )
   }
 
+  // For insert: if a pending insert for the same entity already exists,
+  // update its payload (latest value wins) and return the existing id.
+  if (input.operation === 'insert') {
+    const existing = await query(
+      user,
+      `UPDATE outbound_sync_queue
```
fiddur, line 79: Minor race: this UPDATE…RETURNING + conditional INSERT isn't atomic. Two concurrent `enqueueOutboundSync('insert', …)` calls for the same `(entity_type, entity_id)` can both see `rowCount = 0` and both fall through to the INSERT below — producing the duplicate-flood the dedup is meant to prevent. In practice the calorie-queue worker serialises runs per user, so it's unlikely to fire, but if you want a hard guarantee a partial unique index on `(user, entity_type, entity_id) WHERE status = 'pending' AND operation = 'insert'` + `ON CONFLICT … DO UPDATE` on the INSERT below would make this collision-proof. At minimum the docstring at the top of the function could note that the dedup is best-effort under concurrent writers.

docs/features/calories.md
```diff
@@ -1,128 +1,99 @@
...
+DST is handled via `getLocalDayStart` (re-anchored each chunk iteration to local midnight, not by a constant 24 h step).
 
-### Incremental vs Force Computation
+## Storage
 
-- **Incremental** (default): Checks for existing `aurboda` calorie points in the time range and skips already-computed minutes. Only computes new minutes from fresh HR data.
-- **Force** (`recalculate_calories`): Deletes all existing `aurboda` and `aurboda_gap_fill` calorie points, then recomputes from scratch. Processes in daily chunks to avoid memory issues.
+Both metrics are stored in `time_series` with `source = 'aurboda'`. The `cumulativeSources` filter in `packages/api-spec/src/schemas/common.ts` listed `calories_active` and now also `calories_total` in `aurbodaOnlyMetrics` — so all bucketed/stats/daily-aggregate queries for these metrics use only aurboda's per-minute data. Any rows from `source = 'garmin'` (written by `apps/backend/src/integrations/garmin/process.ts` from the Garmin Connect daily summary) or `source = 'health_connect_aggregate'` (written from the phone's HC aggregate) are stored but filtered out at query time. See [docs/garmin.md](../garmin.md) for the Garmin-direct caveat.
```
fiddur: Small inaccuracy: the source filter that excludes Garmin/HC daily aggregates for these metrics is `aurbodaOnlyMetrics` / `aurbodaOnlySources` (applied via `getSourceFilter` in `apps/backend/src/db/time-series.ts`), not `cumulativeSources`. `cumulativeSources` is the (separate) source list used for other cumulative metrics. Worth rewording to point at the right constant so future readers can find it.

fiddur (commented):
✅ Approved

Solid refactor. The zone-METs model is a well-motivated replacement for Keytel, the full-day BMR floor cleanly justifies switching `calories_total` to aurboda-authoritative, the local-day expansion + DST handling in `computeAndStoreCalories` / `computeAndStoreCaloriesAll` looks correct, the outbound-sync insert dedup avoids the queue flood from full-day rewrites, and the test coverage (unit + the new integration test for outbound dedup) tracks the new shape well. Docs in `docs/features/calories.md` are thorough.

A few small things worth considering — none blocking:

- **`device_timezone` cast in `computeAndStoreCalories`** (`apps/backend/src/services/calorie-computation.ts:259`) is redundant: `UserSettings.device_timezone` is already typed `string | undefined`, so `(settings.device_timezone as string | undefined) ?? undefined` simplifies to `settings.device_timezone`. Also conflicts with the project's "avoid casting" guideline. (Same in pre-existing `getDeviceTimezone` at line 346, outside the diff.)
- **Insert-dedup race in `enqueueOutboundSync`** (`apps/backend/src/db/outbound-sync.ts:66+`): the UPDATE…RETURNING then conditional INSERT pattern is not atomic. Two concurrent inserts for the same `(entity_type, entity_id)` can both see `rowCount = 0`, both INSERT, and you'd end up with two pending rows — exactly the duplication the dedup is meant to prevent. In practice the calorie-queue worker serialises per-user runs at a 5s tick so this is unlikely, but if you want a hard guarantee a unique partial index on `(user, entity_type, entity_id) WHERE status = 'pending' AND operation = 'insert'` + an `ON CONFLICT … DO UPDATE` would close the gap. Worth at least noting in the dedup-behaviour docstring.
- **Hold-forward at day boundaries**: HR samples near the end of day N held forward into the first few minutes of day N+1 are filtered out (`apps/backend/src/services/calorie-computation.ts:310`), and when day N+1 is later recomputed those samples aren't refetched (they're outside `[expandedStart, expandedEnd)`). So the first up-to-4 minutes of day N+1 stay at the BMR floor even if a sample at 23:59 of day N would have applied. Minor edge case; flagging in case it matters for the impulse model.
- **`docs/features/calories.md:69`** says "The `cumulativeSources` filter in `packages/api-spec/src/schemas/common.ts` listed `calories_active`…" — the filter that excludes Garmin/HC is actually `aurbodaOnlyMetrics` (with `aurbodaOnlySources`) via `getSourceFilter` in `db/time-series.ts`; `cumulativeSources` is the (separate) source list used for non-aurboda-only cumulative metrics. Easy one-line clarification.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -124,280 +128,239 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
+ * (e.g. zone-1 start at or below resting HR).
+ */
+const resolveZoneMetsContext = async (
+  user: string,
+  settings: UserSettings,
+  age: number,
+  restingHr: number,
+): Promise<ZoneMetsContext> => {
+  const observedMax = settings.training_load?.observed_hr_max ?? 220 - age
+  let zones = (await getEffectiveHrZones(user)).zones
```
fiddur: The docstring above says "from already-loaded settings", but this call re-fetches user settings: `getEffectiveHrZones` → `getSettings` → `getUserSettings`. Either pass the already-loaded `settings` into a small helper (`resolveZonesFromSettings(settings)`) to honour the docstring, or drop the `settings` parameter and stop pretending it's used for zones. Minor — not blocking — but a second DB round-trip per ingest adds up given how often this path runs after the full-day rewrite change.

fiddur (commented):
✅ Approved

Solid replacement of the Keytel formula with a calibrated zone-METs + BMR model. The maths, types, and tests track each other cleanly, and the documentation (`docs/features/calories.md` and the Garmin note) is now consistent with the new behaviour. A few things worth flagging, none blocking:

- **Write amplification.** `computeAndStoreCalories` now expands every caller window to full local-day boundaries and deletes/re-writes ~1440 × 2 metric rows per day on every trigger. For HR ingestion that fires multiple times per hour (Health Connect / Oura), this is a meaningful jump in DB churn vs. the prior incremental path. The outbound-sync dedup in `enqueueOutboundSync` softens this on the queue side, but the per-minute `time_series` rewrites themselves still happen. Worth keeping an eye on under realistic ingest load — single-user app so likely fine, but easy to forget if multi-user lands later.
- **Outbound-sync insert dedup is documented as non-atomic.** Acknowledged in code; revisit with a partial unique index if a concurrent-writer scenario ever appears in practice.
- **Full-recompute cost.** Per-day chunks × every day with HR data × ~2880 rows/day will be a long-running operation for users with years of HR history. Mostly a heads-up — the doc tells users to run it after deploy, which is the right call.
- **DST handling looks correct** (re-anchoring chunkStart via `getLocalDayStart` and the +26 h snap). Nice touch.
- **Old code removal is clean** — `getTimeSeriesBySource` / `getTimeSeriesWithSource` imports are dropped here and the functions themselves are still callable from elsewhere.

The migration plan (run `recalculate_calories` after deploy, which deletes both `aurboda` and `aurboda_gap_fill`) is correct and the schema change (`bmr_source` replacing `vo2_max_source`) is reflected in both the response schema and tests.

apps/backend/src/services/calorie-computation.ts
```diff
@@ -124,280 +128,240 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
+  //    not just the caller's HR-ingest window.
+  const expandedStart = getLocalDayStart(start, timezone)
+  // end is treated as exclusive; bump just past the last touched minute,
+  // then snap to next local midnight (DST-safe via getLocalDayStart on +26h).
+  const lastTouched = new Date(Math.max(end.getTime() - 1, start.getTime()))
+  const lastDayStart = getLocalDayStart(lastTouched, timezone)
+  const expandedEnd = getLocalDayStart(new Date(lastDayStart.getTime() + 26 * 60 * 60 * 1000), timezone)
+
+  // 3. Weight + height for BMR fallback
+  const weight = await getLatestMetricValue(user, 'weight', expandedEnd)
```
fiddur, line 272: Weight is now only used by the Mifflin-St Jeor fallback inside `resolveBmr`. The unchanged `if (weight === null) return skippedResult('no weight data')` two lines below therefore short-circuits even when a recent `basal_metabolic_rate` lab metric is available — the user would be skipped despite having a perfectly good BMR. Consider moving the null-check into `resolveBmr` (return null only if both lab BMR and a viable Mifflin input set are missing).

apps/backend/src/services/calorie-computation.ts
```diff
@@ -124,280 +128,240 @@ const invalidateTrainingLoadImpulses = async (user: string, fromTime: Date): Pro
...
+
+  // 6. HR data for the full expanded range
+  const hrData = await getTimeSeries(user, 'heart_rate', expandedStart, expandedEnd)
+
+  // 7. Always wipe any prior aurboda rows for the expanded range so this run
+  //    becomes the new ground truth. (No 'force' option needed — every call
+  //    is authoritative for the day(s) it touches.)
+  await deleteTimeSeriesBySource(user, 'calories_active', 'aurboda', expandedStart, expandedEnd)
+  await deleteTimeSeriesBySource(user, 'calories_active', 'aurboda_gap_fill', expandedStart, expandedEnd)
+  await deleteTimeSeriesBySource(user, 'calories_total', 'aurboda', expandedStart, expandedEnd)
```
fiddur, line 294: These three unconditional `deleteTimeSeriesBySource` calls + ~1440 inserts run on *every* HR-ingest trigger, even a small Health Connect batch that adds a handful of HR samples. Worth noting in case incremental syncs end up rewriting the same day dozens of times per day in practice — the recent comment block above explains the intent, just leaving a flag here in case post-deploy I/O numbers ask for an optimisation.

apps/backend/src/services/calories.ts
```diff
@@ -266,42 +167,19 @@ export const computeCaloriesPerMinute = (params: CalorieCalcParams): CalorieData
...
-  ageYears: number,
-  sex: BiologicalSex,
-): number => {
-  let raw: number
-  if (sex === 'male') {
-    raw = (0.634 * avgHr + 0.404 * vo2Max + 0.394 * weightKg + 0.271 * ageYears - 95.7735) / 4.184
-  } else {
-    raw = (0.45 * avgHr + 0.38 * vo2Max + 0.103 * weightKg + 0.274 * ageYears - 59.3954) / 4.184
+export const defaultHrZoneThresholds = (restingHr: number, observedMax: number): HrZoneThresholds => {
+  const hrr = observedMax - restingHr
```
fiddur, line 177: If `observedMax <= restingHr` (badly cached `observed_hr_max`, or a user whose only HR samples were below their `resting_heart_rate` metric), `hrr` goes negative and the zones come back inverted (z5 < z1). `resolveZoneMetsContext`'s `zones[1] <= restingHr` fallback wouldn't catch that. Cheap hardening: clamp `observedMax` to at least `restingHr + 1` (or similar) before computing HRR.

fiddur (commented):
✅ Approved

Solid replacement of the Keytel formula with a zone-METs + BMR model. The shape is good: pure, well-typed helpers in `calories.ts` with focused unit tests, and a thin orchestrator in `calorie-computation.ts` that expands every ingest window to local-day boundaries and re-writes the day from scratch. The day-aligned recompute and DST handling (`+26 h` then snap-to-local-midnight) is a nice fix for the previous 24h-step drift. The outbound-sync `insert` dedup with a documented atomicity caveat is a sensible solution to the duplicate-flood problem; the integration test covers it.

1. **Stale MCP tool description.** `apps/backend/src/mcp/metric-tools.ts:204` still reads *"Uses weight from Health Connect and VO2 max (measured or age/sex fallback)"*. That description isn't in the diff hunks, but it's now wrong and user-visible via tool discovery — please update it (and check the matching REST handler text in `metrics-router.ts` if any) before merge or as a quick follow-up. The schema-level `bmr_source` field is updated correctly.

2. **Eager weight skip can preempt the lab-BMR path.** `computeAndStoreCalories` still does `if (weight === null) return skippedResult('no weight data')` before calling `resolveBmr`. Now that weight is *only* consumed by the Mifflin-St Jeor fallback (not by zone-METs scaling), a user with a recent `basal_metabolic_rate` metric but no weight will be skipped unnecessarily. Pushing the null-check inside `resolveBmr` (skip only if we also have no lab BMR) would close that gap without changing the happy path.

3. **Write amplification on incremental HR sync.** Every `triggerCalorieComputation` call now wipes and re-inserts the full local day(s) it touches — `~1440` rows of `calories_total` plus the active rows, per day. For chatty HR ingests (Health Connect partial syncs, Oura repeated sleep updates), the same day gets rewritten many times. This is intentional per the docs ("every call is authoritative for the day(s) it touches"), and the comment in the source is clear about it; just worth flagging because it's a meaningful behaviour change vs. the previous incremental path. Watch for I/O / training-load-invalidation cost during the first day post-deploy.

4. **N+1 settings fetch in `computeAndStoreCaloriesAll`.** Each per-day chunk re-enters `computeAndStoreCalories`, which re-fetches `getUserSettings(user)` and the various `getLatestMetricValue(...)` lookups. For a full historical recompute (hundreds of days) that's a lot of redundant round-trips. Threading the already-resolved settings / BMR / zone-context through as an optional argument would make `computeAndStoreCaloriesAll` materially faster — easy to do later if it becomes a problem.

5. **Minor:** `defaultHrZoneThresholds` has no guard against `observedMax <= restingHr`. The post-hoc fallback in `resolveZoneMetsContext` only fires when `zones[1] <= restingHr`, so a pathological `observed_hr_max` setting could yield inverted zones. Probably not reachable in practice, but a `Math.max(observedMax, restingHr + N)` would harden it.

None of the above block the PR; (1) is the only one I'd push to land before deploy.

PR #777, opened 2026-05-22
fix(db): hard-delete in deleteTimeSeriesBySource (calorie recompute is broken in prod)

fiddur:
## Summary

PR #775's calorie recompute is currently broken in production. The new model writes per-minute \`calories_total\` and \`calories_active\` with a wipe-and-rewrite pattern: delete prior aurboda rows for the affected local day, then re-insert under the same \`(time, metric, source)\` keys. After deploy, the meals overview shows zero burned calories for today and yesterday, and \`query_metrics calories_total\` returns 0 rows even though the audit log shows \`Computed N calorie points\` (N=4870, 4579, etc.).

Root cause: \`deleteTimeSeriesBySource\` does a **soft** delete (sets \`deleted_at = NOW()\`), and \`insertTimeSeries\`' \`ON CONFLICT (time, metric, source) DO UPDATE SET value = EXCLUDED.value WHERE time_series.deleted_at IS NULL\` skips the update when the existing row is tombstoned — leaving the row invisible. The PK constraint then prevents the new row from being inserted.

The \`WHERE deleted_at IS NULL\` filter on the conflict update exists to preserve user-initiated tombstones (so deleting a manual reading isn't auto-restored by the next background sync). That contract should stay. But the \`deleteTimeSeriesBySource\` callers (calorie recompute, training-load rebuild) want **hard**-delete semantics, not tombstones.

## Changes

- \`deleteTimeSeriesBySource\` now does \`DELETE FROM time_series\` instead of \`UPDATE … SET deleted_at = NOW()\`. The single-row user delete path (\`deleteMetric\`) still soft-deletes — its intent is to preserve the tombstone.
- Docstring expanded to explain the soft-vs-hard contract.
- Regression integration test added: delete + re-insert on the same \`(time, metric, source)\` results in the new value being visible.

## Test plan

- [x] \`pnpm check\` — 0 errors
- [x] All time-series integration tests pass (37 including the new one)
- [ ] After merge + deploy: re-run \`recalculate_calories\` and confirm the meals overview shows non-zero burn for today / yesterday.

### DIFF
--- a/apps/backend/src/db/time-series.integration.test.ts
+++ b/apps/backend/src/db/time-series.integration.test.ts
@@ -678,6 +678,46 @@ describe('Time Series Integration Tests', () => {
 
       expect(count).toBe(0)
     })
+
+    test('subsequent insertTimeSeries on the same (time, metric, source) is visible (no tombstone)', async () => {
+      // Regression: deleteTimeSeriesBySource used to soft-delete (set
+      // deleted_at = NOW()). The ON CONFLICT clause in insertTimeSeries has
+      // `WHERE deleted_at IS NULL`, so re-inserting the same key was
+      // silently skipped and the row stayed invisible. This broke the
+      // calorie recompute's wipe-and-rewrite pattern. Hard-delete now.
+      const user = getTestUser()
+      const t = new Date('2024-02-01T10:00:00Z')
+
+      // 1. Initial insert (e.g. previous recompute pass)
+      await insertTimeSeries(user, [
+        { metric: 'heart_rate', source: 'aurboda', time: t, value: 70 },
+      ])
+
+      // 2. Wipe via deleteTimeSeriesBySource
+      const deleted = await deleteTimeSeriesBySource(
+        user,
+        'heart_rate',
+        'aurboda',
+        new Date('2024-02-01T09:00:00Z'),
+        new Date('2024-02-01T11:00:00Z'),
+      )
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Correct, targeted fix for a real regression. The root cause is accurate: `insertTimeSeries`' ON CONFLICT clause is gated by `WHERE deleted_at IS NULL`, so a soft-deleted row leaves a tombstone that silently blocks any subsequent re-insert at the same `(time, metric, source)` — exactly what the calorie recompute's wipe-and-rewrite pattern needs to survive. Switching to a real `DELETE` is the right call here; the two callers (`calorie-computation.ts` and `training-load/deps.ts`) are both rebuild paths that immediately re-insert, and there are no FKs onto `time_series.id` so hard-delete won't orphan anything.

The docstring is unusually good — it explains *why* this function differs from `deleteTimeSeriesPoint` / `deleteTimeSeriesMetric` and explicitly tells future readers not to flip those to hard-delete. The new integration test reproduces the exact regression scenario (insert → delete → re-insert → expect new value visible) rather than just asserting the row count, which is what you want for a bug like this.

Minor (non-blocking, not in the diff): the section banner at `time-series.ts:374` still says `// Time Series Deletion (soft delete)` but now contains one hard-delete function. Worth tightening the header next time someone touches this file, but not worth churn now.

PR #778, opened 2026-05-24
feat(data): show metrics in Data view with daily rollup

fiddur:
## Summary
- Adds a **Metrics** filter to the Data page. Each metric renders as a single row per day, merged across sources:
  - Stream metrics: `Heart Rate · 00:05 – 16:50 · 60–150 bpm · 500 measurements`
  - Point metrics: `Body Water Mass · 07:59 · 57.2 kg`
  - Cumulative metrics: `Steps · 8523 steps · N measurements`
  - Rows link to the metric's chart.
- Adds an **All off / All on** toggle in the filter sidebar to focus on a single data type.
- Backend: `getTimeSeriesBucketed` and the `BucketMetricStats` api-spec schema now expose `first_time`/`last_time` per metric per bucket, so the listing can show the real measurement window rather than the bucket boundary.

## Test plan
- [x] `pnpm check` clean (typecheck + lint)
- [x] Backend unit tests pass (1782)
- [x] Web unit tests pass (405)
- [x] `time-series.integration.test.ts` passes (37) — covers new `first_time`/`last_time` columns
- [ ] Visual check on aurboda.net: open Data view, confirm Metrics rows appear and toggle works
- [ ] Confirm `body_water_mass` appears for a day with Health Connect data


fiddur (commented):
✅ Approved

Clean addition. The PR plumbs `first_time` / `last_time` through the bucketed-metrics stack (DB query → service → API schema) and uses them to render daily metric rollups in the Data view. The schema additions are picked up consistently in `BucketedMetricData`, `BucketMetricStats`, and `computeContextualHrvBuckets`; tests are updated everywhere they need to be, and a new dedicated test covers the surfaced fields. The new `mkBucket`/`mkStats` helpers in `metrics.test.ts` are a nice cleanup that absorbs the schema growth without exploding the test diff.

Minor observations (non-blocking):

- `metricsToItems` always says `"${count} measurements"` — for cumulative metrics with `count === 1` you'll get `1 measurements` (and the same plural quirk for non-cumulative ranges, but those would normally have count > 1 anyway).
- The metric chart link hardcodes `lookback_days=1`, so clicking a metric for a date other than today will navigate to a chart that doesn't show that date — minor UX wart but consistent with how the chart route is currently parametrised.
- The new `All on/All off` toggle is a nice quality-of-life addition but is technically outside the PR's stated scope; fine to keep, just noting it.

No correctness or test issues spotted.

apps/web/src/pages/Data/index.tsx
```diff
@@ -133,6 +136,65 @@ const reportToItem = (r: Report, multiDay: boolean): DataItem => ({
   type: 'report',
 })
 
+/** Format a numeric value: integer if whole, otherwise 1 decimal. */
+const formatMetricValue = (n: number): string => {
+  const rounded = Math.round(n * 10) / 10
+  return Number.isInteger(rounded) ? rounded.toString() : rounded.toFixed(1)
+}
+
```
fiddur, line 144: Minor: this cast violates the project's 'avoid casting, use proper typing and type guards' rule. `@aurboda/api-spec` already exports `getMetricUnit(metric)` which uses `isValidMetric` internally (and also returns custom-metric units). Importing that instead of casting `metricUnits` would be cleaner and pick up custom metrics for free.

apps/web/src/pages/Data/index.tsx
```diff
@@ -133,6 +136,65 @@ const reportToItem = (r: Report, multiDay: boolean): DataItem => ({
...
+  for (const bucket of response?.buckets ?? []) {
+    for (const [metric, stats] of Object.entries(bucket.metrics)) {
+      const first = new Date(stats.first_time)
+      const last = new Date(stats.last_time)
+      const unit = metricUnit(metric)
+      const unitSuffix = unit ? ` ${unit}` : ''
+      const label = toDisplayName(metric)
+      const parts: string[] = []
+
+      const measurements = `${stats.count} measurement${stats.count === 1 ? '' : 's'}`
```
fiddur, line 160: Nit: for non-cumulative `count === 1` you show the timestamp (good), but the cumulative branch above never shows time when `count === 1` — only the sum and 'measurement(s)'. If that's intentional (cumulative singletons rarely need a time), ignore; otherwise consider pushing `formatTime(first, multiDay)` for symmetry.

fiddur (commented):
✅ Approved

Clean, focused change. The backend extension of `getTimeSeriesBucketed` to expose `first_time`/`last_time` via `MIN(time)`/`MAX(time)` is straightforward and matches a parallel update to `computeContextualHrvBuckets`. Types flow consistently through `BucketedMetricData` → `BucketMetricStats` and into the api-spec schema. Test coverage is good: a new integration assertion verifies the SQL aggregates, a new unit test covers surfacing the new fields, and the `mkBucket`/`mkStats` helpers keep the existing test bodies readable. The Data view integration looks correct — separate query, cancellation on date change, per-metric rendering with sensible branches for cumulative / single-point / range, and a useful 'All on/All off' toggle.

One minor nit (not blocking): `metricUnit` in `apps/web/src/pages/Data/index.tsx` does `(metricUnits as Record<string, string>)[metric] ?? ''`, which violates the 'avoid casting, use proper typing and type guards' guideline. The api-spec package already exports `getMetricUnit(metric)` (uses `isValidMetric` internally and also supports custom metrics) — preferable to import that instead of casting.

Small consistency observation: for cumulative metrics with `count === 1`, the rendered detail omits the timestamp, while non-cumulative `count === 1` shows it. Probably fine, but worth noting if the UX feels inconsistent.

apps/web/src/pages/Data/index.tsx
```diff
@@ -133,6 +136,65 @@ const reportToItem = (r: Report, multiDay: boolean): DataItem => ({
...
+          parts.push(`${formatMetricValue(stats.min)}–${formatMetricValue(stats.max)}${unitSuffix}`)
+        }
+        parts.push(measurements)
+      }
+
+      items.push({
+        color: METRIC_COLOR,
+        detail: parts.join(' · '),
+        end: stats.count > 1 ? last : undefined,
+        href: `/chart?source_type=metric&pattern=${encodeURIComponent(metric)}&lookback_days=1`,
```
fiddur, line 188: `lookback_days=1` is always relative to `now`, so clicking a metric while viewing a past day takes the user to today's chart instead of that day's. The Chart page doesn't support an end-date param today, so a real fix needs a separate change there — just flagging for awareness.

fiddur (commented):
✅ Approved

Clean addition. The change carries `first_time`/`last_time` from the SQL aggregation all the way through `BucketedMetricData` → `BucketMetricStats` → the new `metricsToItems` renderer, and the timestamps from `MIN(time)`/`MAX(time)` give the Data view a sensible time anchor for each metric. The contextual-HRV path was also updated to keep the same shape (now tracking `times` alongside `values` so the in-memory aggregation matches the SQL one), and the schema in `packages/api-spec` makes the new fields required, matching what the backend now always emits.

Test coverage is good: the `mkBucket`/`mkStats` helpers keep the existing tests compact, a new test specifically asserts that the per-metric times surface through `queryMetricsBucketed`, and the integration test verifies the SQL aggregation against a real Postgres bucket. Remember to run `pnpm --filter @aurboda/api-spec build && (cd packages/api-spec && pnpm generate)` so the OpenAPI YAML and Kotlin models pick up the new required fields before Android consumes the bucketed endpoint.

One minor UX note (not blocking): the chart deep-link uses `lookback_days=1`, which is always relative to *now*. If the user is viewing a past day in the Data page and clicks a metric, the chart will show today's data, not that day's. This is a limitation of the Chart page (no date param), so a separate change is needed to fix it properly — flagging only so it doesn't go unnoticed.

PR #779, opened 2026-05-27
feat(food-items): add per-food-item portion sizings (PR1/3)

fiddur:
## Summary

First of 3 PRs adding multi-portion support to food items. This PR introduces the **portions entity + CRUD**; meal logging integration and web UI land in follow-up PRs.

Each food item already has a "base portion" (`default_quantity` + `default_unit`) that the nutrient columns are measured per. This PR adds **additional sizings** stored on a new `food_item_portions` table — each row encodes "X label_unit = Y base_unit" so the meal logger can scale nutrients without unit-conversion guesswork.

Examples (these would be defined per food):

| food | base | portions |
|---|---|---|
| Fylld tortilla | 1 wrap | 2 × wrap → 2 (set as default) |
| Lantmjölk | 100 g | 515 × g → 515; 1 × glas → 515 |
| Choklad | 100 g | 1 × ruta → 3.4; 1 × rad → 13.6 |

### What's in PR1

- New `food_item_portions` table (soft pointer on `food_item_id` — works for per-user OR central shared foods)
- New column `food_items.default_portion_id` (NULL = base portion is the default)
- DB layer: CRUD + batch lookup
- Service layer: CRUD with food-existence guard + "portion belongs to this food" check on set-default
- REST: `GET/POST /food-items/:id/portions`, `PATCH/DELETE /food-items/:id/portions/:portionId`, `PUT /food-items/:id/default-portion`
- MCP tools: `list_food_item_portions`, `add_food_item_portion`, `update_food_item_portion`, `delete_food_item_portion`, `set_default_food_item_portion`
- `FoodItemDetail` now includes `portions: FoodItemPortion[]` (every getDetail path)
- Cascade: deleting a per-user food drops its portions; deleting a portion clears `default_portion_id` if it pointed at the row
- Migration: idempotent `ALTER TABLE food_items ADD COLUMN IF NOT EXISTS default_portion_id UUID` (matches existing pattern in connection.ts)

### What's NOT in PR1 (follow-ups)

- **PR2**: meal logging uses portions (snapshot `food_item_portion_id` + `portion_count` on `meal_food_items`, reformulated scaling, per-user `default_portion_id` for central foods via `shared_food_item_overrides`)
- **PR3**: web UI (food item editor portions section, meal entry portion picker)
- Web UI continues to send legacy `quantity`/`unit` for now — portions exist as metadata but don't yet drive logging
- Android: no work needed (Android only uses Health Connect meal-type enums, no food-item logging UI)

## Test plan

- [x] `pnpm check` clean (backend + web)
- [x] All 2233 backend tests pass (13 new portion integration tests + mock updates for `food-items.test.ts`, `food-item-tools.test.ts`, `food-items-router.test.ts`)
- [ ] CI green
- [ ] Manual: hit `POST /food-items/:id/portions` against staging once deployed, verify it shows in `GET /food-items/:id` payload


apps/backend/src/routes/food-items-router.ts
```diff
@@ -479,6 +509,103 @@ export const createFoodItemsRouter = (authMiddleware: AnyMiddleware, centralDb:
...
+    },
+  )
+
+  router.patch<{ id: string; portionId: string }, FoodItemPortionResponse, UpdateFoodItemPortionBody>(
+    '/:id/portions/:portionId',
+    authMiddleware,
+    validateBody(updateFoodItemPortionBodySchema),
+    async (req, res) => {
+      try {
+        const row = await updatePortion(req.user!, req.params.portionId, req.body)
```
fiddur, line 561: Ownership check happens *after* the update is applied. If a caller passes the wrong `:id` (e.g. portion belongs to food A but URL says food B), `updatePortion` already mutated the row when we discover the mismatch and return 404 — the bad write is persisted. Look up the portion first and verify `food_item_id === req.params.id` before delegating to `updatePortion`.

apps/backend/src/routes/food-items-router.ts
```diff
@@ -479,6 +509,103 @@ export const createFoodItemsRouter = (authMiddleware: AnyMiddleware, centralDb:
...
+        res.status(/not found/i.test(message) ? 404 : 500).json({ error: message, success: false })
+      }
+    },
+  )
+
+  router.delete<{ id: string; portionId: string }, DeleteFoodItemPortionResponse>(
+    '/:id/portions/:portionId',
+    authMiddleware,
+    async (req, res) => {
+      const deleted = await deletePortion(req.user!, req.params.portionId)
```
fiddur, line 580: This handler never checks that the portion belongs to the food item in the URL — it just deletes by portionId. Any caller can `DELETE /food-items/<any-of-their-foods>/portions/<any-of-their-portions>` and it'll succeed. Look up the portion first and 404 if `food_item_id !== req.params.id` before calling `deletePortion`.

apps/backend/src/db/food-items.ts
```diff
@@ -236,6 +238,10 @@ export const updateFoodItem = async (
     setClauses.push(`default_unit = $${idx++}`)
     params.push(input.default_unit)
   }
+  if (input.default_portion_id !== undefined) {
```
fiddur, line 241: Allowing `default_portion_id` through the generic `updateFoodItem` bypasses the "chosen portion belongs to this food" guard that `setDefaultPortion` enforces. A client calling `PATCH /food-items/:id` with `{default_portion_id: <any-uuid>}` can set this to a portion that belongs to a different food (or doesn't exist), and `getDetail` later renders the default as a missing portion. Either reject `default_portion_id` from the generic update path, or perform the same existence/ownership check here before persisting.

fiddur (commented):
🛑 Changes required

Solid schema + service layering for the new portions feature — the soft-pointer design is clearly documented, the integration tests are thorough (cross-food guard, cascade on food delete, default-portion clearing on portion delete, central food rejection), and the api-spec wiring follows the project conventions cleanly.

Two correctness issues in the REST router need fixing before merge, plus one consistency concern around the generic food-item PATCH endpoint:

1. **`PATCH /food-items/:id/portions/:portionId` applies the update before verifying ownership.** The handler calls `updatePortion(...)` first and only then checks `row.food_item_id !== req.params.id`. A caller passing the wrong `:id` will silently mutate the portion and then receive a 404 — the bad write has already happened. Fetch the portion (or just guard via `getFoodItemPortionById`) before delegating to the service, and only update when the parent matches.

2. **`DELETE /food-items/:id/portions/:portionId` never verifies that the portion belongs to `:id` at all.** Any authenticated user can supply any `:id` and the portion will be deleted as long as the portionId exists in their schema. Add the same parent-id guard as on PATCH (and do it *before* deletion, per #1).

3. **`PATCH /food-items/:id` can write an arbitrary `default_portion_id` without the cross-validation that `setDefaultPortion` performs.** The body schema now includes `default_portion_id`, and `updateFoodItem` blindly forwards it — so a client can break the "the chosen portion belongs to this food" invariant by going through the generic endpoint instead of `PUT /:id/default-portion`. Either strip `default_portion_id` from the generic update path, or run the same existence/ownership check in `updateFoodItem` (or its service caller).

Minor: the route handlers do error-message regex matching (`/not found|does not belong/i`) to translate service exceptions into HTTP status codes — fragile, but it's an existing pattern in this file, so flagging it informationally rather than as a blocker.

apps/backend/src/db/food-item-portions.ts
```diff
@@ -0,0 +1,175 @@
...
+  // invariant ("default_portion_id resolves") never points at a deleted row.
+  // Central-target portions have no default pointer to clear (the override
+  // table picks that up in PR2), so this UPDATE is a no-op for those.
+  await query(
+    user,
+    `UPDATE food_items SET default_portion_id = NULL, updated_at = NOW()
+       WHERE default_portion_id = $1`,
+    [id],
+  )
+  const result = await query(user, `DELETE FROM food_item_portions WHERE id = $1`, [id])
```
fiddur: Nit / consistency: the UPDATE on `food_items` and the DELETE on `food_item_portions` run as two separate auto-commits. If the DELETE fails (or the connection drops) after the UPDATE succeeded, the food row's `default_portion_id` has been cleared while the portion is still present — a latent inconsistency that the comment above implies you're trying to avoid. `deleteFoodItem` already uses `BEGIN/COMMIT` for its multi-statement cascade; wrapping this pair the same way would make the invariant actually hold.

apps/backend/src/routes/food-items-router.ts
```diff
@@ -479,6 +510,114 @@ export const createFoodItemsRouter = (authMiddleware: AnyMiddleware, centralDb:
...
+        const fromCentral = await centralDb.getSharedFoodItemById(id)
+        return res.status(fromCentral ? 403 : 404).json({
+          error: fromCentral
+            ? 'Setting a default portion on a central item lives on the per-user override layer (use the override endpoint)'
+            : 'Food item not found',
+          success: false,
+        })
+      }
+      const portionId = req.body?.portion_id ?? null
+      if (portionId !== null && typeof portionId !== 'string') {
```
fiddur: This is the only food-items route that hand-rolls body validation (`typeof portionId !== 'string'`) instead of going through `validateBody(...)`. A tiny `setDefaultPortionBodySchema = z.object({ portion_id: z.string().uuid().nullable() })` in api-spec would match the rest of the router, give the same 400 error shape (field-keyed errors), and also catch non-UUID strings up front rather than letting them fall through to the DB lookup.

packages/api-spec/src/schemas/food-item-portions.ts
```diff
@@ -0,0 +1,75 @@
...
+      description: 'Display quantity for the portion (e.g. 2 in "2 wrap", 1 in "1 glas").',
+    }),
+    label_unit: z.string().min(1).max(100).meta({
+      description: 'Display unit for the portion (e.g. "wrap", "glas", "ruta").',
+    }),
+    base_equivalent: z.number().positive().meta({
+      description:
+        'How much of the food item\'s base unit this whole portion entry equals. Example: a 100 g base with a "1 ruta = 3.4 g" portion stores 3.4 here.',
+    }),
+    sort_order: z.number().int().optional().meta({ description: 'Display order' }),
```
fiddur: Minor: `sort_order`, `created_at`, and `updated_at` are marked `.optional()` here, but the DB always populates them and `serializePortion` in the router always emits them. Tightening these to required would let clients narrow without `?? 0` / `?? ''` ceremony at every callsite.

fiddur (commented):
✅ Approved

Nicely-scoped PR1/3: a new `food_item_portions` table, db/service/router/MCP wiring, and the `default_portion_id` pointer on `food_items` — with thoughtful soft-pointer design (per-user OR central foods), pre-mutation ownership guards on the per-portion endpoints, and a deliberate decision to keep `default_portion_id` out of the generic PATCH (covered by a regression test). Coverage is solid: db integration tests, service integration tests, router unit tests for the ownership/strip guards, and the MCP mocks are updated.

Migration looks correct: `ALTER TABLE food_items ADD COLUMN IF NOT EXISTS default_portion_id UUID` for existing users, the column is added to the CREATE TABLE in `schema/meals.ts` for new users, the new table goes into `tableCreationOrder`, and `cleanTestDb` includes it. Schema export via `@aurboda/api-spec` follows the documented snake_case + `.meta({id, description})` conventions.

A couple of small things worth considering (non-blocking — left inline):

- `deleteFoodItemPortion` does an UPDATE-then-DELETE without a transaction. The window is tiny, but if the DELETE fails after the UPDATE commits, the food row's `default_portion_id` is gone with the portion still present. Other delete paths in this codebase (`deleteFoodItem`) wrap multi-statement cascades in `BEGIN/COMMIT`; same treatment here would be consistent.
- The new `PUT /:id/default-portion` route is the only food-items endpoint that hand-rolls request body validation (`typeof portionId !== 'string'`) instead of using a Zod schema + `validateBody`. A tiny `setDefaultPortionBodySchema` in api-spec would match the rest of the router and give callers the same 400 shape.
- `foodItemPortionSchema` declares `sort_order`, `created_at`, and `updated_at` as `.optional()`, but the DB always returns them and `serializePortion` always emits them. Tightening these to required would let clients rely on the fields without narrowing.

None of those block the PR — the behavior is correct as written and the path is well-tested. Verdict: approved.

fiddur (commented):
✅ Approved

Solid PR1/3. The schema, db, service, REST and MCP layers all line up cleanly with the project's `api-spec` single-source-of-truth pattern, and the test coverage is good — integration tests for the soft-pointer behaviour, cascade-on-food-delete, the cross-food default-portion guard, and route-level ownership guards for PATCH/DELETE. Schema-stripping `default_portion_id` from the generic PATCH (forcing callers to the dedicated `/default-portion` endpoint that validates ownership) is a nice defence-in-depth choice. The `deleteFoodItemPortion` BEGIN/UPDATE/DELETE/COMMIT works because of the per-user-DB design (one cached `Client` per user); worth a mental note that nesting this inside another transaction would break, but no current caller does so.

Things to keep in mind, not blockers:

- **Soft pointer to central foods can be orphaned.** A user-created portion that targets a central `shared_food_items` id will remain in the per-user DB even if the central row is renamed/removed (or merged). The cascade comment in `food-item-portions.ts` only covers per-user food deletion. Worth a follow-up note in the docs or a sync/cleanup pass when the override layer lands in PR2.

- **`base_equivalent` semantics vs. nutrient scaling.** Nutrient totals on `food_items` are per `(default_quantity, default_unit)` (e.g. per 100 g), but `base_equivalent` is a count of base units (e.g. 3.4 g, not 0.034). Any later consumer that scales nutrients via portions must divide by `default_quantity`, not just multiply by `base_equivalent`. Worth a one-line comment on `base_equivalent` in the schema, or a helper in the service that does the conversion so callers can't get this wrong.

- **Every `getDetail()` now issues an extra `SELECT` for portions** (parallel with sensitivities, so no added latency, but extra DB load). Fine for now; if detail fetches become hot, consider batching portions into the food list/detail queries similarly to how `getPortionsByFoodItemIds` already supports.

- Minor: the REST PATCH/DELETE routes do a pre-mutation `getFoodItemPortionById`, then the service does it again on its own — two SELECTs per call. Acceptable, but could be tightened later by passing the prefetched row into the service.

apps/backend/src/services/food-items.ts
```diff
@@ -425,9 +428,15 @@ export const createFoodItemsService = (centralDb: CentralDb): FoodItemsService =
...
     // Sensitivity flags work for both per-user and central items — the
     // junction's food_item_id is a soft pointer, so a user can flag a
-    // central LSV item just as easily as one of their own.
-    const sensitivityFlags = await getFoodItemSensitivities(user, id)
+    // central LSV item just as easily as one of their own. Portions share
+    // the same soft-pointer design, so the same id resolves regardless of
+    // which library the food itself lives in.
+    const [sensitivityFlags, portionRows] = await Promise.all([
+      getFoodItemSensitivities(user, id),
+      listPortionsForFoodItem(user, id),
```
fiddur, line 442: Minor: this adds an unconditional `listPortionsForFoodItem` SELECT to every `getDetail()` call, even for foods that have no portions. Parallelised with sensitivities so no serial latency cost, but it doubles the DB hits for the (likely common) no-portion case. Worth considering a `hasPortions` flag on the food row or a single combined query if detail becomes hot.

apps/backend/src/db/food-item-portions.integration.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const after = await getFoodItemById(user, food.id)
+    expect(after?.default_portion_id).toBeUndefined()
+  })
+
+  test('getPortionsByFoodItemIds groups by food id', async () => {
+    const user = getTestUser()
+    const a = await upsertFoodItem(user, { name: 'A' })
+    const b = await upsertFoodItem(user, { name: 'B' })
+    const c = await upsertFoodItem(user, { name: 'C-no-portions' })
+    await insertFoodItemPortion(user, {
```
fiddur, line 98: Nit: this dynamic `import('./food-items.ts')` can just be a top-level static import alongside the existing `import { deleteFoodItem, getFoodItemById, upsertFoodItem } from './food-items.ts'` — easier to grep and consistent with the rest of the file.

packages/api-spec/src/schemas/food-item-portions.ts
```diff
@@ -0,0 +1,96 @@
...
+    }),
+    label_quantity: z.number().positive().meta({
+      description: 'Display quantity for the portion (e.g. 2 in "2 wrap", 1 in "1 glas").',
+    }),
+    label_unit: z.string().min(1).max(100).meta({
+      description: 'Display unit for the portion (e.g. "wrap", "glas", "ruta").',
+    }),
+    base_equivalent: z.number().positive().meta({
+      description:
+        'How much of the food item\'s base unit this whole portion entry equals. Example: a 100 g base with a "1 ruta = 3.4 g" portion stores 3.4 here.',
```
fiddur: Consider clarifying here that `base_equivalent` is a *count of the food's base unit* (e.g. 3.4 g for a 100 g base), not a fraction of `default_quantity`. The example does say this, but a consumer scaling nutrients will need to divide by `default_quantity` — easy to get wrong. A one-line note ("to scale nutrients: nutrient * base_equivalent / default_quantity") would prevent the future foot-gun in PR2/PR3.

fiddur (commented):
✅ Approved

Well-structured PR1/3 introducing per-food-item portion sizings end-to-end (schema → DB → service → REST + MCP → api-spec types). The defence-in-depth design is good throughout:

- `food_item_portions.food_item_id` as a deliberate soft pointer is documented and supported by `ensureFoodExists` in the service, which checks both per-user and central libraries.
- `setDefaultPortion` validates `portion.food_item_id === foodItemId` before writing — and `updateFoodItem` returning null on a central-id rejection acts as a second guard.
- `deleteFoodItemPortion` wraps the pointer-nullification + row delete in a transaction so the `food_items.default_portion_id` invariant can't be left dangling on partial failure. The shared `getDbForUser` cache means the BEGIN/COMMIT really do run on the same connection.
- `deleteFoodItem` already runs in a transaction and now correctly cascades `food_item_portions` alongside `food_item_sensitivities`.
- REST `PATCH`/`DELETE /:id/portions/:portionId` enforce that the portion's `food_item_id` matches the URL `:id` before any mutation — the test cases for cross-food URL mismatch are a nice touch.
- `updateFoodItemBodySchema` deliberately excludes `default_portion_id` (forcing callers to the dedicated `PUT /:id/default-portion` endpoint that runs the ownership check), and there's a test that explicitly verifies the generic PATCH drops it.

Integration tests against testcontainers cover the meaningful invariants (sort order, cascade behaviour, default-pointer clearing on portion delete, group-by lookup).

Minor observations (none blocking):

1. `getPortionsByFoodItemIds` is exported from `db/index.ts` but only exercised by its own integration test — presumably staged for PR2/3 (list-detail batching). Worth removing from the barrel if it isn't actually consumed elsewhere this PR cycle, otherwise fine to leave.
2. The new `food_item_portions` table has no DB-level `CHECK (label_quantity > 0)` / `CHECK (base_equivalent > 0)`; positivity is only enforced via Zod at the API boundary. Consistent with how other tables in this project work, so flagging only as something to keep in mind if a future direct-insert path (e.g. sync/import) bypasses the service.
3. No `docs/` update accompanying the new portion model — acceptable for an incremental PR1/3, but worth syncing `docs/features/meals.md` (or a new `docs/features/food-portions.md`) by the time the final PR lands, per the project's documentation policy.
4. The schema's `idx_food_item_portions_food (food_item_id, sort_order, id)` doesn't cover the `ORDER BY sort_order, created_at, id` tail; Postgres will do a small in-memory sort on each food's portion list. Volume is negligible for the foreseeable future, so a non-issue.

LGTM.

PR #781, opened 2026-05-27
feat(food-items): meal logging via portion id (PR2/3)

fiddur:
## Summary

Second of 3 PRs adding multi-portion support. Builds on [PR1 #779](https://github.com/fiddur/aurboda/pull/779) to wire portions into meal logging.

Backend-only — web UI lands in PR3.

### What's in PR2

**DB**
- `meal_food_items`: new `food_item_portion_id UUID` + `portion_count DOUBLE PRECISION`. Legacy rows stay valid; scaling falls back to existing `quantity`/`unit` path.
- `shared_food_item_overrides`: new `default_portion_id UUID`. Lets a user preselect a portion on a central LSV food without forking it.

**api-spec**
- `foodItemInputSchema` / `foodItemSchema`: add `food_item_portion_id` + `portion_count`.
- `FoodItemDetail`: add `effective_default_portion_id` (resolves per-user direct OR central via override).
- `SharedFoodItemOverride` + `SetBody`: add `default_portion_id` (string sets, null clears, omitted leaves).

**Services**
- `meals.ts`: when `food_item_portion_id` is set, scale by `portion_count × portion.base_equivalent / default_quantity`. Legacy `quantity`/`unit` columns get populated from the portion's label (so display fallback survives portion deletion).
- Resnapshot path refetches the portion and rescales; if the portion was deleted since logging, falls back to legacy scaling against the recorded `q`/`u`.
- `food-items.ts`: `applySharedOverrides` now decorates central rows with the user override's `default_portion_id` when set (NULL = no override → fall through to central).
- Router serializer emits `effective_default_portion_id`.
- MCP `set_shared_food_item_override` at-least-one-field guard accepts `default_portion_id`.

### Tests

- 4 new meal integration tests: portion-scaled snapshot, cross-food rejection, missing portion_count rejection, resnapshot preserves portion link.
- 2 new override integration tests: `default_portion_id` round-trip independent of icon, clear-via-null reverts.
- All 2243 backend tests pass.

### Backwards compat

- Legacy `quantity`/`unit` input continues to work — only when both `food_item_portion_id` AND `portion_count` are present does the portion path activate.
- Existing meals continue to read with `food_item_portion_id` NULL on every row.

## Test plan

- [x] `pnpm check` clean (backend + web)
- [x] Backend tests green
- [ ] CI green


fiddur (commented):
🛑 Changes required

PR2/3 wires up portion-based meal logging end-to-end: junction columns, schema, service-layer scaling/resnapshot, per-user portion override on central items, and a `effective_default_portion_id` projection on `FoodItemDetail`. The core scaling math (`portion_count × base_equivalent / default_quantity`) is correct, the integration tests cover the happy path + the two main rejection paths + resnapshot, and the override migration / round-trip is solid.

A few concerns worth addressing before merge:

1. **Portion-validation errors throw instead of returning a `MealResult`** (`apps/backend/src/services/meals.ts:396/399/404`). The rest of `addMeal` / `updateMealById` returns `{ success: false, error }` so the router can 400. Throwing here means the router's `addMeal(...)` call rejects (becomes a 500 / unhandled rejection) — and worse, `dbUpsertMeal` runs *before* `syncFoodItemsToJunction`, so a request with an invalid `food_item_portion_id` will leave behind an orphan meal row with no items. Either validate the portions up front (before the meal row is written) or return a `MealResult` failure so the route layer maps it to a 400 cleanly.

2. **Reference branch of `serializeDetail` drops `effective_default_portion_id`** (`apps/backend/src/routes/food-items-router.ts:135-146`). The composite branch and the plain-atomic branch both surface it; only the reference-enriched branch omits it. A per-user reference-type food item can still have its own portions + `default_portion_id`, so the UI loses the preselect for that subtype. (Line is outside this PR's diff, so flagged here rather than inline.)

3. **Schema doesn't enforce `portion_count` when `food_item_portion_id` is set** (`packages/api-spec/src/schemas/meals.ts`, `foodItemInputSchema`). The descriptions say it's required, but there's no `.refine()` — so the only enforcement is the runtime throw in `resolvePortionForInput`, which (per #1) is the wrong layer. Adding a refine like `setSharedFoodItemOverrideBodySchema` already uses would give a clean 400 via `validateBody`.

4. **Resnapshot does an N+1 `getFoodItemPortionById` lookup per matching link.** Bounded (only links pointing at the food being resnapshotted), so not blocking, but a single batch by id would be cheap.

Non-blocking nits: the `usingPortion` predicate is recomputed in both `computeScale` and `buildScaledJunctionItem`; extracting it once and passing it (or `scale` + `usingPortion`) would tighten the pair.

apps/backend/src/services/meals.ts
```diff
@@ -345,6 +379,33 @@ const resolveCanonical = async (
...
+ */
+const resolvePortionForInput = async (
+  user: string,
+  fi: FoodItemInput,
+  canonical: MergedFoodItem,
+): Promise<FoodItemPortionRow | null> => {
+  if (!fi.food_item_portion_id) return null
+  const portion = await getFoodItemPortionById(user, fi.food_item_portion_id)
+  if (!portion) {
+    throw new Error(`Portion not found: ${fi.food_item_portion_id}`)
```
fiddur, line 428: These three throws in `resolvePortionForInput` (`Portion not found`, `does not belong`, `portion_count must be …`) bubble out of `addMeal` / `updateMealById`, which are declared `Promise<MealResult>`. The router handler at `routes/meals-router.ts:130` only branches on `result.success === false` → a thrown Error turns into a 500 (or unhandled rejection) instead of a 400.

Worse: `addMeal` calls `dbUpsertMeal` before `syncFoodItemsToJunction`, so a request with a portion that targets a different food will persist an orphan meal row with no items before the throw fires. Consider either: (a) returning `{ success: false, error }` from the calling sites so the route can map to 400 cleanly, or (b) pre-validating all portion ids in `addMeal` / `updateMealById` before the meal row is written.

packages/api-spec/src/schemas/meals.ts
```diff
@@ -117,6 +125,14 @@ export const foodItemInputSchema = z
       description:
         'ID of a canonical food item (use the `id` returned by search_food_items). Preferred over `name` — binding by ID avoids creating per-user duplicates of items already in the shared library.',
     }),
+    food_item_portion_id: z.string().uuid().optional().meta({
+      description:
+        'Optional portion id (from food_item_portions). When set, nutrients are scaled by `portion_count × portion.base_equivalent / food.default_quantity` and the row\'s display label is derived from the portion. When omitted, falls back to the legacy `quantity` + `unit` scaling.',
+    }),
```
fiddur, line 131: The description says `portion_count` is required when `food_item_portion_id` is set, but the object isn't `.refine()`d to enforce it — so the only check is the runtime throw in `resolvePortionForInput`, which currently surfaces as a 500. Adding a refine here (mirroring `setSharedFoodItemOverrideBodySchema`'s pattern) would let `validateBody` reject the request with a 400 before the meal row is ever upserted.

apps/backend/src/services/meals.ts
```diff
@@ -499,31 +561,43 @@ export async function resnapshotMealsForFoodItem(
...
         }
-        for (const field of NUTRIENT_FIELD_NAMES) {
-          const v = link[field]
-          if (typeof v === 'number') passThrough[field] = v
+        rowsUpdated++
+        // Refetch the portion if the row was logged via the portion path.
+        // If it's been deleted since, fall back to legacy scaling — the
+        // recorded quantity/unit still encode the user's intent.
+        const portion = link.food_item_portion_id
+          ? await getFoodItemPortionById(user, link.food_item_portion_id)
```
fiddur: `getFoodItemPortionById` is called once per link inside the `links.map(async …)` — N portion fetches per meal during resnapshot. Since every link that needs a refetch points at the same food, a single batched fetch (or a small Map keyed by portion id) would collapse this to one query per meal. Not blocking, but trivial to do here.

apps/backend/src/services/meals.ts
```diff
@@ -296,19 +320,29 @@ const round2 = (n: number): number => Math.round(n * 100) / 100
...
 export const buildScaledJunctionItem = (
   fi: FoodItemInput,
   canonical: MergedFoodItem,
   sortOrder: number,
   effectiveValues: Record<string, number>,
+  portion: FoodItemPortionRow | null = null,
 ): Record<string, unknown> => {
-  const scale = computeScale(fi, canonical)
+  const scale = computeScale(fi, canonical, portion)
+  const usingPortion = portion !== null && typeof fi.portion_count === 'number' && fi.portion_count > 0
```
fiddur: Minor: `usingPortion` is also recomputed inside `computeScale` (line 301) with the same predicate. Consider computing it once in `buildScaledJunctionItem` and either passing it into `computeScale` or having `computeScale` return `{ scale, usingPortion }` so the two pieces of logic stay in sync.

packages/api-spec/src/schemas/food-items.ts
```diff
@@ -259,8 +267,12 @@ export const setSharedFoodItemOverrideBodySchema = z
       description:
         'Override icon for the central item. String sets the icon, null hides the central icon, omitted leaves the column unchanged.',
     }),
+    default_portion_id: z.string().uuid().nullable().optional().meta({
```
fiddur, line 275: The route handler for `PUT /food-items/:id/override` writes this field straight through `setSharedFoodItemOverride` with no check that the portion exists or belongs to the central food being overridden. Compare with `setDefaultPortion` for per-user items, which rejects unknown/foreign portions. A user can currently set their override's `default_portion_id` to any UUID — including a portion that belongs to a different food — and the inconsistency only surfaces later when they try to log a meal and `resolvePortionForInput` rejects it. Please add the same ownership validation in the override write path (skipping the check when `default_portion_id === null`).

fiddur (commented):
🛑 Changes required

Solid PR — the portion-based logging pipeline is cleanly factored: `prepareFoodItems` pre-validates every input before any DB write (eliminating the orphan-meal bug, with an integration test that pins the regression), and the dual-path `computeScale` keeps the legacy `quantity`/`unit` flow alongside the new portion flow. Snapshot of `quantity`/`unit` from the portion label gives a clean fallback if the portion is later deleted. Test coverage on the meals side is strong: scaling math, ownership rejection, missing `portion_count`, and resnapshot rescaling after a recipe edit are all covered.

Two issues worth addressing before merge:

1. **Missing ownership validation on `shared_food_item_overrides.default_portion_id`.** `PUT /food-items/:id/override` (and `set_shared_food_item_override`) accept any UUID as `default_portion_id` and write it straight through `setSharedFoodItemOverride` (`apps/backend/src/db/shared-food-item-overrides.ts` via `OVERRIDE_FIELDS`). The mirror endpoint for per-user items, `setDefaultPortion` (`apps/backend/src/services/food-item-portions.ts:82`), validates that the portion exists and `portion.food_item_id === foodItemId` before writing. The override path should do the same — otherwise the UI's `effective_default_portion_id` can resolve to a portion that belongs to a different food, and meal-log then fails with the unrelated-portion error from `resolvePortionForInput`. A `clear` (`default_portion_id: null`) should keep skipping the lookup.

2. **`PATCH /meals/:id` maps the new validation errors to 404.** `apps/backend/src/routes/meals-router.ts:154` returns `404` for every `updateMealById` failure. With this PR, that handler can now return errors like `"Portion ... does not belong to food item ..."` or `"portion_count must be a positive number ..."` from the new `prepareFoodItems` catch — these are 400-class, not 404. `POST /meals` correctly returns 400 (line 131). Recommend distinguishing the not-found case from validation errors on the update path (e.g. an error code, or a sentinel string the route can match).

Minor: the cast `link.portion_count as number | undefined` in `resnapshotMealsForFoodItem` (meals.ts ~line 663) is redundant — `MealFoodItemLink.portion_count` is already `number | undefined`.

apps/backend/src/services/meals.ts
```diff
@@ -437,6 +546,24 @@ export async function addMeal(user: string, input: AddMealInput): Promise<MealRe
...
+    input.food_items !== undefined && input.food_items !== null && input.food_items.length > 0
+  if (replacingWithItems) {
+    try {
+      // Source on the existing meal is needed for canonical resolution
+      // (auto-creates per-user items if not found). Look it up once.
+      const existing = await dbGetMealById(user, id)
+      if (!existing) return { error: 'Meal not found', success: false }
+      prepared = await prepareFoodItems(user, input.food_items!, existing.source)
+    } catch (err) {
+      return { error: err instanceof Error ? err.message : 'Invalid food items', success: false }
```
fiddur: `updateMealById` now returns `success: false` for portion validation errors ("Portion ... does not belong ...", "portion_count must be ..."), but the PATCH route in `meals-router.ts:154` maps every failure to HTTP 404. The same error from `addMeal` correctly becomes 400 (line 131). Consider distinguishing not-found from validation errors here (e.g. throw a typed/coded error or return an explicit kind) so the route can pick the right status.

apps/backend/src/services/meals.ts
```diff
@@ -345,26 +397,81 @@ const resolveCanonical = async (
...
+    )
+  }
+  if (typeof fi.portion_count !== 'number' || fi.portion_count <= 0) {
+    throw new Error('portion_count must be a positive number when food_item_portion_id is set')
+  }
+  return portion
+}
+
+/**
+ * Pre-resolve every food item + portion BEFORE any DB writes, so a bad
```
fiddur: Minor: this isn't strictly true. `prepareFoodItems` → `resolveCanonical` → `findOrCreate` can auto-create a per-user food item by name as a side effect, *before* `resolvePortionForInput` runs. So a request like `{ name: "BrandNewFood", food_item_portion_id: <bogus> }` will leave a freshly minted per-user food behind even though the meal row never lands. Self-healing on retry (the second call's `findOrCreate` finds the row), but the comment overstates the guarantee.

apps/backend/src/services/meals.ts
```diff
@@ -495,6 +641,15 @@ export async function resnapshotMealsForFoodItem(
   const canonical = detail.item
   const effective = getEffectiveNutrients(detail)
 
+  // Pre-fetch every current portion for this food once, instead of one
+  // round-trip per refreshed link. A link whose food_item_portion_id is
+  // absent from this map points at a portion that has since been deleted —
+  // we treat that as portion=null and fall back to legacy scaling against
```
fiddur: "Still encode the user's intent" is optimistic. When a portion is deleted, the link's `quantity` is `count × label_quantity` (e.g. 3) and `unit` is `label_unit` (e.g. `'ruta'`) — so the legacy `computeScale` sees a unit mismatch against the canonical `'g'` and returns `scale=1`, which then writes the full canonical 100g nutrient values to the row. Resnapshot makes the row *more* wrong than it was. Out-of-scope to fix here, but worth a follow-up — e.g. snapshot `base_equivalent` into the junction at write time so the row remains self-sufficient after portion deletion.

fiddur (commented):
✅ Approved

## Summary

PR2 of the food-portions feature plumbs portion-based meal logging end-to-end: the junction table gains `food_item_portion_id` + `portion_count`, `shared_food_item_overrides` gains a `default_portion_id` override, and the meals service grows a pre-validate / pure-write split so a bad portion id no longer leaves orphan meal rows behind. The API surface and MCP tool both gain matching schemas and an ownership guard, and `effective_default_portion_id` is surfaced on `FoodItemDetail` for the UI to preselect.

The design is clean — soft pointers on the junction, override-then-central fallthrough for the effective default, an `errorCode` discriminator that lets the route layer split 404/400 instead of papering everything over with 404. Test coverage is good: integration tests for the scale formula, the portion-ownership reject, the no-orphan-meal regression, and the resnapshot-rescale path; router tests for the 400/404 split and the override-PUT ownership guard.

A couple of minor observations (left inline) — they're not blockers:

- The "BEFORE any DB writes" comment on `prepareFoodItems` is *almost* true: `resolveCanonical → findOrCreate` can still auto-create a per-user food item by name before portion validation runs, so a bad portion id alongside a brand-new food name will leave the new food behind even though the meal row is gone. Idempotent on retry (name lookup finds it), but the comment is slightly aspirational.
- After a portion is hard-deleted, junction rows still reference it. Resnapshot then falls back to the legacy quantity/unit path, where `unit='ruta'` vs canonical `'g'` triggers `scale=1` and the row's nutrients get rewritten to the canonical 100g values — i.e. resnapshot makes the row *more* wrong, not less. Edge case (user has to delete a portion they previously logged with), and out-of-scope for this PR, but worth a follow-up: either snapshot `base_equivalent` into the junction at write time, or skip the resnapshot for orphaned-portion rows.

Verdict: approved.

fiddur (commented):
🛑 Changes required

Solid PR overall — clean separation of validate-then-write in `prepareFoodItems` / `writePreparedToJunction` fixes the orphan-meal regression cleanly, the resnapshot deleted-portion guard is well-reasoned, the 404/400 errorCode split is small and well-tested, and the schema-level refine on `food_item_portion_id` ⇔ `portion_count` catches the common misuse. Tests look comprehensive (4 new portion-meal integration tests + override round-trip + router-level error mapping + portion ownership guards on both REST and MCP).

**One blocking issue** (see inline): introducing `default_portion_id` to `setSharedFoodItemOverride` lets a user create an override row by sending `{default_portion_id: ...}` alone. That INSERT leaves `icon` at its column default (NULL), and the read-side in `applySharedOverrides` already treats `icon: null` as "explicit hide". Net effect: setting a default portion on a central food silently strips that food's central icon for the user. The DB-layer round-trip test you added (`default_portion_id override round-trips independently of icon`) only verifies the column round-trip, not the service-layer presentation, so the regression isn't caught.

The minimal fixes I can think of:
- distinguish "never set" from "explicit null" for icon (extra `icon_set` boolean column, or COALESCE to a sentinel), or
- in `applySharedOverrides`, only spread `icon` when it was meaningfully set (mirror what you did for `default_portion_id`: `...(override.icon !== undefined && override.icon !== null ? { icon: override.icon } : {})`) — but this changes the existing "null hides icon" semantics, so worth confirming.

A couple of smaller things worth a look but not blocking are in the inline comments.

apps/backend/src/services/food-items.ts
```diff
@@ -317,7 +317,14 @@ const applySharedOverrides = async (
   return items.map((item) => {
     const override = overrides.get(item.id)
     if (!override) return item
-    return { ...item, icon: override.icon ?? undefined }
+    // For default_portion_id, NULL on the override means "no override" — fall
+    // through to whatever the central row has (currently nothing). Only set
+    // the field when the user explicitly chose a portion.
+    return {
+      ...item,
+      icon: override.icon ?? undefined,
```
fiddur: Regression: a user who sets only `default_portion_id` (e.g. via `PUT /food-items/:id/override` with `{default_portion_id: "…"}`) gets an override row inserted with `icon = NULL` (the DB column has no default and the INSERT in `setSharedFoodItemOverride` only lists explicitly-provided columns). Here, `override.icon ?? undefined` then collapses that incidental NULL into `icon: undefined`, which spreads onto `...item` and **strips the central food's icon** from the response. Before PR2 this couldn't happen because the body refine required `icon` to be supplied. The existing "null icon = user explicitly hid the central icon" semantics now overlap with "user never touched icon", and there's no way for the read path to tell them apart. Either persist the distinction in the DB (`icon_set` flag, or a sentinel) or only spread `icon` when it was actually set by the user — mirroring the pattern you used right below for `default_portion_id`. Worth a service-level test asserting the central icon survives a `default_portion_id`-only override.

apps/backend/src/services/meals.ts
```diff
@@ -437,6 +569,28 @@ export async function addMeal(user: string, input: AddMealInput): Promise<MealRe
...
+  const hasFoodItemsUpdate = input.food_items !== undefined
+  const replacingWithItems =
+    input.food_items !== undefined && input.food_items !== null && input.food_items.length > 0
+  if (replacingWithItems) {
+    try {
+      // Source on the existing meal is needed for canonical resolution
+      // (auto-creates per-user items if not found). Look it up once.
+      const existing = await dbGetMealById(user, id)
+      if (!existing) return { error: 'Meal not found', errorCode: 'not_found', success: false }
+      prepared = await prepareFoodItems(user, input.food_items!, existing.source)
```
fiddur, line 588: Minor: this fetches the meal once here (for `existing.source`) and then `dbUpdateMeal` does it again under the hood at line 594 — and there's a small TOCTOU window between them. Not blocking, but if `dbUpdateMeal` already has the source on the row it touched, threading it back through would let you drop this read. Otherwise consider an explicit comment that the duplication is intentional.

apps/backend/src/routes/food-items-router.ts
```diff
@@ -482,6 +495,19 @@ export const createFoodItemsRouter = (authMiddleware: AnyMiddleware, centralDb:
           success: false,
         })
       }
+      // When the user sets default_portion_id, enforce the same ownership
+      // invariant as the per-user `set_default_food_item_portion` path:
+      // the portion must exist AND belong to this central food. A `null`
+      // clears the override and skips the check.
+      if (req.body.default_portion_id) {
```
fiddur, line 503: Nit: `if (req.body.default_portion_id)` treats `null` (clear) the same as `undefined` (leave-alone), which is what you want, but the error message returned at line 506 (`'default_portion_id does not belong to this food item'`) is also used when the portion id is simply unknown / doesn't resolve at all (`getFoodItemPortionById` returns null). That's misleading for that case — consider distinguishing 'portion not found' from 'portion targets a different food', since the user-facing fix is different.

fiddur (commented):
✅ Approved

✅Approved

Solid PR2 with carefully-considered edge cases and thorough test coverage.

**Strengths**

- **`icon_overridden` discriminator** (`shared_food_item_overrides`) is the right solution to the column-default-NULL ambiguity introduced by adding a second override field — without it, a `default_portion_id`-only override row would silently strip the central icon. The schema comment, db comment, migration backfill, and the regression test (`food-items.test.ts` line 681) all line up.
- **Orphan-meal regression fix** in `addMeal`: pre-validating portions in `prepareFoodItems` before the meal row is upserted closes a real footgun where a bad portion id used to leave an item-less meal behind. Covered by the integration test `rejects portion that does not belong to the food` which explicitly asserts `allMeals.length === 0`.
- **`errorCode` discriminator** (`not_found` vs `invalid`) cleanly splits 404 vs 400 in `PATCH /meals/:id`. The route maps unknown codes to 500, which is the safe default.
- **Resnapshot preserves snapshots for deleted portions** — the new `passThroughLink` correctly avoids rescaling rows whose `(quantity, unit)` are portion-label values that don't match the canonical default_unit (the old code would have fallen back to scale=1 and overwritten with full per-default-quantity nutrients). The comment in `meals.ts:693-699` captures the reasoning well, and the integration test makes it stick.
- **Pre-fetching portions once via `listPortionsForFoodItem`** in `resnapshotMealsForFoodItem` is a nice batch-read improvement over per-link round-trips.
- **Portion ownership guards** are mirrored consistently across REST (`PUT /food-items/:id/override`) and MCP (`set_shared_food_item_override`), with `null` correctly short-circuiting the lookup.

**Notes / non-blocking**

- `prepareFoodItems` acknowledges its own caveat in the docstring: `resolveCanonical` → `findOrCreate` can still side-effect a per-user food row by name BEFORE the portion check fails, so `{ name: "BrandNewFood", food_item_portion_id: <bogus> }` does leave a freshly-created food behind even though the meal is rejected. The meal-row guarantee (the load-bearing one) holds, and a retry sees the existing row via findOrCreate. Documented trade-off — fine to leave as-is.
- `setSharedFoodItemOverride(user, id, { default_portion_id: null })` against a non-existent row will insert a row whose every override field is NULL and `icon_overridden=FALSE`, i.e. effectively a no-op row that survives until DELETEd. Harmless (the read path passes through central for every field) but slightly noisy. Probably not worth adding a guard.
- `effective_default_portion_id` for central items relies on `applySharedOverrides` decorating `default_portion_id` from the user's override row; nothing verifies the referenced portion still exists. A deleted portion would leave the client with a dangling preselection id. Likely fine for PR2 if PR3 picks up portion deletion semantics; worth keeping in mind.
- Minor: `(detail.item.default_portion_id as string | undefined) ?? undefined` in `food-items-router.ts` is a redundant cast (the named field is already `string | undefined`), but it matches the existing style in `db/food-items.ts:39` to work around the broad index signature on `FoodItemEntity`. No change needed.

PR #782, opened 2026-05-27
feat(food-items): web UI portions editor on FoodItemDetail (PR3a/4)

fiddur:
## Summary

First web UI slice for the food-portions feature. After [PR1 #779](https://github.com/fiddur/aurboda/pull/779) (entity) and [PR2 #781](https://github.com/fiddur/aurboda/pull/781) (meal-logging integration), this PR lets users define and manage portions from the food-item editor.

### What's in PR3a

- New **Portions section** in `FoodItemDetail` (per-user foods only; shared/central foods get the override-layer treatment in a later PR).
- Each row: editable `label_quantity` / `label_unit` / `base_equivalent` (auto-save on blur, matching the page's existing commit pattern).
- Default-portion radio per row (calls `PUT /food-items/:id/default-portion`). "Base" radio clears (revert to base portion).
- Inline "Add portion" row.
- API hooks: `addFoodItemPortionApi`, `updateFoodItemPortionApi`, `deleteFoodItemPortionApi`, `setDefaultPortionApi` — thin axios wrappers in `state/api/meals.ts`.
- CSS with dark-mode styles.

### Not yet in PR3a (follow-ups)

- **PR3b** (next): meal-entry portion picker — after picking a food, show portions; on portion pick, replace quantity+unit with a `× count` input; preselect via `effective_default_portion_id`.
- **PR4**: drop legacy `meal_food_items.quantity`/`unit` columns once dual-writing is no longer needed.
- Override-layer default portion on central shared foods (small UI add).

Portions defined here are already usable via REST/MCP today (backend landed in PR2). The web meal entry still uses the legacy `quantity`/`unit` path until PR3b.

## Test plan

- [x] `pnpm check` clean (web)
- [x] All 405 web tests pass
- [ ] CI green
- [ ] Manual: open a food's editor, add "1 ruta = 3.4 g" on a 100 g chocolate, set as default, verify GET returns it with effective_default_portion_id

### DIFF
--- a/apps/web/src/pages/FoodItems/FoodItemDetail.tsx
+++ b/apps/web/src/pages/FoodItems/FoodItemDetail.tsx
@@ -1,6 +1,7 @@
 import {
   type FoodItemDetail as ApiFoodItemDetail,
   type FoodItemIngredient,
+  type FoodItemPortion,
   NUTRIENT_FIELDS,
   type NutrientFieldDef,
 } from '@aurboda/api-spec'
@@ -12,6 +13,12 @@ import { ConfirmButton } from '../../components/ConfirmButton'
 import { FoodItemAutocomplete } from '../../components/FoodItemAutocomplete'
 import { IconInput } from '../../components/IconInput'
 import { type IngredientRow, IngredientList } from '../../components/IngredientList'
+import {
+  addFoodItemPortionApi,
+  deleteFoodItemPortionApi,
+  setDefaultPortionApi,
+  updateFoodItemPortionApi,
+} from '../../state/api/meals'
 import { auth } from '../../state/auth'
 import { isEmoji, isIconPath, isUrl } from '../../utils/emojiLookup'
 import './FoodItemDetail.css'
@@ -470,6 +477,8 @@ export function FoodItemDetail() {
         commitDefaultUnit={commitDefaultUnit}
       />
 
+      {!isShared && <PortionsSection item={item} foodItemId={id} />}
+
       <CompositeOrAtomicSection
         item={item}
         isShared={isShared}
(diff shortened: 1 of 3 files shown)


apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -568,6 +577,237 @@ function DefaultMetaRow({
...
+      setError(null)
+      invalidate()
+    },
+  })
+
+  // Effective default is exposed by the server (resolves override layer for
+  // central items); for per-user items it mirrors the row column. UI only
+  // shows the radio when the food has portions to choose from.
+  const effectiveDefault = (item as ApiFoodItemDetail & { effective_default_portion_id?: string })
+    .effective_default_portion_id
```
fiddur: This cast is unnecessary — `effective_default_portion_id` is part of `FoodItemDetail` already (`packages/api-spec/src/schemas/food-items.ts:173`). Replace with `const effectiveDefault = item.effective_default_portion_id`. Project standard: "Avoid casting, use proper typing and type guards."

apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -568,6 +577,237 @@ function DefaultMetaRow({
...
+  // Local state lets the user edit without each keystroke racing to the API.
+  const [lq, setLq] = useState(String(portion.label_quantity))
+  const [lu, setLu] = useState(portion.label_unit)
+  const [be, setBe] = useState(String(portion.base_equivalent))
+
+  useEffect(() => {
+    setLq(String(portion.label_quantity))
+    setLu(portion.label_unit)
+    setBe(String(portion.base_equivalent))
+  }, [portion.id, portion.label_quantity, portion.label_unit, portion.base_equivalent])
```
fiddur: Edge case: if a user types into one field, then edits + blurs another field, the resulting mutation/invalidate refreshes `portion` props and this effect resets all three local states — silently dropping the unsaved input on the first field. Consider resetting only the field whose corresponding portion value actually changed (three smaller effects), or skipping the reset for fields that are currently focused/dirty.

apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -568,6 +577,237 @@ function DefaultMetaRow({
...
+      <input
+        type="number"
+        step="0.01"
+        value={be}
+        onInput={(e) => setBe((e.target as HTMLInputElement).value)}
+        onBlur={commit('base_equivalent')}
+      />
+      <span class="fi-portion-base-unit">{baseUnit}</span>
+      <button type="button" class="btn-link fi-portion-del" onClick={onDelete}>
+        Delete
```
fiddur: Other destructive actions in this file use `ConfirmButton` — a single misclick here deletes the portion without any confirmation. Worth using the same pattern for consistency.

fiddur (commented):
🛑 Changes required

Solid first cut at the portions editor — the per-row state + auto-save-on-blur pattern matches the rest of `FoodItemDetail`, validation reverts cleanly on bad input, and the wiring of the four mutations + invalidate is straightforward.

A couple of things to address before merging:

- **Unnecessary type assertion.** `effective_default_portion_id` is already on `ApiFoodItemDetail` (see `packages/api-spec/src/schemas/food-items.ts:173`), so the `item as ApiFoodItemDetail & { effective_default_portion_id?: string }` cast is dead weight and conflicts with the project rule "Avoid casting, use proper typing and type guards." Just read `item.effective_default_portion_id` directly.
- **Per-row `useEffect` clobbers unsaved sibling fields.** If the user types into `label_quantity` (no blur), then edits + blurs `label_unit`, the resulting mutation/invalidate refreshes the row's props; the effect's deps include `label_quantity` (unchanged) so most of the time it's safe, but if any of the three fields actually changes via that mutation the effect resets *all three* local states and silently drops the pending input on the other field. Either narrow each input's reset to its own dep, or gate the reset behind a "not currently focused/dirty" check.

Minor things worth considering (non-blocking):

- The Delete button bypasses the project's `ConfirmButton` pattern used elsewhere in this file — a stray click destroys a portion (and any meals referencing it must rely on the junction snapshot fallback). Consider matching the rest of the page.
- `PortionsSection` is hidden entirely for shared (central) items, but the backend exposes an override layer for them. Fine to defer (the PR title is PR3a/4), but worth a TODO so it isn't forgotten.
- All four mutations share a single `error` string — a delete failure will clear an add-form error message and vice versa. Probably fine for now, but the add-row error in particular is the one users most need persisted while they fix the inputs.

apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -568,6 +580,252 @@ function DefaultMetaRow({
...
+      </header>
+
+      <ul class="fi-portions-list">
+        <li class="fi-portion-row fi-portion-base">
+          <label class="fi-portion-default">
+            <input
+              type="radio"
+              name="default-portion"
+              checked={!effectiveDefault}
+              onChange={() => defaultMutation.mutate(null)}
```
fiddur: When `effectiveDefault` is already undefined this radio is already `checked`, but onChange will still fire if the user clicks it, sending a redundant `PUT /default-portion` with `portion_id: null`. Cheap to short-circuit: `onChange={() => effectiveDefault && defaultMutation.mutate(null)}`.

apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -568,6 +580,252 @@ function DefaultMetaRow({
...
+    }
+    const raw = field === 'label_quantity' ? lq : be
+    const parsed = parseFloat(raw)
+    if (!(parsed > 0)) {
+      // Revert; the schema rejects non-positive values.
+      if (field === 'label_quantity') setLq(String(portion.label_quantity))
+      else setBe(String(portion.base_equivalent))
+      return
+    }
+    if (parsed !== portion[field]) onUpdate({ [field]: parsed })
```
fiddur, line 794: Strict `!==` against a floating-point server value works for the common cases but is fragile in principle (e.g. `parseFloat('0.1') + 0` is fine, but any future arithmetic on `parsed` could drift). Not a blocker — just flagging in case portion math grows.

apps/web/src/state/api/meals.ts
```diff
@@ -175,6 +179,56 @@ export const previewMergeFoodItemsApi = async (
   return response.data.data
 }
 
+// ─── Food item portions ─────────────────────────────────────────────────────
+
+export const addFoodItemPortionApi = async (
+  foodItemId: string,
+  body: AddFoodItemPortionBody,
+): Promise<FoodItemPortion> => {
```
fiddur, line 187: On a non-2xx response axios rejects before this check runs, so `response.data.error` only ever surfaces for the 2xx-with-no-data case. Consistent with the rest of this file; just noting that errors from the backend's `{ error: '...' }` body actually come back via axios's rejection, not this branch.

fiddur (commented):
✅ Approved

Adds a Portions editor section on the per-user FoodItemDetail page plus four supporting API client wrappers in `state/api/meals.ts`. The implementation is self-contained, follows the existing FoodItemDetail commit-on-blur pattern, and the api-spec types/schemas it depends on are already in place.

### Strengths
- The three independent `useEffect`s in `PortionRow` to avoid clobbering sibling field edits during an invalidate is a deliberate, well-justified fix (and the inline comment explains it).
- Validation in `addMutation.mutationFn` mirrors the server zod schema (`positive()` / non-empty `label_unit`) and `commit()` reverts to the server value on invalid input rather than silently sending a bad PATCH.
- `addError` / `rowError` are split so an unrelated row failure doesn't blank out the message the user needs to see while fixing their draft.
- `ConfirmButton` is used for delete, matching the rest of FoodItemDetail.
- The deliberate scoping to per-user items only (`!isShared && <PortionsSection ...>`) with a TODO for the override path is the right call — keeps this PR focused.

### Minor observations (non-blocking)
- No frontend tests are added for `PortionsSection` / `PortionRow`. `CLAUDE.md` asks for unit tests "when reasonable"; the validation/commit logic in `commit()` and the add-row validator would be cheap to cover. Worth doing in a follow-up if not here.
- The other food-item mutations in this file invalidate `['foodItem', id]` *and* `['foodItems']` (the list query) — the new `invalidate()` helper in `PortionsSection` only hits the detail key. If the list view ever surfaces portion-derived info (effective default, portion count), it'd go stale until the next refetch. Acceptable for now since the list doesn't show portions, but worth a comment.
- `parseFloat` is locale-insensitive — Swedish-locale users typing `1,5` will get `1` silently. Existing site pattern, so not a blocker, but the wider site eventually needs a decimal-parse helper.
- Update/delete/default mutations don't disable their controls while `isPending`; double-clicks during slow networks can fire duplicate requests. The backend is idempotent enough that this is a UX nit, not a correctness bug.

fiddur (commented):
✅ Approved

Solid, focused PR3a slice. The portions editor is well-scoped: it is gated by `!isShared` (with an explicit TODO for the central/override layer), uses `useMutation` consistently, separates add-row vs. row-level errors (good touch — keeps draft-input feedback alive while other failures happen), and the per-field `useEffect`s in `PortionRow` correctly avoid clobbering unsaved edits when a sibling field's invalidate round-trips. API helpers in `meals.ts` follow the existing pattern (axios + typed response + token header + sane error fallbacks).

A few minor things worth noting (not blocking):

- **Default-portion radio symmetry**: the Base radio short-circuits a redundant PUT when already default (commit 7fb322b), but the per-portion `<input type="radio" onChange={onSetDefault}>` in `PortionRow` does not guard on `isDefault`. With a controlled radio onChange does not normally fire when state is unchanged, but if the Base case warranted the guard, the per-portion case is the same shape — consider making them consistent.
- **Help-text fallback**: when `default_unit` is undefined, `baseUnit` falls back to the literal string `'base unit'`, which makes the help sentence read "…measured in base unit". A food item without a default unit is unusual but possible; tightening the fallback (or hiding the section) would be nicer.
- **Add error message**: `'label_quantity, label_unit, and base_equivalent are all required and positive'` surfaces api-spec field names to end users — fine for now, but worth replacing with user-language if this section gets polish.
- **Disabled state during pending**: only the Add button disables on `isPending`. Row inputs/radios stay live during update/delete/default-set mutations; the commit-on-blur diff check makes this safe, but rapid clicks on the default radio can fire overlapping PUTs.

No tests for the new component, which matches the rest of `FoodItemDetail.tsx` (no existing component tests for it), so I'm not flagging it. PR3a as defined looks ready.

PR #783, opened 2026-05-27
feat(meals): meal-entry portion picker (PR4/4)

fiddur:
## Summary

Final web-UI slice of the food-portions feature. PR3 ([#782](https://github.com/fiddur/aurboda/pull/782)) added the portions editor on `FoodItemDetail`; this PR wires the meal-entry row to use those portions when logging.

### What's in PR4

- **Portion picker** in `MealDetail`'s food row: after picking a food, a `<select>` shows "Base + each defined portion". Hidden when the food has no portions (legacy quantity/unit path stays).
- **Quantity input switching**: when a portion is selected, the row swaps to a single "Count × {portion label}" input. When Base, the existing quantity + unit pair is shown.
- **Auto-pin to effective default**: once the food's detail arrives, the row auto-selects `effective_default_portion_id` once per food id change. Existing meals being edited keep their pinned portion (the effect skips when already pinned).
- **Live nutrient display** uses the same portion formula as the backend: `nutrient × count × base_equivalent / canonical.default_quantity`.
- **Save payload** sends `{ food_item_portion_id, portion_count }` for pinned rows and the legacy `{ quantity, unit }` for base rows.

### Plumbing

- New `fetchFoodItemDetailApi` in `state/api/meals.ts`, cached by react-query (keyed on `food_item_id` — same key the FoodItemDetail page uses, so the cache is shared).
- `mealItems.ts`:
  - `FoodItemEdit` gains `food_item_portion_id`, `portion_count`, and a `portion` snapshot.
  - `editItemsToBody` switches payload shape based on whether the row is pinned.
  - `mealItemsToEdit` round-trips the new fields.
  - `scaleNutrient` accepts an optional `portionScale` argument.
- `FoodItemRow` extracted into `PortionPicker`, `QuantityInputs`, and a `useAutoDefaultPortion` hook to stay under the project's cyclomatic-complexity limit.
- CSS for the picker + portion-unit label, dark-mode styles.

### Backwards compat

- Foods without portions: UI unchanged (no picker shown).
- Existing meals logged via the legacy path: render with the quantity+unit inputs they already had.
- Save payload is one of two shapes per row; the backend (from PR2) accepts both.

## Test plan

- [x] `pnpm check` clean (web)
- [x] All 405 web tests pass
- [ ] CI green
- [ ] Manual: log a meal of "3 ruta" of a chocolate that defines a "ruta" portion; verify count input + live kcal scaling; verify on reload the row stays pinned to the portion.

### DIFF
--- a/apps/web/src/pages/Meals/MealDetail.tsx
+++ b/apps/web/src/pages/Meals/MealDetail.tsx
@@ -657,16 +657,22 @@ export function MealDetail() {
 
   const commitItems = (next: FoodItemEdit[]) => {
     setItems(next)
-    // Filter rows that are mid-edit on a portion (pinned to a portion but
-    // the count input is currently empty/zero). Without this, such a row
-    // would fall through editItemsToBody's portion fork and briefly get
-    // re-saved as a legacy-quantity row, then re-pinned when the count
-    // refills. We skip the offending row only (not the whole save), so
-    // edits to other rows still flush.
-    const saveable = next.filter(
-      (fi) => !(fi.food_item_portion_id && !(typeof fi.portion_count === 'number' && fi.portion_count > 0)),
-    )
-    const body = { food_items: editItemsToBody(saveable) }
+    // Skip the entire save when any row is in a transient portion-editing
+    // state (pinned to a portion but the count input is currently
+    // empty/zero). Filtering the offending row out of the body would let
+    // updateMealById's wholesale replace semantic silently delete the row
+    // server-side until the count refills — and `flush()` on unmount would
+    // commit the row-less body, losing the row entirely. Skipping the
+    // schedule means other-row edits wait for the offending row to be
+    // completed; preferable to silent server-side data loss.
+    if (
+      next.some(
+        (fi) => fi.food_item_portion_id && !(typeof fi.portion_count === 'number' && fi.portion_count > 0),
+      )
+    ) {
+      return
+    }
+    const body = { food_items: editItemsToBody(next) }
     // De-dupe consecutive identical bodies — useAutoDefaultPortion's
     // backfill on load propagates through here with the same
     // food_item_portion_id + portion_count the server already has, and
--- a/apps/web/src/pages/Meals/MealDetail.tsx
+++ b/apps/web/src/pages/Meals/MealDetail.tsx
@@ -249,6 +249,14 @@ const portionToSnapshot = (p: FoodItemPortion) => ({
   base_equivalent: p.base_equivalent,
 })
 
+const MACRO_LABELS: Record<'calories' | 'protein' | 'carbs' | 'fat' | 'fiber', (v: number) => string> = {
+  calories: (v) => `${v} kcal`,
+  protein: (v) => `P ${v}g`,
+  carbs: (v) => `C ${v}g`,
+  fat: (v) => `F ${v}g`,
+  fiber: (v) => `Fib ${v}g`,
+}
+
 function FoodItemRow({
   item,
   index,
@@ -277,6 +285,25 @@ function FoodItemRow({
       food_item_portion_id: p.id,
       portion_count: keepCount ? (item.portion_count ?? 1) : 1,
       portion: portionToSnapshot(p),
+      // For existing portion-pinned rows loaded via mealItemsToEdit, `ref`
+      // holds the per-entry snapshot (ref.calories = already-scaled, ref.quantity
+      // = portion_count × label_quantity), not the canonical (per default_quantity)
+      // values that scaleNutrient's portion-path formula requires. Refresh ref
+      // from the food's detail so the formula
+      //   nutrient × count × base_equivalent / ref.quantity
+      // evaluates correctly. (For freshly-picked foods, ref is already canonical
+      // from FoodItemEntity, and keepCount=false, so this branch is harmless.)
+      ref:
+        keepCount && detail
+          ? {
+              calories: detail.calories as number | undefined,
+              carbs: detail.carbs as number | undefined,
+              fat: detail.fat as number | undefined,
+              fiber: detail.fiber as number | undefined,
+              protein: detail.protein as number | undefined,
+              quantity: detail.default_quantity,
+            }
+          : item.ref,
     }),
   )
 
@@ -285,13 +312,6 @@ function FoodItemRow({
       ? { count: item.portion_count, base_equivalent: item.portion.base_equivalent }
       : undefined
 
-  const MACRO_LABELS: Record<'calories' | 'protein' | 'carbs' | 'fat' | 'fiber', (v: number) => string> = {
-    calories: (v) => `${v} kcal`,
-    protein: (v) => `P ${v}g`,
-    carbs: (v) => `C ${v}g`,
-    fat: (v) => `F ${v}g`,
-    fiber: (v) => `Fib ${v}g`,
-  }
   const macros = (['calories', 'protein', 'carbs', 'fat', 'fiber'] as const).map((f) => ({
     field: f,
     value: scaleNutrient(item.ref, f, item.quantity, portionScale),
--- a/apps/web/src/pages/Meals/MealDetail.tsx
+++ b/apps/web/src/pages/Meals/MealDetail.tsx
@@ -541,6 +541,9 @@ export function MealDetail() {
   const itemsFlusherRef = useRef(
     createDebouncedFlusher<UpdateMealBody>(FOOD_ITEM_DEBOUNCE_MS, (body) => saveRef.current(body)),
   )
+  // Serialized last-saved food_items body — used by commitItems to de-dupe
+  // identical-payload schedules so on-load backfills don't trigger writes.
+  const lastItemsBodyRef = useRef<string | null>(null)
   const mealTypeFlusherRef = useRef(
     createDebouncedFlusher<string>(FOOD_ITEM_DEBOUNCE_MS, (v) => saveRef.current({ meal_type: v })),
   )
@@ -573,7 +576,11 @@ export function MealDetail() {
     // explicitly opted out of a slot, so treat it as a custom "other" rather
     // than silently defaulting to lunch.
     setMealType(meal.meal_type ?? DEFAULT_CUSTOM_TYPE)
-    setItems(mealItemsToEdit(meal.food_items))
+    const initialItems = mealItemsToEdit(meal.food_items)
+    setItems(initialItems)
+    // Seed the de-dupe ref so the first commitItems call (typically from
+    // useAutoDefaultPortion's backfill) doesn't fire a no-op write.
+    lastItemsBodyRef.current = JSON.stringify({ food_items: editItemsToBody(initialItems) })
     setFlags(meal.sensitivities ?? [])
     setMacros({
       calories: meal.calories,
@@ -632,7 +639,23 @@ export function MealDetail() {
 
   const commitItems = (next: FoodItemEdit[]) => {
     setItems(next)
-    itemsFlusherRef.current.schedule({ food_items: editItemsToBody(next) })
+    // Skip the save when any row is in a transient portion-editing state
+    // (pinned to a portion but the count input is currently empty). Without
+    // this guard the row would fall through editItemsToBody's portion fork
+    // and briefly get re-saved as a legacy-quantity row, then re-pinned
+    // once the user finishes typing the new count.
+    if (next.some((fi) => fi.food_item_portion_id && !(typeof fi.portion_count === 'number' && fi.portion_count > 0))) {
+      return
+    }
+    const body = { food_items: editItemsToBody(next) }
+    // De-dupe consecutive identical bodies — useAutoDefaultPortion's
+    // backfill on load propagates through here with the same
+    // food_item_portion_id + portion_count the server already has, and
+    // there's no reason to round-trip a no-op write per detail-page open.
+    const serialized = JSON.stringify(body)
+    if (serialized === lastItemsBodyRef.current) return
+    lastItemsBodyRef.current = serialized
+    itemsFlusherRef.current.schedule(body)
   }
 
   const commitName = () => {
--- a/apps/web/src/pages/Meals/MealDetail.tsx
+++ b/apps/web/src/pages/Meals/MealDetail.tsx
@@ -1,4 +1,4 @@
-import type { UpdateMealBody } from '@aurboda/api-spec'
+import type { FoodItemDetail as ApiFoodItemDetail, FoodItemPortion, UpdateMealBody } from '@aurboda/api-spec'
 
 import { NUTRIENT_FIELDS } from '@aurboda/api-spec'
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
@@ -11,6 +11,7 @@ import { FoodItemAutocomplete } from '../../components/FoodItemAutocomplete'
 import {
   addFoodItemApi,
   deleteMealApi,
+  fetchFoodItemDetailApi,
   fetchMeal,
   fetchSensitivityFlags,
   type FoodItemEntity,
@@ -127,6 +128,110 @@ function MealFlagsEditor({
   )
 }
 
+const parseNum = (v: string) => (v === '' ? undefined : parseFloat(v))
+
+/**
+ * After a food is picked, auto-select its effective default portion exactly
+ * once per food id change. Skips when the row is already pinned (e.g.
+ * editing an existing meal where the portion was already chosen).
+ */
+const useAutoDefaultPortion = (
+  item: FoodItemEdit,
+  detail: ApiFoodItemDetail | undefined,
+  applyPortion: (p: FoodItemPortion) => void,
+): void => {
+  const lastAppliedFoodId = useRef<string | undefined>(undefined)
+  useEffect(() => {
+    if (!detail || !item.food_item_id) return
+    if (lastAppliedFoodId.current === item.food_item_id) return
+    lastAppliedFoodId.current = item.food_item_id
+    if (item.food_item_portion_id) return
+    const def = detail.effective_default_portion_id
+    if (!def) return
+    const p = detail.portions?.find((pp) => pp.id === def)
+    if (p) applyPortion(p)
+  }, [detail?.id, item.food_item_id])
+}
+
+function PortionPicker({
+  selectedId,
+  portions,
+  baseLabel,
+  onPick,
+}: {
+  selectedId: string | undefined
+  portions: FoodItemPortion[]
+  baseLabel: string
+  onPick: (portionId: string) => void
+}) {
+  if (portions.length === 0) return null
+  return (
+    <select
+      class="food-portion-picker"
+      value={selectedId ?? ''}
+      onChange={(e) => onPick((e.target as HTMLSelectElement).value)}
+    >
+      <option value="">Base ({baseLabel})</option>
+      {portions.map((p) => (
+        <option key={p.id} value={p.id}>
+          {p.label_quantity} {p.label_unit}
+        </option>
+      ))}
+    </select>
+  )
+}
+
+function QuantityInputs({
+  item,
+  onUpdate,
+}: {
+  item: FoodItemEdit
+  onUpdate: (patch: Partial<FoodItemEdit>) => void
+}) {
+  if (item.food_item_portion_id && item.portion) {
+    return (
+      <>
+        <input
+          type="number"
+          step="0.1"
+          value={item.portion_count ?? ''}
+          placeholder="Count"
+          class="food-num-input"
+          onInput={(e) => onUpdate({ portion_count: parseNum((e.target as HTMLInputElement).value) })}
+        />
+        <span class="food-portion-unit">
+          × {item.portion.label_quantity} {item.portion.label_unit}
+        </span>
+      </>
+    )
+  }
+  return (
+    <>
+      <input
+        type="number"
+        step="0.1"
+        value={item.quantity ?? ''}
+        placeholder="Qty"
+        class="food-num-input"
+        onInput={(e) => onUpdate({ quantity: parseNum((e.target as HTMLInputElement).value) })}
+      />
+      <input
+        type="text"
+        value={item.unit ?? ''}
+        placeholder="Unit"
+        class="food-unit-input"
+        onInput={(e) => onUpdate({ unit: (e.target as HTMLInputElement).value })}
+      />
+    </>
+  )
+}
+
+const portionToSnapshot = (p: FoodItemPortion) => ({
+  label_quantity: p.label_quantity,
+  label_unit: p.label_unit,
+  base_equivalent: p.base_equivalent,
+})
+
 function FoodItemRow({
   item,
   index,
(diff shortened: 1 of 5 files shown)


apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -127,6 +128,110 @@ function MealFlagsEditor({
...
+  item: FoodItemEdit,
+  detail: ApiFoodItemDetail | undefined,
+  applyPortion: (p: FoodItemPortion) => void,
+): void => {
+  const lastAppliedFoodId = useRef<string | undefined>(undefined)
+  useEffect(() => {
+    if (!detail || !item.food_item_id) return
+    if (lastAppliedFoodId.current === item.food_item_id) return
+    lastAppliedFoodId.current = item.food_item_id
+    if (item.food_item_portion_id) return
```
fiddur: This early-return is what breaks editing an existing meal whose row already has `food_item_portion_id` set. `mealItemsToEdit` doesn't (and can't — the meal response has no `label_quantity`/`label_unit`/`base_equivalent`) populate the `portion` snapshot, so reopening such a meal leaves `item.portion` undefined permanently:

- `QuantityInputs` falls through to the legacy Qty/Unit inputs (it requires `food_item_portion_id && portion`), but anything typed there is dropped on save because `editItemsToBody` still picks the portion path.
- `scaleNutrient`'s `portionScale` is undefined, so macros are scaled by the legacy formula (and the stored `quantity` for a portion-saved row is typically null, so values come out unscaled).

Consider dropping this guard and instead branching: if `item.food_item_portion_id` is already set, backfill `portion` from `detail.portions.find(p => p.id === item.food_item_portion_id)` (without touching `portion_count`); otherwise apply the effective default as today.

apps/web/src/pages/Meals/mealItems.ts
```diff
@@ -39,6 +60,8 @@ export const mealItemsToEdit = (
 ): FoodItemEdit[] =>
   (items ?? []).map((fi) => ({
     food_item_id: fi.food_item_id,
+    food_item_portion_id: fi.food_item_portion_id,
+    portion_count: fi.portion_count,
```
fiddur, line 74: We carry `food_item_portion_id` + `portion_count` over from the server, but the meal response has no portion snapshot fields, so `portion` (label_quantity/label_unit/base_equivalent) stays undefined on the edit row. Combined with the guard in `useAutoDefaultPortion`, this means existing portion-based rows never end up with a valid `portion` snapshot and render via the legacy path. Either the snapshot needs to be backfilled from the food-item detail query in the row component, or the meal response/loader needs to surface the portion's display fields.

fiddur (commented):
🛑 Changes required

Nice refactor of the row (extracted PortionPicker/QuantityInputs, single-patch `update`, the auto-default-portion hook), and the schema/payload split between portion-path and legacy-path is clean. But the load path for **existing** meal items with a portion is broken:

- `mealItemsToEdit` carries over `food_item_portion_id` + `portion_count` but never populates the `portion` snapshot (`label_quantity`, `label_unit`, `base_equivalent`) — the meal response doesn't include those fields.
- `useAutoDefaultPortion` short-circuits as soon as `item.food_item_portion_id` is set, so it never backfills the snapshot from `detail.portions` either.

Consequences when reopening a meal whose row uses a portion:
  1. `QuantityInputs` falls through to the legacy `Qty`/`Unit` inputs (the `item.food_item_portion_id && item.portion` guard is false). The user sees free-form qty/unit fields, but anything they type into them is silently dropped on save — `editItemsToBody` still emits the portion path because `food_item_portion_id` + `portion_count` are both present.
  2. `scaleNutrient` falls back to the legacy `nutrient × currentQuantity / ref.quantity` path. For portion-only saves the stored `quantity` is typically null/undefined, so the displayed macros end up unscaled (canonical values) or undefined.
  3. The `PortionPicker` itself displays the right selection — so the UI looks half-correct, which makes the bug easy to miss in casual testing but very confusing when actually editing.

The fix is to backfill `portion` from `detail.portions` when `food_item_portion_id` is set but `portion` is missing. Most natural place is inside the row's `useEffect`-style hook (lift the early-return so it can also backfill the snapshot for an already-pinned portion), or as a separate effect that watches `detail` + `item.food_item_portion_id`.

Minor: the inline 5-deep ternary at the macros render (`MealDetail.tsx` ~344–352) is a regression in readability vs. the previous five-line block — a small `{calories: 'kcal', protein: 'P', ...}` label map would be clearer. Not blocking.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -127,6 +128,127 @@ function MealFlagsEditor({
...
+    if (lastAppliedFoodId.current === item.food_item_id) return
+    lastAppliedFoodId.current = item.food_item_id
+    if (item.food_item_portion_id) {
+      // Already pinned. Backfill the snapshot if we're missing it (loading
+      // an existing meal). If the portion id no longer resolves (deleted
+      // since logging), leave the row alone — the user will see legacy
+      // qty/unit and can re-pick.
+      if (item.portion) return
+      const p = detail.portions?.find((pp) => pp.id === item.food_item_portion_id)
+      if (p) applyPortion(p, /* keepCount */ true)
```
fiddur, line 163: Backfilling the `portion` snapshot alone causes the inline macros to display wrong for existing portion-pinned meal rows. `mealItemsToEdit` captures `ref.calories = scaled_calories` and `ref.quantity = portion_count × label_quantity` from the meal-entry snapshot, but `scaleNutrient`'s portion path expects `ref` to be canonical (`per default_quantity` nutrient over `default_quantity`). After backfill the formula collapses to `scaled_C × (base_equivalent / label_quantity)` — e.g. a `1 ruta = 3.4 g` portion will show 3.4× the actual kcal/P/C/F/Fib. Since `detail` already has `default_quantity` and the canonical nutrient values, refresh `ref` here as well (e.g. `update({ ..., ref: { calories: detail.calories, protein: detail.protein, …, quantity: detail.default_quantity } })`) so the portion-path math works for loaded rows too.

apps/web/src/pages/Meals/mealItems.ts
```diff
@@ -66,23 +89,49 @@ export const mealItemsToEdit = (
...
   currentQuantity: number | undefined,
+  portionScale?: { count: number; base_equivalent: number },
 ): number | undefined => {
   if (!ref) return undefined
   const baseValue = ref[field]
   if (typeof baseValue !== 'number') return undefined
   if (ref.quantity === undefined || ref.quantity === 0) return baseValue
+  if (portionScale) {
+    return (
+      Math.round(((baseValue * portionScale.count * portionScale.base_equivalent) / ref.quantity) * 10) / 10
```
fiddur, line 142: This formula is only correct when `ref` carries canonical values (per-`default_quantity` nutrient + `quantity = default_quantity`). For rows produced by `mealItemsToEdit` from an existing portion-pinned meal, `ref` instead holds the per-entry snapshot (`ref.calories = scaled_C`, `ref.quantity = portion_count × label_quantity`), so this evaluates to `scaled_C × (base_equivalent / label_quantity)` — wrong by `base_equivalent/label_quantity`. Either normalize `ref` at load time (the auto-default effect already has `detail` available) or document and enforce that callers must pass a canonical `ref` when supplying `portionScale`.

apps/web/src/pages/Meals/mealItems.ts
fiddur, line 74: Worth adding test coverage for the portion fields here and in `editItemsToBody`'s portion branch — `mealItems.test.ts` only covers the legacy path. A round-trip test (`mealItemsToEdit` then `scaleNutrient` with portion params) would have caught the wrong-scale display bug on load.

fiddur (commented):
🛑 Changes required

Nice extraction of `PortionPicker` / `QuantityInputs` and the `useAutoDefaultPortion` hook — the flow reads clearly and `handleFoodPick` correctly resets all the portion fields so a new food can re-apply its default.

One blocking correctness issue: when an existing portion-pinned meal row is loaded, the inline macros displayed under the row come out wrong by a factor of `portion.base_equivalent / portion.label_quantity`.

The meal response serializes a portion-pinned junction row with `quantity = portion_count × label_quantity` and `calories` (and other nutrient columns) as the already-scaled snapshot value (see `apps/backend/src/services/meals.ts` ~L359 / `linksToFoodItems`). `mealItemsToEdit` then captures `ref.calories = fi.calories` (already scaled) and `ref.quantity = fi.quantity` (= `count × label_quantity`). The new commit 0675c68 backfills `item.portion` so the row now renders via the portion path in `scaleNutrient`:

```
baseValue × count × base_equivalent / ref.quantity
= scaled_C × count × BE / (count × label_quantity)
= scaled_C × (BE / label_quantity)
```

For a "1 ruta = 3.4 g" portion, the displayed kcal/P/C/F/Fib will be 3.4× the actual value as soon as the user opens the meal. (Before this commit the row fell through to the legacy path and displayed correctly, just with awkward qty+unit inputs.)

The portion-path formula assumes `ref` is canonical (`ref.calories = per-default_quantity nutrient`, `ref.quantity = default_quantity`). For freshly-picked foods (`handleFoodPick`) that's true, since `ref` comes straight from `FoodItemEntity`. For loaded existing meals it isn't. `useAutoDefaultPortion` already has `detail` in hand and so could refresh `ref` from the canonical values (`detail.calories`, …, `detail.default_quantity`) at the same time as it backfills the snapshot.

Secondary notes (non-blocking):
- The auto-default effect fires on every load of an existing portion-pinned meal and triggers a debounced PATCH carrying the same `food_item_portion_id`/`portion_count` payload — semantically a no-op but a wasted round trip per row per visit. Worth considering a same-payload guard in `commitItems` or only scheduling when the resulting body actually differs.
- `mealItems.test.ts` already covers the legacy path well; the new portion branches (`editItemsToBody` portion fork, `scaleNutrient` portion path, `mealItemsToEdit` portion mapping) have no tests. The bug above would have been caught by a single test asserting `scaleNutrient(ref, 'calories', undefined, { count, base_equivalent })` for a backfilled-existing-meal ref.
- `MACRO_LABELS` is re-allocated on every `FoodItemRow` render; trivial to lift to module scope alongside `portionToSnapshot`.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -140,68 +270,125 @@ function FoodItemRow({
...
+      // = portion_count × label_quantity), not the canonical (per default_quantity)
+      // values that scaleNutrient's portion-path formula requires. Refresh ref
+      // from the food's detail so the formula
+      //   nutrient × count × base_equivalent / ref.quantity
+      // evaluates correctly. (For freshly-picked foods, ref is already canonical
+      // from FoodItemEntity, and keepCount=false, so this branch is harmless.)
+      ref:
+        keepCount && detail
+          ? {
+              calories: detail.calories as number | undefined,
```
fiddur: Nit: these `as number | undefined` casts are no-ops — `detail.calories` etc. are already typed `number | undefined` via `nutrientFieldsSchema`. CLAUDE.md asks to avoid casts; you can just write `calories: detail.calories,` (same for the four below).

fiddur (commented):
✅ Approved

Solid finisher for the portions series. The portion picker / count UI cleanly toggles between the portion-path and legacy quantity+unit-path rendering, and the autosave keeps writing the correct shape on each path. The two trickiest spots — backfilling the `portion` snapshot when loading an existing portion-pinned meal row, and refreshing `ref` to the canonical (per `default_quantity`) values so the live-scaling formula matches the backend — are well-commented and have targeted unit tests in `mealItems.test.ts`. The new portion round-trip tests and the `scaleNutrient` portion-path tests pin the contract nicely.

A few smaller observations, none blocking:

- The `as number | undefined` casts on `detail.calories`/`carbs`/`fat`/`fiber`/`protein` (MealDetail.tsx ~lines 299-303) are no-ops — the api-spec types are already `number | undefined` from `nutrientFieldsSchema`. CLAUDE.md asks to avoid casts; these can just be dropped.
- `useAutoDefaultPortion`'s effect deps `[detail?.id, item.food_item_id]` omit `item.food_item_portion_id`/`item.portion`, but the `lastAppliedFoodId` ref guards against unwanted re-runs and the comments explain the intent. One subtle consequence: if the cached detail is invalidated and re-fetched after the user has interacted (e.g., the food's `effective_default_portion_id` changes server-side), the default won't re-apply on the same row. Acceptable, but worth keeping in mind if defaults ever start changing mid-session.
- Minor: when the count input is momentarily cleared (`portion_count` → undefined) while autosave fires, `editItemsToBody` falls through to the legacy path and sends `quantity`/`unit` from the row, which can flip the row's persisted shape briefly until the user finishes typing. Probably fine for autosave, but if it ever shows up in history/audit logs it's worth a debounce gate.

apps/web/src/pages/Meals/mealItems.ts
```diff
@@ -66,23 +89,49 @@ export const mealItemsToEdit = (
...
   items
     .filter((fi) => fi.food_item_id && fi.name.trim())
-    .map((fi) => ({
-      food_item_id: fi.food_item_id,
-      name: fi.name,
-      quantity: fi.quantity,
-      unit: fi.unit,
-    }))
+    .map((fi) =>
+      fi.food_item_portion_id && typeof fi.portion_count === 'number'
```
fiddur, line 103: When the user clears the count input mid-edit, `portion_count` is `undefined` while `food_item_portion_id` is still set, so this falls through to the legacy branch and emits a payload without `food_item_portion_id`. After the 600 ms debounce the meal briefly gets re-saved as a legacy-quantity row, then re-pinned once the user finishes typing. Consider either keeping the previous valid `portion_count` until a new positive number arrives, or skipping the autosave when `food_item_portion_id` is set but `portion_count` is missing — the row is in a transient state, not a deliberate mode switch.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -140,68 +270,125 @@ function FoodItemRow({
...
+    queryKey: ['foodItem', item.food_item_id],
+  })
 
-  const kcal = scaleNutrient(item.ref, 'calories', item.quantity)
-  const prot = scaleNutrient(item.ref, 'protein', item.quantity)
-  const carbs = scaleNutrient(item.ref, 'carbs', item.quantity)
-  const fat = scaleNutrient(item.ref, 'fat', item.quantity)
-  const fiber = scaleNutrient(item.ref, 'fiber', item.quantity)
-  const hasMacros = [kcal, prot, carbs, fat, fiber].some((v) => typeof v === 'number')
+  useAutoDefaultPortion(item, detail, (p, keepCount) =>
```
fiddur, line 288: Heads-up: on load of an existing portion-pinned meal this `applyPortion(keepCount=true)` propagates through `update` → `onChange` → `commitItems`, which schedules a debounced PUT 600 ms later. The payload is identical to what's already on the server (only the local `portion` snapshot/`ref` change), so it's a no-op write but still a network round-trip on every detail-page open. Could be sidestepped by applying the snapshot/ref refresh through a setter that doesn't go through `commitItems`.

fiddur (commented):
✅ Approved

Solid PR4/4 — the meal-entry portion picker hangs together cleanly. The `useAutoDefaultPortion` effect correctly handles both the "freshly picked food → auto-apply default portion" and the "loading an existing portion-pinned row → backfill the snapshot" paths, and the inline WHY comments around the `ref` refresh are exactly the kind that pay off later. Splitting `QuantityInputs` / `PortionPicker` / `MACRO_LABELS` out of `FoodItemRow` keeps the render readable. The new `scaleNutrient` portion path mirrors the backend formula and is well-tested, including the regression that motivated the canonical-ref refresh.

Two small things to consider (not blocking):

- **Transient mode-flop while editing `portion_count`.** When the user clears the count input to retype it, the row keeps `food_item_portion_id` but `portion_count` becomes `undefined`. `editItemsToBody` then drops into the legacy branch and emits `{ food_item_id, name, quantity, unit }` with no `food_item_portion_id`. After the 600 ms debounce that briefly persists the meal as a legacy-quantity row before the next keystroke re-pins it. Worth either keeping the previous `portion_count` until a valid new one arrives, or suppressing the save when the row is in an obviously transient state (portion id set, count empty).
- **Idempotent autosave on load for portion-pinned rows.** When a portion-pinned meal loads, `useAutoDefaultPortion` calls `applyPortion(keepCount=true)`, which goes through `update` → `onChange` → `commitItems` → `itemsFlusherRef.current.schedule(...)`. The payload is byte-identical to what's on the server (only the `portion` snapshot and `ref` change locally), so a redundant PUT fires 600 ms after every open. Cheap, but easy to skip — e.g. apply the snapshot/ref refresh via a direct setter that doesn't go through `commitItems`.

Also a couple of stylistic nits worth knowing about but not worth churning over: the parse helpers (`parseNum`) still surface `NaN` for non-numeric input — same pattern that already exists elsewhere, just more places now where a NaN `portion_count` would be sent and rejected by the `.positive()` schema. And `useAutoDefaultPortion`'s effect closes over `item`/`detail` without listing them in the dep array — fine under oxlint, but worth re-checking if exhaustive-deps ever gets turned on.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -127,6 +128,135 @@ function MealFlagsEditor({
...
+      // qty/unit and can re-pick.
+      if (item.portion) return
+      const p = detail.portions?.find((pp) => pp.id === item.food_item_portion_id)
+      if (p) applyPortion(p, /* keepCount */ true)
+      return
+    }
+    const def = detail.effective_default_portion_id
+    if (!def) return
+    const p = detail.portions?.find((pp) => pp.id === def)
+    if (p) applyPortion(p, /* keepCount */ false)
```
fiddur, line 174: This branch fires for *any* row without `food_item_portion_id` once detail arrives, including legacy rows loaded from an existing meal. Because `applyPortion(p, /* keepCount */ false)` resets `portion_count` to `1` and keeps the snapshot-style `ref`, a legacy row originally logged as e.g. `50 g` of a food whose effective default portion is `1 ruta = 3.4 g` is silently recomputed as `1 × 3.4 g` and then autosaved through `commitItems`. That's a silent ~15× nutrient drop on meal open, persisted to the server.

Consider gating this on "row was created this session" — e.g. set a non-persisted `_isNew: true` in `handleAdd` and require it here — or skip when the row has a server-provided `quantity`. Either way the auto-default behaviour should not touch rows that arrived via `mealItemsToEdit`.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -445,7 +639,23 @@ export function MealDetail() {
 
   const commitItems = (next: FoodItemEdit[]) => {
     setItems(next)
-    itemsFlusherRef.current.schedule({ food_items: editItemsToBody(next) })
+    // Skip the save when any row is in a transient portion-editing state
+    // (pinned to a portion but the count input is currently empty). Without
+    // this guard the row would fall through editItemsToBody's portion fork
+    // and briefly get re-saved as a legacy-quantity row, then re-pinned
+    // once the user finishes typing the new count.
+    if (next.some((fi) => fi.food_item_portion_id && !(typeof fi.portion_count === 'number' && fi.portion_count > 0))) {
```
fiddur: Minor: this short-circuits saves for *all* rows whenever any single row is in transient-count state. If the user clears one row's portion count and then edits a different row, those other edits won't autosave until the empty count is filled in. Consider filtering the offending row(s) out of the body instead of returning early, so the rest of the meal still flushes.

fiddur (commented):
🛑 Changes required

Clean factoring of the row-level portion picker, with good unit coverage for the new `scaleNutrient` portion path and the `mealItemsToEdit` ↔ `editItemsToBody` round-trip. The transient-row guard and de-dupe ref in `commitItems` are sensible mitigations for the load-time autosave noise.

One blocking concern, though — the auto-default branch in `useAutoDefaultPortion` doesn't distinguish a freshly-added row from an existing legacy meal row loaded from the server. Both arrive with `food_item_portion_id === undefined`, so when the user has set an `effective_default_portion_id` for the food, opening any old legacy-logged meal silently rewrites the row to `portion_count = 1` of that default portion and triggers an autosave with the new representation. For a row originally logged as e.g. "50 g chocolate" against a default portion of "1 ruta ≈ 3.4 g", this turns 50 g into 3.4 g — a ~15× nutrient drop persisted to the backend without user interaction. Please scope the branch-2 apply to rows that are actually fresh in this session (e.g. tag rows from `handleAdd` with a `_isNew` flag and only auto-apply when set), or skip the auto-apply whenever the loaded row already has a `quantity` from the server.

A couple of smaller things flagged inline.

apps/web/src/pages/Meals/MealDetail.tsx
```diff
@@ -445,7 +657,24 @@ export function MealDetail() {
...
     setItems(next)
-    itemsFlusherRef.current.schedule({ food_items: editItemsToBody(next) })
+    // Filter rows that are mid-edit on a portion (pinned to a portion but
+    // the count input is currently empty/zero). Without this, such a row
+    // would fall through editItemsToBody's portion fork and briefly get
+    // re-saved as a legacy-quantity row, then re-pinned when the count
+    // refills. We skip the offending row only (not the whole save), so
+    // edits to other rows still flush.
+    const saveable = next.filter(
+      (fi) => !(fi.food_item_portion_id && !(typeof fi.portion_count === 'number' && fi.portion_count > 0)),
```
fiddur: This per-row filter drops the row entirely from the saved body when `portion_count` is empty. Because `updateMealById` in the backend replaces `food_items` wholesale (see `writePreparedToJunction` / `setMealFoodItems(user, id, [])` in `apps/backend/src/services/meals.ts`), a debounced save that fires while the user is paused with an empty count will remove that row from the meal server-side until the count refills — and `flush()` on unmount commits the row-less body, so closing the page mid-edit loses the row. With a single-row meal it clears the meal entirely (`food_items: []`). Skipping the whole `schedule()` call when any row is mid-edit (the pre-fix behaviour the commit message refers to) avoids the data-loss window; the cost is just that other-row edits wait for the offending row to be completed, which seems preferable to silently deleting it on the server.

fiddur (commented):
🛑 Changes required

Overall a well-thought-through PR with thorough doc comments and good test additions for `scaleNutrient` portion-path scaling and the `mealItemsToEdit`/`editItemsToBody` round-trip. The `_isNew` flag, `useAutoDefaultPortion`, and the de-dupe of identical food_items bodies all hang together coherently, and the macros refactor (`MACRO_LABELS`) cleans up the render path nicely.

One concern worth addressing before merging:

**Per-row transient-guard can silently delete a row on the server.** In `commitItems` (apps/web/src/pages/Meals/MealDetail.tsx), `saveable` filters portion-pinned rows whose `portion_count` is empty/zero out of the request body, and `editItemsToBody(saveable)` is then sent to the backend. But `updateMealById` in `apps/backend/src/services/meals.ts` treats `food_items` as a full replacement (`writePreparedToJunction` / `setMealFoodItems(user, id, [])`). So if the user clears a portion count and pauses for >600ms (the debounce), an interim save fires that *removes* that row from the meal on the server until the count refills. If only one row exists and they blank its count, the save becomes `{food_items: []}` and clears the whole meal — losing the row entirely if they then navigate away (flush on unmount commits the row-less body). The inline comment frames this as “keep edits to other rows still flowing”, but the side effect is data loss, not just a brief visual flicker. Consider skipping the entire schedule when any row is mid-edit (the simpler pre-fix behaviour), or preserving the offending row's last-saved payload so dropping the local edit doesn't drop the server row.

Nothing else stood out: the de-dup seed handling, `useAutoDefaultPortion`'s `lastAppliedFoodId` ref, the `ref` refresh from canonical detail for portion-backfill rows, and the new tests all look correct.

fiddur (commented):
✅ Approved

Nice closing PR for the portion picker. The two-mode `QuantityInputs` + dropdown UI is clean and the supporting state machine in `useAutoDefaultPortion` is carefully reasoned (the inline comments do a lot of the heavy lifting). The portion-path formula in `scaleNutrient` correctly mirrors the backend (`nutrient × count × base_equivalent / ref.quantity`), and because the snapshot's nutrient/quantity ratio is invariant, both canonical and per-entry snapshot refs give correct display — so the ref-refresh on backfill is belt-and-suspenders but harmless. Round-trip tests + portion-path tests give good coverage of the new logic.

A few smaller observations (not blocking):

- `useAutoDefaultPortion`'s effect reads `item.food_item_portion_id`, `item.portion`, `item._isNew`, and the freshly-rebuilt `applyPortion` closure, but its deps are only `[detail?.id, item.food_item_id]`. The `lastAppliedFoodId` ref makes the omission intentional and safe today, but a future maintainer enabling react-hooks/exhaustive-deps will either get noisy warnings or be tempted to add deps that break the guard. An explicit `// eslint-disable-next-line react-hooks/exhaustive-deps` (plus a one-liner explaining the ref is the gate) would lock that intent in.
- `lastAppliedFoodId.current = item.food_item_id` is set unconditionally before the backfill/default branches, so the name is slightly misleading — it's really "last seen food id, won't retry." If the food's portions change mid-session (e.g. user adds one in another tab and detail re-fetches with a new portion list), the effect won't reapply the default. Likely fine — user can re-pick — but worth being aware of.
- `FoodItemRow` is rendered with `key={i}` in `FoodItemsEditor` (pre-existing). With the new per-row `useRef` + per-row react-query subscription this means a removal of any row in the middle shifts state across rows: the ref/query state previously bound to row N now serves row N's content. The `lastAppliedFoodId !== new food_item_id` re-runs the effect, and the existing guards (`if (item.food_item_portion_id) ...`) make the wrong outcomes rare, but a stable `key` (e.g. a client-side row id) would make this more robust as the row component takes on more local state.
- The de-dupe via `JSON.stringify` is sensitive to key ordering — works today because `editItemsToBody` constructs the object literally in a fixed key order, just worth noting if anyone refactors `editItemsToBody` to spread fields.

None of these are blockers — overall this looks ready to ship.

PR #785, opened 2026-05-29
feat(food-items): portion unit-model — quantity-in-unit, drop label_quantity

fiddur:
## Summary

Reframes food-item portions as **named units with a per-unit conversion to the base unit**, replacing the old "count × label_quantity" model. The number a user enters when logging is now simply the **quantity in the selected unit**.

Backend + web ship together because removing `label_quantity` from the shared `@aurboda/api-spec` type is a breaking change for the web build.

### Model change
- A portion is now `1 label_unit = base_equivalent base_units`; the unit label never contains a number. `label_quantity` is dropped everywhere (schema, api-spec, db, service, routes, MCP, tests). Existing rows are normalised on migrate via `base_equivalent /= label_quantity`.
- The base unit is itself a pickable unit (`base_equivalent = 1`). Scaling stays `entered_qty × base_equivalent / default_quantity` — e.g. `58 g` on a 100 g base ⇒ ×0.58; `1 wrap` on a `2 wrap` base ⇒ ×0.5; `1 glas (=515 g)` ⇒ ×5.15.
- Separate **default logging amount**: new `default_log_quantity` on `food_items` and `shared_food_item_overrides`; `effective_default_quantity` surfaced on the detail response; the default-portion endpoint + MCP tool now take `{ portion_id, quantity }`.
- **Backfill** a base unit for every food so all foods have at least one unit to log in: per-user → `1 portion`, central shared → `100 g` (only where missing).

### Web UI
- Meal-entry row reordered to **`[food] [number] [unit ▾] [×]`**, the redundant `× 1 unit` span removed, unit labels shown bare, the number prefilled from the food's default amount, and an **"item details"** link added after the per-row macros.
- Food-item portions editor drops the `label_quantity` input (`1 [unit] = [amount] [base unit]`) and gains a **default-amount** field.

### Docs
- `docs/features/meals.md` updated for the unit model.

## Verification
- `pnpm check` (backend + web + api-spec): **0 type errors, 0 lint errors**.
- Backend tests: **2253 passed**. Web tests: **409 passed**.
- api-spec rebuilt + `pnpm generate` run (OpenAPI + Kotlin).

### DIFF
--- a/apps/backend/src/routes/food-items-router.ts
+++ b/apps/backend/src/routes/food-items-router.ts
@@ -101,6 +101,19 @@ const serializeFoodItem = (
   return result as FoodItemEntity
 }
 
+/**
+ * Resolve the default logging quantity, expressed in the unit named by the
+ * effective default portion. Prefer an explicit `default_log_quantity`;
+ * otherwise fall back to the base `default_quantity` only for the base unit
+ * (no default portion) — for a non-base default unit fall back to 1 of that
+ * unit, never the base quantity (which would be a number in the wrong unit).
+ */
+export const resolveEffectiveDefaultQuantity = (
+  effectiveDefaultPortionId: string | undefined,
+  defaultLogQuantity: number | undefined,
+  baseQuantity: number | undefined,
+): number | undefined => defaultLogQuantity ?? (effectiveDefaultPortionId ? 1 : baseQuantity)
+
 const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
   const base = serializeFoodItem(detail.item)
   const sensitivities = detail.sensitivities ?? []
@@ -110,12 +123,11 @@ const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
   // for central items applySharedOverrides decorated it from the user's
   // shared_food_item_overrides row (when the user picked a portion).
   const effective_default_portion_id = (detail.item.default_portion_id as string | undefined) ?? undefined
-  // Resolved default *quantity*: the per-user / override default_log_quantity
-  // when set, else the base default_quantity. applySharedOverrides decorates
-  // central items' default_log_quantity from the user's override row.
-  const effective_default_quantity =
-    (detail.item.default_log_quantity as number | undefined) ??
-    (detail.item.default_quantity as number | undefined)
+  const effective_default_quantity = resolveEffectiveDefaultQuantity(
+    effective_default_portion_id,
+    detail.item.default_log_quantity as number | undefined,
+    detail.item.default_quantity as number | undefined,
+  )
   // Composite branch: ingredient list + derived totals.
   if (detail.ingredients) {
     return {
@@ -645,7 +657,7 @@ export const createFoodItemsRouter = (authMiddleware: AnyMiddleware, centralDb: 
         })
       }
       try {
-        await setDefaultPortion(user, id, req.body.portion_id, req.body.quantity)
+        await setDefaultPortion(user, id, req.body.portion_id, req.body.quantity ?? null)
       } catch (err) {
         const message = err instanceof Error ? err.message : 'Failed to set default portion'
         return res.status(/not found|does not belong/i.test(message) ? 400 : 500).json({
--- a/apps/backend/src/routes/food-items-router.ts
+++ b/apps/backend/src/routes/food-items-router.ts
@@ -110,6 +110,12 @@ const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
   // for central items applySharedOverrides decorated it from the user's
   // shared_food_item_overrides row (when the user picked a portion).
   const effective_default_portion_id = (detail.item.default_portion_id as string | undefined) ?? undefined
+  // Resolved default *quantity*: the per-user / override default_log_quantity
+  // when set, else the base default_quantity. applySharedOverrides decorates
+  // central items' default_log_quantity from the user's override row.
+  const effective_default_quantity =
+    (detail.item.default_log_quantity as number | undefined) ??
+    (detail.item.default_quantity as number | undefined)
   // Composite branch: ingredient list + derived totals.
   if (detail.ingredients) {
     return {
@@ -126,6 +132,7 @@ const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
       is_shared,
       portions,
       effective_default_portion_id,
+      effective_default_quantity,
       sensitivities,
     }
   }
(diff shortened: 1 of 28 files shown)


apps/backend/src/routes/food-items-router.ts
```diff
@@ -110,6 +110,12 @@ const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
   // for central items applySharedOverrides decorated it from the user's
   // shared_food_item_overrides row (when the user picked a portion).
   const effective_default_portion_id = (detail.item.default_portion_id as string | undefined) ?? undefined
+  // Resolved default *quantity*: the per-user / override default_log_quantity
+  // when set, else the base default_quantity. applySharedOverrides decorates
+  // central items' default_log_quantity from the user's override row.
+  const effective_default_quantity =
```
fiddur: `default_quantity` is the **base-unit** amount, but `effective_default_portion_id` may name a different unit. When a non-base portion is the default and `default_log_quantity` is NULL (pre-existing foods migrated without the new column, or MCP/override callers passing `quantity: null`), this returns the base quantity expressed in the wrong unit. The schema documents `effective_default_quantity` as being "in the unit named by effective_default_portion_id", and the meal logger prefills it directly — so a `100 g`-base food with a `glas` default unit prefills `100 glas` (regression from the prior `1`). Only fall back to `default_quantity` when the effective default is the base unit; otherwise default to `1` (or leave absent).

fiddur (commented):
🛑 Changes required

Solid, well-documented refactor reframing portions as named units (drop `label_quantity`, add per-unit `base_equivalent` + a `default_log_quantity` prefill). Schema/MCP/REST/web/docs are consistently updated, the `label_quantity` removal is complete (no dangling TS references), and the per-unit `base_equivalent / label_quantity` normalization in the migration is correct. Tests are updated alongside. I found one correctness bug and one data-migration gap worth resolving before merge.

**1. (blocking) `effective_default_quantity` falls back to the base-unit quantity even when the default unit is a non-base portion.** In `serializeDetail` the resolved quantity is `default_log_quantity ?? default_quantity`. But `default_quantity` is the *base*-unit amount, while `effective_default_portion_id` may name a different unit. The schema doc states `effective_default_quantity` is "in the unit named by effective_default_portion_id", so the fallback violates its own contract whenever a non-base portion is the default and `default_log_quantity` is NULL. This is reachable:
  - Per-user foods that had `default_portion_id` set before this PR — after migration their new `default_log_quantity` column is NULL, so the resolver returns the base quantity in the wrong unit.
  - MCP `set_default_food_item_portion` / central-item overrides called with `portion_id` set + `quantity: null`.

  The web meal logger then prefills it verbatim (`MealDetail.tsx` `useAutoDefaultPortion` → `apply(p, count ?? 1)`), so e.g. a `100 g`-base food with a `glas` default unit prefills `100 glas` instead of `1 glas` — a regression vs. the prior behaviour that prefilled `1`. The fallback to `default_quantity` should only apply when the effective default unit is the base unit (`effective_default_portion_id` absent); otherwise default to `1` (one of the named unit) or leave it absent.

**2. (please verify) Historical meal-junction `portion_count` is not migrated alongside `base_equivalent`.** The migration divides `base_equivalent` by the old `label_quantity` (per-unit reframe), but meal junction rows logged against a portion store `portion_count` under the *old* "count of portion entries" semantics. After migration, `resnapshotMealsForFoodItem` recomputes `portion_count × base_equivalent / default_quantity` with the new (smaller) `base_equivalent`, which silently rescales nutrients for any meal logged against a portion whose `label_quantity` was ≠ 1 (e.g. the documented `Fylld tortilla 2 wrap`). The stored junction `quantity` column already holds `portion_count × label_quantity`, so a matching `UPDATE` (set `portion_count = quantity` for portion-linked rows) inside the guarded migration would close the gap. If no production data has `label_quantity ≠ 1` this is moot — worth confirming.

Minor: the `default-portion` endpoint now requires `quantity` in the body (nullable but not optional); fine for the internal web/MCP callers but a breaking change for any direct API consumer sending only `portion_id`.

apps/backend/src/routes/food-items-router.ts
fiddur: Fixed in e037194d. `effective_default_quantity` now resolves via `resolveEffectiveDefaultQuantity`: it only falls back to the base `default_quantity` when the effective default is the base unit; for a non-base default unit with no explicit `default_log_quantity` it defaults to `1` of that unit (never the base quantity in the wrong unit). Added unit tests covering all three branches. Also reframed historical `meal_food_items.portion_count` in the migration (set `= quantity` for portion-linked rows) so resnapshot no longer rescales nutrients for portions with `label_quantity ≠ 1`, and made the default-portion body `quantity` optional for back-compat.

fiddur (commented):
✅ Approved

Reviewed the `default_log_quantity` / portion unit-model changes across the layers I could read cleanly:

- **`packages/api-spec/src/schemas/food-items.ts`** — `default_log_quantity` is added to `foodItemEntitySchema` (positive, optional), `effective_default_quantity` to `foodItemDetailSchema`, and `default_log_quantity` (nullable) to both `sharedFoodItemOverrideSchema` and `setSharedFoodItemOverrideBodySchema`. The `.refine()` on the override-body schema is correctly extended so `default_log_quantity` also satisfies the "at least one field" rule. Descriptions are clear and consistent with the snake_case convention. Looks correct.
- **`apps/backend/src/db/food-items.ts`** — the new column is added consistently in all three places: `FOOD_ITEM_COLUMNS`, the `mapFoodItemRow` mapper (`(row.default_log_quantity as number | null) ?? undefined`), and the `updateFoodItem` conditional SET clause. Consistent and correct.
- **`apps/backend/src/services/food-items.ts`** — `applySharedOverrides` layers `default_log_quantity` onto central items with a `!= null` guard, matching the existing override-merge pattern. `routes/food-items-router.ts` imports the new portion types cleanly.

These changes are internally consistent and I found no blocking issues in them.

**Verification caveat — please rely on PR CI as the source of truth.** I could not run a meaningful local `pnpm check`: this worktree has no `node_modules` installed (`pnpm` reported "Local package.json exists, but node_modules missing"), so the `@aurboda/api-spec` build failed (can't resolve `zod`) and the backend type-check then produced 438 errors — 406 of them plain `TS2307: Cannot find module ...` and the remaining ~32 all downstream cascades of unresolved `zod`/api-spec types (e.g. `{ [x: string]: any }` params in `mcp/*-tools.ts`, missing props on `ActivityResult` / `ResolvedTrainingLoadSettings`, etc.), spread across many subsystems unrelated to this PR. None are attributable to this diff, so I did not treat them as findings. I also could not read a few changed files cleanly (`mcp/food-item-tools.ts`, `schema/meals.ts`, `services/meals.ts`, the `food-item-portions` db/service files, and the web `FoodItemDetail`/`MealDetail` changes).

Given the PR drops `label_quantity` and reworks portions, please just confirm against green CI that: (1) the schema/DDL defines the `default_log_quantity` column the new SELECT/INSERT/UPDATE code references, and (2) no remaining consumer still references the removed `label_quantity`. If CI is green, this is good to merge.

PR #786, opened 2026-05-31
feat(web): food-item editor — base at top, default as number + unit dropdown

fiddur:
## Summary

Tidies the food-item edit UI, which used "default" in two confusingly-similar places.

- **Top meta row → "Base:"** — relabelled from "Default:". This edits the nutrient-density basis (`default_quantity` / `default_unit`), with `100` / `g` placeholders. (No more radio buttons for choosing the base.)
- **Units box → single "Default: [number] [unit ▾]" control** — the same shape as the meal logger. Replaces the number-only "Default amount" field *and* the per-row "default" radio buttons. The dropdown lists the base unit plus every named unit; picking one sets `default_portion_id`, the number sets `default_log_quantity`. Changing the unit clears the quantity so the server resolves a sensible default for that unit (1 of a named unit, or the base quantity) rather than carrying a base-unit amount into a different unit.
- **Unit rows** lose their radio — they're just `1 [unit] = [amount] [base]` + Delete.

Web-only; the `PUT /food-items/:id/default-portion` endpoint already accepts `{ portion_id, quantity }`.

## Verification
- `pnpm --filter aurboda-web check`: **0 type errors, 0 lint errors**.
- Web tests: **409 passed**.
- Formatted with oxfmt; only the two `FoodItemDetail.{tsx,css}` files changed.

### DIFF
--- a/apps/web/src/pages/FoodItems/FoodItemDetail.tsx
+++ b/apps/web/src/pages/FoodItems/FoodItemDetail.tsx
@@ -551,7 +551,7 @@ function DefaultMetaRow({
     <div class="fi-meta">
       {item.source && <span class="fi-source">Source: {item.source}</span>}
       <span class="fi-default-edit">
-        Default:
+        Base:
         {isShared ? (
           <span class="fi-default-readonly">
             {item.default_quantity ?? '—'} {item.default_unit ?? ''}
@@ -562,14 +562,14 @@ function DefaultMetaRow({
               type="number"
               step="0.1"
               value={defaultQuantity}
-              placeholder="Qty"
+              placeholder="100"
               onInput={(e) => setDefaultQuantity((e.target as HTMLInputElement).value)}
               onBlur={commitDefaultQuantity}
             />
             <input
               type="text"
               value={defaultUnit}
-              placeholder="Unit"
+              placeholder="g"
               onInput={(e) => setDefaultUnit((e.target as HTMLInputElement).value)}
               onBlur={commitDefaultUnit}
             />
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/FoodItems/FoodItemDetail.tsx
```diff
@@ -580,44 +580,73 @@ function DefaultMetaRow({
...
       />
-      <span class="fi-default-amount-hint">in the selected default unit, prefilled when logging</span>
+      <select
+        class="fi-default-unit"
+        value={effectiveDefault ?? ''}
+        onChange={(e) => {
+          // Keep the current amount — only the unit changes.
+          const t = qty.trim()
+          const n = t === '' ? null : parseFloat(t)
+          onCommit((e.target as HTMLSelectElement).value || null, n !== null && n > 0 ? n : null)
```
fiddur: Minor/optional: if the qty input currently holds an invalid value (e.g. the user typed `abc` or a negative number without blurring) and they then switch the unit, `n !== null && n > 0` is false, so this commits `null` as the quantity — silently wiping a previously-saved valid amount. `commitQty` deliberately resets-without-committing on invalid input; switching units behaves differently. Consider falling back to `effectiveQty` (or leaving the quantity untouched) instead of committing `null` when the typed value is invalid.

fiddur (commented):
✅ Approved

Clean, self-contained refactor of the food-item default editor: the base row label becomes "Base:", `DefaultAmount` becomes `DefaultLoggingAmount` (number input + unit dropdown), and the per-row/base radio buttons for selecting the default unit are replaced by the dropdown. The new `onCommit(portionId, quantity)` signature threads through `defaultMutation` cleanly, the units list is now conditionally rendered only when portions exist, and `effectiveDefault`/`effectiveQty` resolution is unchanged. I couldn't run `pnpm check` (node_modules isn't installed in this worktree), so the type/lint check was reasoned through manually — please confirm CI is green.

Two non-blocking notes:
- **Dead CSS**: removing the radio/base-row markup leaves `.fi-portion-base` and `.fi-portion-default` (and their references in the responsive media query around CSS lines 652-653) unused. Worth dropping in a follow-up to keep the stylesheet tidy. (These lines aren't in this diff, so flagged here rather than inline.)
- **Minor edge case** on the unit dropdown's `onChange` (see inline comment) where an invalid value in the qty box at switch time silently clears a previously-saved quantity.

apps/web/src/pages/FoodItems/FoodItemDetail.tsx
fiddur: Good catch — fixed. Switching the unit now only commits null when the field is empty (intentional clear); a transient/invalid value falls back to the last effective quantity, so a saved amount is never silently wiped.

fiddur (commented):
✅ Approved

Clean, well-documented UI refactor of the food-item detail page (`FoodItemDetail.tsx` + CSS only).

- `DefaultMetaRow` is relabeled `Base:` with `100`/`g` placeholders — consistent with the base/default split.
- The old radio-button default selection (a `Base` row plus a radio per portion) is replaced by `DefaultLoggingAmount`: a number input + a unit dropdown (base unit as the empty-value option, portions by id). `PortionRow` correctly drops its now-unused `isDefault`/`onSetDefault` props, and the portions list only renders when portions exist.
- The `onCommit(portionId, quantity)` signature is threaded consistently through to `defaultMutation` / `setDefaultPortionApi`.

The tricky edge cases are handled well and clearly commented (and match the two follow-up fix commits): clearing the number commits a `null` quantity while keeping the selected unit; a non-positive/invalid number resets instead of committing; and switching the unit preserves the current amount, falling back to the last effective quantity so a transient invalid value can't wipe a saved one. State sync via the `effectiveQty` `useEffect` is correct.

Non-blocking notes:

1. Switching the unit deliberately keeps the raw numeric amount (e.g. a base default of `100 g` becomes `100` of a portion unit). The comment documents this as intended; just confirm it's the desired UX, since the number's meaning changes with the unit.
2. This interactive default-selection logic (quantity carried across unit switches, invalid-input handling) has no automated test — there's no test file for this page. Given how subtle the commit logic is, a small unit/component test would be worth adding, but it's not a blocker.

PR #787, opened 2026-06-01
feat(activity-detail): show distance, pace, cadence, HR averages

fiddur:
## Summary
- Activity detail summary previously only showed Active Calories — add Distance, Avg Pace, Avg Cadence, Avg HR, and Max HR.
- Fields are rendered only when present on the activity (already returned by the detail endpoint via `ActivityComputedMetrics`).
- Avg Pace uses `avg_pace` when set; otherwise derives it from `avg_speed`.

## Test plan
- [x] `pnpm check` (web + backend + api-spec) — passes
- [x] `vitest run src/pages/EntityDetail/` — 44 tests pass (8 new format-utils tests)
- [ ] Verify on a Running activity (e.g. `/detail/activity/merged:6742de9d-...`): Distance, Avg Pace, Avg Cadence appear
- [ ] Verify on an activity without these fields (e.g. a sleep entry) that the rows are omitted cleanly

### DIFF
--- a/apps/web/src/pages/EntityDetail/format-utils.test.ts
+++ b/apps/web/src/pages/EntityDetail/format-utils.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, test } from 'vitest'
+
+import { formatCadence, formatDistance, formatPace } from './format-utils'
+
+describe('formatDistance', () => {
+  test('formats meters under 1km without decimals', () => {
+    expect(formatDistance(850)).toBe('850 m')
+    expect(formatDistance(0)).toBe('0 m')
+    expect(formatDistance(999)).toBe('999 m')
+  })
+
+  test('formats km with two decimals at and above 1000m', () => {
+    expect(formatDistance(1000)).toBe('1.00 km')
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped change. The new `formatDistance`/`formatPace`/`formatCadence` helpers are extracted as pure functions and covered by focused unit tests with good edge cases (sub-km vs km, pace derived from speed, leading-zero seconds, rounding, undefined handling). The pace fallback logic (`avg_pace > 0` else derive from `avg_speed`) is correct, and all referenced fields (`distance`, `avg_pace`, `avg_speed`, `avg_cadence`, `avg_hr`, `max_hr`, `avg_hrv`) exist as optional numbers in the api-spec `Activity` schema.

The `index.tsx` refactor cleanly replaces three separately-wrapped `entity-fields` blocks with a single data-driven `ActivityStats` component that renders only the rows whose data is present, returning `null` when empty. Functional style, no classes, types are explicit — consistent with the project conventions.

Note: I couldn't run the test suite in this review checkout (vitest not installed), but I traced each test case against the implementation and they all hold. No blocking issues.

PR #789, opened 2026-06-01
refactor(activity-detail): unified summary table

fiddur:
## Summary
- Activity detail read-only summary previously stacked separate \`.entity-fields\` blocks, producing uneven row spacing and misaligned value columns (e.g. Time/Duration tight, then a visible step before Avg Pace/Cadence).
- Render all rows (Time, Duration, Distance, Avg Pace, Avg Cadence, Avg HR, Max HR, Active Calories, Asleep, Avg HRV, Notes) as a single \`<table class="activity-stats-table">\` — semantically correct for tabular data, with naturally aligned label/value columns and uniform row padding.
- Row builder extracted to \`activityStats.ts\` with focused unit tests (inclusion logic, custom \`durationLabel\` for sleep, notes row). \`EditableActivityFields\` in read mode now renders only the title h2; edit mode is unchanged.

## Test plan
- [x] \`pnpm check\` — 0 errors across backend / web / api-spec
- [x] \`vitest run src/pages/EntityDetail/\` — 51 tests pass (7 new)
- [ ] Visual check on a Running activity: rows have consistent spacing and values column-align
- [ ] Visual check on a Sleep activity: \"In Bed\" label appears in the table; \"Asleep\" row shows
- [ ] Edit mode still works (inputs render, save flow intact)

### DIFF
--- a/apps/web/src/pages/EntityDetail/EditableActivityFields.tsx
+++ b/apps/web/src/pages/EntityDetail/EditableActivityFields.tsx
@@ -4,7 +4,7 @@
  */
 import { ActivityTypePicker } from '../../components/ActivityTypePicker'
 import { IconPreview } from '../../components/IconPreview'
-import { formatDateTime, formatDuration, formatTime } from './format-utils'
+import { formatDuration } from './format-utils'
 
 export interface ActivityDraft {
   activity_type: string
@@ -17,13 +17,10 @@ export interface ActivityDraft {
 
 interface EditableActivityFieldsProps {
   title: string
-  displayStart: Date
-  displayEnd?: Date
-  notes?: string
   isEditing: boolean
   draft: ActivityDraft
   onDraftChange: (draft: ActivityDraft) => void
-  /** Label for the duration row (e.g. "In Bed" for sleep). Defaults to "Duration". */
+  /** Label for the duration row in edit mode (e.g. "In Bed" for sleep). Defaults to "Duration". */
   durationLabel?: string
   /** Icon (emoji or URL) to display next to the title. */
   icon?: string
@@ -33,9 +30,6 @@ interface EditableActivityFieldsProps {
 
 export const EditableActivityFields = ({
   title,
-  displayStart,
-  displayEnd,
-  notes,
   isEditing,
   draft,
   onDraftChange,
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped refactor that unifies the read-only activity summary into a single semantic table. The `buildActivityStatRows` helper is extracted into its own module with good unit coverage, and the Time/Duration/Notes rows previously rendered by `EditableActivityFields` are now folded into the same table — logic is preserved verbatim (formatter calls, ordering, and presence checks are unchanged). I verified:

- All formatter signatures (`formatDistance`, `formatPace`, `formatCadence`, `formatMinutesAsHM`, `formatDateTime`, `formatTime`, `formatDuration`) match their callers, and the test expectations (`2.77 km`, `7h 5m`, `31m`, bpm rows) are correct.
- No leftover references to the removed `ActivityStats` component, the inlined `buildActivityStatRows`, or the dropped imports (`formatMinutesAsHM`, `formatCadence`, `formatDistance`, `formatPace`) remain in `index.tsx`.
- `getUserNotesContent` returns `string`, satisfying the non-optional `notes` field of `BuildActivityStatRowsInput`.
- `<table>` with `<th scope="row">` is good semantics for key/value pairs, and `white-space: pre-wrap` on `td` correctly preserves multi-line notes. The new table colors (`#6b7280`, `#374151`) match the existing `.field-label`/`.field-value` values, so it stays visually consistent.

I could not run `pnpm check`/`vitest` — dependencies are not installed in this worktree and the npm cache is read-only — so type-check/test execution wasn't confirmed locally, but CI will cover it.

Minor (non-blocking) notes:
- The file-level doc comment in `EditableActivityFields.tsx` (lines 1-3) still says the component renders "title, time, duration, and notes in either read mode ... or edit mode". In read mode it now renders only the title; worth updating to avoid drift.
- Behavioural change worth confirming is intentional: in read-only view the Notes row now appears at the very bottom of the unified table (after Avg HRV) rather than directly under Time/Duration as before, and the whole stats block now renders after the schema-data fields rather than before them. Looks deliberate for the "unified table" goal.
- `activityStats.test.ts` uses `as Activity` to build fixtures, which is at odds with the repo's "avoid casting" guidance, but is a pragmatic and common pattern for partial test mocks — fine as-is.

fiddur (commented):
✅ Approved

Solid, well-scoped refactor. The read-only Time/Duration/Notes rendering is moved out of `EditableActivityFields` and merged with the metric rows into a single `buildActivityStatRows` helper rendered as a semantic `<table>`. The extracted helper is pure and now has good unit coverage (`activityStats.test.ts`), and I verified each expected value against the underlying formatters (`formatDuration` 09:57→10:28 = `31m`, `formatDistance(2770)` = `2.77 km`, `formatMinutesAsHM(425)` = `7h 5m`, etc.). Import cleanup is correct (no leftover unused imports; `formatTime`/`formatDateTimeLocal`/`Activity` are all still used), `EditableActivityFields` has only this one call site, and the new CSS colors (`#6b7280`/`#374151`) match the existing `.field-label`/`.field-value` colors so there's no theming inconsistency. The `<th scope="row">` markup and `white-space: pre-wrap` on the value cell are nice touches (multi-line notes now render correctly).

1. The file-level doc comment at the top of `EditableActivityFields.tsx` ("...renders activity title, time, duration, and notes in either read mode (plain text) or edit mode...") is now stale for read mode — read mode renders only the title. Worth updating to match the new inline comment at lines 149-150.

2. Behavior change worth confirming is intentional: because the Time/Duration rows moved from `EditableActivityFields` into the read-only `ActivityStatsTable`, they now render *below* `SchemaDataFields` rather than directly under the title. This is consistent with the "unified summary table" goal, but it does reorder the read-only layout — just confirm that's the intended visual order.

PR #790, opened 2026-06-04
fix(lastfm): trigger deduction rules on background and MCP scrobble syncs

fiddur:
## Problem

A scrobble-based deduction rule (the user's **Vocal Training** rule, matching vocal-exercise tracks by `Sam West` et al.) had only been applied intermittently for weeks. Qualifying scrobbles existed in the data (e.g. May 26, Jun 2, Jun 3) with no resulting `vocal_training` activity, while a dry-run of the rule over the last 90 days said **17** activities should exist vs. only ~8 actually present.

## Root cause

Deduction rules run only when something triggers an evaluation. For scrobbles the only trigger is the `onActivitySynced` notifier — and it was wired into **just one** of the three Last.fm sync paths:

| Sync path | Triggered deduction? |
|---|---|
| REST `POST /sync/lastfm` (web "sync" button) | ✅ |
| Background auto-sync (`sync-provider.ts`, fires lazily before tag/daily-summary queries) | ❌ |
| MCP `sync_lastfm` tool | ❌ |

The background auto-sync is the path that runs in normal use, so newly-scrobbled tracks landed in `raw_records`/`music_scrobble` activities **without ever being run through the rules**. Tags only appeared when an explicit web sync or a rule edit happened to force a re-evaluation. (The MCP tool even advertised "applies auto-tagging rules" but didn't.)

## Fix

Wire `onActivitySynced` into the sync-provider config and pass the notifier to the MCP sync tool, so **every** sync path evaluates deduction rules over the synced window — matching what the REST route already does.

- `sync-provider`: fire the notifier over `[last_sync_time, now]` when a sync ingests new scrobbles (tight window → less churn than a blanket 30-day re-eval on every 30-min auto-sync).
- `sync_lastfm` MCP tool: fire the notifier over the synced window, mirroring the REST route.
- Export `DEFAULT_SYNC_HISTORY_DAYS` for the first-sync fallback window.
- Hoist the `activityNotifier`/`deductionQueue` in `api.ts` so the notifier can be injected into `createSyncProvider`.
- Unit tests for both paths (fires only on success with new scrobbles; skips within the sync threshold).

Existing gap data was already backfilled separately via `evaluate_deduction_rules`.

## Notes / follow-up

- The same trigger gap likely exists for **other** auto-synced providers (Garmin/Oura/RescueTime) whose activities feed `activity`/`screentime` conditions — scoped out of this PR, worth a follow-up.
- `docs/lastfm.md` is broadly stale (predates the tags→deduction-rules/activities migration); this PR only corrects the two sync sections it touches.

## Test plan

- [x] `pnpm --filter aurboda-backend check` (tsgo + oxlint) — passes, 0 new errors
- [x] New unit tests pass (`sync-provider.test.ts`, `sync-tools.test.ts`)
- [ ] CI green

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -201,7 +201,7 @@ const main = async () => {
 
   // Create sync provider for auto-syncing data before queries. onActivitySynced
   // lets background scrobble syncs trigger deduction rules (e.g. auto-tagging),
-  // matching what the REST /sync routes already do.
+  // like the REST /sync routes — here fired only when a sync ingests new data.
   const syncProvider = createSyncProvider({
     garmin,
     getLastFmApiKey: () => centralDb.getLastFmApiKey(),
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -186,16 +186,8 @@ const main = async () => {
       centralDb.upsertStravaAthleteMapping(stravaAthleteId, username),
   })
 
-  // Create sync provider for auto-syncing data before queries
-  const syncProvider = createSyncProvider({
-    garmin,
-    getLastFmApiKey: () => centralDb.getLastFmApiKey(),
-    oura,
-  })
-
-  // Initialize shared pg-boss instance and job queues (before MCP mount)
-  const boss = await createPgBoss()
-
+  // Deduction queue is assigned once pg-boss is up (below). The notifier closes
+  // over the variable, so it starts enqueuing as soon as the queue exists.
   let deductionQueue: DeductionQueue | null = null
   const activityNotifier: ActivityNotifier = (user, activityType, start, end, sourceRuleId) => {
     deductionQueue?.enqueueEvaluation({
@@ -206,6 +198,20 @@ const main = async () => {
       window_start: start.toISOString(),
     })
   }
+
+  // Create sync provider for auto-syncing data before queries. onActivitySynced
+  // lets background scrobble syncs trigger deduction rules (e.g. auto-tagging),
+  // matching what the REST /sync routes already do.
+  const syncProvider = createSyncProvider({
+    garmin,
+    getLastFmApiKey: () => centralDb.getLastFmApiKey(),
+    oura,
+    onActivitySynced: activityNotifier,
+  })
+
+  // Initialize shared pg-boss instance and job queues (before MCP mount)
+  const boss = await createPgBoss()
+
   const engineDeps = createDefaultEngineDeps(activityNotifier)
 
   if (boss) {
--- a/apps/backend/src/integrations/lastfm/sync.ts
+++ b/apps/backend/src/integrations/lastfm/sync.ts
@@ -19,7 +19,7 @@ import { getSyncState, insertActivities, insertRawRecord, upsertSyncState } from
 import { lastfmClient } from './client.ts'
 
 /** Default start date for historical sync (30 days back) */
-const DEFAULT_SYNC_HISTORY_DAYS = 30
+export const DEFAULT_SYNC_HISTORY_DAYS = 30
 
 /** Result of a sync operation */
 export interface LastFmSyncResult {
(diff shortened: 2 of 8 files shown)


fiddur (commented):
✅ Approved

Solid, well-targeted fix. Background auto-sync (`createSyncProvider`) and the MCP `sync_lastfm` tool previously ingested scrobbles without triggering deduction rules, unlike the REST `/sync/lastfm` route. This PR wires the shared `ActivityNotifier` (`onActivitySynced` / `deps.onActivityMutated`) into both paths, with guards (`status === 'success' && scrobbles_processed > 0`) and a correct synced window. The export of `DEFAULT_SYNC_HISTORY_DAYS` (= 30, matching the REST route's `DEFAULT_SYNC_LOOKBACK_MS`) keeps the lookback consistent. New unit tests cover the trigger/no-trigger/failure/threshold-skip cases, and the docs update is accurate (verified `docs/features/deduction-rules.md` and the `music_scrobble` activity type exist).

The closure ordering in `api.ts` is correct: `syncProvider` is created before `deductionQueue` is assigned, but `activityNotifier` reads `deductionQueue` lazily at call time (and no sync fires before the server is listening), so the comment is accurate.

I could not run the new tests — dependencies aren't installed and this sandbox's pnpm/npm store is read-only — but both test files are correct by inspection.

- The window-start logic differs between the two new paths. The background sync (`sync-provider.ts:143`) precisely mirrors `syncLastFmData`'s window (`last_sync_time` for incremental, else 30 days). The MCP tool (`sync-tools.ts:211`) always uses 30 days when no `start_date` is given, even when an incremental sync only fetched since `last_sync_time` — so deduction re-evaluates a wider window than was actually ingested. This matches the existing REST route's behavior and is idempotent, so it's a reasonable choice; just noting the asymmetry.
- The "matching the REST /sync/lastfm route" comments are slightly imprecise: the REST route triggers `onActivitySynced` unconditionally (no success/`scrobbles_processed` guard), whereas the new paths only trigger on success with new scrobbles. The new behavior is arguably better; the comment just overstates the parity.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. The two paths that previously ingested scrobbles without ever firing deduction rules — the background auto-sync (`syncLastFmIfNeeded`) and the `sync_lastfm` MCP tool — now invoke the same `ActivityNotifier` the REST `/sync/lastfm` route already used, closing a real parity gap (CLAUDE.md: REST and MCP should have equivalent capabilities).

Correctness checks pass:
- The api.ts reordering is sound: `activityNotifier` closes over the `let deductionQueue` binding (read at call time via `?.`), so creating `syncProvider` with `onActivitySynced: activityNotifier` before pg-boss/`deductionQueue` is assigned is safe — events simply no-op until the queue exists. The inline comment documents this.
- The deduction window in the background path (`syncState?.last_sync_time ?? subDays(now, DEFAULT_SYNC_HISTORY_DAYS)`) faithfully mirrors `syncLastFmData`'s own incremental-vs-first-sync window logic, and the MCP path's `start_date ? new Date(start_date) : subDays(now, 30)` matches the REST route's `DEFAULT_SYNC_LOOKBACK_MS` (also 30 days). Exporting `DEFAULT_SYNC_HISTORY_DAYS` to share the constant is the right call.
- Gating on `result.status === 'success' && result.scrobbles_processed > 0` is valid against the `LastFmSyncResult` type and correctly avoids re-evaluating rules when nothing new arrived.
- All `registerSyncTools`/`createSyncProvider` call sites are updated; the new params are optional so nothing breaks.

Test coverage is good: both new files exercise the fire/no-fire branches (new scrobbles, zero scrobbles, error status, and within-threshold skip for the background path) and assert the `'*'` activity type and window bounds. I could not execute the suite (node_modules not installed in this worktree), but the tests are logically consistent with the implementation and mocks.

One minor, non-blocking observation: the docs say "Every sync path … triggers evaluation … over the just-synced window," but the new MCP/background paths gate on `scrobbles_processed > 0` while the pre-existing REST route fires unconditionally on success. The difference is benign (the new paths are slightly more conservative), but if you want true behavioral parity you could either gate the REST route the same way or note the distinction in the docs. Out of scope for this PR.

PR #791, opened 2026-06-07
fix(web): meals day total honors meal-level macros over food-item aggregate

fiddur:
## Problem

On `/meals`, the **Day Totals** calorie sum disagreed with the sum of the visible meal cards. For one real day the header showed **1653 kcal** while the cards summed to **2290 kcal**.

Root cause: `aggregateDayNutrients` summed each meal's `meal.nutrients`, which is aggregated **from per-food-item data**. When a meal has a food item with no per-item nutrition (`nutrient_data_incomplete: true` — e.g. a home-cooked dish logged with a manual meal-level estimate but no per-ingredient breakdown), that item contributes 0 to `meal.nutrients`. But each meal card shows the meal-level `meal.calories` (the manual estimate), so the two views diverged.

Concretely, the dinner that day had `meal.calories = 760` (estimate) but `meal.nutrients.calories = 122.82` (only the rice had per-item data). `760 − 122.82 = 637.18`, exactly the gap between 2290.6 and 1653.42.

## Fix

- Extract `aggregateDayNutrients` into `apps/web/src/pages/Meals/dayNutrients.ts` (pure, testable).
- For the macro fields (`calories`, `protein`, `carbs`, `fat`, `fiber`), prefer the meal-level value over the per-food-item aggregate when they differ — so the day total matches the cards.
- Micronutrients have no meal-level override and stay summed from the aggregate. The existing "totals may be understated" notice still covers those.

## Tests

New `dayNutrients.test.ts` (7 cases), including a reproduction of the real day (asserts 2290.6, not 1653.42), micronutrient fallback, lower-than-aggregate override, and no-nutrients skip.

`pnpm check` passes (0 errors); all Meals tests green.

### DIFF
--- a/apps/web/src/pages/Meals/dayNutrients.test.ts
+++ b/apps/web/src/pages/Meals/dayNutrients.test.ts
@@ -62,8 +62,16 @@ describe('aggregateDayNutrients', () => {
     expect(aggregateDayNutrients(meals).calories).toBe(122.82)
   })
 
-  test('skips meals without nutrients', () => {
+  test('includes meals with a meal-level macro but no nutrients object', () => {
+    // A manually logged meal with no itemized food items has no `nutrients`
+    // (the backend only attaches it when junction links exist), but its
+    // calories show on the card and must count toward the day total.
     const meals = [makeMeal({ calories: 100 }), makeMeal({ calories: 50, nutrients: { calories: 50 } })]
+    expect(aggregateDayNutrients(meals).calories).toBe(150)
+  })
+
+  test('contributes nothing for a meal with neither nutrients nor meal-level macros', () => {
+    const meals = [makeMeal({}), makeMeal({ calories: 50, nutrients: { calories: 50 } })]
     expect(aggregateDayNutrients(meals).calories).toBe(50)
   })
 
--- a/apps/web/src/pages/Meals/dayNutrients.test.ts
+++ b/apps/web/src/pages/Meals/dayNutrients.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, test } from 'vitest'
+
+import type { Meal } from '../../state/api'
+
+import { aggregateDayNutrients } from './dayNutrients'
+
+const makeMeal = (partial: Partial<Meal>): Meal => ({ time: new Date(0), ...partial })
+
+describe('aggregateDayNutrients', () => {
+  test('sums macros across meals when meal-level and aggregate agree', () => {
+    const meals = [
+      makeMeal({ calories: 300, protein: 10, nutrients: { calories: 300, protein: 10 } }),
+      makeMeal({ calories: 200, protein: 5, nutrients: { calories: 200, protein: 5 } }),
+    ]
+    const totals = aggregateDayNutrients(meals)
+    expect(totals.calories).toBe(500)
+    expect(totals.protein).toBe(15)
+  })
+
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped fix. The day total now swaps the per-food-item aggregate for the meal-level macro (calories/protein/carbs/fat/fiber) on each meal, so the Day Totals stay consistent with the per-meal cards even when a meal is `nutrient_data_incomplete`. Micronutrients correctly remain summed from `meal.nutrients`. Extracting the logic into a pure, unit-tested `dayNutrients.ts` module aligns with the project's testability guidance.

I replicated `aggregateDayNutrients` and ran all seven test cases — every one passes, and the arithmetic in the "real day" case (2290.6) and the micronutrient case (sodium 290.96) checks out. The swap is also robust to float drift between `meal.calories` and the aggregate, since it replaces rather than adjusts.

1. **Meals with a meal-level macro but no `nutrients` object are still excluded** from the day total via `if (!meal.nutrients) continue`. A manually-logged meal with `calories` set but zero food items (backend `attachFoodItems` only sets `nutrients` when junction links exist) renders its calories on the card (`MealCaloriesLine` shows whenever `meal.calories !== undefined`) but contributes 0 to the Day Total. This is pre-existing behavior, explicitly codified by the "skips meals without nutrients" test, but it's the same card-vs-day-total inconsistency class this PR targets, reached via a different path — worth a sanity check that it's intended / out of scope.

2. Summing guards on `val > 0` while the swap subtracts `nutrientVal` unconditionally. These only diverge for negative nutrient values (zero subtracts cleanly), which don't occur in practice — no action needed.

Couldn't run `pnpm check`/`pnpm test` in this worktree (deps not installed and the sandbox blocks the pnpm store symlink), so the TypeScript narrowing of `meal.nutrients` through the second loop and the import ordering are assumed to have passed local CI.

fiddur (commented):
✅ Approved

Clean, well-targeted fix. The day-total aggregation now honors the manual-override-aware meal-level macros (`meal.calories` etc.) over the per-food-item `meal.nutrients` aggregate, which fixes the understated day total when a meal is flagged `nutrient_data_incomplete` (a food item with no nutrition contributes 0 to the aggregate but the meal card still shows the estimated macro). The follow-up commit also correctly counts manually-logged meals that have a meal-level macro but no `nutrients` object at all.

The algorithm is sound: it sums all positive nutrients first, then for each macro field replaces this meal's aggregate contribution (subtract `nutrientVal`, add `mealVal`) only when they differ — so the adjustment is per-meal and accumulates correctly across meals, micros stay summed, and it falls back to the aggregate when no meal-level macro is present. Extracting the helper into its own `dayNutrients.ts` module with co-located unit tests aligns well with the project's testability guidance.

Verification: all 8 unit tests pass, `pnpm typecheck` is clean (after building `@aurboda/api-spec`), and lint reports 0 errors (the warnings shown are pre-existing on unchanged lines). The test suite covers the agree case, override>aggregate, the real-day reproduction, micronutrient pass-through, fallback, no-`nutrients` meals, empty meals, and a lower-than-aggregate override.

Minor (non-blocking) observation: step 2 subtracts `nutrientVal` unconditionally while step 1 only adds nutrient values that are `> 0`. This is only a discrepancy if a nutrient value were negative, which doesn't occur in practice (backend `aggregateNutrients` filters `val > 0`), so no change needed.

PR #793, opened 2026-06-08
feat: event-outcome correlation engine with exposure correction + Explore UI (#792)

fiddur:
Implements #792: a generic, statistically-correct correlation engine plus an Explore UI.

### Event-outcome mode (the core of #792)
For presence-only outcomes like `back_pain`/`fissure_pain` (rows only on bad days), averaging the metric is meaningless. New mode on `get_generic_correlation` (`outcome.type: "event"`):
- **Onset-collapsing** — a multi-day flare collapses to one onset (`collapse_gap_days`, default 3).
- **Exposure/base-rate correction** — over a known-day denominator, compares onset rate among days exposed to a recent trigger vs unexposed.
- **Reverse conditional headline** — reports `P(recent trigger | onset)` (the user's actual question) beside the base rate.
- **Proper stats** — relative risk + 95% CI, exact 1-df chi-squared p-value (`erfc`), Fisher's exact fallback for small cells. Also **fixes** the old bogus `exp(-0.5·χ²)` p-value used by existing tools.
- **Regime scoping** (`period_start`/`period_end`) and a **`known` vs `all` denominator** toggle.
- **Performance** — day-level linear sweep; 2200-day windows run in <500ms (no timeout).

### Continuous mode
New `get_metric_correlation` / `POST /correlations/continuous`: Pearson + Spearman over two aligned daily series with a day-lag shift — answers "how does carb intake affect my sleep?".

### Nutrition + selectors
- `getDailyNutrientTotals` db query (UTC-day macro sums, with integration tests).
- A `Selector` abstraction resolving tag/activity/metric/nutrition/productivity to events + daily series + known days.
- `list_correlation_selectors` / `GET /correlations/selectors` discovery endpoint.

### Web
- Correlations page rebranded **Analyze** with an **Explore** tab: trigger/outcome selector builders, lag windows, regime dates, denominator toggle; results table (P(trigger|onset), base rate, RR + CI, p, n) and a continuous scatter. New nav entry.

## API surface
Per the chosen approach, the event-outcome capability **extends the existing tools** (additive only — new optional fields/response blocks, no breaking changes to existing consumers or generated Kotlin).

## Tests
- Stats vs known values (chi²/Fisher/RR-CI), onset-collapsing, exposure base-rate, selector resolution, continuous alignment.
- Synthetic acceptance-criteria fixture reproducing the ejaculation↔back_pain weak signal (~1.5×, n.s.) and a no-timeout multi-year check.
- Full backend (1843) + web (425) unit suites green; backend & web typecheck/lint clean.

## Notes / follow-ups
- Reproducing the exact production result depends on real Hälsa/Oura data on `valhall`; unit tests use a synthetic fixture matching the issue's numbers.
- Nutrition selectors currently cover the 5 meal-level macros; micronutrients are a natural follow-up.


apps/backend/src/services/correlations/event-outcome.ts
```diff
@@ -0,0 +1,260 @@
...
+): string[] => {
+  if (conditions.length === 0) return []
+  const sortedPerCondition = conditions.map((c) => c.eventDays.map(dayNumber).sort((a, b) => a - b))
+  const result: string[] = []
+  for (const day of candidateDays) {
+    const d = dayNumber(day)
+    let allMet = true
+    for (let i = 0; i < conditions.length; i++) {
+      const windowDays = Math.max(1, conditions[i].windowDays)
+      const count = countInRange(sortedPerCondition[i], d - windowDays + 1, d)
```
fiddur, line 168: `countInRange` over `eventDays` counts the number of distinct *event days* in the window, so `min_count` here means "≥N days with the event," not "≥N occurrences." The `api-spec` schema documents `min_count` as "Minimum occurrences within the window," and the legacy `getGenericCorrelation` path counts raw occurrences (`triggerEvents.filter(...).length`). For the default `min_count: 1` it's identical; the two diverge only when there are multiple same-day occurrences. Given the engine is day-granular this interpretation is defensible — but consider aligning the schema wording so the two trigger paths don't quietly mean different things for `min_count > 1`.

fiddur (commented):
✅ Approved

Strong, well-structured PR. The event-outcome engine, the statistical primitives, and the selector layer are all pure/unit-testable with good coverage, and the new `getDailyNutrientTotals` has an integration test as required by CLAUDE.md. API/MCP/REST parity is honoured (`get_metric_correlation` ↔ `POST /correlations/continuous`, `list_correlation_selectors` ↔ `GET /correlations/selectors`), schemas live in `api-spec`, snake_case is consistent, and the docs in `docs/correlations.md` are excellent. The contingency-table math in `computeEventOutcome` is correct (a/b/c/d sum to `known`, onsets ⊆ known), the chi-squared/Fisher selection and Wald RR CI are sound, and replacing the old `exp(-0.5·χ²)` p-value approximation with `erfc(√(χ²/2))` is a genuine correctness fix (no existing tests assert the old values, so nothing breaks).

- **`min_count` semantics differ from the legacy path / schema docs.** In `compoundTriggerDays`, a condition is satisfied when at least `minCount` of its *event days* fall in the window — but the schema documents `min_count` as "Minimum occurrences within the window," and the legacy `getGenericCorrelation` path counts raw occurrences. For the default `min_count: 1` these are identical; they only diverge when there are multiple same-day occurrences. Since the engine is day-granular this is arguably the only sensible interpretation, but the doc wording ("occurrences") is now slightly misleading. Inline note added.
- **`getEventOutcomeCorrelation` (single-trigger) is exported from `index.ts` but not wired to any REST route or MCP tool** — event-outcome is only reachable via the generic-correlation `event` outcome path. It looks like dead/forward-looking code; consider wiring it or dropping it.
- **`resolveWindow` (explore.ts) mixes timezones in its default branch:** explicit `period_start`/`period_end` are parsed as UTC (`T00:00:00.000Z`), but the no-arg defaults use local-time `new Date()` + `setHours`/`setDate`, after which `enumerateDays` reinterprets via `toISOString()` (UTC). On a non-UTC server the default window can be off by a day. Harmless if the server runs UTC, but worth normalising.
- **Nutrition trigger + non-event outcome silently yields nothing:** the legacy `getGenericCorrelation` branch has no `'nutrition'` trigger case, so such a combination produces an empty result rather than an error. Edge combination, but a guard or validation would be friendlier.
- **Default `denominator: 'known'` is a footgun for the flagship presence-only use case** (e.g. `back_pain`), where `known` collapses the denominator to flare days only. This is clearly documented and user-selectable, so just flagging it.

docs/correlations.md
```diff
@@ -0,0 +1,109 @@
...
+- Fisher's exact test (two-sided) is used automatically for small expected cell
+  counts.
+- Relative-risk CIs use the Wald log-RR standard error; they are omitted when an
+  outcome cell count is zero.
+
+## Performance
+
+Analysis is day-level with linear exposure sweeps, so multi-year regimes
+(2000+ days) run without timing out.
+```
```
fiddur: Stray trailing code fence — this ``` has no matching opener (the doc ends on the previous prose line), so it renders as a dangling/empty code block. Drop it.

fiddur (commented):
✅ Approved

Solid, well-structured feature. The statistical core in `stats.ts` is correct and properly tested: the 1-df chi-squared p-value via `erfc(√(χ²/2))` (fixing the old `exp(-0.5·χ²)` approximation), two-sided Fisher's exact for small expected cells, risk ratio with a Wald log-RR CI, and Pearson/Spearman with tie-aware ranking. The event-outcome engine (onset-collapsing, lag-window exposure via binary search, known-day denominator, reverse-conditional headline) is internally consistent — the 2×2 table cells sum to the known-day total and map correctly onto the epidemiology convention used by `riskRatio`/`significance2x2`. UTC-day bucketing is consistent across selectors, the SQL nutrient totals (`TO_CHAR(... AT TIME ZONE 'UTC')`), and the engines. Test coverage is strong (stats, event-outcome, continuous, selectors, generic, plus DB integration tests for the new `getDailyNutrientTotals`/`getMealLogCompletedInRange`). API/MCP parity is maintained and api-spec is the single source of truth. Note: the OpenAPI/Kotlin artifacts under `packages/api-spec/generated/` are gitignored, so not regenerating them in the PR is expected, not a gap.

Non-blocking observations:

- **MCP casts vs. project standard.** `apps/backend/src/mcp/correlation-tools.ts` casts `triggers as TriggerCondition[]`, `outcome as OutcomeConfig`, and `trigger/outcome as Selector` to bridge the api-spec-inferred types to the parallel hand-written service types (selectors.ts `Selector`, types.ts `OutcomeConfig`/`TriggerCondition`). This is the kind of casting CLAUDE.md asks to avoid; the router passes the same values without casts, so consider having the services consume the api-spec types directly (as the api-spec guidance suggests) to drop the casts. Pre-existing pattern, so non-blocking.

- **Explore trigger picker can hold an out-of-range kind.** In `Explore.tsx`, `triggerSelector` persists across mode switches. If a user picks a `metric` trigger in Continuous mode then switches to Event mode (where `metric` isn't an allowed trigger kind), the `<select>` shows a value with no matching option and `buildRequest` silently falls back to `{ pattern: '', type: 'tag' }`, whose empty regex matches every activity → every day becomes a trigger day (base_rate ≈ 1, garbage result). Consider resetting/clamping the trigger selector to an allowed kind when the mode changes.

- **`period.days` off-by-one (display only).** `resolveWindow` sets `start` to `endDay 00:00 − periodDays·day` and `end` to `endDay 23:59:59.999`, so `days = round((end−start)/86_400_000)` reports `periodDays + 1`. The analysis window itself is correct; only the echoed `days` is one high.

Neither blocks merge.

fiddur (commented):
✅ Approved

Solid, well-engineered PR. The new event-outcome engine (onset-collapsing, exposure correction, 2×2 contingency table → relative risk + Wald CI + chi-squared/Fisher) is statistically sound: the contingency mapping `a/b/c/d` is correct, all cells are provably non-negative, and the reverse-conditional/base-rate headline matches the docs. Replacing the old `exp(-0.5·χ²)` p-value approximation with `erfc(√(χ²/2))` is a genuine correctness fix that also improves `get_event_probability` and the legacy generic tag path. The pure engines (`computeEventOutcome`, `computeContinuous`, `stats.ts`) are dependency-free and thoroughly unit-tested, including degenerate tables, multi-year performance, and a realistic weak-signal scenario; `selectors.ts`/`generic.ts` are covered with mocked-db tests, and the new `getDailyNutrientTotals`/`getMealLogCompletedInRange` paths have integration tests. Conventions are followed well: `typedRouter()` with proper api-spec generics (no `req.body` casts), snake_case throughout, `.shape` for MCP, functional style, UTC day-bucketing consistent between SQL (`AT TIME ZONE 'UTC'`) and TS, and the `getDailyNutrientTotals` dynamic SQL is safe (column names come from the fixed `NUTRIENT_KEYS` whitelist, values parameterized). Docs and MCP/REST parity are in place.

Non-blocking notes:

- **Reported `period.days` is off-by-one for trailing-N requests.** In `explore.ts` `resolveWindow`, the trailing-N window spans `[endDay − periodDays, endDay]` inclusive, which is `periodDays + 1` calendar days, and `days = round((end − start)/MS_PER_DAY)` then reports `periodDays + 1` (e.g. 91 for a "last 90 days" request). The explicit start/end path is exact (Jan 1–31 → 31). This only affects the displayed `period.days` count, not the analysis, but you may want to subtract a day from the start or report `periodDays` directly.

- **Redundant casts in the MCP layer.** `correlation-tools.ts` uses `outcome as Selector` / `trigger as Selector` (and `outcome as OutcomeConfig` / `triggers as TriggerCondition[]`), whereas the REST router passes the same api-spec body types into the same service functions without any cast — implying the casts are unnecessary and they brush against the "avoid casting" guideline. Worth dropping them (or aligning the service signatures to the api-spec types) if they compile clean without.

- Minor UI inconsistency: in continuous mode `productivity_app` is offered as a trigger kind but omitted from the outcome picker's `allowedKinds`. Harmless, just asymmetric.

None of these block merge.

fiddur (commented):
✅ Approved

Solid, well-tested PR. The new event-outcome correlation engine is cleanly factored into pure, dependency-free functions (`computeEventOutcome`, `computeContinuous`, `compoundTriggerDays`) with thorough unit tests, and the statistical primitives in `stats.ts` (erfc-based χ² p-value, Fisher's exact, Wald-CI relative risk) are correct against the known reference values the tests assert. I verified the 2×2 contingency cells in `computeEventOutcome` can't go negative (unexposed onsets ≤ unexposed days, exposed onsets ≤ exposed days), the binary-search exposure/range helpers are correct, and the onset-collapsing matches its documentation. Replacing the old `exp(-0.5·χ²)` approximation with `erfc(√(χ²/2))` is a genuine correctness fix for the pre-existing `chiSquaredTest`. API-spec is the single source of truth (schemas → router via `typedRouter` → MCP via `.shape`), snake_case is consistent, the new `getDailyNutrientTotals` DB function has integration tests, and docs are added and accurate.

Non-blocking observations:

1. **Denominator default for presence-only metrics.** The default `denominator: 'known'` is degenerate for the canonical issue-#792 use case (a presence-only metric like `back_pain` only logs rows on bad days, so 'known' days ≈ onset days). Users must know to switch to 'all'. This is documented in `docs/correlations.md` and exposed as a UI toggle, so it's an intentional tradeoff — but the Explore intro text doesn't explain the denominator, so it's an easy footgun for the very scenario the feature targets. Consider defaulting the UI to 'all' when the outcome source is a metric, or adding a one-line hint near the toggle.

2. **Casts at the MCP boundary** (`outcome as Selector`, `trigger as Selector` in `correlation-tools.ts`) slightly conflict with the project's "avoid casting, use type guards" guideline, but they mirror the pre-existing `triggers as TriggerCondition[]` pattern and bridge two structurally-identical (api-spec-inferred vs hand-written) types, so they're consistent with the surrounding code.

3. **Minor:** `period.days` for a trailing window reports `periodDays + 1` (e.g. requesting "last 90 days" yields `days: 91`) because the range spans `[endDay − periodDays, endDay]` inclusive. It matches the actual enumerated-day count, so it's defensible, just slightly surprising versus the requested number.

I couldn't execute the test suite (no `node_modules` in this worktree), but the test files are well-aligned with the implementations and the mocks cover the functions each path touches.

apps/backend/src/services/correlations/explore.ts
```diff
@@ -0,0 +1,187 @@
...
+export const getCompoundEventOutcome = async (
+  user: string,
+  triggers: TriggerCondition[],
+  outcome: EventOutcomeConfig,
+  lagWindows: string[],
+  options: { periodStart?: string; periodEnd?: string; periodDays?: number; denominator?: 'known' | 'all' },
+): Promise<CompoundEventOutcomeResult> => {
+  const { start, end } = resolveWindow(options)
+  const collapseGapDays = outcome.collapse_gap_days ?? 3
+  const denominator = options.denominator ?? 'known'
```
fiddur, line 135: Non-blocking: the `'known'` default is degenerate for presence-only metrics (e.g. `back_pain`), where the metric only logs rows on flare days — so the known-day denominator collapses to roughly the onset days themselves and the base-rate/relative-risk comparison becomes meaningless. That's exactly the issue-#792 use case. It's documented and toggleable in the UI, but since the Explore intro doesn't mention the denominator, consider defaulting to `'all'` when `outcome.source === 'metric'` (or surfacing a hint), so the headline scenario works without the user knowing to flip it.

apps/backend/src/services/correlations/generic.ts
```diff
@@ -65,7 +67,19 @@ export async function getGenericCorrelation(
   lagWindows: string[] = ['24h', '48h', '7d'],
   periodDays: number = 90,
   sync?: SyncProvider,
+  options: GenericCorrelationOptions = {},
```
fiddur, line 70: `options.periodStart`/`options.periodEnd` are only consulted on the `event` path (via `getGenericEventOutcome` → `resolveWindow`). For `tag`/`metric`/`productivity` outcomes the window below is always `new Date()` minus `periodDays`, so a caller passing `period_start`/`period_end` with a non-event outcome silently gets last-`periodDays` results. This conflicts with the schema, which exposes these fields for all generic-correlation bodies and documents `period_days` as "Ignored when period_start is given." Either apply the regime params in the averaging path too, or validate/document that they're event-only.

fiddur (commented):
🛑 Changes required

Strong, well-engineered PR. The statistical core (`stats.ts`) is correct and a genuine improvement: `erfc`/`chiSquaredPValue1df` replace the old `exp(-0.5·χ²)` approximation, Fisher's exact test kicks in for small expected cells, and the Wald RR/CI math is right. The event-outcome onset-collapsing + exposure-correction model is clearly documented and the 2×2 contingency table (`a/b/c/d`) is constructed soundly (all onsets are pre-filtered to known days, so `d ≥ 0`). Test coverage is excellent — unit tests for the stats primitives, the onset/exposure engine (including realistic weak-signal and multi-year-performance cases), continuous correlation, selectors, and the generic event-outcome route, plus integration tests for the new `getDailyNutrientTotals`. Schema/type/MCP/REST parity all look consistent, the new `GET /correlations/selectors` route isn't shadowed by `/activity-impact/:activity`, and the SQL in `getDailyNutrientTotals` correctly whitelists columns and buckets by UTC.

One issue worth addressing before merge:

- **`period_start`/`period_end` are silently ignored for non-`event` outcomes in `getGenericCorrelation`.** When `outcome.type` is `tag`/`metric`/`productivity`, the function computes the window from `new Date()` minus `periodDays` and never consults `options.periodStart`/`options.periodEnd` — only the `event` path (via `resolveWindow`) honors them. This contradicts the schema, where `period_start`/`period_end` are advertised on `genericCorrelationBodySchema` for all outcomes and `period_days` is documented as "Ignored when period_start is given." The Explore UI only sends `period_start` in event mode, so it's unaffected, but an MCP/REST consumer scoping a tag/metric correlation to a regime would get last-90-days results instead. Either honor the regime params in the averaging path too, or constrain/document that they only apply to event outcomes. (Detail noted inline.)

Minor (non-blocking): in `Explore.tsx` the event-mode trigger kinds are duplicated (module-level `EVENT_TRIGGER_KINDS` vs. the inline array in the render), and the continuous-mode outcome picker omits `productivity_app`.

fiddur (commented):
✅ Approved

Solid, well-engineered feature. The event-outcome (onset-collapsing + exposure correction) and continuous engines are implemented as pure, dependency-free functions with thorough unit tests, and the statistical primitives (erfc/chi-squared 1df, Fisher's exact, risk ratio + Wald CI, Pearson/Spearman) are validated against known reference values in `stats.test.ts`. The earlier inaccurate `exp(-0.5·χ²)` p-value is correctly replaced. Architecture follows the repo standards well: shared logic in `services/`, schemas as the single source of truth in `@aurboda/api-spec`, REST router passes typed bodies without casts, the new DB function `getDailyNutrientTotals` has integration tests, and UTC day-bucketing is consistent across nutrients/metrics/events. Docs (`docs/correlations.md`) are clear and accurate.

I verified the 2×2 contingency construction in `computeEventOutcome`, the binary-search exposure sweep, onset-collapsing, the known-day denominator filtering, and the degenerate-table guards (empty/zero-margin tables return p=1 and null RR without crashing). No blocking correctness issues found.

Non-blocking observations (no action required to merge):

1. **Default denominator for presence-only metrics via MCP/REST.** The web UI cleverly defaults `denominator` to `all` when the event-outcome source is a metric (`setEventOutcomeSource`), but the API/MCP default is `known`. For the exact #792 use case (a presence-only metric like `back_pain` that only logs on bad days), an agent calling `get_generic_correlation` without specifying `denominator` gets `known`, which collapses the denominator to roughly the onset days and makes the base-rate comparison degenerate. This is documented in `docs/correlations.md`, but it might be worth nudging the agent toward `denominator: "all"` for presence-only metrics in the `get_generic_correlation` tool description / `eventOutcomeSchema` so the headline result isn't silently meaningless.

2. **Nutrition trigger + non-event outcome throws at runtime.** `getGenericCorrelation` throws `'Nutrition triggers are only supported with an "event" outcome'` for the averaging path. The schema permits that combination (no cross-field validation), so a client can trigger a 500 rather than a 400. A refinement on `genericCorrelationBodySchema` (or a typed error) would make this a clean client error.

3. **Minor duplication.** `enumerateDays` exists in both `selectors.ts` and `explore.ts`, and `mean`/`stddev` are duplicated between `stats.ts` and `utils.ts` (utils now also re-uses stats for chi-squared). Worth consolidating eventually, but harmless.

Nice touches: the `min_count` day-granular-vs-raw-occurrence semantic difference between the event and averaging paths is documented in the schema; the SQL column interpolation in `getDailyNutrientTotals` is guarded by the `NUTRIENT_KEYS` whitelist with a comment; and TZ-safe date formatting (`TO_CHAR ... AT TIME ZONE 'UTC'`) is used consistently.

PR #796, opened 2026-06-08
fix(health-connect): dedupe insertRawRecords batch to avoid 21000 (#772)

fiddur:
Closes #772.

## Problem
A Health Connect re-sync chunk can include the same `ExerciseSessionRecord` twice (same `metadata.id`). `insertRawRecords` fed the whole batch into one `ON CONFLICT (source, record_type, external_id) DO UPDATE`, so two rows hitting the same key raised `21000: ON CONFLICT DO UPDATE command cannot affect row a second time` — aborting the batch and blocking sync (Sara's case). #770 fixed this for `insertActivities` but the HC path runs `insertRawRecords` first.

## Fix
- Add a small shared `dedupeLastWins(items, keyFn)` helper (the pattern #770 inlined into `insertActivities`; the issue suggested extracting it).
- `insertRawRecords` now dedupes the batch by `(source, record_type, external_id)` (last write wins), keeping rows with a NULL `external_id` as-is since they're distinct in the unique index.

## Tests
- Unit test for `dedupeLastWins` (last-wins, null keys never deduped, order preserved).
- Integration tests: a batch with two records sharing `external_id` no longer throws and collapses to one row (second wins); two NULL-`external_id` rows are both kept.

### DIFF
--- a/apps/backend/src/db/dedupe.test.ts
+++ b/apps/backend/src/db/dedupe.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, test } from 'vitest'
+
+import { dedupeLastWins } from './dedupe.ts'
+
+describe('dedupeLastWins', () => {
+  test('keeps the last item per key, preserving first-seen order', () => {
+    const items = [
+      { k: 'a', v: 1 },
+      { k: 'b', v: 2 },
+      { k: 'a', v: 3 },
+      { k: 'c', v: 4 },
+      { k: 'b', v: 5 },
+    ]
+    const result = dedupeLastWins(items, (i) => i.k)
+    expect(result).toEqual([
+      { k: 'a', v: 3 },
+      { k: 'b', v: 5 },
+      { k: 'c', v: 4 },
+    ])
+  })
+
+  test('never dedupes items whose key is null', () => {
+    const items = [
+      { k: null, v: 1 },
+      { k: null, v: 2 },
+      { k: 'x', v: 3 },
+      { k: 'x', v: 4 },
+    ]
+    const result = dedupeLastWins(items, (i) => i.k)
+    expect(result).toEqual([
+      { k: null, v: 1 },
+      { k: null, v: 2 },
+      { k: 'x', v: 4 },
+    ])
+  })
+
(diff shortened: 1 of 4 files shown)


fiddur (commented):
✅ Approved

Solid, focused fix for the `21000: ON CONFLICT DO UPDATE command cannot affect row a second time` error in `insertRawRecords`.

**Correctness verified:**
- The dedupe key `(source, record_type, external_id)` exactly matches the table's conflict target (`CONSTRAINT unique_source_record UNIQUE (source, record_type, external_id)` in `schema/metrics.ts`), so it collapses precisely the rows that would otherwise trip 21000.
- NULL/`undefined` `external_id` is handled correctly: the constraint is a plain `UNIQUE` (no `NULLS NOT DISTINCT`), so Postgres treats NULL external_ids as distinct and they never collide — `dedupeLastWins` returns `null` for them and keeps them all. The `== null` loose check also covers the `undefined` case, which matters since `RawRecord.external_id` is optional and HC's `data.metadata?.id` can be `undefined`.
- Last-write-wins with first-seen ordering matches the upsert's `EXCLUDED.data` semantics and is consistent with the existing `insertActivities` dedupe (#770).

**Tests:** Good coverage — the `dedupe.test.ts` unit tests exercise ordering, null keys, empty input, and passthrough; the integration tests cover both the 21000-avoidance (last wins) and the NULL-`external_id` keep-all cases, with `beforeEach(cleanTestDb)` isolation. Extracting `dedupeLastWins` into a reusable, unit-tested helper is a nice improvement over the inline `Map` logic.

**Optional (non-blocking):** The pipe-delimited composite key could in theory collide if a field contained `|` (e.g. `source='a|b'` + `id='c'` vs `source='a'` + `id='b|c'`), which would silently over-dedupe. In practice `source`/`record_type` are controlled values and HC `external_id`s are UUIDs, and this mirrors the established `insertActivities` pattern, so it's fine as-is — just noting it. Likewise, refactoring `insertActivities` to reuse this helper would be a reasonable future DRY cleanup but isn't warranted here since it splits into two conflict keys with different semantics.

PR #797, opened 2026-06-08
fix(calories): convert height metres to cm before Mifflin-St Jeor BMR (#776)

fiddur:
Closes #776.

## Problem
The `height` metric is stored in **metres** (canonical unit `'m'` in `common.ts:246`; Health Connect ingestion writes `heightInMeters` straight to the value). But `computeAndStoreCalories` passed it to `resolveBmr` as `height_cm`. So any user without a lab `basal_metabolic_rate` (i.e. the Mifflin-St Jeor fallback) got a BMR computed with a height ~100× too small.

Demo profile (78.9 kg, 1.8 m, male, 41 y/o):
- correct ≈ 10·78.9 + 6.25·**180** − 5·41 + 5 = **1714 kcal/day**
- buggy ≈ 10·78.9 + 6.25·**1.8** − 5·41 + 5 = **600 kcal/day**

## Fix
Convert the `height` metric (metres) to centimetres at the call site before passing it to `resolveBmr`.

## Test
Unit test for the Mifflin path (mocked db: height 1.8 m, weight 78.9 kg, no lab BMR, no HR) asserting the stored `calories_total` BMR floor ×1440 lands in 1500–2000 kcal/day — it would have asserted ~600 and failed before the fix.

### DIFF
--- a/apps/backend/src/services/calorie-computation-bmr.test.ts
+++ b/apps/backend/src/services/calorie-computation-bmr.test.ts
@@ -1,0 +1,69 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest'
+
+import * as dbIndex from '../db/index.ts'
+import * as timeSeries from '../db/time-series.ts'
+import { computeAndStoreCalories } from './calorie-computation.ts'
+
+vi.mock('../db/index.ts', () => ({
+  enqueueOutboundSync: vi.fn(),
+  getUserSettings: vi.fn(),
+  upsertUserSettings: vi.fn(),
+}))
+
+vi.mock('../db/time-series.ts', () => ({
+  deleteTimeSeriesBySource: vi.fn(),
+  getMetricTimeRange: vi.fn(),
+  getTimeSeries: vi.fn(),
+  insertTimeSeries: vi.fn(),
+}))
+
+describe('computeAndStoreCalories — Mifflin-St Jeor BMR from height metric', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Correct, minimal fix for a real unit bug. The `height` metric is canonically stored in metres (`packages/api-spec/src/schemas/common.ts:246` declares `height: 'm'`, and Health Connect ingest writes `heightInMeters` directly at `apps/backend/src/db/health-connect.ts:327`), while `estimateBmrMifflinStJeor` expects centimetres (`6.25 * heightCm`). Passing the raw metres value as `height_cm` made the Mifflin-St Jeor fallback compute ~600 kcal/day instead of ~1714. The `* 100` conversion with a null guard is the right fix, the comment explains why, and `resolveBmr`'s `height_cm <= 0` check is preserved. I confirmed this is the only consumer of the height metric feeding a BMR calculation, so there are no other sites with the same bug.

The new regression test correctly isolates the BMR floor (no HR data → every `calories_total` minute equals BMR/min) and asserts a realistic band that the buggy path would fail. Traced by hand it passes (1714 kcal/day). I could not execute it here because dependencies aren't installed in this worktree and the sandbox blocks fetching them, but the logic checks out.

Two non-blocking notes:
- Historical `calories_total`/`calories_active` rows computed while the bug was live still hold the wrong (~600 kcal/day BMR floor) values; they won't self-correct until a full recompute (`computeAndStoreCaloriesAll`) runs for affected users. Worth a one-time recompute if not already planned.
- The test's age is derived from `new Date()` via `calculateAge`, so it's mildly time-dependent — but the 1500–2000 assertion band tolerates many years of age drift, so it won't flake.

PR #798, opened 2026-06-08
fix(food-items): FoodItemDetail.portions is always [] when empty (#780)

fiddur:
Closes #780.

## Problem
`GET /api/food-items/:id` returned `"portions": null` (not `[]`) when a food item had zero portions — a shape-contract violation. `getDetail` set `portions = portionRows.length > 0 ? portionRows : undefined`, and the router's `detail.portions?.map(...)` then yielded null/absent. Any consumer doing `detail.portions.map(...)` / `.length` throws.

## Fix
Always return the array (empty when none):
- service `getDetail`: `const portions = portionRows`
- router `serializeDetail`: `(detail.portions ?? []).map(serializePortion)` (defensive boundary)

So the detail contract is `portions: FoodItemPortion[]` on every path. The dedicated `GET /:id/portions` endpoint already returned `[]`; this aligns the detail payload.

## Test
`getDetail` unit test asserting `portions` is `[]` for an item with no portions.

### DIFF
--- a/apps/backend/src/routes/food-items-router.ts
+++ b/apps/backend/src/routes/food-items-router.ts
@@ -117,7 +117,8 @@ export const resolveEffectiveDefaultQuantity = (
 const serializeDetail = (detail: ServiceFoodItemDetail): FoodItemDetail => {
   const base = serializeFoodItem(detail.item)
   const sensitivities = detail.sensitivities ?? []
-  const portions = detail.portions?.map(serializePortion)
+  // Always an array so the response is `portions: []`, never null/absent (#780).
+  const portions = (detail.portions ?? []).map(serializePortion)
   const is_shared = detail.is_shared
   // For per-user items default_portion_id is the column on the food row;
   // for central items applySharedOverrides decorated it from the user's
--- a/apps/backend/src/services/food-items.test.ts
+++ b/apps/backend/src/services/food-items.test.ts
@@ -212,6 +212,17 @@ describe('createFoodItemsService.getById', () => {
   })
 })
 
+describe('createFoodItemsService.getDetail portions', () => {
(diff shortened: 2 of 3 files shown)


fiddur (commented):
✅ Approved

Correct, focused fix. `getDetail` now returns `portions: []` instead of `undefined` when a food item has no portions, and `serializeDetail` is hardened with `(detail.portions ?? [])` so the REST response always carries `portions: []`. I verified the `undefined` → `[]` change is side-effect-free: the only consumer of `ServiceFoodItemDetail.portions` is `serializeDetail` itself — the MCP tools (`mcp/food-item-tools.ts`) and `services/meals.ts` query `listPortionsForFoodItem` directly rather than reading `detail.portions`, so no truthiness-based branch flips. The new unit test is sound: `listPortionsForFoodItem` is mocked to `[]` at module level and there's no `clearMocks`/`mockReset` in the vitest config or setup file, so the default holds for the new case.

One non-blocking follow-up: the code comments promise that "consumers can rely on `.map`/`.length` without a null check (#780)", but `foodItemDetailSchema.portions` in `packages/api-spec/src/schemas/food-items.ts:174` is still `.optional()`. Since api-spec is the single source of truth, the generated TS/Kotlin types still type `portions` as optional/nullable — and indeed web consumers still defensively null-check it (`apps/web/src/pages/FoodItems/FoodItemDetail.tsx:666` does `item.portions ?? []`, `MealDetail.tsx` uses `detail.portions?.find(...)`). To fully deliver the stated type-level contract, drop `.optional()` on the schema and regenerate (OpenAPI/TS/Kotlin) per the api-spec workflow. The runtime behavior is already correct as-is.

PR #799, opened 2026-06-08
fix(activities): include comments in GET /activities/:id so Notes render (#794)

fiddur:
Closes #794.

## Problem
The activity-detail page loads via `fetchActivityById` → `GET /activities/:id`, but that response omitted `comments`. So `getUserNotesContent(activity)` returned `''`, the unified stats table's **Notes row never rendered**, and edit mode showed an **empty Notes textarea** even when the activity had a user note. (Notes still showed via the separate NotesSection, so impact was limited, but the table row + edit field were dead — and #789's unit test passed in isolation while the integration path was broken.)

## Fix (chose "populate comments" — fixes both symptoms)
Attach `comments` to the activity-detail response on both paths:
- **plain UUID**: `getCommentsMap(user, 'activity', [realId])`
- **merged**: collect + dedupe comments across all merged sources (mirrors the list path)

The response schema already permits `comments` and the web layer already reads `activity.comments`, so no schema or frontend change is needed — the Notes row and the edit-mode textarea now reflect the real note.

## Tests
New `activities-router.test.ts` (supertest): `GET /activities/:id` carries the user note as `comments`, and returns `[]` when there are none. The comment-attaching logic itself is already covered by `queries/activities.test.ts`.

### DIFF
--- a/apps/backend/src/routes/activities-router.ts
+++ b/apps/backend/src/routes/activities-router.ts
@@ -50,6 +50,7 @@ import {
 import {
   computeActivityDetailMetrics,
   getActivityFullDetail,
+  getCommentsMap,
   parseActivityId,
   parseMetricsParam,
   queryActivities,
@@ -111,9 +112,21 @@ const buildMergedResponse = async (
     })
   }
 
+  // Collect user notes/comments across every merged source (deduped by id) so
+  // the detail page's Notes row reflects the real note (#794).
+  const commentLookupIds = overlapping.map((a) => a.id).filter((id): id is string => Boolean(id))
+  const commentsMap = await getCommentsMap(user, 'activity', commentLookupIds)
+  type CommentEntry = NonNullable<ReturnType<typeof commentsMap.get>>[number]
+  const seenComments = new Map<string, CommentEntry>()
+  for (const id of commentLookupIds) {
+    for (const c of commentsMap.get(id) ?? []) seenComments.set(c.id, c)
+  }
+  const comments = [...seenComments.values()]
+
   return {
     ...metrics,
     activity_type: activity.activity_type,
+    comments,
     data: Object.keys(mergedData).length > 0 ? mergedData : activity.data,
     end_time: activity.end_time?.toISOString(),
     id: activity.id,
@@ -468,11 +481,16 @@ export const createActivitiesRouter = (
       }
     }
 
+    // Attach user notes/comments so the detail page's Notes row and edit-mode
+    // notes textarea reflect the real note (#794).
+    const commentsMap = await getCommentsMap(user, 'activity', [realId])
+
     // Plain UUID: return raw single activity (no overlap lookup)
     res.json({
       data: {
         ...activityMetrics,
         activity_type: activity.activity_type,
+        comments: commentsMap.get(realId) ?? [],
         data: activity.data,
         deleted_at: activity.deleted_at?.toISOString(),
         end_time: activity.end_time?.toISOString(),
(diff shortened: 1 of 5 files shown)


apps/backend/src/routes/activities-router.ts
```diff
@@ -111,9 +112,21 @@ const buildMergedResponse = async (
     })
   }
 
+  // Collect user notes/comments across every merged source (deduped by id) so
+  // the detail page's Notes row reflects the real note (#794).
+  const commentLookupIds = overlapping.map((a) => a.id).filter((id): id is string => Boolean(id))
```
fiddur, line 118: Non-blocking: this dedupe-by-note-id loop is the same logic as `enrichActivity` in `services/queries/activities.ts` (`[a.id, ...source_ids]` → `Map` by `c.id`). Worth extracting into a shared helper (e.g. in `queries/types.ts` next to `getCommentsMap`) so the merged-detail and list paths can't drift. Fine to leave for a follow-up.

fiddur (commented):
✅ Approved

Clean, correctly-targeted fix for #794. The Notes row on the activity detail page reads `activity.comments`, and that page fetches `GET /activities/:id` via `fetchActivityById` (not `/full`), so attaching `comments` to this endpoint's response is exactly the right place.

Verification notes:
- **Schema-compatible**: `comments` is already an optional field on `activitySchema` (which `ActivityDetailResponse` wraps), and the `CommentSummary` shape returned by `getCommentsMap` is structurally assignable to `commentSchema`. No type or OpenAPI change is required, and the inline-literal `comments: commentsMap.get(realId) ?? []` won't trip an excess-property check.
- **Merged path dedup** mirrors the existing `enrichActivity` pattern (collect comments across all source ids, dedupe by note id). It's actually slightly more complete here since `getOverlappingActivities` already returns the full merged group (winner + siblings).
- **Plain vs merged split is consistent**: merged rows are navigated to with the `merged:` prefix (list results set `id` to `merged:${a.id}`), which routes to `buildMergedResponse` and picks up sibling comments; a plain single-source activity correctly only resolves its own notes.
- **Tests** cover the populated and empty-notes cases for the plain path and assert the `getCommentsMap('tester', 'activity', [id])` call. I was unable to execute the suite (the sandbox pnpm store is read-only, so dependencies won't install), but the test logic matches the handler.

Non-blocking: the dedupe-by-note-id block is now duplicated between `enrichActivity` (services/queries/activities.ts) and `buildMergedResponse`. Could be lifted into a small shared helper later, and the merged path's comment behavior isn't unit-tested here (only the plain path is). Neither blocks merge.

fiddur (commented):
✅ Approved

Clean, well-scoped fix for #794. `GET /activities/:id` now attaches user notes as `comments` on both the plain-UUID path (`getCommentsMap(...).get(realId) ?? []`) and the merged path (collected across all overlapping source rows and deduped by note id), so the detail page's Notes row and edit-mode textarea render the real note.

Verified:
- The new `dedupeCommentsForIds` helper is an exact, behavior-preserving extraction of the dedup logic previously inlined in `enrichActivity`, now shared by the list path and the merged-detail response — good de-duplication of logic per the project's normalized/DRY guidance.
- `activitySchema.comments` already exists in `@aurboda/api-spec`, and `CommentSummary` (required `created_at`/`updated_at`) is assignable to the optional `Comment` schema, so the typed router stays sound.
- Barrel re-exports `getCommentsMap`/`dedupeCommentsForIds` once each (no duplicate-export hazard).
- `getNotesByEntityIds` short-circuits on empty id arrays, so the merged path is safe even if the overlap set were empty.
- Tests cover the plain path with and without notes and assert the `getCommentsMap('tester', 'activity', [id])` call signature.

Non-blocking nit: the merged-detail path (`buildMergedResponse`) has no direct test of its comment aggregation, though it reuses the same shared helper the list-path tests already exercise. Consider adding a merged-case test if convenient. (Could not run the test/typecheck locally — node_modules is not installed in this review worktree — so the above is from static analysis; CI should confirm.)

PR #800, opened 2026-06-09
fix(correlations): don't block analysis on live external syncs (#792)

fiddur:
## Problem

Issue #792's `get_metric_correlation` (and `get_generic_correlation`) **timed out on every call** — see the issue comment, where even a **14-day dense window** hung. Since a tiny window has trivial DB work, this was never data volume: it's a **fixed per-call cost**.

## Root cause

`getContinuousCorrelation` (and both branches of `getGenericCorrelation`) `await Promise.all([...])` of **four external syncs** — Oura tags, Oura sessions, RescueTime, calendars — before reading any data:

```ts
if (sync) await Promise.all(triggerSyncs(sync, user))
```

A first or stale sync makes live HTTP calls that can take many seconds — long enough to blow the MCP/HTTP request timeout no matter how small the analysis window. `list_correlation_selectors` worked precisely because it does **no** sync. All the underlying DB queries (`getTimeSeries`, `getDailyNutrientTotals`, `getAllActivitiesInRange`) are date-bounded and indexed.

## Fix

Fire the syncs **fire-and-forget** so data is warm for the *next* request without blocking *this* one — the exact pattern already used in `services/queries/productivity.ts` (`void sync.syncRescueTimeIfNeeded(user)`). Correlation analysis is retrospective over historical data, so the freshest few minutes are irrelevant to the result.

- New shared helper `triggerCorrelationSyncs(sync, user)` (`background-sync.ts`), used by both the continuous (`explore.ts`) and generic (`generic.ts`) engines.
- Uses `Promise.allSettled` + `void` so a rejected sync can never become an unhandled rejection.
- Unit test asserts it returns synchronously without awaiting slow syncs.

## Testing

- `pnpm check` — 0 errors
- `pnpm exec vitest run src/services/correlations/` — 65 passed (incl. 3 new)

## Out of scope / follow-up

This fixes the reported timeouts (incl. the 14-day case). For very long multi-year windows there's a secondary cost: `getAllActivitiesInRange` fetches all activities then filters by regex in JS — worth a follow-up if multi-year tag→metric calls still feel slow. The event-outcome / base-rate half of #792 also remains open per the issue comment.

### DIFF
--- a/apps/backend/src/services/correlations/background-sync.test.ts
+++ b/apps/backend/src/services/correlations/background-sync.test.ts
@@ -4,16 +4,18 @@ import type { SyncProvider } from '../queries/index.ts'
 
 import { triggerCorrelationSyncs } from './background-sync.ts'
 
-/** A SyncProvider whose methods record calls and never resolve, to prove the
- *  helper does not await them. */
+/** A full SyncProvider whose methods record their call and never resolve, to
+ *  prove the helper does not await them. Built as a complete literal (rather
+ *  than a cast partial) to honour the repo's "avoid casting" guideline. */
 const makeHangingSync = (): { sync: SyncProvider; calls: string[] } => {
   const calls: string[] = []
   const never = () => new Promise<void>(() => {})
-  const sync = {
-    syncOuraIfNeeded: vi.fn((_u: string, t: string) => {
-      calls.push(`oura:${t}`)
+  const sync: SyncProvider = {
+    syncOuraIfNeeded: vi.fn((_user: string, dataType: 'tags' | 'sessions') => {
+      calls.push(`oura:${dataType}`)
       return never()
     }),
+    syncGarminIfNeeded: vi.fn(() => never()),
     syncRescueTimeIfNeeded: vi.fn(() => {
       calls.push('rescuetime')
       return never()
@@ -22,7 +24,8 @@ const makeHangingSync = (): { sync: SyncProvider; calls: string[] } => {
       calls.push('calendars')
       return never()
     }),
-  } as unknown as SyncProvider
+    syncLastFmIfNeeded: vi.fn(() => never()),
+  }
   return { sync, calls }
 }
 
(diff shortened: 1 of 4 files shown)


fiddur (commented):
✅ Approved

Clean, focused fix for #792. The change converts the previously-awaited `Promise.all([...syncs])` in the correlation engines into a fire-and-forget call, so a first/stale external sync (Oura/RescueTime/calendars) can no longer blow the MCP/HTTP request timeout on a retrospective analysis. It also de-duplicates the identical 4-call sync block that lived in three places into a shared `triggerCorrelationSyncs` helper, which is good.

Correctness looks solid:
- The trigger points are preserved exactly — the event path triggers via `getGenericEventOutcome`, the non-event path via `getGenericCorrelation` (after the early `return` at line 75), and `getContinuousCorrelation` keeps its own call. No double-trigger, no path loses its sync.
- Using `Promise.allSettled` is actually an improvement over the `void Promise.all([...])` pattern in sibling modules (`queries/tags.ts`, `queries/daily-summary.ts`): with `Promise.all`, a rejected sync becomes an unhandled rejection, whereas `allSettled` absorbs it. Good defensive choice.
- `SyncProvider` imports remain in use in both `explore.ts` and `generic.ts` (function signatures), and the removed `triggerSyncs` helper has no remaining references.
- The deliberate tradeoff (the triggering request may see slightly stale data; data is fresh for the next request) is clearly documented in the new module's header comment.

I could not run `pnpm check`/tests in this worktree (dependencies are not installed — no vitest/tsgo binaries), so the build/test status is unverified here, but the new tests read correctly and CI will cover it.

Minor, non-blocking notes:
- The header docstring says it mirrors `services/queries/productivity.ts`, but the closest analog is really `daily-summary.ts`/`tags.ts` (`void Promise.all`) — and this helper deliberately differs/improves by using `allSettled`. Harmless, but the reference is slightly off.
- The third test ('does not surface a rejected sync as an unhandled rejection') has no explicit assertion; it relies on the runner flagging unhandled rejections. That's fine as documentation, just weak as a test.

apps/backend/src/services/correlations/background-sync.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    }),
+    syncRescueTimeIfNeeded: vi.fn(() => {
+      calls.push('rescuetime')
+      return never()
+    }),
+    syncCalendarsIfNeeded: vi.fn(() => {
+      calls.push('calendars')
+      return never()
+    }),
+  } as unknown as SyncProvider
```
fiddur: Non-blocking: `as unknown as SyncProvider` is a bit at odds with the repo guideline "Avoid casting, use proper typing and type guards." Sibling test `hrv-activities.test.ts` builds a full `SyncProvider` literal (all methods) instead of casting a partial. Pragmatic for a partial mock here, but worth matching the existing pattern if you want to keep tests cast-free.

fiddur (commented):
✅ Approved

Clean, well-scoped fix for the correlation timeout (#792). The change replaces the blocking `await Promise.all([...])` auto-sync in the correlation engines with a fire-and-forget `triggerCorrelationSyncs` helper, and centralizes the previously-duplicated 4-sync trigger (explore.ts + generic.ts x2) into one place.

Correctness checks pass:
- The behavior shift to non-blocking is intentional and matches the existing query-layer pattern (`queries/tags.ts`, `daily-summary.ts`, `activities.ts` all use `void Promise.all`). The new helper goes one step further with `Promise.allSettled`, which is strictly safer.
- Every `*IfNeeded` method in `sync-provider.ts` already swallows its own errors via try/catch, so the prior awaited `Promise.all` never actually rejected — no error-surfacing behavior is lost; only the blocking wait is removed.
- No double-trigger: the event-outcome path early-returns in `getGenericCorrelation` (line 75) before reaching the line-106 trigger, so only `getGenericEventOutcome` fires it; `getCompoundEventOutcome` never triggered syncs itself. Even if it overlapped, the staleness gate + `allSettled` make it idempotent.
- `triggerSyncs` is fully removed with no stale references; `SyncProvider` remains used as a param type in both files.
- Tests honor the repo's "avoid casting" guideline by building full `SyncProvider` literals.

- The helper docstring says it attaches "a defensive catch," but the implementation uses `Promise.allSettled` (which collects rather than catches). Slightly inaccurate wording.
- The third test ('does not surface a rejected sync as an unhandled rejection') asserts on a *separate* `Promise.allSettled` it constructs itself, so it doesn't directly verify the helper's absorption — it would pass regardless of the helper. It still exercises the rejection path, so it's a weak guard rather than a wrong one.

Note: dependencies are not installed in this review worktree, so I could not execute the new test file; the logic was verified by reading.

PR #803, opened 2026-06-09
feat(correlations): in-UI guidance & guardrails for Explore (#792)

fiddur:
First of the #792 follow-up UI workstreams (implementing them one PR at a time). Frontend-only, additive guidance to make the Explore form hard to misuse.

## Problem
The biggest footgun: picking **Event-onset** mode with a **continuous** outcome (e.g. `sleep_score`) collapses every day into ~1 onset and returns nonsense. There was no in-UI explanation of when to use which mode, and the controls (collapse gap, denominator, …) are easy to misread.

## Changes
- **Mode help text** — "pick mode by your OUTCOME" (presence-only → Event onset; value-every-day → Continuous).
- **Mismatch warning** — shown when Event-onset is chosen with an outcome that looks continuous.
  - Heuristic (frontend-only, no backend change): built-in device/biometric metrics (`sleep_score`, `hrv_rmssd`, `weight`) are dense; presence-only symptom metrics (`back_pain`, `fissure_pain`) are **custom** and therefore not in the built-in `MetricType` enum. So a built-in metric in event mode is almost certainly a mode mismatch — and the custom metrics that genuinely belong in event mode are correctly left un-warned. Uses `isValidMetric` from api-spec.
- **Tooltips** on lag windows, collapse gap, denominator, regime (exact copy from the issue).
- **Preset examples** that populate the whole form to teach by doing: *Carbs → sleep (Continuous)*, *Ejaculation → back-pain onset (Event)*, *Sauna → back-pain onset (Event)*.

The heuristic + copy are extracted to a pure `exploreGuidance` module (no signals/JSX) with **unit tests** (6 cases: built-in flagged, custom not flagged, whitespace, continuous mode, tag outcome, empty).

## Testing
- `pnpm --filter aurboda-web check` — 0 errors
- `vitest run exploreGuidance.test.ts` — 6 passed

## Remaining #792 workstreams (next PRs, one at a time)
- Binary→continuous inference (group means, Welch t / Mann–Whitney, effect size, plain-language verdict)
- Nutrition completeness filter (`n_complete` vs `n_aligned`)
- Visualization upgrades (box/violin for binary, axis labels, regression overlay, RR-CI bars)
- Trigger selector cleanup: retire the legacy "tag" kind (tags merged into activities)

### DIFF
--- a/apps/web/src/pages/Correlations/Explore.tsx
+++ b/apps/web/src/pages/Correlations/Explore.tsx
@@ -95,6 +95,8 @@ const applyEventPreset = (pattern: string) => {
   eventOutcomeValue.value = 'back_pain'
   lagWindowsInput.value = '24h,48h,72h,7d'
   collapseGapDays.value = 3
+  // Scope to the behavioural regime the #792 worked example uses (the date the
+  // relevant protocol/era began); a leak is rare and meaningful only from here.
   periodStart.value = '2019-09-28'
   periodEnd.value = ''
 }
--- a/apps/web/src/pages/Correlations/Explore.tsx
+++ b/apps/web/src/pages/Correlations/Explore.tsx
@@ -17,6 +17,7 @@ import {
   fetchCorrelationSelectors,
   fetchGenericCorrelation,
 } from '../../state/api'
+import { eventOutcomeLooksContinuous, MODE_HELP, TOOLTIPS } from './exploreGuidance'
 
 type Mode = 'event' | 'continuous'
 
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/Correlations/Explore.tsx
```diff
@@ -82,6 +83,40 @@ const fmt = (value: number | null | undefined, digits = 2): string =>
...
+
+/** Apply a back-pain-onset event preset for the given trigger tag. */
+const applyEventPreset = (pattern: string) => {
+  setMode('event')
+  triggerSelector.value = { kind: 'tag', pattern }
+  setEventOutcomeSource('metric')
+  eventOutcomeValue.value = 'back_pain'
+  lagWindowsInput.value = '24h,48h,72h,7d'
+  collapseGapDays.value = 3
+  periodStart.value = '2019-09-28'
```
fiddur, line 100: Minor / optional: `2019-09-28` is an unexplained magic date. Since this preset is a fixed "worked example" for the owner's own data this is fine, but a one-word comment on why this date (e.g. when back-pain logging began) would help future readers.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend-only enhancement to the correlation Explore form: worked-example presets, mode-selection help text, control tooltips, and a "this outcome looks continuous" guardrail.

The mismatch heuristic (`eventOutcomeLooksContinuous`) is well-factored into a pure, signal/JSX-free module so it can be unit-tested in isolation, and the test covers the meaningful edge cases (mode, source, whitespace, empty). The logic is correct: built-in metrics are all continuous, so flagging a built-in metric chosen as an event outcome is sound; custom presence-only symptom metrics (back_pain, fissure_pain) are correctly left un-warned. The documented false-negative (a *custom* continuous metric won't be flagged) is an acceptable, explicitly-noted trade-off for a non-blocking advisory.

The new module-level helpers and `PRESETS` follow the file's existing module-signal/module-helper convention rather than introducing a new pattern, and the warning reads signals during render so it stays reactive. No API/MCP/DB surface is touched, so the parity and DB-integration-test rules don't apply.

Note: I could not execute the tests — `pnpm install` fails in this sandbox (read-only pnpm store) — but the assertions are verifiably correct against the `validMetrics` list. One minor, non-blocking observation inline.

fiddur (commented):
✅ Approved

Frontend-only UI guidance for the Correlations Explore form: worked-example presets, mode help text, control tooltips, and a soft "this outcome looks continuous" warning. Solid, focused change.

Verified locally:
- `pnpm test exploreGuidance.test.ts` → 6/6 pass.
- `pnpm check` (tsgo + oxlint) → 0 errors; the only warnings touching `Explore.tsx` are pre-existing (`buildRequest()!` at line 340, outside this diff).
- `oxfmt --check` clean on all three changed TS files.

Things done well:
- The mismatch heuristic is correctly extracted into a pure, signal/JSX-free module (`exploreGuidance.ts`) and unit-tested in isolation — exactly the testability-first approach the project standards ask for.
- `eventOutcomeLooksContinuous` reuses `isValidMetric` from `@aurboda/api-spec` rather than duplicating a metric list, consistent with the single-source-of-truth rule.
- The warning reads signals inside the component render, so it stays reactive in line with the rest of this page's module-signal pattern; presets reuse `setMode`/`setEventOutcomeSource` so the trigger-kind clamp and denominator default stay correct.

Non-blocking observation: the heuristic only warns on built-in metrics, so a *custom* continuous metric (e.g. a daily 1–10 mood log) used in event mode won't be flagged (false negative). This is intentional and documented in the code comment ("Custom metrics … are left un-warned") — a conservative choice to avoid false positives — so no change needed.

PR #804, opened 2026-06-09
feat(correlations): binary-trigger group comparison & plain-language verdicts (#792)

fiddur:
Second #792 follow-up workstream (one PR at a time). Full-stack: backend stats + api-spec + web.

## Problem
For a **binary/presence trigger** (e.g. `hot_bath` → `sleep_score`) every aligned point sits at x=0 or x=1, so a Pearson r is misleading. Users actually mean *"how much does X change Y?"*. Continuous results also had no plain-language read, making a tiny-n r easy to over-interpret.

## Changes
**Backend `stats.ts`** (new, unit-tested against t-tables / hand calcs):
- `welchTTest` — Welch's two-sample t-test; Student-t two-sided p via a new `regularizedIncompleteBeta` (continued fraction) on the existing `logGamma`.
- `mannWhitneyU` — tie-corrected normal approximation + rank-biserial effect size.
- `cohensD` (pooled) and a `twoGroupComparison` builder.

**`continuous.ts`** — split the aligned outcome by trigger presence (>0) and attach `group_comparison` (means, difference, Cohen's d, Welch, Mann–Whitney, `trigger_is_binary`). Null when there's no split. Flows automatically into **both** the REST `get_metric_correlation` and MCP responses — no new endpoint, parity preserved.

**api-spec** — `GroupComparison` / `WelchResult` / `MannWhitneyResult` schemas added to `ContinuousCorrelationData`.

**Web** — when the trigger is binary the group comparison is the headline (correlation flagged as misleading); otherwise a plain-language strength verdict + small-sample caution sits beside r/ρ. Verdict helpers are pure and unit-tested.

Visualization of these (box/violin, regression overlay, RR-CI bars) is the **next** workstream (#4) — this PR is the inference layer.

## Testing
- backend `vitest run src/services/correlations/` — 82 passed (incl. 18 new stats/group-comparison cases)
- web `vitest run exploreGuidance.test.ts` — 11 passed

- Nutrition completeness filter (`n_complete` vs `n_aligned`)
- Visualization upgrades (box/violin, axis labels, regression overlay, RR-CI bars)

### DIFF
--- a/apps/backend/src/services/correlations/continuous.ts
+++ b/apps/backend/src/services/correlations/continuous.ts
@@ -4,7 +4,17 @@
  * questions like "how does carb intake affect my sleep the next day?" work.
  */
 
-import { pearson, spearman } from './stats.ts'
+import { type TwoGroupComparison, pearson, spearman, twoGroupComparison } from './stats.ts'
+
+/**
+ * Group comparison for a binary/presence trigger: how the continuous outcome
+ * differs between days the trigger was present vs absent. A Pearson r on a 0/1
+ * trigger is misleading, so this answers "how much does X change Y?" directly.
+ */
+export interface GroupComparison extends TwoGroupComparison {
+  /** True when every aligned trigger value is exactly 0 or 1. */
+  trigger_is_binary: boolean
+}
 
 export interface AlignedPoint {
   /** Trigger day (YYYY-MM-DD). */
@@ -26,6 +36,11 @@ export interface ContinuousResult {
   spearman: number | null
   /** Aligned series (for plotting); capped to avoid huge payloads. */
   series: AlignedPoint[]
+  /**
+   * Present-vs-absent group comparison of the outcome. Null when the trigger is
+   * never present or always present (no split to compare).
+   */
+  group_comparison: GroupComparison | null
 }
 
 export interface ContinuousInput {
@@ -74,5 +89,25 @@ export const computeContinuous = (input: ContinuousInput): ContinuousResult => {
     pearson: pearson(triggerValues, outcomeValues),
     spearman: spearman(triggerValues, outcomeValues),
     series,
+    group_comparison: computeGroupComparison(triggerValues, outcomeValues),
+  }
+}
+
+/**
+ * Split the outcome by whether the trigger was present (value > 0) on the same
+ * aligned day and compare the two groups. Returns null when there is nothing to
+ * compare (the trigger is present on every day or on none).
+ */
+const computeGroupComparison = (triggerValues: number[], outcomeValues: number[]): GroupComparison | null => {
+  const withValues: number[] = []
+  const withoutValues: number[] = []
+  for (let i = 0; i < triggerValues.length; i++) {
+    if (triggerValues[i] > 0) withValues.push(outcomeValues[i])
+    else withoutValues.push(outcomeValues[i])
+  }
+  if (withValues.length === 0 || withoutValues.length === 0) return null
+  return {
+    ...twoGroupComparison(withValues, withoutValues),
+    trigger_is_binary: triggerValues.every((v) => v === 0 || v === 1),
   }
 }
--- a/apps/web/src/pages/Correlations/Explore.tsx
+++ b/apps/web/src/pages/Correlations/Explore.tsx
@@ -356,10 +356,7 @@ function ContinuousResults({ data }: { data: ContinuousCorrelationData }) {
         n={data.n} · Pearson r={fmt(data.pearson)} · Spearman ρ={fmt(data.spearman)} · lag {data.lag_days}d
       </p>
       {!showGroupAsHeadline && (
-        <p class="explore-verdict">
-          {describeCorrelationStrength(data.pearson)} correlation
-          {caution && <span class="explore-caution"> · {caution}</span>}
-        </p>
+        <p class="explore-verdict">{describeCorrelationStrength(data.pearson)} correlation</p>
       )}
       {gc && (
         <>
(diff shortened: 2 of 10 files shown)


apps/web/src/pages/Correlations/Explore.tsx
```diff
@@ -309,11 +344,33 @@ function ContinuousResults({ data }: { data: ContinuousCorrelationData }) {
...
+  // is binary the group comparison is the headline rather than the correlation.
+  const showGroupAsHeadline = gc !== null && gc.trigger_is_binary
+  const caution = sampleCaution(data.n)
+
   return (
     <div class="explore-results">
       <p class="explore-summary">
         n={data.n} · Pearson r={fmt(data.pearson)} · Spearman ρ={fmt(data.spearman)} · lag {data.lag_days}d
       </p>
+      {!showGroupAsHeadline && (
```
fiddur, line 358: Minor (non-blocking): when `showGroupAsHeadline` is true (binary trigger), this whole verdict block is hidden, which also drops the `sampleCaution(data.n)` warning. A present-vs-absent split on few days is exactly the case where "very small sample — treat as anecdotal" matters most, yet it won't be shown there. Consider surfacing `caution` independently of the correlation-strength line (e.g. inside `GroupComparisonPanel` or as a sibling) so it appears in both headline modes.

fiddur (commented):
🛑 Changes required

Solid, well-tested feature. The new statistics primitives in `stats.ts` (Welch's t-test, Mann–Whitney U with tie correction, Cohen's d, the regularized incomplete beta / two-sided Student-t p-value) are numerically correct — I verified the math by hand against the unit tests and ran the suites. The `group_comparison` is built from the same aligned trigger/outcome arrays used for the correlation, the present/absent split and `trigger_is_binary` flag are sound, and edge cases (empty groups, fewer than 2 values, zero variance, all-tied ranks) are guarded so nothing divides by zero or returns NaN. The schema change flows through correctly (`group_comparison` is required-nullable and the router spreads the service result verbatim, so the response always conforms), and the frontend plain-language verdict helpers have sensible, tested thresholds.

Verification run locally:
- `vitest run stats.test.ts continuous.test.ts` → 40 passed
- `vitest run exploreGuidance.test.ts` → 11 passed (after building `@aurboda/api-spec`)
- `pnpm --filter aurboda-backend check` and `pnpm --filter aurboda-web check` → 0 errors (warnings are all pre-existing in unrelated files)

One standards gap blocks approval:

**Docs not updated for the new endpoint output.** `docs/correlations.md` (Continuous mode, ~lines 93–94) still states the endpoint "returns Pearson and Spearman coefficients plus the aligned series for plotting" — it now also returns `group_comparison` (means, difference, Cohen's d, Welch t-test, Mann–Whitney U) and the UI surfaces plain-language verdicts. The project standards are explicit here: "When changing APIs ... or data models, update the corresponding docs to reflect the new behavior." Please add a short paragraph to the Continuous mode / Statistics notes section describing `group_comparison` and when it appears (binary vs continuous trigger). The file isn't in this PR's diff so I couldn't anchor an inline comment.

Non-blocking observation noted inline regarding where the small-sample caution is shown.

fiddur (commented):
✅ Approved

Solid, well-tested addition. The binary-trigger group comparison (`twoGroupComparison`: Welch t-test, Mann–Whitney U with tie correction, Cohen's d) and the plain-language verdict/caution helpers are implemented with standard, correct numerics and good edge-case handling (empty/single-element groups, zero-variance groups, tie correction). I verified locally:

- Backend `stats.test.ts` + `continuous.test.ts` (40 tests) pass; web `exploreGuidance.test.ts` (11 tests) pass.
- `tsgo --noEmit` is clean for both `apps/backend` and `apps/web`; oxlint reports 0 errors on the changed files (remaining warnings are pre-existing non-null-assertion warnings in untouched files).
- The hand-checked statistical formulas match the tests (e.g. `studentTTwoSidedP(2.228, 10) ≈ 0.05`, Welch t/df for the documented hand calculation, rank-biserial ±1 on fully separated groups).
- `group_comparison` is spread through `getContinuousCorrelation` and reaches both the REST `/continuous` route and the `get_metric_correlation` MCP tool, so API/MCP parity holds. The api-spec `groupComparisonSchema`/`welchResultSchema`/`mannWhitneyResultSchema` match the backend interfaces field-for-field, and `GroupComparison` is re-exported from the package root for the web import.
- Generated OpenAPI/Kotlin artifacts are build-time only (not version-controlled), so nothing needs regenerating in this PR.

Docs and schema descriptions are kept in sync, and the UI correctly demotes the misleading Pearson r to a group comparison when the trigger is binary. One minor, non-blocking semantic note inline about the present/absent split using `> 0`.

apps/backend/src/services/correlations/continuous.ts
```diff
@@ -74,5 +89,25 @@ export const computeContinuous = (input: ContinuousInput): ContinuousResult => {
...
+/**
+ * Split the outcome by whether the trigger was present (value > 0) on the same
+ * aligned day and compare the two groups. Returns null when there is nothing to
+ * compare (the trigger is present on every day or on none).
+ */
+const computeGroupComparison = (triggerValues: number[], outcomeValues: number[]): GroupComparison | null => {
+  const withValues: number[] = []
+  const withoutValues: number[] = []
+  for (let i = 0; i < triggerValues.length; i++) {
+    if (triggerValues[i] > 0) withValues.push(outcomeValues[i])
```
fiddur, line 105: Non-blocking: the split uses `> 0` for "present", but the doc comment and schema describe "absent (value 0)". For the triggers used here (nutrition amounts, activity/productivity counts, tag presence — all non-negative) this is correct, and an always-positive continuous trigger conveniently yields `null` (no day is 0). The only gap is a hypothetical trigger metric that goes negative on some days: those days would be bucketed as "absent" rather than "present". Fine to leave as-is given the trigger domain, just worth being aware of.

PR #805, opened 2026-06-09
feat(correlations): nutrition completeness filter for continuous mode (#792)

fiddur:
Third #792 follow-up workstream (one PR at a time). Full-stack: db + api-spec + REST + MCP + web.

## Problem
Logging completeness limits nutrition analyses: many days are marked "completed" but carry only meal *flags*, not full macros. These sum to 0 in `getDailyNutrientTotals` and **contaminate nutrient correlations** — inflating `n` with noisy zeros (the issue's `carbs → sleep_score` n=104 was really ~40–60 full-nutrition days).

## Changes
- **db** `getNutritionCompleteDaysInRange` — UTC days with ≥1 meal that has real macros (non-null `calories`), distinguishing complete days from flag-only ones. **Integration-tested** (3 cases) against real Postgres.
- **selectors** — `resolveNutrition` also returns `completeDays`.
- **continuous.ts** — report `n_complete` (of the `n` aligned pairs, how many are nutrition-complete on every nutrition side; `null` when no side is nutrition) and optionally drop incomplete pairs (`requireComplete`). Complexity kept under budget via a small `pairIsComplete` helper.
- **api-spec** — `nutrition_completeness` (`'all' | 'complete_only'`, default `all`) on the request; `n_complete` on the response. Threaded through **REST** and **MCP** (`get_metric_correlation`), so parity holds.
- **web** — a "Nutrition days: All / Full macros only" toggle (shown only when a side is nutrition) and `n=104 (47 with full nutrition)` in the summary.
- **docs** — `docs/correlations.md` documents `n_complete` and the filter.

## Definition
A day is **nutrition-complete** when it has at least one meal with non-null `calories` (real food logged). Flag-only meals (logged with no macros) leave `calories` NULL. This is independent of the event-mode `denominator` knob.

## Testing
- backend `vitest run src/services/correlations/` — 84 passed (incl. new completeness cases)
- new meals integration test — 3 passed (testcontainers)

- Visualization upgrades (box/violin, axis labels, regression overlay, RR-CI bars)

### DIFF
--- a/apps/backend/src/db/meals.ts
+++ b/apps/backend/src/db/meals.ts
@@ -181,24 +181,30 @@ export const getDailyNutrientTotals = async (
 }
 
 /**
- * UTC days in the inclusive [start, end] range that have *real* nutrition
- * logged — at least one meal with a non-null `calories` value.
+ * UTC days in the inclusive [start, end] range that have a *real* logged value
+ * for `nutrient` — at least one meal whose `nutrient` column is non-null.
  *
- * Distinguishes nutrition-complete days from flag-only days (a meal logged with
- * no macros), which otherwise sum to 0 in getDailyNutrientTotals and contaminate
- * nutrient correlations by inflating n with noisy zeros. Bucketed by UTC date to
- * match the correlation daily matrix.
+ * Keyed on the specific nutrient (not a single calories proxy) so completeness
+ * matches what's being correlated: a day that logged calories but never tracked
+ * fiber is complete for `calories` but not for `fiber`. Distinguishes
+ * nutrition-complete days from flag-only days (a meal logged with no macros),
+ * which otherwise sum to 0 in getDailyNutrientTotals and contaminate nutrient
+ * correlations by inflating n with noisy zeros. Bucketed by UTC date to match
+ * the correlation daily matrix.
  */
 export const getNutritionCompleteDaysInRange = async (
   user: string,
+  nutrient: NutrientKey,
   start: Date,
   end: Date,
 ): Promise<string[]> => {
+  // Column name comes from the fixed NUTRIENT_KEYS whitelist, so interpolation is safe.
+  const column = NUTRIENT_KEYS.includes(nutrient) ? nutrient : 'calories'
   const result = await query(
     user,
     `SELECT DISTINCT TO_CHAR((time AT TIME ZONE 'UTC')::date, 'YYYY-MM-DD') AS date
        FROM meals
-       WHERE time >= $1 AND time <= $2 AND calories IS NOT NULL
+       WHERE time >= $1 AND time <= $2 AND ${column} IS NOT NULL
        ORDER BY date`,
     [start, end],
   )
--- a/apps/backend/src/db/meals.ts
+++ b/apps/backend/src/db/meals.ts
@@ -180,6 +180,31 @@ export const getDailyNutrientTotals = async (
   return totals
 }
 
+/**
+ * UTC days in the inclusive [start, end] range that have *real* nutrition
+ * logged — at least one meal with a non-null `calories` value.
+ *
+ * Distinguishes nutrition-complete days from flag-only days (a meal logged with
+ * no macros), which otherwise sum to 0 in getDailyNutrientTotals and contaminate
+ * nutrient correlations by inflating n with noisy zeros. Bucketed by UTC date to
+ * match the correlation daily matrix.
+ */
+export const getNutritionCompleteDaysInRange = async (
+  user: string,
+  start: Date,
+  end: Date,
+): Promise<string[]> => {
+  const result = await query(
+    user,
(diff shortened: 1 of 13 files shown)


fiddur (commented):
✅ Approved

Solid, well-tested addition. The nutrition-completeness concept is threaded cleanly from the DB (`getNutritionCompleteDaysInRange`) through the selector resolver, the pure `computeContinuous` core, the service params, MCP tool, REST router, response schema, and the Explore UI — with unit tests for the pure logic and integration tests for the new query. Naming/parity conventions from CLAUDE.md are respected (snake_case at the boundary, `.meta({ id, description })` on the new schema, `.shape` in MCP), and the docs are updated.

- `completeDays` is always a subset of `knownDays` (a calories-bearing meal is also a meal counted by `getDailyNutrientTotals`), so the sets stay consistent in `computeContinuous`.
- When `nutrition_completeness: 'complete_only'` is sent but neither side is nutrition, both complete-sets are null, `pairIsComplete` returns true for every pair, and nothing is filtered — graceful no-op. `n_complete` is correctly null in that case (`tracksCompleteness` false).
- The new required `n_complete` field doesn't break existing `computeContinuous` tests (they assert per-field, not whole-object), and the generated OpenAPI/Kotlin artifacts aren't git-tracked so no regen is missing from the diff.
- The JSX summary line renders correctly in both branches (the `· Pearson` separator lives outside the conditional; the leading space is inside the template literal).

One non-blocking design note (left inline): completeness is keyed solely on `calories IS NOT NULL`, used as a proxy for "real macros" regardless of which nutrient is being correlated. This is documented and reasonable, but means a day with calories logged but a null/absent value for the *specific* nutrient (e.g. `fiber`) still counts as complete while contributing a 0 — so `complete_only` won't fully suppress noisy zeros for the rarer macros. Acceptable as-is given how meal logging fills macros together; flagging for awareness.

No blocking issues found.

apps/backend/src/db/meals.ts
```diff
@@ -180,6 +180,31 @@ export const getDailyNutrientTotals = async (
...
+export const getNutritionCompleteDaysInRange = async (
+  user: string,
+  start: Date,
+  end: Date,
+): Promise<string[]> => {
+  const result = await query(
+    user,
+    `SELECT DISTINCT TO_CHAR((time AT TIME ZONE 'UTC')::date, 'YYYY-MM-DD') AS date
+       FROM meals
+       WHERE time >= $1 AND time <= $2 AND calories IS NOT NULL
```
fiddur: Non-blocking: completeness is defined as `calories IS NOT NULL`, used as a single proxy for "real macros" across all nutrients. For a correlation on a rarer macro (e.g. `fiber`), a day that logged calories but no fiber still counts as complete here while `getDailyNutrientTotals` contributes a 0 for fiber — so `complete_only` won't fully remove noisy zeros for that nutrient. This is documented and likely fine in practice (macros are usually logged together), just noting the proxy's edge.

fiddur (commented):
✅ Approved

Solid, well-tested addition of a per-nutrient completeness filter to continuous correlations. The feature is correctly threaded through every layer (DB → service → REST/MCP → schema → web).

**Verified correct:**
- `getNutritionCompleteDaysInRange` reuses the exact UTC bucketing and range predicate of `getDailyNutrientTotals`, so `completeDays` align with the correlation daily matrix. The column whitelist (`NUTRIENT_KEYS.includes` + `NutrientKey` typing) keeps interpolation safe, and `insertMeal` storing `calories ?? null` means the `IS NOT NULL` filter genuinely excludes flag-only meals. Integration tests cover per-nutrient keying, any-meal-completes-day, range bounds, and flag-only exclusion.
- `computeContinuous`: `n_complete` is `null` only when neither side tracks completeness; an empty `completeDays` array correctly yields a `0` count rather than `null` (truthy `[]` → empty Set). `requireComplete` drops incomplete pairs before they reach values/series/group_comparison, and the count is independent of the series cap. Unit tests exercise all three paths.
- `completeDays` is a subset of `knownDays` and threshold-independent (threshold only affects `eventDays`, unused in continuous mode).
- `computeContinuous` is the only constructor of `ContinuousResult`, so the new mandatory `n_complete` field doesn't break other call sites. The response schema field is always populated; generated OpenAPI/Kotlin are build-time only (not git-tracked), so no regeneration commit is required.
- Frontend toggle is gated on `continuousHasNutrition()`, and `n_complete !== null` renders `0` correctly.

**Non-blocking nit:** the `nutritionCompleteness` signal isn't reset when both selectors switch away from nutrition, so a stale `'complete_only'` may still be sent. This is harmless — the backend no-ops it (no `completeDays` ⇒ no filtering, `n_complete` null) and the UI hides the indicator — and arguably preserves the user's choice if they switch back. Module-level Preact signals match the established pattern in this file.

Docs in `docs/correlations.md` are accurate and thorough. No changes required.

PR #806, opened 2026-06-09
feat(correlations): chart upgrades — box plot, regression overlay, RR bars (#792)

fiddur:
Fourth #792 follow-up workstream (one PR at a time). Mostly frontend SVG, plus a small supporting backend stat.

## Problem (from the #792 UX comment)
The Explore charts were misleading/illegible:
- a **binary trigger** scatter collapses to two vertical lines at x=0/x=1;
- charts had **no axis labels or units**;
- continuous scatters had **no regression line** and no on-plot stats;
- event-onset had a table but **no visual** of effect size per lag.

## Changes
- **Box plot** for binary triggers (present vs absent) with per-group n, instead of a meaningless scatter.
- **Scatter + regression line** for continuous↔continuous, annotated `r / ρ / n / p`.
- **Axis labels with units** on every chart, derived per selector kind (metric units from api-spec, kcal/g for nutrients, count/min otherwise).
- **Relative-risk chart** for event-onset: per-lag RR points with 95% CI whiskers and an RR=1 reference line; significant lags highlighted.
- Backend: the continuous response now includes **`pearson_p`** (two-sided significance of the Pearson r via the t-transform, reusing `studentTTwoSidedP`) so the scatter's `p` annotation is real. Schema + REST + MCP flow automatically.

Chart maths (regression, quartiles, axis labelling) live in a pure `exploreCharts` module with unit tests; the SVG components stay presentational.

## Testing
- web `vitest run src/pages/Correlations/` — 20 passed (9 chart helpers + 11 guidance)
- backend `vitest run src/services/correlations/` — 86 passed (incl. new `pearson_p` cases)

### DIFF
--- a/apps/backend/src/services/correlations/continuous.test.ts
+++ b/apps/backend/src/services/correlations/continuous.test.ts
@@ -20,7 +20,40 @@ describe('computeContinuous', () => {
 
     expect(result.n).toBe(4)
     expect(result.pearson).toBeCloseTo(1, 5)
+    expect(result.pearson_p).toBe(0) // perfect correlation
     expect(result.spearman).toBeCloseTo(1, 5)
+  })
+
+  test('reports a Pearson p-value for a noisy correlation', () => {
+    const days = Array.from(
+      { length: 20 },
+      (_, i) => new Date(Date.parse('2024-01-01T00:00:00Z') + i * 86_400_000).toISOString().split('T')[0],
+    )
+    // Weak/none relationship -> p should be a real probability in (0, 1].
+    const trigger = seriesMap(days.map((d, i) => [d, i % 5]))
+    const outcome = seriesMap(days.map((d, i) => [d, (i * 7) % 3]))
+    const result = computeContinuous({
+      triggerDaily: trigger,
+      outcomeDaily: outcome,
+      triggerKnown: days,
+      outcomeKnown: days,
+      lagDays: 0,
+    })
+    expect(result.pearson_p).not.toBeNull()
+    expect(result.pearson_p!).toBeGreaterThan(0)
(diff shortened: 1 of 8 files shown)


fiddur (commented):
✅ Approved

Solid, well-tested visualization PR. The new `pearson_p` is computed via the standard t-transform for Pearson significance (`t = r·√((n−2)/(1−r²))` on `n−2` df) with correct edge handling (null for `n < 3` or no variance, `0` for `|r| ≥ 1`), wired through the shared api-spec schema so both REST and MCP pick it up, and covered by new backend tests. The frontend chart math (`linearRegression`, `fiveNumberSummary`, `describeSelectorAxis`) is extracted into a JSX-free `exploreCharts.ts` module with good unit tests — nice for testability. Docs are updated. CI (Node check/test/build, Android) is green.

I verified the referenced types line up with the schemas: `LagExposureResultData.p_value` is non-nullable (so `l.p_value < 0.05` in `RrBars` is safe), `relative_risk`/`ci_low`/`ci_high` are nullable and correctly guarded, and `metricUnits`/`isValidMetric` are exported from the api-spec root. The empty-series case is fine — points and the regression line are guarded so no NaN coordinates reach the DOM.

A couple of non-blocking cosmetic notes (no change required):
- The box plot's quartiles are computed from `data.series`, which is capped at 1000 points, while the `Absent/Present (n=…)` labels use the full `group_comparison` counts. These only diverge above 1000 aligned days, so it's a harmless edge-case mismatch.
- The OLS regression overlay draws between `xMin`/`xMax` using fitted y-values, which can fall slightly outside `[yMin, yMax]` when there are y-outliers; the line would then be clipped at the SVG edge rather than mis-rendered. Purely visual.

Nothing blocking — approving.

PR #807, opened 2026-06-09
feat(correlations): selectable HRV-context metric (HR / stress, not just HRV) (#792)

fiddur:
Fifth #792 follow-up workstream (one PR at a time), addressing your request: *"I would like to be able to select e.g. stress_level or even heart rate instead of HRV. I (and other users) may not have continuous HRV throughout the day."*

## Problem
The "HRV context" analysis computed its productivity `correlation_coefficient` against HRV only. Without continuous HRV that correlation is null/sparse — even though the table already shows HR and stress.

## Changes
- **`context_metric`** parameter (`hrv_rmssd` default, `heart_rate`, `stress_level`) selects which denser autonomic metric drives the productivity correlation. All three means (HRV/HR/stress + baseline deltas) remain in the response/table; only the correlation follows the choice. The chosen metric is **echoed** as `context_metric`.
- **api-spec**: `HrvContextMetric` enum on the query + MCP input schemas, echoed on `HrvActivitiesData`.
- **service**: correlate productivity scores against the chosen metric's in-window values.
- **REST + MCP**: threaded through; the MCP tool description now explains the option.
- **web**: a **Context** dropdown in the HRV tab and a metric-aware explanation of `r`.
- **docs + tests**.

## Testing
- backend `vitest run src/services/correlations/` — 88 passed (incl. new `context_metric` cases)
- web `vitest run src/pages/Correlations/` — 20 passed

## Remaining #792 workstream (last PR)
- Trigger selector cleanup: retire the legacy "tag" kind (tags merged into activities)

### DIFF
--- a/apps/backend/src/services/correlations/hrv-activities.ts
+++ b/apps/backend/src/services/correlations/hrv-activities.ts
@@ -2,6 +2,8 @@
  * HRV-activities correlation analysis.
  */
 
+import type { HrvContextMetric } from '@aurboda/api-spec'
+
 import type { SyncProvider } from '../queries/index.ts'
 import type {
   ActivityCorrelation,
@@ -22,6 +24,7 @@ export async function getHrvActivitiesCorrelation(
   user: string,
   periodDays: number = 30,
   sync?: SyncProvider,
+  contextMetric: HrvContextMetric = 'hrv_rmssd',
 ): Promise<HrvActivitiesResult> {
   const end = new Date()
   end.setHours(23, 59, 59, 999)
@@ -49,6 +52,11 @@ export async function getHrvActivitiesCorrelation(
     getAllActivitiesInRange(user, start, end),
   ])
 
+  // The series the productivity correlation is computed against. HRV is the
+  // default, but heart rate / stress are denser when continuous HRV is sparse.
+  const contextData =
+    contextMetric === 'heart_rate' ? hrData : contextMetric === 'stress_level' ? stressData : hrvData
+
   // Calculate baseline stats
   const baselineHrvValues = hrvData.map(([, v]) => v)
   const baselineHrValues = hrData.map(([, v]) => v)
@@ -59,13 +67,21 @@ export async function getHrvActivitiesCorrelation(
   // === Productivity correlations by category ===
   const productivityByCategory = new Map<
     string,
-    { hrvValues: number[]; hrValues: number[]; stressValues: number[]; minutes: number; scores: number[] }
+    {
+      hrvValues: number[]
+      hrValues: number[]
+      stressValues: number[]
+      contextValues: number[]
+      minutes: number
+      scores: number[]
+    }
   >()
 
   for (const record of productivity) {
     const category = record.resolved_category?.join(' > ') || record.category || 'Uncategorized'
     if (!productivityByCategory.has(category)) {
       productivityByCategory.set(category, {
+        contextValues: [],
         hrValues: [],
         hrvValues: [],
         minutes: 0,
@@ -76,17 +92,19 @@ export async function getHrvActivitiesCorrelation(
     const cat = productivityByCategory.get(category)!
     cat.minutes += record.duration_sec / 60
 
-    // Get HRV/HR during this productivity window
+    // Get HRV/HR/stress during this productivity window (all shown in the table).
     const hrvInWindow = getDataInRange(hrvData, record.start_time, record.end_time)
     const hrInWindow = getDataInRange(hrData, record.start_time, record.end_time)
     const stressInWindow = getDataInRange(stressData, record.start_time, record.end_time)
+    const contextInWindow = getDataInRange(contextData, record.start_time, record.end_time)
     cat.hrvValues.push(...hrvInWindow)
     cat.hrValues.push(...hrInWindow)
     cat.stressValues.push(...stressInWindow)
+    cat.contextValues.push(...contextInWindow)
 
     if (record.productivity !== undefined && record.productivity !== null) {
-      // Add productivity score for correlation calculation - one score per HRV value
-      cat.scores.push(...hrvInWindow.map(() => record.productivity!))
+      // One productivity score per context-metric value, for the correlation.
+      cat.scores.push(...contextInWindow.map(() => record.productivity!))
     }
   }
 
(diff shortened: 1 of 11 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped feature that makes the autonomic metric behind the productivity `correlation_coefficient` selectable (`hrv_rmssd` default, `heart_rate`, `stress_level`).

**Strengths**
- Fully backward compatible: with the default `hrv_rmssd`, `contextData === hrvData`, so `contextInWindow`, `cat.scores`, and the Pearson inputs are identical to the previous code path — no behavior change for existing callers.
- Good API/MCP parity per the project standards: the REST router and the MCP tool both accept and forward `context_metric`, the schema is defined with `.meta({ id, description })`, exported from `schemas/index.ts`, added to query/data/MCP-input schemas, and consumed as a type across web + backend.
- All three means (`mean_hrv`/`mean_hr`/`mean_stress`) stay populated regardless of the chosen metric, matching the docs and the new test.
- Docs (`docs/correlations.md`) and a unit test were added, and the chosen metric is echoed back via `context_metric` and surfaced in the UI copy. CI (Node test/check/build, Android) is green.

**Minor (non-blocking)**
- The test `correlates productivity against the chosen context metric` only asserts the three means; it doesn't actually verify the correlation coefficient was derived from the selected series (its own comment notes r is null because the productivity score is constant). The name slightly overstates what's covered — consider a follow-up case with a varying score that checks `correlation_coefficient` differs per `context_metric`.
- Pre-existing (not introduced here): the correlation is only computed when `contextValues.length === scores.length`, i.e. it silently yields `null` for any category that contains a productivity record without a score. Carried over unchanged, just noting it.

apps/backend/src/services/correlations/hrv-activities.ts
```diff
@@ -97,10 +115,10 @@ export async function getHrvActivitiesCorrelation(
     const stats = calculateHrvStats(data.hrvValues, data.hrValues, data.minutes, data.stressValues)
     const statsWithDelta = addBaselineDelta(stats, baseline)
 
-    // Calculate correlation between productivity score and HRV
+    // Correlation between productivity score and the chosen context metric.
     const correlation =
-      data.scores.length >= 3 && data.hrvValues.length === data.scores.length
-        ? pearsonCorrelation(data.scores, data.hrvValues)
+      data.scores.length >= 3 && data.contextValues.length === data.scores.length
+        ? pearsonCorrelation(data.scores, data.contextValues)
```
fiddur, line 121: Non-blocking: the coefficient is now computed against `contextValues`, but the row's `sample_count`/`mean_hrv` (from `calculateHrvStats` on `hrvValues`) still reflect HRV. With `context_metric=heart_rate` and sparse HRV, a category can surface `sample_count: 0` / `mean_hrv: null` alongside a non-null `correlation_coefficient` — which is the intended behavior, but the displayed sample count no longer corresponds to the series the coefficient was derived from. Consider clarifying this in the UI/help text.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The `context_metric` selector is threaded consistently through the api-spec schema, service, REST router, MCP tool, and web frontend, and the snake_case/`.meta({ id })` conventions are followed. The default (`hrv_rmssd`) reproduces the previous behavior exactly — `contextData === hrvData`, so both the correlation input and the equality guard (`contextValues.length === scores.length`) are byte-for-byte equivalent to the old HRV-based path, making this fully backwards-compatible. Passing `undefined` for the optional param correctly falls through to the default in both the REST and MCP call sites.

The new tests are good: the sign-flip test (HRV → r>0.9, HR → r<-0.9 on the same windows) genuinely proves the correlation uses the selected series, and asserting `mean_hr`/`mean_hrv` stay populated regardless confirms the three means are independent of the chosen context. Docs are updated. Generated OpenAPI/Kotlin artifacts are gitignored, so no regeneration needs to be committed (CI rebuilds them).

I couldn't execute the tests locally (deps not installed and the sandbox npm cache is read-only), but I traced the new test logic by hand against `getDataInRange`/`pearsonCorrelation` and it checks out.

Non-blocking observation: each productivity row's `sample_count`/`mean_hrv` remain HRV-based (via `calculateHrvStats`), while `correlation_coefficient` is now computed against the chosen context series. For a user with sparse HRV who selects `heart_rate`, a row can show `sample_count: 0` / `mean_hrv: null` yet a non-null correlation derived from the denser HR series. That's the intended point of the feature, but the UI's sample count no longer matches the basis of the coefficient — worth a brief clarification in the table/help text at some point.

fiddur (commented):
✅ Approved

Clean, well-scoped feature that makes the HRV-context productivity correlation computed against a selectable autonomic metric (`hrv_rmssd` default, `heart_rate`, `stress_level`). The change is faithful end-to-end and respects the project's parity rules.

**What I verified**
- **Backward compatibility:** with the default `contextMetric === 'hrv_rmssd'`, `contextData === hrvData` and `contextInWindow === hrvInWindow`, so the correlation (`pearsonCorrelation(scores, contextValues)`) and the `scores.length === contextValues.length` guard are exactly equivalent to the old `hrvValues`-based logic. No behavior change for existing callers.
- **Parity & wiring:** the new `context_metric` flows through all layers — `hrvContextMetricSchema` (re-exported via `schemas/index.ts`), the query schema, the data schema, the MCP input schema, the router, the service signature, the frontend query key/fetch, and the `HrvContextMetric` type re-export in `state/api/types.ts`. All `getHrvActivitiesCorrelation` call sites are updated; `context_metric` is correctly threaded and the optional/undefined param falls back to the default.
- **Tests:** the new `derives the productivity correlation from the chosen context series` test is sound — I hand-computed it (HRV r≈+1.00, HR r≈-1.00; `mean_hrv` = (40+42+60+62)/4 = 51, `mean_hr` = (70+72+50+52)/4 = 61), and the `getDataInRange` binary search correctly selects the two in-window points per productivity record. The sign-flip assertion genuinely proves the selected series drives the coefficient.
- **Docs/UI:** `docs/correlations.md` and the in-page help text accurately describe that the HRV/HR/Stress columns remain each metric's own mean independent of the chosen context, while only `r` follows the context selection.

**Non-blocking observation (pre-existing, not introduced here):** the correlation pairs each intra-window sample with the window's single productivity score (`contextInWindow.map(() => record.productivity!)`), i.e. pseudoreplication that treats within-window samples as independent observations. This was already the case with HRV; switching to a denser metric like `heart_rate` will inflate the effective n much more, pushing `r` toward extreme values and making it look stronger than the underlying between-window signal warrants. Since no p-value is reported for the productivity `r` and the methodology pre-dates this PR, it's not a blocker — but worth a follow-up to consider aggregating per-window means before correlating.

No blocking issues found. (Generated artifacts under `packages/api-spec/generated/` are gitignored, so their absence from the diff is expected; CI regenerates them.)

PR #808, opened 2026-06-09
feat(correlations): retire the legacy "tag" selector in the Explore picker (#792)

fiddur:
Final #792 follow-up workstream, addressing your note: *"the selector for trigger included 'tag' and other things that were long since merged into Activity. I don't even know what that selection would do."*

## Problem
The trigger/outcome picker offered both **tag** and **activity**. Tags were long ago merged into activities, and both kinds resolve identically (match `activity_type` by regex), so the duplicate was pure confusion.

## Changes (UI + docs only)
- Drop **tag** from the picker's selectable kinds; default to **activity**, relabelled **"activity / tag"** so it's clear it covers former tags. Its autocomplete already lists every `activity_type`.
- The event-outcome source's **tag** option is relabelled **"activity / tag"** (placeholder → `activity_type regex`).
- Presets and the form default now use the `activity` kind.
- The backend schema still accepts the `tag` kind/source for older REST + MCP callers — nothing breaking, this is a UI cleanup. Documented in `docs/correlations.md`.

## Testing
- web `vitest run src/pages/Correlations/` — 20 passed
- `pnpm --filter aurboda-web check` — 0 errors

## This completes the #792 follow-up series
All six workstreams from the issue's UX comments are now done (guidance/guardrails, binary→continuous inference, nutrition completeness, visualizations, HRV context metric, and this tag-selector cleanup).

### DIFF
--- a/apps/web/src/pages/Correlations/Explore.tsx
+++ b/apps/web/src/pages/Correlations/Explore.tsx
@@ -673,7 +673,7 @@ export function ExploreTab() {
                 <option value="tag">activity / tag</option>
               </select>
               <input
-                list="metric-options"
+                list={eventOutcomeSource.value === 'metric' ? 'metric-options' : 'activity-type-options'}
                 placeholder={
                   eventOutcomeSource.value === 'metric' ? 'metric (e.g. back_pain)' : 'activity_type regex'
                 }
@@ -683,6 +683,11 @@ export function ExploreTab() {
               <datalist id="metric-options">
                 {selectors?.metrics.map((m) => (
                   <option value={m.value}>{m.label}</option>
+                ))}
+              </datalist>
+              <datalist id="activity-type-options">
+                {selectors?.activity_types.map((a) => (
+                  <option value={a.value}>{a.label}</option>
                 ))}
               </datalist>
             </div>
--- a/apps/web/src/pages/Correlations/Explore.tsx
+++ b/apps/web/src/pages/Correlations/Explore.tsx
@@ -32,7 +32,7 @@ type Mode = 'event' | 'continuous'
 
 // --- Form state (module signals, mirroring the rest of this page) ---
 const mode = signal<Mode>('event')
-const triggerSelector = signal<CorrelationSelector>({ kind: 'tag', pattern: '' })
+const triggerSelector = signal<CorrelationSelector>({ kind: 'activity', pattern: '' })
 const outcomeSelector = signal<CorrelationSelector>({ kind: 'metric', metric: '' })
 const eventOutcomeSource = signal<'metric' | 'tag'>('metric')
 const eventOutcomeValue = signal('')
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/Correlations/Explore.tsx
```diff
@@ -661,11 +670,13 @@ export function ExploreTab() {
...
               >
                 <option value="metric">metric</option>
-                <option value="tag">tag</option>
+                <option value="tag">activity / tag</option>
               </select>
               <input
                 list="metric-options"
-                placeholder={eventOutcomeSource.value === 'metric' ? 'metric (e.g. back_pain)' : 'tag regex'}
+                placeholder={
+                  eventOutcomeSource.value === 'metric' ? 'metric (e.g. back_pain)' : 'activity_type regex'
```
fiddur, line 678: Nice that the placeholder is now `activity_type regex`, but the input below still binds `list="metric-options"` (only metric suggestions). So when the source is "activity / tag", the autocomplete suggests metrics that don't match the placeholder. The trigger `SelectorPicker` autocompletes `activity_types` here — it'd be consistent to point this datalist at `selectors?.activity_types` too. Non-blocking (it's a free-text regex field, and the mismatch predates this PR), just flagging since the relabel makes it more noticeable.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend change that retires the redundant `tag` kind from the Explore picker in favor of `activity`. I verified the load-bearing claim against the backend: trigger `type: 'tag'` and `type: 'activity'` resolve through identical logic (both regex-match `activity_type` in `generic.ts`), and event-outcome `source: 'tag'` resolves via the same activity path (`resolveSelector` maps both `tag`/`activity` to `resolveActivity` in `selectors.ts`). So keeping `source: 'tag'` for the event outcome while relabeling it "activity / tag" is correct and produces the same results. All `'tag'` → `'activity'` substitutions (default trigger, preset, metric fallback, picker default) are sound, and dropping `'tag'` from `EVENT_TRIGGER_KINDS`/`ALL_SELECTOR_KINDS` removes it from the dropdown while the api-spec schema still accepts it for backward compatibility. The docs addition is accurate. The `KIND_LABELS` lookup is type-safe.

One minor, non-blocking UX inconsistency (pre-existing, but surfaced by this PR's relabel): the event-outcome value input's autocomplete still only suggests metrics even when the source is "activity / tag". The trigger `SelectorPicker` correctly autocompletes `activity_types`, so the event-outcome side is now the odd one out — see the inline note. Not a correctness issue (the field is a free-text regex), so I'm approving.

fiddur (commented):
✅ Approved

Clean, well-scoped UI change that retires the redundant `tag` selector from the Explore picker. I verified the core premise against the backend: `apps/backend/src/services/correlations/generic.ts` handles `type: 'activity'` and `type: 'tag'` identically — both regex-match against `activity_type` via `matchesPattern` (with the comment "Tags are now activities"). So defaulting triggers to `activity`, dropping `tag` from `EVENT_TRIGGER_KINDS`, and sending `type: 'activity'` in the metric-fallback are all behavior-preserving. The backend still accepts `tag`, so older callers are unaffected, matching the new doc note.

Nice catch on the event-outcome autocomplete: the base version always sourced the tag-outcome input from `metric-options`, but the backend matches tag outcomes against `activity_type` (`generic.ts:110`/`139`), so switching to the new `activity-type-options` datalist (and the `activity_type regex` placeholder) is a genuine fix, not just cosmetics. The `KIND_LABELS` map and all the `'activity'` substitutions are type-safe.

Non-blocking observations (no action required):
- `SelectorPicker` still carries `value.kind === 'tag'` branches and a `selectors?.tags` reference (around lines 235, 249-250). Since `tag` is no longer in any `allowedKinds` and nothing sets a selector to that kind, this is now dead/defensive code — harmless, could be pruned later.
- The schema description at `packages/api-spec/src/schemas/correlations.ts:402` still reads "regex for tags, exact match for activity types", which is now inaccurate (the backend regex-matches both). Pre-existing and outside this diff, but worth tidying when convenient.

Note: I could not run the full `pnpm check` because `node_modules` is not installed in this review worktree; the review is based on diff inspection plus backend cross-checks.

PR #809, opened 2026-06-12
feat(food-items): duplicate a food item into an editable copy

fiddur:
## What

Adds a **duplicate food item** function and UI so you can copy a recipe and then tweak it — e.g. swap a single ingredient — without re-entering everything.

## How it works

`duplicateFoodItem` (`services/food-items.ts`) creates a fresh per-user `manual` copy named `"<name> (copy)"`, deduped to `"(copy 2)"`, `"(copy 3)"`, … so it never silently overwrites an existing item via the `name_lower` upsert conflict. It copies:

- nutrient columns + `default_quantity` / `default_unit` / `icon`
- composite **ingredients** (then re-caches derived nutrients)
- **portions** (fresh ids), with `default_portion_id` / `default_log_quantity` remapped onto the new portion rows
- the **reference** pointer (atomic items)
- **sensitivity** flag assignments

`source` / `source_id` provenance is deliberately dropped — the copy is a user-authored fork (`source: 'manual'`).

Works on **per-user and central** sources: duplicating a central (e.g. Livsmedelsverket) item gives you an **editable per-user fork** — a clean way to base a custom recipe on a canonical entry.

## Surface (API/MCP parity)

- **REST:** `POST /food-items/:id/duplicate` → the new copy's detail
- **MCP:** `duplicate_food_item` tool
- **Web:** a `Duplicate` button on the food item detail header (next to Re-snapshot/Delete, shown for shared items too) and a `Duplicate` action on each Food Items list row. Both navigate to the new copy for editing.

## Tests

- Integration (`food-items-duplicate.integration.test.ts`, testcontainers): atomic copy, composite ingredients + independence, portions + default-portion remap, sensitivities, reference, name dedup, central fork, unknown id → null
- Router endpoint tests: 404 + 200 happy path
- `pnpm check` green (0 errors); 94 food-item tests pass

## Docs

- `docs/features/meals.md`: added the duplicate REST endpoint + MCP tool to the food-items lists

### DIFF
--- a/apps/backend/src/services/food-items.ts
+++ b/apps/backend/src/services/food-items.ts
@@ -640,6 +640,13 @@ export const clearCompositeNutrientCache = async (
  * with — and thus silently overwrites via the name_lower upsert conflict — an
  * existing item. Only the per-user library is checked; central names live in
  * a separate database and never conflict with per-user inserts.
+ *
+ * The check is best-effort against concurrency: two duplicate calls racing on
+ * the same source could each see the same name free, and the second's upsert
+ * (ON CONFLICT (name_lower) DO UPDATE) would then overwrite the first's row
+ * rather than insert a new one. Per-user, single-actor DBs make this vanishingly
+ * rare — same as the concurrent-setIngredients note above — so we don't pay for
+ * a fail-and-retry insert path here.
  */
 const findAvailableCopyName = async (user: string, baseName: string): Promise<string> => {
   const first = `${baseName} (copy)`
--- a/apps/backend/src/services/food-items.ts
+++ b/apps/backend/src/services/food-items.ts
@@ -22,8 +22,13 @@ import {
   findCompositeParentsOfIngredient,
   type FoodItemIngredientRow,
   getIngredients as dbGetIngredients,
+  setIngredients as dbSetIngredients,
 } from '../db/food-item-ingredients.ts'
-import { type FoodItemPortionRow, listPortionsForFoodItem } from '../db/food-item-portions.ts'
+import {
+  type FoodItemPortionRow,
+  insertFoodItemPortion,
+  listPortionsForFoodItem,
+} from '../db/food-item-portions.ts'
 import {
   findOrCreateFoodItem as findOrCreateUserFoodItem,
   getFoodItemById as getUserFoodItemById,
@@ -33,9 +38,14 @@ import {
   type MergeFoodItemResult,
   mergeFoodItems as dbMergeFoodItems,
   searchFoodItems as searchUserFoodItems,
+  setFoodItemReference as dbSetFoodItemReference,
   updateFoodItem as dbUpdateFoodItem,
+  upsertFoodItem,
 } from '../db/food-items.ts'
-import { getFoodItemSensitivities } from '../db/sensitivities.ts'
+import {
+  getFoodItemSensitivities,
+  setFoodItemSensitivities as dbSetFoodItemSensitivities,
+} from '../db/sensitivities.ts'
 import { getSharedFoodItemOverridesByIds } from '../db/shared-food-item-overrides.ts'
 
 /**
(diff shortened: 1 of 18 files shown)


apps/web/src/pages/FoodItems/index.tsx
```diff
@@ -46,6 +46,15 @@ export function FoodItems() {
     onSuccess: () => queryClient.invalidateQueries({ queryKey: ['foodItems'] }),
   })
 
+  const duplicateMutation = useMutation({
```
fiddur, line 53: Non-blocking: this list-page `duplicateMutation` has no `onError`, so a failed duplicate (network/500) silently does nothing for the user. The detail-page mutation (`FoodItemDetail.tsx`) surfaces the error via `setSaveError`. Consider adding a comparable error surface here for consistency.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. `duplicateFoodItem` is implemented as shared service logic and exposed via both the REST endpoint (`POST /food-items/:id/duplicate`) and the `duplicate_food_item` MCP tool, satisfying the API/MCP parity requirement, with docs updated in `docs/features/meals.md`. The rest of the diff is pure oxfmt formatting (calorie-computation.ts, activityStats.ts, schemas, docs) and is harmless.

The copy logic is careful in the right places:
- `findAvailableCopyName` checks the per-user library and picks `(copy)`, `(copy 2)`, … so the `name_lower` upsert conflict can never silently overwrite an existing item.
- Portions are recreated with fresh ids and `default_portion_id` / `default_log_quantity` are remapped onto the copy's own portions.
- Composite ingredients are copied then re-cached via `cacheCompositeNutrients`, which fully overwrites every nutrient column — so copying the cached nutrient values in `buildFoodItemCopyInput` is genuinely harmless (the comment is accurate).
- The reference pointer is copied only for atomic items (`ingredients.length === 0`), and central sources are forked into editable `source: 'manual'` per-user rows.

Test coverage is excellent: integration tests cover atomic, composite (with independence checks), portions/default remap, sensitivities, reference, name dedup, central fork, and the not-found case; router tests cover the 404 and 200 paths and assert the upsert payload; MCP parity is in place.

Two non-blocking observations (not requested as changes):
- `duplicateFoodItem` performs several sequential writes (upsert → setIngredients → cache → portions → defaults update → reference → sensitivities) outside a single transaction. A mid-way failure would leave an orphaned partial copy. This matches the codebase's existing style (each underlying op is internally atomic) and a retry just produces a `(copy 2)`, so it's acceptable — flagging only for awareness.
- Minor UX inconsistency between the two web entry points (see inline comment).

apps/web/src/pages/FoodItems/index.tsx
fiddur: Done in a3f52035 — added an `onError` that shows a dismissible error banner on the list page, matching the detail page's error surface.

fiddur (commented):
✅ Approved

Solid, well-tested feature. `duplicateFoodItem` is cleanly decomposed and correct: atomic copies carry nutrients/defaults/icon; composites re-create the ingredient list and re-cache derived nutrients; portions are recreated with fresh ids and the default portion is remapped onto the copy's own rows; the reference pointer (atomic-only) and sensitivity flags are copied; and the `(copy)`/`(copy N)` dedup avoids the `name_lower` upsert silently overwriting an existing row. The reference-enrichment interaction is handled correctly — `getDetail` returns the raw `self` row as `detail.item`, so inherited reference values aren't baked into the copy's own columns, and the pointer is re-established instead. API/MCP parity is honored, docs (`meals.md`) are updated, and coverage is strong (8 integration tests + 2 router tests). Frontend wiring is consistent with existing mutation patterns and surfaces errors on both the list and detail pages.

- **TOCTOU on the copy name.** `findAvailableCopyName` checks availability, then `upsertFoodItem` runs `ON CONFLICT (name_lower) DO UPDATE`. Two concurrent duplicates of the same source (e.g. from two of the repo's clones / sessions) could pass the check and have the second one *update* the first's row rather than insert. Very low risk for this single-user app and the in-page button-disable covers rapid double-clicks, but it's the exact silent-overwrite the helper exists to prevent. Flagged inline.
- **List page shares one `duplicateMutation`**, so `disabled={duplicateMutation.isPending}` disables every row's Duplicate button while any one is running. Minor UX. Flagged inline.
- **Scope:** the PR bundles unrelated oxfmt-only reformatting across several files (`calorie-computation.ts`, `activityStats.ts`, `admin.ts`, `food-item-portions.ts`, `meals.ts`, docs) from the `pnpm fix` commit. I verified these are formatting-only with no behavior change, but they do mix concerns into a feature PR.

apps/backend/src/services/food-items.ts
```diff
@@ -620,6 +630,158 @@ export const clearCompositeNutrientCache = async (
...
+  centralDb: CentralDb,
+  sourceId: string,
+): Promise<FoodItemDetail | null> => {
+  const service = createFoodItemsService(centralDb)
+  const detail = await service.getDetail(user, sourceId)
+  if (!detail) return null
+  const source = detail.item
+
+  const name = await findAvailableCopyName(user, source.name as string)
+  const created = await upsertFoodItem(user, buildFoodItemCopyInput(source, name))
```
fiddur, line 757: Non-blocking: there's a TOCTOU gap between `findAvailableCopyName` and this upsert. `upsertFoodItem` runs `ON CONFLICT (name_lower) DO UPDATE`, so if two duplicates of the same source race (the repo runs in multiple clones/sessions per CLAUDE.md), both can see `"X (copy)"` as free and the second upsert will overwrite the first's row instead of inserting a new one — the exact silent overwrite `findAvailableCopyName` is meant to prevent. Low risk for a single-user app, but worth a note; a unique-name retry or letting the insert fail-and-retry on conflict would close it.

apps/web/src/pages/FoodItems/index.tsx
```diff
@@ -132,6 +156,15 @@ export function FoodItems() {
                 <td class="fi-num">{item.fiber ?? '—'}</td>
                 <td class="fi-source">{item.source ?? '—'}</td>
                 <td class="fi-row-actions">
+                  <button
+                    type="button"
+                    class="btn-secondary"
+                    onClick={() => duplicateMutation.mutate(item.id)}
+                    disabled={duplicateMutation.isPending}
```
fiddur: Minor UX: this is a single shared `duplicateMutation`, so `disabled={duplicateMutation.isPending}` disables the Duplicate button on *every* row while any one duplicate is in flight. Fine functionally; if you want per-row disabling you'd need to track the pending source id (e.g. compare against `duplicateMutation.variables`).

apps/backend/src/services/food-items.ts
fiddur: Good catch. Closed the loop by documenting it as a best-effort limitation in `findAvailableCopyName` (3a8961e2) — mirroring the existing concurrent-`setIngredients` note. For this single-actor, per-user app the race (two sessions duplicating the *same* source within the same instant) is vanishingly rare, so I opted not to add a fail-and-retry insert path.

apps/web/src/pages/FoodItems/index.tsx
fiddur: Fixed in 3a8961e2 — now `disabled={duplicateMutation.isPending && duplicateMutation.variables === item.id}`, so only the in-flight row is disabled.

fiddur (commented):
✅ Approved

Solid, well-tested feature. `duplicateFoodItem` is cleanly decomposed (name-dedup, copy-input builder, portion remap, default-portion update) and the ordering is correct: create the row, then set ingredients + re-cache derived nutrients, then recreate portions with fresh ids and remap `default_portion_id`/`default_log_quantity`, then reference + sensitivities. It correctly handles both per-user and central sources (central fork → editable `source: 'manual'` copy), and the reference is only copied for atomic items. The copy-name dedup normalizes the same way (`toLowerCase().trim()`) as the `name_lower` upsert conflict key, so the `(copy)` / `(copy N)` walk genuinely avoids the silent ON CONFLICT overwrite it documents.

API/MCP parity is respected (MCP tool returns the raw service detail like `get_food_item`; the REST route serializes via `serializeDetail`), docs in `meals.md` are updated, and the frontend wires the button on both the list and detail pages with sensible per-row disabling (`mutation.variables === item.id`) and error surfacing on the list page (which navigates away on success). Test coverage is good — 8 integration cases asserting copy independence, composite re-derivation, fresh portion ids, dedup, and the central fork, plus router 404/200 cases.

Non-blocking observations:

- **Mixed-in formatting churn.** Several files unrelated to the feature (`calorie-computation.ts`, `EntityDetail/activityStats.ts`, `schemas/admin.ts`, `schemas/food-item-portions.ts`, `schemas/meals.ts`, and a few docs) carry pure oxfmt reflow. It's consistent with the project's `pnpm fix` workflow and the dedicated style commit, just adds review noise.
- **No enclosing transaction.** The duplicate performs ~6 sequential writes (upsert → setIngredients → cacheCompositeNutrients → copy portions → update defaults → reference/sensitivities) without wrapping them in one transaction, so a mid-way failure could leave a partial copy. This matches the existing service patterns (e.g. `set_food_item_ingredients` does setIngredients + cache separately) and is low-risk for a single-actor per-user DB, so not a blocker — just worth being aware of.

Couldn't execute the suite (node_modules absent in this worktree); review is by reading. The integration tests rely on testcontainers and should be run in CI.

PR #810, opened 2026-06-12
feat(food-items): portion units for recipe ingredients

fiddur:
## What

Recipe **ingredients** could only use a free-text unit — you couldn't pick a defined portion the way meal items can. So in a copied recipe you couldn't say "2 brödkaka". This adds portion/unit selection to recipe ingredients, mirroring how meals already work.

## How

An ingredient can now carry `food_item_portion_id` + `portion_count` (a portion of the *ingredient* food). When set, its nutrient contribution scales by:

```
portion_count × portion.base_equivalent / ingredient.default_quantity
```

— the same formula meal items use. The legacy `quantity` + `unit` path still works; for the portion path both columns are filled from the portion (`portion_count` + `label_unit`) as a display fallback, while scaling re-derives live from the portion id. The portion must belong to the ingredient food — validated in a shared `prepareIngredientInputs` used by **both** the REST route and the MCP tool.

## Changes

- **DB**: nullable `food_item_portion_id` + `portion_count` on `food_item_ingredients` (table def + idempotent `ALTER … ADD COLUMN IF NOT EXISTS` migration), round-tripped through `set`/`getIngredients`.
- **api-spec**: `foodItemIngredientSchema` gains the portion fields + two refines (portion_count required with a portion id; quantity required without one). Resolved schema returns them.
- **service**: `computeIngredientScale` / `aggregateNutrientsFromIngredients` use the resolved portion; `getDetail` resolves each ingredient's portion; `duplicateFoodItem` carries the portion fields.
- **web**: the ingredient editor swaps the free-text unit input for a portion **picker** (base unit + the ingredient food's portions) and carries the typed number across unit switches — reusing the meal `UnitPicker`/`QuantityInput` pattern.
- **docs**: `meals.md` composite + MCP sections.

## Tests

- Integration (testcontainers): DB round-trip of portion columns, portion scaling end-to-end (`2 brödkaka` → 140 kcal), legacy still scales, portion-ownership rejection, cached row columns.
- Unit: `aggregateNutrientsFromIngredients` portion scaling + deleted-portion incomplete flag; route tests for portion validation (400 not-belonging / 200 persists with filled display columns).
- `pnpm check` green (0 errors).

## Note

`food_item_portion_id` is a soft pointer (no FK), consistent with `meal_food_items` and `food_item_portions` — portions can target per-user or central foods.

### DIFF
--- a/apps/backend/src/db/connection.ts
+++ b/apps/backend/src/db/connection.ts
@@ -1124,6 +1124,18 @@ export const migrateSchema = async (user: string) => {
     )
   }
 
+  // food_item_ingredients: per-ingredient portion selection, mirroring
+  // meal_food_items. Both nullable so legacy ingredients (logged with free-form
+  // quantity + unit before portions existed) keep working — the scaling falls
+  // back to (quantity, unit) when food_item_portion_id is NULL.
+  if (existingTableNames.has('food_item_ingredients')) {
+    await query(db, `ALTER TABLE food_item_ingredients ADD COLUMN IF NOT EXISTS food_item_portion_id UUID`)
+    await query(
+      db,
+      `ALTER TABLE food_item_ingredients ADD COLUMN IF NOT EXISTS portion_count DOUBLE PRECISION`,
+    )
+  }
+
   // meal_food_items: drop the FK on food_item_id since the canonical row
   // may now live in the central shared_food_items table (different
   // database). The food_item_name + food_item_icon columns existed in
--- a/apps/backend/src/db/food-item-ingredients.ts
+++ b/apps/backend/src/db/food-item-ingredients.ts
@@ -14,7 +14,7 @@
 import { query } from './connection.ts'
 
 const COLUMNS =
-  'id, parent_food_item_id, ingredient_food_item_id, quantity, unit, sort_order, created_at, updated_at'
+  'id, parent_food_item_id, ingredient_food_item_id, quantity, unit, food_item_portion_id, portion_count, sort_order, created_at, updated_at'
 
 export interface FoodItemIngredientRow {
   id: string
@@ -22,6 +22,10 @@ export interface FoodItemIngredientRow {
   ingredient_food_item_id: string
   quantity: number
   unit?: string
+  /** When set, the ingredient is measured in this portion (unit); scaling uses portion_count × base_equivalent. */
+  food_item_portion_id?: string
+  /** The count entered in the portion's unit (paired with food_item_portion_id). */
+  portion_count?: number
   sort_order: number
   created_at: Date
   updated_at: Date
@@ -31,6 +35,8 @@ export interface FoodItemIngredientInput {
   ingredient_food_item_id: string
   quantity: number
   unit?: string
+  food_item_portion_id?: string
+  portion_count?: number
   sort_order?: number
 }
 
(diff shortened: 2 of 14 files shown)


fiddur (commented):
✅ Approved

Solid, well-tested feature that adds per-ingredient portion units to composite recipes, closely mirroring the existing meal-portion path (`resolvePortionForInput` / `buildScaledJunctionItem` / `computeScale`). I checked the scaling math, schema, migration, validation ordering, and back-compat and found no blocking issues.

**What I verified:**
- **Scaling parity** — `computeIngredientScale`'s portion branch (`portion_count × base_equivalent / default_quantity`) matches `meals.ts` `computeScale`, and correctly flips `nutrient_data_incomplete` when a referenced portion no longer resolves (deleted unit). Both the read path (`getDetail`) and the cache path (`cacheCompositeNutrients`) resolve portions, and integration tests cover both (140 kcal / 5.6 g, cached 300 kcal).
- **Validation** — `prepareIngredientInputs` enforces that the portion exists and belongs to the ingredient food, and that `portion_count` is positive, before persisting; both the REST route and MCP tool wrap it and surface a 400 / error. Cycle check runs first, as before.
- **Schema** — the refactor to a bare `foodItemIngredientFields` object (so `.refine()` on the request schema and `.object({...fields, name, icon})` on the response schema both work) is necessary and correct, since the refined request schema is now a `ZodEffects` that lacks `.extend()`. The `.object().refine().refine().meta({id})` shape matches the already-shipping `foodItemInputSchema` in `meals.ts`, so OpenAPI model generation is unaffected. Making `quantity` optional is a safe relaxation — the column is NOT NULL and `prepareIngredientInputs` always fills `quantity` (= `portion_count` on the portion path), so responses always carry it.
- **Migration** — `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` for existing DBs plus the new columns in `schema/meals.ts` `mealsTables` for fresh DBs; both nullable, preserving legacy quantity/unit rows. `setIngredients` correctly bumps to 7 columns/row inside its existing transaction.
- **Frontend** — `IngredientList` fetches each ingredient food's detail (react-query cached by id) for its portions + base unit; `persist` sends the portion path (portion id + count, omitting quantity so the backend derives it) vs. the legacy path. `duplicateFoodItem` copies the portion columns.

**Minor, non-blocking notes (no action required):**
- The unit picker commits immediately and *carries the number* across a unit switch — switching a "2 brödkaka" (≈70 g) ingredient to the base unit "g" yields "2 g", not "70 g", silently changing the recipe's derived totals. This is the documented, intentional behavior and matches the meal-logging UI, so it's fine; just flagging it as a UX subtlety.
- `getDetail` resolves portions via one `getFoodItemPortionById` per distinct portion id (parallel). Fine for typical recipes; if detail ever gets hot, a `getPortionsByIds` batch would remove the fan-out (the sibling `getPortionsByFoodItemIds` already exists for the food-id case).
- No frontend component test for the new `UnitPicker` / carry logic, but that's consistent with the repo (apps/web has no `.test.tsx` files); backend coverage is thorough.

Docs (`docs/features/meals.md`) are updated for both the data model and the MCP tool surface. Generated OpenAPI/Kotlin artifacts are gitignored, so no regenerated files are expected in the diff.

PR #812, opened 2026-06-15
fix(locations): stop a stale GPS fix from swallowing travel + destination (#811)

fiddur:
Fixes #811.

## The bug

An unknown (`"Somewhere"`) place visit was built by merging *every* consecutive unmatched fix purely by name, anchored at the first fix and reported at its coordinates. So a single stale GPS breadcrumb on the way to an un-named destination collapsed the entire intervening block — **travel time and the real destination** — into one long stay sitting at the **wrong (travel-waypoint) coordinates**. That's why the ER trip showed one `Somewhere` 14:50→22:54 at a point near Lönnåsen instead of the hospital, and why the hospital couldn't be attached.

## The fix (`apps/backend/src/services/locations.ts`)

**`getPlaceVisits`** — an unknown visit now breaks when a fix:
- moves **beyond the cluster radius (200 m) from the visit's running centroid** (real movement), or
- arrives after a **> 90-min gap with no fixes** (no evidence of continuous presence).

It also reports the unknown visit at its **running centroid**, not the first (possibly stale) fix. So the dwell at the destination surfaces as **its own stay at its own coordinates** — promotable/renameable.

**`mergeShortUnknownVisits`** — a short unknown blip is absorbed into a neighbour only when **genuinely contiguous** (close in *both* space and time). A brief GPS glitch still stitches back into the stay it interrupted, but a chain of travel fragments can no longer keep extending a neighbouring stay (which would otherwise re-create the swallowing).

Also refactored the visit loop into small helpers (`classifyFix` / `continuesVisit` / `extendVisit` / `finalizeVisit`).

## Behaviour after the fix (the reported day)

- Morning `Hökås` ends ~11:12 (no longer stretched across the drive).
- The hospital dwell is its own `Somewhere` stay at the hospital coordinates (~7 h) — attachable as a named place.
- The travel waypoint and drive fixes no longer form a stay.

## Tests

- Unit: new space/time gating in `mergeShortUnknownVisits` (drops a far travel fragment; refuses to bridge a long time gap) — existing glitch-merge tests still pass.
- Integration: `getPlaceVisits` reproduction of the #811 ER-trip scenario (stale waypoint → ~7 h hospital dwell → home), asserting the hospital is its own stay at its coordinates, sized to the dwell (not the full block), and the morning home stay isn't extended.
- `pnpm check` green (0 errors).

## Note / possible follow-up

This fixes the user-facing `places`/visit segmentation (the reported symptom). The lower-level `detectStays` clustering already breaks on distance but does not break on long *time* gaps; hardening that (so a multi-hour no-data gap doesn't inflate a detected-location's stay duration) could be a follow-up if desired.

### DIFF
--- a/apps/backend/src/services/locations-place-visits.integration.test.ts
+++ b/apps/backend/src/services/locations-place-visits.integration.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Integration test for getPlaceVisits — issue #811.
+ *
+ * A stale GPS fix on the way to an un-named destination used to be merged with
+ * every later "unknown" fix (by the shared "Somewhere" name), producing one
+ * long stay anchored at the first (travel) fix that swallowed the drive and the
+ * real destination. getPlaceVisits must instead break unknown visits on
+ * movement / long gaps, so the destination surfaces as its own stay at its own
+ * coordinates (promotable to a named place).
+ */
+import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest'
+
+import type { Location } from '../db/types.ts'
+
+import { insertLocations, insertNamedLocation } from '../db/locations.ts'
+import { cleanTestDb, getTestUser, startTestDb, stopTestDb } from '../test/db-test-helper.ts'
+import { getPlaceVisits } from './locations.ts'
+
+const CONTAINER_TIMEOUT = 120_000
+
+const HOME = { lat: 57.66, lon: 12.6 } // "Hökås", named
+const HOSPITAL = { lat: 57.72, lon: 12.92 } // un-named destination, ~25 km east
+
+const at = (hour: number, minute: number, lat: number, lon: number): Location => ({
+  lat,
+  lon,
+  regions: [],
+  source: 'owntracks',
+  time: new Date(Date.UTC(2026, 5, 14, hour, minute, 0)),
+})
+
+const haversine = (a: { lat: number; lon: number }, b: { lat: number; lon: number }): number => {
+  const R = 6371000
+  const dLat = ((b.lat - a.lat) * Math.PI) / 180
+  const dLon = ((b.lon - a.lon) * Math.PI) / 180
+  const h =
+    Math.sin(dLat / 2) ** 2 +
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Solid, well-targeted fix for #811. The refactor of `getPlaceVisits` into `classifyFix` / `continuesVisit` / `extendVisit` / `finalizeVisit` is clean and preserves the prior identity-key behavior (named_location_id still splits same-named places), while adding the two guards that matter for unknown stays: a centroid break radius and a max-gap cap. The running-centroid math in `extendVisit` is a correct incremental mean (`point_count` initialized to 1), and `finalizeVisit` correctly drops the bookkeeping field from the `PlaceVisit` output.

The `mergeShortUnknownVisits` contiguity gating (space ≤ 200m AND time ≤ 30min) is the right complement — it stops a chain of brief travel fragments from repeatedly extending a neighbouring stay. I traced all pre-existing tests (they use a single shared coordinate and zero gaps, so they stay contiguous and their expectations are unchanged) and both new unit tests, plus the integration scenario, and they all hold: the stale 14:50 waypoint and the travel fixes split off and get dropped, the hospital surfaces as a single ~420-min stay at its own centroid, and the morning home stay ends at 11:12.

I could not execute the suite in this environment (the sandbox blocks writing to the pnpm store outside the worktree, so `pnpm install` fails with EROFS), so this is based on a code/trace review rather than a live run; the integration test additionally needs Docker/testcontainers.

Two non-blocking design notes worth being aware of (no change required):

- **Sparse sampling at a real unknown destination.** A genuine long stay at an un-named place sampled less often than every 90 minutes (e.g. OwnTracks significant-location mode, or phone idle) will break into multiple sub-90-min segments, each of which may be short enough to be dropped — losing a real stay. The 90-min cap is a reasonable heuristic for the common case, but it does trade away the sparse-sampling tail.

- **Centroid drift / mid-stay outliers.** A single GPS outlier > 200m mid-cluster, or slow wandering within a band wider than 200m, will fragment one unknown stay into pieces (the outlier dropped as a 0-min visit). This is consistent with `DEFAULT_CLUSTER_RADIUS_METERS` used elsewhere, so it's an acceptable and intentional consistency choice — just flagging the behavior.

Nice test coverage on both the unit and integration layers, and the comments explaining the why (with the issue reference) are excellent.

PR #813, opened 2026-06-30
feat(sharing): shared dashboards + public user pages (backend)

fiddur:
## Summary

Phase 1 (backend) of the social features. Lets a user publish **read-only, independently-editable** copies of a dashboard under a public namespace, viewable by anyone (anonymous or cross-instance) without auth.

- Public profile: `<public-base>/u/<username>` (lists public shares)
- A shared dashboard: `<public-base>/u/<username>/<slug>` (public *or* unlisted-by-slug)

### Federation identity
A user's global identity is their full public **base URL**, not `user@host` — so an instance can run over http or under a sub-path of another site (`http://some.thing/with/other/things/u/fiddur`). Absolute URLs are built from the configured public base (`WEB_HOST`), preserving sub-paths.

### Hard backend security
The public viewer can only ever receive the exact data the dashboard's own widgets render:
- public endpoints take **no** data-shaping params from the request;
- the resolver (`services/shared-dashboard-data.ts`) calls the same user-scoped services as the authed dashboard, with params taken **only** from the stored widget config;
- each widget returns a **minimal projection** (no raw rows / notes / locations);
- quick-link hrefs are stripped.

## What's included
- **api-spec** `schemas/social.ts`: CRUD bodies/responses, public profile + public dashboard responses, per-widget `widget_data` (keyed by widget id).
- **DB**: per-user `shared_dashboards` table (registered in schema + migrate loop), CRUD with url-safe slug generation + collision retry, integration tests.
- **Resolver** + integration test (runs every widget type against real PG; asserts it only takes `(user, config)`).
- **Routers**: authed `/shared-dashboards` CRUD; unauthenticated `/public/:username/dashboards` and `/public/:username/:slug`.
- **MCP parity**: `list/create/update/delete_shared_dashboard`.
- **Docs**: `docs/features/sharing.md` (+ link from `dashboard.md`).

## Forward-compat
`widget_data` is keyed by stable widget id and responses carry each widget's `id`+`type`, so a later per-chart endpoint (`/u/<user>/<slug>/<widgetId>`) and cross-dashboard embedding are a thin addition.

## Not in this PR
Frontend (widget presentational/container split, public pages, management UI) — Phases 2 & 3. Kotlin model regen — deferred until Android consumes these models.

## Testing
- `pnpm check` clean in `apps/backend` and `packages/api-spec`.
- New integration tests pass (testcontainers): `db/shared-dashboards.integration.test.ts`, `services/shared-dashboard-data.integration.test.ts`.


apps/backend/src/services/shared-dashboard-data.ts
```diff
@@ -0,0 +1,264 @@
...
+  const { end, start } = lookbackRange(config.lookback_days ?? 7)
+  const types = await getAllActivityTypeNames(user)
+  const activities = await queryActivities(user, types, start, end)
+  return {
+    data: {
+      activities: activities.map((a) => ({
+        activity_type: a.activity_type,
+        end_time: a.end_time,
+        start_time: a.start_time,
+      })),
```
fiddur: This over-exposes the owner's data relative to what the widget renders. `ActivitySummaryWidget` only shows three aggregates (workout count/minutes, avg sleep hours/nights, meditation count/minutes) and respects `show_workouts`/`show_sleep`/`show_meditation`. Here you fetch **all** activity types via `getAllActivityTypeNames` and ship the full per-activity list (type + exact start/end times) to anonymous viewers — including activity types outside the three summary categories, and ignoring the visibility toggles. That's a complete activity timeline over the wire, which breaks the PR's stated guarantee that a public viewer only receives "the exact data the dashboard's own widgets render." Aggregate this server-side to the same counts/averages the widget displays (filtered by category and the show_* flags) instead of returning raw rows.

docs/features/sharing.md
```diff
@@ -0,0 +1,93 @@
...
+| `DELETE /shared-dashboards/:id` | Delete (its slug stops resolving) |
+
+Public (unauthenticated):
+
+| Method & path                          | Purpose                                   |
+| -------------------------------------- | ----------------------------------------- |
+| `GET /public/:username/dashboards`     | List a user's **public** shared dashboards |
+| `GET /public/:username/:slug`          | View one shared dashboard + resolved data |
+
+The same CRUD capability is available over MCP as `list_shared_dashboard`,
```
fiddur: Tool name typo: the registered MCP tool is `list_shared_dashboards` (plural), not `list_shared_dashboard`.

fiddur (commented):
🛑 Changes required

Solid, well-documented backend for shared dashboards + public pages. The architecture is sound: the public routes take no viewer-supplied data-shaping params, the security boundary lives in one place (`resolveDashboardData(user, config)`), most resolvers emit minimal aggregate projections, and `quick_link` hrefs are neutralized. I verified the public username validation (`/^[a-z][a-z0-9_]{2,30}$/` + `RESERVED_USERNAMES`) exactly matches signup, so the username — which selects the per-user database — can't be abused for catalog/role injection. DB-layer and resolver are covered by integration tests.

**One blocking issue — `activity_summary` over-exposes data, contradicting the PR's stated guarantee.** The authenticated `ActivitySummaryWidget` only ever renders three aggregates (workout count + total minutes, avg sleep hours + nights, meditation count + total minutes) and honors the `show_workouts`/`show_sleep`/`show_meditation` toggles. But `resolveActivitySummary` ships the owner's **entire** activity timeline to anonymous viewers: every activity type (via `getAllActivityTypeNames`, not just the three summary categories) with exact `start_time`/`end_time` per activity, ignoring the visibility toggles. For a health app that tracks sensitive activity types and has explicit sensitivity flags, this is a real privacy regression and directly violates the doc's promise that "a public viewer can only ever receive the exact data the dashboard's own widgets render." Resolve this server-side to the same aggregates the widget shows (filtered by category + the show_* flags) rather than emitting the raw list. Every other widget resolver is consistent with its widget; this is the lone exception.

**Non-blocking notes:**
- `resolveDashboardData` fans out over every widget in a stored config with no cap. An account holder could publish a config with many expensive widgets (e.g. `correlation`) and repeatedly hit the unauthenticated, `private`/60s-cached public URL to amplify load on their own per-user DB connection. Consider bounding widget count or rate-limiting the public endpoints.
- The public routes are the first unauthenticated callers of `getDbForUser`, which caches a persistent `pg.Client` per user in a module-level map. Worth confirming this is acceptable (it's bounded to real usernames, since a missing DB throws before caching).
- No router-level tests for `public-shares-router` (the security-critical `isValidPublicUsername`, missing-DB 404 handling, and `sanitizeConfig` are all unexported and untested). Given this is the public attack surface, a couple of supertest-style cases would be worthwhile.
- `GET /public/:username/dashboards` returns 200 (with an empty list) for any existing user even with zero public shares, vs 404 for non-existent — an enumeration oracle. Likely fine since usernames are public profile identifiers by design, just flagging.
- Minor: the final `throw new Error('Failed to generate a unique shared dashboard slug')` in `createSharedDashboard` is unreachable — the last attempt rethrows the original unique-violation error.

apps/backend/src/routes/public-shares-router.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (!isValidPublicUsername(username)) {
+        return res.status(404).json({ error: 'Dashboard not found', success: false })
+      }
+      try {
+        const record = await getSharedDashboardBySlug(username, slug)
+        if (!record) {
+          return res.status(404).json({ error: 'Dashboard not found', success: false })
+        }
+        const widgetData = await resolveDashboardData(username, record.config)
+        res.setHeader('Cache-Control', 'private, max-age=60')
```
fiddur: `private` here prevents shared/CDN caches from storing this response, which is counter to the goal of bounding fan-out on cached public requests — and this content is intentionally public/anonymous. Consider `public, max-age=60` (or `s-maxage`). Also note the list endpoint (`/public/:username/dashboards`) sets no `Cache-Control` at all; a consistent policy across both public endpoints would be cleaner. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-tested backend for shared dashboards + public user pages. The security model is the heart of this PR, and it holds up under scrutiny:

- **Username safety verified.** `isValidPublicUsername` uses exactly the signup regex `^[a-z][a-z0-9_]{2,30}$` plus the shared `RESERVED_USERNAMES` list (confirmed identical against `auth-routes.ts`). The username only ever flows into `userDbName()` (a `pg.Client` `database` connection param, not interpolated SQL) and `SET ROLE %L` (pg-format literal-escaped) — no injection surface, with defense-in-depth on top. A non-existent user surfaces as `3D000` and is mapped to 404.
- **Resolver is a clean boundary.** `resolveDashboardData(user, config)` takes no viewer input (enforced by an arity test), reads parameters exclusively from the stored config, and each widget resolver returns only the aggregated projection the widget renders — never raw rows. The `activity_summary` integration test explicitly asserts no `activities` array leaks. Slug lookups are parameterized; `quick_link` hrefs are neutralized; `sanitizeConfig` is non-mutating (tested).
- **Wiring checks out.** Spot-verified the service call shapes the resolver depends on: `getEffectiveHrZones` (`zones[1..5]`), `queryActivities` (string timestamps — `durationMinutes`'s typing matches), `getChartData` (`activity_type_id ?? tag_definition_id`, so mapping `config.tag_definition_id → activity_type_id` is correct), and `chartDataBucketSchema` ({bucket_start, value}). The three activity-summary categories are disjoint (`isExerciseActivityType` excludes sleep/meditation). `migrateSchema` creates `shared_dashboards` for existing users via `tableCreationOrder`, and generated api-spec artifacts are gitignored (built in CI), so nothing is missing there.

No blocking issues. A few non-blocking notes for consideration:

1. **Cache-Control on the public detail endpoint** uses `private, max-age=60` on intentionally-public, anonymous content — `private` prevents shared/CDN caching, which works against the stated goal of bounding fan-out on cached public requests. The list endpoint (`/public/:username/dashboards`) sets no cache header at all. Consider `public` and a consistent policy across both. (inline note)
2. **Unauthenticated traffic can trigger a per-user schema migration.** For a user created before this table existed who hasn't reconnected, the first public hit throws `42P01` and `query()`'s retry runs `migrateSchema` for that user — a relatively heavy operation now reachable by anonymous callers. It's coalesced (once per user), idempotent, and only for existing users, so it's bounded; just worth being aware that this PR first exposes that retry path to unauthenticated requests.
3. **Minor MCP/REST parity gap.** REST exposes `GET /shared-dashboards/:id` but there's no `get_shared_dashboard` MCP tool. Functionally covered by `list_shared_dashboards` (returns full configs), so not a real capability gap, just noting against the parity guideline.

apps/backend/src/routes/public-shares-router.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+import type { DashboardConfig, PublicProfileResponse, PublicSharedDashboardResponse } from '@aurboda/api-spec'
+
+import { RESERVED_USERNAMES } from '../api/auth-routes.ts'
+import { getSharedDashboardBySlug, listPublicSharedDashboards } from '../db/index.ts'
+import { buildProfileUrl, buildShareUrl } from '../services/share-urls.ts'
+import { resolveDashboardData } from '../services/shared-dashboard-data.ts'
+import { type TypedRouter, typedRouter } from '../typed-router.ts'
+
+const USERNAME_RE = /^[a-z][a-z0-9_]{2,30}$/
```
fiddur: This regex duplicates the signup validation in `auth-routes.ts` (`/^[a-z][a-z0-9_]{2,30}$/`). Because this gate also decides which per-user database an anonymous request connects to, the two must stay in lockstep — if signup's rule is ever loosened without updating this copy, usernames this gate was meant to block could slip through. You already import `RESERVED_USERNAMES` from `auth-routes.ts`; consider exporting and reusing the regex (or a shared `isValidUsername`) from there too rather than re-declaring it.

fiddur (commented):
✅ Approved

Solid, well-tested backend for shared dashboards + public pages. The security boundary is carefully designed and holds up: the public, unauthenticated routes validate `username` against the same regex as signup before using it as the per-user DB selector (so the `aurboda_<username>` database name can't be manipulated), all SQL is parameterized, `resolveDashboardData(user, config)` takes no viewer input, each resolver returns only a minimal projection (no raw activity rows/notes/locations), quick-link hrefs are neutralized, and the slug-generation retry handles unique-violation collisions. Integration tests exercise the CRUD and every widget type against a real Postgres, and `sanitizeConfig` is verified non-mutating. Express 5 forwards the `throw error` rethrows to the error middleware, so those are safe.

A few non-blocking considerations:

1. **Duplicated username regex (defense-in-depth).** `USERNAME_RE` in `public-shares-router.ts` re-declares the signup regex from `auth-routes.ts`. Since this gate now also selects which database a fully-anonymous request connects to, drift between the two is a latent risk: loosening the signup rule later without updating this copy could let usernames through that the gate was meant to reject. It already imports `RESERVED_USERNAMES` from `auth-routes.ts` — exporting and reusing the regex (or a shared `isValidUsername`) from the same place would keep the two in lockstep.

2. **`Cache-Control: public, max-age=60` vs. revocation.** `docs/features/sharing.md` states deleting a share "immediately makes its slug return 404," but with a public/shared-cache TTL of 60s a deleted or just-made-unlisted dashboard can still be served from a browser/CDN cache for up to a minute after revocation. For a privacy feature that's a minor but real staleness window worth being aware of (and worth softening the doc wording, or lowering the TTL for the per-share endpoint).

3. **No rate limiting on the new unauthenticated, DB-touching surface.** These are the first anonymous endpoints that open/serve from per-user databases and fan out up to `MAX_RESOLVED_WIDGETS` (60) widget resolvers, each issuing several queries. The widget cap and 60s caching help, but an attacker varying usernames/slugs bypasses the cache and can trigger repeated `pg.Client` connection attempts. There's no global rate limiter in the repo today, so this is a pre-existing gap rather than a regression — just flagging that this PR meaningfully expands the unauthenticated attack surface.

apps/backend/src/services/shared-dashboard-data.ts
```diff
@@ -0,0 +1,313 @@
...
+            count: sleep.length,
+          }
+        : null,
+    },
+    type: 'activity_summary',
+  }
+}
+
+const resolveHrZones = async (user: string): Promise<WidgetData> => {
+  const { end, start } = lookbackRange(7)
```
fiddur: `resolveHrZones` hardcodes a 7-day window and ignores the widget's `config.lookback_days` (the `hr_zones` config schema defines it, default 7). An owner who set e.g. `lookback_days: 30` would see their authenticated widget differ from the public view. Other resolvers honor the config (`resolveSparklineCard` uses `config.lookback_days ?? 30`); consider passing `config` here and using `lookbackRange(config.lookback_days ?? 7)` for consistency. Minor — only matters when the owner customized the window.

fiddur (commented):
✅ Approved

Solid, well-structured backend for shared dashboards + public pages. The security boundary is thoughtfully designed and matches the documented guarantee:

- The username (which selects the per-user DB) is validated against the same strict `USERNAME_REGEX` used at signup before it's ever used to target a database — the `isValidUsername` refactor correctly preserves the separate `RESERVED_USERNAMES` check in `/signup` (line 137-140), so behavior is unchanged.
- Slug lookups are parameterized; the resolver takes only `(user, config)` with no viewer-supplied channel; each widget resolver returns a minimal projection; `activity_summary` emits only aggregates (verified by test); quick-link hrefs are neutralized; fan-out is bounded by `MAX_RESOLVED_WIDGETS` with a logged warning; and `Cache-Control: public, max-age=60` is appropriate for anonymous content (404s are not cached since the header is set only on the success path).
- Express 5 (`^5.1.0`) auto-forwards async handler rejections, so the `throw error` re-raise in the public routes is safe.
- Good test coverage of the security-critical helpers (`isValidUsername`, `sanitizeConfig`) and DB/resolver integration. Schemas are exhaustive — the 9 widget literals match the resolver and `nullData` switches. Generated OpenAPI/Kotlin artifacts are gitignored, so nothing is left stale.

I could not run `pnpm check` (the sandbox's read-only pnpm store blocks `pnpm install` in this throwaway worktree), so the build is unverified here, but a static pass over imports, discriminated unions, and types found no compile concerns; CI will confirm.

Non-blocking notes (not requiring changes):
- `/public/:username/dashboards` returns 200 (possibly empty) for an existing user vs 404 for a non-existent one, allowing username enumeration. Acceptable here since usernames are already public identifiers in the share URLs.
- The response returns the full sanitized `config` (all widgets) but `widget_data` is capped at 60 entries, so a >60-widget dashboard would render some widgets without data. Bounded and logged; fine given real dashboards are far smaller.

One minor inline note on `resolveHrZones` below.

fiddur (commented):
✅ Approved

Solid, well-structured backend for shared dashboards + public pages. The security boundary is the heart of this PR and it holds up well:

- **Username → DB selection is safe.** Public routes never touch `req.user`; they resolve the owner DB by username, gated through the shared `isValidUsername` (same regex + reserved-name list as signup) *and* `pg-format`'s `SET ROLE %L` escaping in `getDbForUser`. A missing DB (`3D000`) and an invalid username both collapse to a 404, so existence isn't leaked beyond what the public profile URL already implies.
- **No viewer input reaches the owner's data.** `resolveDashboardData(user, config)` takes only the stored config; each resolver returns a minimal projection (aggregates/means/series points, never raw rows with notes/titles/locations), `activity_summary` emits only counts/minutes and respects visibility toggles, and `quick_link` hrefs are neutralized via `sanitizeConfig`. The `(user, config)`-only signature is even asserted in a test. Per-request fan-out is bounded by `MAX_RESOLVED_WIDGETS`, and a widget resolver throwing degrades to a null-data entry rather than failing the whole response.
- **Slug-as-capability** (56 bits, base64url) is reasonable for unlisted shares, and the expensive resolution path is only reached after a valid slug lookup, so the unauthenticated DoS surface is limited. `Cache-Control: public` is safe here since the cache key includes the secret slug and responses aren't viewer-specific.

Correctness spot-checks all passed: every imported service symbol exists and the positional calls (`queryMetrics`, `getPeriodSummary`, `getActivityImpact`, `getTrend`, `getChartData`) match their signatures; `hr.zones[1..5]` matches `HrZoneThresholds`; `ActivityResult.start_time/end_time` are strings so `durationMinutes`' `new Date(...)` is fine. Tests (unit + integration), docs, schemas, and MCP/REST parity are all in good shape, and the generated OpenAPI/Kotlin artifacts are correctly gitignored.

Two minor, non-blocking observations (neither needs to gate merge):
1. When a config exceeds `MAX_RESOLVED_WIDGETS`, the echoed `sanitizeConfig(record.config)` still contains all widgets while `widget_data` only covers the first 60 — the frontend will render the overflow widgets with no data. Edge case (real dashboards are far smaller, and it's logged), but consider truncating the echoed config to match, or documenting the asymmetry.
2. `resolveDashboardData` keys by `widget.id` via `Object.fromEntries`, so any duplicate widget id across sections would silently collapse (last wins). Owner-generated ids are presumably unique, so this is only a theoretical concern.

PR #814, opened 2026-06-30
feat(sharing): split dashboard widgets into View + container (frontend)

fiddur:
## Summary

Phase 2 of the social features (frontend foundation). Refactors each dashboard widget into a presentational `*View` (renders from a resolved data payload) plus a thin container that fetches and maps to that payload. The home dashboard is behaviour-identical; a new `PublicWidgetRenderer` reuses the Views to render a shared dashboard read-only from server-resolved `widget_data` (consumed by Phase 3).

Follows #813 (backend), now merged.

## What's included
- Extracted Views: `MetricCardView`, `SparklineCardView`, `TrendChartView`, `BarChartView`, `CorrelationView`, `ActivitySummaryView`, `HrZonesView`, `GoalProgressView`, and a non-navigational `QuickLinkView`; shared `TrendIndicator`.
- Each container keeps its existing loading/error behaviour, then renders its View — no change to the home dashboard.
- Chart views (`bar`/`trend`) link to the full chart page only when given an `href` (home); the public view omits it.
- `ActivitySummaryView` renders from the per-category aggregates the backend now returns (exercise/sleep/meditation, null = hidden), matching the privacy-aligned resolver.
- `PublicWidgetRenderer` maps a widget + its `widget_data` entry to the right View; quick links render as static (non-navigational) tiles.

## Not in this PR
Public pages (`/u/:username`, `/u/:username/:slug`) and the `/shared-dashboards` management UI — Phase 3.

## Testing
- `pnpm check` clean in `apps/web` and `apps/backend`; existing widget tests pass.

### DIFF
--- a/apps/web/src/components/widgets/MetricCardWidget.tsx
+++ b/apps/web/src/components/widgets/MetricCardWidget.tsx
@@ -1,8 +1,12 @@
 /**
  * MetricCardWidget - Displays a single metric value with optional trend indicator.
+ *
+ * Split into a presentational `MetricCardView` (renders from a resolved data
+ * payload) and a `MetricCardWidget` container that fetches and maps to that
+ * payload. The public shared-dashboard renderer reuses `MetricCardView`.
  */
 
-import type { MetricCardConfig } from '@aurboda/api-spec'
+import type { MetricCardConfig, MetricCardData } from '@aurboda/api-spec'
 
 import { useQuery } from '@tanstack/react-query'
 import { endOfDay, startOfDay, subDays } from 'date-fns'
@@ -14,25 +18,7 @@ import {
   type BaselineData,
   type PeriodMetricStats,
 } from '../../state/api'
-
-// Trend indicator component
-function TrendIndicator({ value, inverse = false }: { value: number | null; inverse?: boolean }) {
-  if (value === null) return null
-
-  const isPositive = inverse ? value < 0 : value > 0
-  const arrow = value > 0 ? '\u2191' : value < 0 ? '\u2193' : '\u2192'
-  const className = isPositive ? 'trend-positive' : value === 0 ? 'trend-neutral' : 'trend-negative'
-
-  return (
-    <span class={`trend-indicator ${className}`}>
-      {arrow} {Math.abs(value).toFixed(1)}%
-    </span>
-  )
-}
-
-interface MetricCardWidgetProps {
-  config: MetricCardConfig
-}
+import { TrendIndicator } from './TrendIndicator'
 
 // Baseline metrics that come from the baseline API
 const baselineMetrics = ['hrv_7day', 'hrv_30day', 'rhr_7day', 'rhr_30day'] as const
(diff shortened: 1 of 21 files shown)


apps/web/src/components/widgets/MetricCardWidget.tsx
```diff
@@ -71,56 +37,104 @@ const metricToApiMetric: Record<string, string> = {
...
+        display: value !== null ? Math.round(value).toLocaleString() : null,
+        dynamicSubtitle: max !== null ? `Max: ${Math.round(max).toLocaleString()}` : undefined,
       }
     case 'zone2_weekly':
-      return {
-        subtitle: undefined,
-        trend: stats.change_from_previous_period_percent ?? null,
-        value: avg !== null ? Math.round((avg * 7) / 60) : null,
-      }
+      return { display: value !== null ? String(Math.round((value * 7) / 60)) : null }
```
fiddur: Minor behavioral change for `zone2_weekly`: the old code returned a numeric `value` that the renderer formatted with `value.toFixed(1)` (e.g. `5.0`), whereas this now returns `String(Math.round(...))` (e.g. `5`). The `steps`/default arms preserve their old formatting, so this is the one display difference in an otherwise behavior-preserving split. If dropping the `.0` is intentional, fine — just flagging it since the PR is framed as a pure View/container split.

fiddur (commented):
✅ Approved

Clean, well-scoped refactor that splits each dashboard widget into a presentational `*View` and a fetching container, with a new `PublicWidgetRenderer` that drives the views from the server-resolved `WidgetData` payload (no fetching). I verified the new api-spec data types (`MetricCardData`, `SparklineCardData`, `HrZonesData`, `ActivitySummaryData`, etc.) exist and match the field usage, that `DashboardWidget` is a discriminated union so `PublicWidgetRenderer`'s `switch (widget.type)` correctly narrows `widget.config` per arm, that the `HrZones` array↔object threshold conversion round-trips consistently (`hr_zone_start[i]` ↔ `toThresholds` key `i+1`), and that `GoalProgressData.goals` reuses the same `goalProgressSchema` that `fetchGoalsProgress` returns. The widened `SummarizableActivity` type in `activitySummary.ts` is still satisfied by the existing test's `Activity` fixtures, so the unit test continues to compile/pass.

Notes (non-blocking):
- `PublicWidgetRenderer` is exported from `index.ts` but not yet consumed anywhere in the web app — presumably wired in a follow-up PR. Worth confirming the public shared-dashboard page lands before relying on it.
- One small unintended display change for the `zone2_weekly` metric (see inline comment).

I couldn't run `pnpm check` here because the review worktree is on a read-only filesystem for pnpm's store (install fails with EROFS) — that's an environment limitation, not a PR issue. Please confirm CI is green.

fiddur (commented):
✅ Approved

Clean, mechanical refactor that splits each dashboard widget into a presentational `*View` and a fetching container, plus a new `PublicWidgetRenderer` that reuses the Views with server-resolved `widget_data`. The data shapes line up with the api-spec `*Data` schemas added in #813, and the split preserves behavior.

Verified locally:
- `pnpm --filter aurboda-web typecheck` passes; `test` passes (452/452); `lint` reports 0 errors (the 255 warnings are pre-existing, all in unrelated `Meals/index.tsx`).
- Behavior preservation spot-checks all hold: default title formulas (`BarChartView`/`TrendChartView`/`CorrelationView`) match the originals; `MetricCardView.computeDisplay` keeps the per-metric formatting (steps locale string, `zone2_weekly` `.toFixed(1)`, default `.toFixed(1)`, baseline via default branch) and the count/max subtitle rules; `ActivitySummaryView` null-category = hidden semantics match the old `show_*` toggles (card still renders at count 0); HrZones thresholds (object↔array) and zone seconds round-trip correctly; Sparkline series/value/trend round-trip correctly; query gating (`enabled: isBaseline` / `!isBaseline`) is unchanged.
- `PublicWidgetRenderer` covers all 9 widget types, matching `WidgetRenderer` and the `dashboard.ts` config union.

Non-blocking suggestions:
- No tests were added for the new presentational Views or `PublicWidgetRenderer`. The split makes them pure and easy to test (each renders from a static `*Data` payload), so a few render assertions — especially for `MetricCardView`'s metric-specific formatting and `ActivitySummaryView`'s null-category hiding — would lock in the behavior this PR was careful to preserve. The existing `summarizeActivities` test still passes against the widened `SummarizableActivity` input.
- The live containers reserialize through the wire shape (e.g. Sparkline `Date`→ISO→`Date`, HrZones thresholds object→array→object). It's harmless and intentional to keep one shared View, just noting the minor redundant work on the authenticated path.

PR #815, opened 2026-06-30
feat(sharing): public pages + shared-dashboards management UI

fiddur:
## Summary

Phase 3 (final) of the social features — the user-facing public pages and management UI. Builds on #813 (backend) and #814 (widget split), both merged.

- **`/shared-dashboards`** (authed, sidebar "Share"): create a shareable copy from your current home dashboard or a blank one, rename (auto-saves on blur), toggle **public/unlisted**, copy the share link, delete.
- **`/u/:username`** (public): a user's profile listing their public dashboards.
- **`/u/:username/:slug`** (public): read-only render of a shared dashboard via `PublicWidgetRenderer` using the server-resolved `widget_data` — **no per-widget fetching**, so the hard backend security boundary holds.
- Public `/u/...` pages render **without app chrome** (header/sidebar/footer) for a clean anonymous view.
- `state/api/social.ts`: authed CRUD + unauthenticated public fetchers.

## End-to-end is now complete
With this, the shared-dashboard feature works front to back: create/manage shares, public + unlisted visibility, federation-ready absolute URLs, and the privacy-preserving server-side data resolver from Phase 1.

## Not in this PR (follow-up)
Per-widget editing of a shared dashboard beyond seeding from the home dashboard (would extract the Dashboard grid/editor into a reusable component). Noted in `docs/features/sharing.md`.

## Testing
- `pnpm check` clean in `apps/web`; full web suite (452 tests) passes.
- Manual verification to do after deploy: create a share, mark public, open `/u/fiddur` and `/u/fiddur/<slug>` logged-out; confirm an unlisted share is reachable by link but absent from the profile.


apps/web/src/pages/SharedDashboards/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    onSuccess: invalidate,
+  })
+
+  const handleRename = (e: Event) => {
+    const value = (e.target as HTMLInputElement).value.trim()
+    if (value && value !== dashboard.name) updateMutation.mutate({ name: value })
+  }
+
+  const handleCopy = async () => {
+    await navigator.clipboard.writeText(dashboard.share_url)
```
fiddur: `navigator.clipboard` is only defined in secure contexts. `social.ts` explicitly notes instances may be served over plain http, where `navigator.clipboard` is `undefined` and this line throws a `TypeError` as an unhandled promise rejection — the "Copied!" state never shows and the user gets no feedback. Consider guarding (`if (navigator.clipboard) { ... } else { /* fallback */ }`) or wrapping in try/catch.

fiddur (commented):
✅ Approved

Solid, well-scoped frontend addition. The public pages (`/u/:username`, `/u/:username/:slug`), the owner-facing `/shared-dashboards` management UI, and the `social.ts` API client all line up correctly with the `@aurboda/api-spec` types I verified (`SharedDashboard`, `Create/UpdateSharedDashboardBody`, `Public*Response`, `SectionType`, `defaultDashboardConfig`). A few things I checked that are correct:

- Chrome-hiding via `useLocation().path.startsWith('/u/')` correctly strips Header/Sidebar/Footer for public routes while keeping them for `/shared-dashboards`.
- The `is_public` / rename toggles send only the changed field, and the backend `updateSharedDashboard` only writes defined columns, so partial updates won't clobber `config`.
- User-supplied `name`/`username`/`slug` are rendered as escaped JSX text and `encodeURIComponent`-ed in hrefs; the external View link has `rel="noopener noreferrer"`.
- `PublicWidgetRenderer`, `fetchDashboard`, and the widget exports all exist and match their usages.

Non-blocking nits: (1) the JSX in `index.tsx` (lines 73-121) is inconsistently indented after the rewrap — `pnpm fix`/`oxfmt` would normalize it (CI `check` doesn't run the formatter, so it won't fail, but worth tidying); (2) neither `createMutation` nor the row mutations surface errors to the user, so a failed create/rename/delete is silent — a minor UX gap. See the inline note on the clipboard call.

fiddur (commented):
✅ Approved

Clean, well-structured frontend-only PR adding the public sharing pages (`/u/:username`, `/u/:username/:slug`), the `/shared-dashboards` management UI, and the `social` API client. I verified the key contracts:

- **API parity**: all six client calls in `social.ts` map exactly to existing backend routes (`/shared-dashboards` CRUD in `shared-dashboards-router.ts`, `/public/:username/dashboards` and `/public/:username/:slug` in `public-shares-router.ts`). Public calls correctly omit the `Authorization` header.
- **Layout**: hiding `Header`/`Sidebar`/`Footer` for `/u/` paths works with the flexbox shell (`.app-content` has `flex: 1`, no fixed sidebar offset), so public pages render full-width without misalignment.
- **Types**: the `!query.data.config` guard correctly narrows the destructured `config` in `PublicDashboard` (confirmed compiles under `--strict`). Importing the runtime constant `defaultDashboardConfig` from api-spec follows existing precedent (`metricUnits`, `validMetrics`, etc.) — the "types only" rule is about Zod validation schemas, not shared constants.
- **Safety**: `encodeURIComponent` is applied on username/slug in URLs, the clipboard write is guarded for non-secure (http) contexts with a `window.prompt` fallback, and the external `View` link uses `rel="noopener noreferrer"`. `QuickLinkView` renders a non-navigational static tile (the `quick-link-static` class is real).

Minor, non-blocking notes (no change required):
- On a failed `updateMutation` (rename on blur / toggle Public), the user sees an `alert` but the input/checkbox visual state isn't explicitly reset to the server value until the next refetch — acceptable since there's no optimistic cache write, just slightly janky on error.
- No unit tests are added for the new pages or `social.ts`, but this is consistent with the web app's convention of testing only pure-logic helpers (e.g. the `Timeline/*.test.ts` and `periodSummary.test.ts` files), not presentational page components or thin axios wrappers.

Docs (`docs/features/sharing.md`) are updated to describe the new web UI. Looks good to merge.

PR #816, opened 2026-06-30
fix(sync): trigger deduction rules on Garmin/Oura/RescueTime auto-sync

fiddur:
## Problem

Follow-up to #790. That PR fixed the Last.fm path; the same trigger gap remained for the other auto-synced providers.

Deduction rules run only when something fires the `onActivitySynced` notifier. The background auto-sync paths in `sync-provider.ts` — `syncGarminIfNeeded`, `syncOuraIfNeeded`, `syncRescueTimeIfNeeded` — called their sync functions directly and never fired it. So data ingested through the lazy auto-sync (which runs before tag/daily-summary/productivity queries) never ran through deduction rules:

- **Garmin / Oura** → `activity`-condition rules (sleep, exercise, meditation, etc.)
- **RescueTime** → `screentime_category` / `activity_data` rules (e.g. the "Saraswati TV" rule on `computer_active`)

Only the REST `/sync/*` routes triggered evaluation. `docs/features/deduction-rules.md` already states *"Rule evaluation is debounced per-user when triggered by data syncs"* — this PR makes that true for the auto-sync paths.

## Fix

Add a shared `triggerDeductionAfterSync` helper in the sync provider and wire it into all three methods. It fires `onActivitySynced(user, '*', windowStart, now)` only when the sync **succeeded and ingested new records**, with the window starting at the prior `last_sync_time` (what the sync actually fetched) or a 30-day fallback on first sync — mirroring the Last.fm fix and the REST routes' lookback.

- All three providers return a uniform `SyncResult { records_processed, status }`, so the gate (`status === 'success' && records_processed > 0`) is shared.
- Unit tests for each path: fires on success-with-records, no-fire on zero/error, and skip cases (disabled Garmin type, missing RescueTime key).

## Scope / notes

- Calendars (`syncCalendarsIfNeeded`) intentionally left out — not in the requested set and its rule relevance is marginal; can follow up if wanted.
- No docs change: the deduction-rules doc already documents sync-triggered evaluation; this aligns the auto-sync paths with it.
- Merged latest `origin/develop` (72 commits), reinstalled deps, rebuilt api-spec.

## Test plan

- [x] `pnpm --filter aurboda-backend check` (tsgo + oxlint) — 0 errors
- [x] `sync-provider.test.ts` — 12 tests pass (4 Last.fm + 4 Garmin + 2 Oura + 2 RescueTime)
- [ ] CI green


apps/backend/src/services/sync-provider.ts
```diff
@@ -64,6 +65,23 @@ export interface SyncProviderConfig {
...
+  // last_sync_time (what this sync fetched) or, on a first sync, falls back to a
+  // 30-day lookback matching the REST /sync routes.
+  const triggerDeductionAfterSync = (
+    user: string,
+    priorSyncState: SyncState | null,
+    result: { records_processed: number; status: string },
+  ): void => {
+    if (result.status !== 'success' || result.records_processed === 0) return
+    const end = new Date()
+    const start = priorSyncState?.last_sync_time ?? subDays(end, DEFAULT_SYNC_HISTORY_DAYS)
```
fiddur: Minor / non-blocking: on a **first** sync (no prior `syncState`) the fallback window is 30 days, but Garmin and Oura actually ingest 90 days of history on first sync (`DEFAULT_SYNC_HISTORY_DAYS = 90` in their respective `sync.ts` modules; the value imported here is the Last.fm one, which is 30). So on the very first sync, deduction rules would only be evaluated over the most recent 30 of the 90 freshly-imported days. This matches the existing REST `/sync` routes (`DEFAULT_SYNC_LOOKBACK_MS = 30 days`), so it's consistent rather than a regression — but the comment's "30-day lookback matching the REST /sync routes" glosses over the fact that the integrations themselves backfill 90 days. Worth a follow-up if first-sync rule coverage matters; fine to ship as-is otherwise.

fiddur (commented):
✅ Approved

Clean, focused fix. Extracting `triggerDeductionAfterSync` and wiring it into the Garmin/Oura/RescueTime auto-sync paths closes the same trigger gap the Last.fm path already handled, and it correctly fires only on `status === 'success'` with `records_processed > 0`. Nice touches: it reuses the `syncState` already fetched for the threshold check (no extra DB round-trip), uses the prior `last_sync_time` as the window start, and is wired in production via `onActivitySynced: activityNotifier` in `api.ts`. The three `SyncResult` shapes (Garmin/Oura/RescueTime) are all structurally compatible with the `{ records_processed; status }` parameter. Test coverage is solid (success / no-records / errored / disabled / no-key), and I verified locally: all 12 tests pass, `tsgo` typecheck is clean, and no new lint warnings in the changed files.

One non-blocking note on the first-sync fallback window (inline). Not a regression — it mirrors the existing REST `/sync` behavior — so approving as-is.

apps/backend/src/services/sync-provider.ts
fiddur: Good catch — fixed in 8c0b3eac. The shared helper now falls back to `FIRST_SYNC_FALLBACK_DAYS = 90` (the longest provider backfill; over-covering RescueTime/Last.fm is a harmless idempotent no-op) instead of the Last.fm 30, so first-sync deduction covers all 90 freshly-imported days. The window comment is corrected and a first-sync test was added.

fiddur (commented):
✅ Approved

Solid, focused change that brings Garmin/Oura/RescueTime auto-sync to parity with the existing Last.fm path by firing `onActivitySynced` over the freshly-synced window. The logic is correct:

- `triggerDeductionAfterSync` only fires on `status === 'success'` with `records_processed > 0`, matching the Last.fm guard.
- It captures the *prior* `syncState` (fetched before the sync ran), so `last_sync_time` is the correct lower bound of what an incremental sync fetched, and `?? subDays(end, FIRST_SYNC_FALLBACK_DAYS)` handles the first-sync case.
- I verified `FIRST_SYNC_FALLBACK_DAYS = 90` against the actual backfills: Garmin `DEFAULT_SYNC_HISTORY_DAYS = 90`, Oura `= 90`, RescueTime `= 30` — so the comment is accurate and 90 covers the widest backfill.
- All three provider `SyncResult` types are structurally compatible with the helper's `{ records_processed; status }` param, and each function always returns a result object, so the guard can't throw.

Tests are well-structured and cover the meaningful branches (new records → fires, zero records → skips, error → skips, first-sync 90-day window, disabled data type, missing RescueTime key). I couldn't execute the suite — this worktree has no `node_modules` installed and the sandbox is offline/read-only — but the assertions and mocks read correctly against the implementation.

Minor, non-blocking observation (not introduced by this PR — inherited from the Last.fm pattern): incremental Garmin/Oura syncs fetch from `last_sync_time - INCREMENTAL_SYNC_OVERLAP_DAYS` (2 days) to catch retroactive edits, but the deduction window starts exactly at `last_sync_time`. Retroactively-edited records re-ingested within that 2-day overlap won't be re-evaluated. Since evaluation is idempotent and that range was already covered on the prior sync, this is harmless in practice and consistent with the existing design — flagging only for awareness.

PR #818, opened 2026-06-30
fix(sharing): blank board empty + dark-mode colors

fiddur:
Polish fixes for the just-shipped shared-dashboards feature.

## 1. "Create blank" produced a populated board
`Create blank` seeded from `defaultDashboardConfig` (the populated default layout). Now it uses an empty config `{ version: 1, sections: [] }`. `Create from my dashboard` still copies the home layout.

## 2. Dark-mode colors broken on the new pages
The new pages hardcoded light colors and the board-name input had no explicit text color, so in dark mode the title was invisible (white-on-white focused, dark-on-dark blurred) and rows/links/inputs didn't adapt. Added `@media (prefers-color-scheme: dark)` blocks + explicit input colors for SharedDashboards, PublicProfile, and PublicDashboard. (PublicDashboard widget cards already inherit Dashboard's dark styles.)

## Testing
- `pnpm check` clean in `apps/web`.
- Manual (dark mode): board name readable focused + blurred on /shared-dashboards; /u/<user> profile + /u/<user>/<slug> render with correct contrast.

## Follow-up (separate PR)
Editing a shared board's widgets (and adding a chart to a shared board from the chart page) — see notes; a blank board is only useful once editing lands.

### DIFF
--- a/apps/web/src/pages/PublicDashboard/style.css
+++ b/apps/web/src/pages/PublicDashboard/style.css
@@ -27,3 +27,14 @@
 .public-dashboard .quick-link-static {
   cursor: default;
 }
+
+@media (prefers-color-scheme: dark) {
+  .public-dashboard .public-muted,
+  .public-dashboard .public-loading {
+    color: #9ca3af;
+  }
+
+  .public-dashboard .public-attribution {
+    color: #9ca3af;
+  }
+}
--- a/apps/web/src/pages/PublicProfile/style.css
+++ b/apps/web/src/pages/PublicProfile/style.css
@@ -39,3 +39,25 @@
   border-color: #8b5cf6;
   background: #faf5ff;
 }
+
+@media (prefers-color-scheme: dark) {
+  .public-profile h1 {
+    color: #f3f4f6;
+  }
+
+  .public-profile .public-muted,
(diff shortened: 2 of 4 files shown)


fiddur (commented):
✅ Approved

Small, focused fix that does two things cleanly:

1. **"Create blank" produces an empty dashboard.** Replacing `defaultDashboardConfig` with `{ sections: [], version: 1 }` is correct — it satisfies `dashboardConfigSchema` (`version: z.literal(1)`, `sections` array), the runtime import of `defaultDashboardConfig` was properly dropped in favor of a type-only `DashboardConfig` import (matching the frontend's type-only convention), and `PublicDashboard` renders an empty `sections-grid` without error.

2. **Dark-mode CSS** for the public dashboard, public profile, and shared-dashboards pages. All selectors map to real class names in the corresponding components, the color values mirror their light-mode counterparts, and `prefers-color-scheme: dark` is consistent with the app-wide dark-mode approach (no manual theme toggle exists).

No blocking issues. The minor detail that `emptyConfig` is allocated even on the seed-from-home path is negligible and not worth changing. No tests are included, which is reasonable for a CSS/trivial-logic change.

PR #819, opened 2026-06-30
feat(sharing): edit shared boards in place + target them from charts

fiddur:
## Summary
Addresses two gaps reported while using the shared-dashboards feature: **you couldn't edit a created board's widgets**, and from a chart you could only add to your home dashboard.

- **Edit in place on `/u/<you>/<slug>`** — when you're logged in and viewing your *own* shared board (token present + profile username === your username), an **Edit** toggle appears with the same add/remove/move-widget + section controls as the home dashboard; changes save via `updateSharedDashboard`. Owners see **live** widget data here (their cookie is available); non-owners keep the read-only public render.
- **Chart "Add to dashboard" → dashboard selector** — pick the target dashboard (home or any shared board), then the section.
- **`EditableDashboard` extraction** — the section/widget grid + editing controls are now one reusable component used by both the home Dashboard and shared boards (no duplication). Home Dashboard behaviour is unchanged.

This makes the "Create blank" board (#818) actually useful — you can now populate it.

## Notes
- Owner detection is client-side for UX only; the backend already enforces auth on `updateSharedDashboard`, and the public read endpoint stays unchanged.
- Stacked conceptually on #818 (blank + dark-mode fixes); no file overlap. Will merge `develop` in before finalizing.

## Testing
- `pnpm check` clean in `apps/web`; full suite (452 tests) passes.
- Manual: on `/u/<you>/<slug>` toggle Edit, add/move/remove widgets + sections, confirm they persist and show on the public view; from a chart, add to a shared board via the new selector; confirm a logged-out viewer still sees read-only.

### DIFF
--- a/apps/web/src/components/EditableDashboard/index.tsx
+++ b/apps/web/src/components/EditableDashboard/index.tsx
@@ -1,0 +1,296 @@
+/**
+ * EditableDashboard - the section/widget grid with optional inline editing,
+ * shared by the home Dashboard and the owner's view of a shared dashboard.
+ *
+ * Controlled component: it renders `config` and calls `onChange(next)` for every
+ * edit (add/remove/move widget, add/delete section). The parent owns
+ * persistence (home dashboard vs. a shared dashboard) and the edit toggle.
+ */
+import type { DashboardConfig, DashboardSection, DashboardWidget, SectionType } from '@aurboda/api-spec'
+
+import { useState } from 'preact/hooks'
+
+import { DashboardEditor } from '../DashboardEditor'
+import { WidgetRenderer } from '../widgets'
+
+const generateSectionId = () => `section-${Date.now()}-${Math.random().toString(36).slice(2, 9)}`
+
+function DashboardSectionComponent({
+  section,
+  isEditing,
+  onRemoveWidget,
+  onMoveWidget,
+  onAddWidgetClick,
+  onDeleteSection,
+}: {
+  section: DashboardSection
+  isEditing: boolean
+  onRemoveWidget?: (widgetId: string) => void
+  onMoveWidget?: (widgetId: string, direction: 'up' | 'down') => void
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped change. The inline section/widget/add-section UI is extracted from `Dashboard/index.tsx` into a new controlled `EditableDashboard` component and reused by both the home dashboard and a new owner-editable shared-dashboard view, plus the chart "Add to dashboard" dialog gains a target selector (home vs. a shared dashboard). I traced the refactor against the original and it's faithful — handlers, grid classes, collapse default, and add/move/remove/delete logic are preserved.

Correctness and security look sound:
- `isOwner = Boolean(auth.value.token) && auth.value.user === username` gates the editable view, and that view is only ever populated from the authenticated user's own `listSharedDashboards()` data (server-enforced auth). A logged-in user visiting their own URL with a slug that isn't theirs falls back gracefully to the read-only public view. The exact-match username comparison is consistent with the case-sensitive username handling elsewhere in the codebase.
- Owner edits use `WidgetRenderer` (live data) while non-owners keep the read-only `PublicWidgetRenderer` snapshot, matching the updated docs.
- Types/imports/API signatures all line up (`updateSharedDashboard`, `listSharedDashboards`, `DashboardConfig`).

One non-blocking observation (not introduced by this PR): edits are persisted per-action against the currently-rendered `config` with no optimistic update, so the rendered config only changes after each save's round-trip completes. Two rapid successive edits (a second edit issued before the first save returns) would be computed from the stale pre-save config and the later save could clobber the earlier one. This is the exact same pattern the home dashboard already used on `develop`, now also applied to shared dashboards — worth a future follow-up (optimistic update or debounced single save) but not a regression and low-probability given each edit requires UI interaction. No tests were added, which is consistent with the existing (untested) dashboard components and the project's backend-focused test emphasis.

PR #820, opened 2026-06-30
feat(challenges): federated competitions backend

fiddur:
## Summary

Phase 1 (backend + api-spec) of **Challenges** — federated, cross-instance competitions on a single built-in metric or activity type (cumulative total) over a date span. Builds on the sharing foundation (PRs #813–#819).

### Federation-first, one model
Every member is identified by their base-URL profile and contributes data through a capability **data endpoint**; "same instance" is just the aggregator reading a member's data in-process instead of over HTTP. One join protocol, one data shape.

### Protocol
- `GET /.well-known/aurboda` — discovery (`product`, `version`, `federation`, `api_base`).
- `GET /public/:username/:slug` — **type-dispatched**: a shared dashboard *or* a challenge spec (with `join_token` + public member list). Slugs are unique across both per user.
- `POST /public/:username/:slug/members` — register-back; gated by `join_token`, and the host probes the member's data endpoint before accepting.
- `GET /public/:username/:slug/standings` — host aggregation (local members in-process; remote members pulled with a 5-min TTL cache; failed fetch → last-known flagged `stale`).
- `GET /challenge-data/:username/:token` — a member's own series for one challenge (minimal projection).

Join-by-URL is local-or-federated; the host is always a member.

### Included
- **api-spec**: `challenges.ts`, extended `/status`, `type` discriminator on the public dashboard response, `BaseResponse`.
- **DB**: per-user `challenges` / `challenge_members` / `challenge_participations`; slug uniqueness across dashboards+challenges.
- **Services**: spec→`getChartData` resolver, federation HTTP client + unified `joinChallenge`, standings aggregation.
- **Routers**: authed `/challenges` (CRUD, members, participations, join); unauth public + data + `.well-known`.
- **MCP**: `create/list/update/delete_challenge`, `join_challenge`.
- **Tests**: DB CRUD (7) + standings integration (local + mocked-remote + cache + minimal projection).
- **Docs**: `docs/features/challenges.md`.

### Notes
- **Trust model**: a member's instance is trusted to report honest numbers (Strava-style); capability tokens + unguessable slugs are the gates. Documented.
- **v1 simplifications**: cumulative total; UTC buckets (exact totals over a fixed window); on-demand pull + TTL cache. Timezone-local bucketing, background polling, goals/teams, signed instance-key auth are out of scope.
- **Deployment**: `/.well-known/aurboda` must be routed to the backend at the web base (standard federation; noted in docs). Dev clones hit the backend directly.

### Not in this PR
Phase 2 — frontend: Challenges page (create/join), public challenge view (race chart + leaderboard), join flows, sidebar.

## Testing
- `pnpm check` clean (backend + api-spec); new integration tests pass; OpenAPI generation succeeds with the new schemas.
- Cross-instance manual test (3 local clones) to do during/after review.


apps/backend/src/routes/public-shares-router.ts
```diff
@@ -69,32 +93,121 @@ export const createPublicSharesRouter = (webHost: string): TypedRouter => {
...
+      }
+      try {
+        const challenge = await getChallengeBySlug(username, slug)
+        if (!challenge) return res.status(404).json({ error: 'Challenge not found', success: false })
+        if (req.body.join_token !== challenge.join_token) {
+          return res.status(403).json({ error: 'Invalid join token', success: false })
+        }
+        // Probe the member's data endpoint before accepting them.
+        try {
+          await fetchMemberData(req.body.data_endpoint_url)
```
fiddur, line 160: SSRF: this fetches an arbitrary, attacker-supplied URL from an *unauthenticated* endpoint (the `join_token` is public to anyone who fetched the spec), and `getChallengeStandings` re-fetches it on every `?refresh=1` standings call. `fetchMemberData` uses a plain `axios.get` with no allowlist / private-IP filtering / redirect cap, so a member can be registered pointing at `http://169.254.169.254/...`, `localhost`, or internal hosts. This is outside the documented "trust honest numbers" model. Suggest rejecting private/loopback/link-local targets (and limiting redirects) on all outbound federation fetches.

apps/backend/src/routes/public-shares-router.ts
```diff
@@ -69,32 +93,121 @@ export const createPublicSharesRouter = (webHost: string): TypedRouter => {
...
+        } catch {
+          return res.status(400).json({ error: 'Member data endpoint unreachable', success: false })
+        }
+        await upsertChallengeMember(username, challenge.id, {
+          data_endpoint_url: req.body.data_endpoint_url,
+          display_name: req.body.display_name,
+          identity_base_url: req.body.identity_base_url,
```
fiddur, line 167: `identity_base_url` is taken from the request with no binding to the registering instance, and `upsertChallengeMember` keys on `(challenge_id, identity_base_url)`. A caller who has the slug+`join_token` can pass the host's own profile URL (or another member's) and the `ON CONFLICT DO UPDATE` will overwrite that row — flipping `kind` to `remote`, nulling `local_user`, and setting an attacker-controlled `data_endpoint_url`, hijacking that member's standing. At minimum, refuse register-back that would overwrite an existing `kind='local'` member.

apps/backend/src/db/challenges.ts
```diff
@@ -0,0 +1,468 @@
...
+): Promise<ChallengeParticipationRecord | null> => {
+  const result = await query<ParticipationRow>(
+    user,
+    `SELECT ${PARTICIPATION_COLUMNS} FROM challenge_participations WHERE data_token = $1`,
+    [token],
+  )
+  return result.rows.length ? mapParticipation(result.rows[0]) : null
+}
+
+export const getParticipationByUrl = async (
```
fiddur, line 466: `getParticipationByUrl` is exported but never called anywhere. The join flow (`joinChallenge`) always calls `createChallengeParticipation`, so joining the same challenge URL twice creates duplicate participations, each with its own live `data_token`. Either use this helper to make join idempotent (return/refresh the existing participation) or drop the dead function. It's also missing an integration test, as are `getChallengeById` and `deleteChallengeParticipation` (CLAUDE.md requires one per db function).

fiddur (commented):
🛑 Changes required

Solid, well-structured first cut at federated challenges: schemas in api-spec, shared service logic reused by REST + MCP, per-DB storage, and good integration tests. The data model and CRUD plumbing look clean and match the docs. A few things should be addressed before merge, mostly around the unauthenticated federation surface.

**Security (federation register-back):** `POST /public/:username/:slug/members` is unauthenticated and the `join_token` gate is weak — that token is returned to *anyone* who fetches `GET /public/:username/:slug`, so it proves "I read the spec", not identity. Two concrete problems flow from the untrusted body:

1. **SSRF.** The host probes `req.body.data_endpoint_url` (and re-fetches it on every standings call) with a plain `axios.get` — no allowlist, private/loopback/link-local filtering, or redirect limits. An attacker who knows a slug can register a member pointing at `http://169.254.169.254/...`, `http://localhost:...`, or internal hosts, then trigger fetches at will via the unauthenticated `GET /public/:username/:slug/standings?refresh=1`. The documented v1 trust model ("trust a member's instance to report honest numbers") covers fabricated *values*, not server-side requests to internal infrastructure. Recommend blocking private/loopback/link-local IP ranges (and capping redirects) on all outbound federation fetches.

2. **Member identity is unauthenticated and overwritable.** `upsertChallengeMember` keys on `(challenge_id, identity_base_url)` and the caller supplies `identity_base_url` with no binding to who's actually registering. A caller can pass the host's own profile URL (or another member's) and the `ON CONFLICT DO UPDATE` will flip that existing row to `kind='remote'`, `local_user=NULL`, and an attacker-controlled `data_endpoint_url` — hijacking the host's (or another member's) standing. At minimum, reject register-back that targets an existing `kind='local'` member.

Also note `?refresh=1` on the public standings endpoint lets an unauthenticated caller force N parallel 8s outbound fetches + DB writes per request — an amplification/DoS lever worth gating.

**Correctness — join is not idempotent.** `getParticipationByUrl` is defined and exported but never used; `joinChallenge` always calls `createChallengeParticipation`, so joining the same challenge twice creates duplicate participations (each with its own live `data_token`), and the local-join member upsert is idempotent while the participation side is not. Either wire up the dedup via `getParticipationByUrl` or remove the dead function.

**Tests / standards.** CLAUDE.md requires every db function to have an integration test; `getChallengeById`, `deleteChallengeParticipation`, and `getParticipationByUrl` are currently uncovered (the latter also being dead code).

Minor: for `source_type: 'activity_type'`, `getChartData` only consults `pattern`, never `activity_type_id`, so a spec carrying only `activity_type_id` yields an empty series — worth confirming the create path always sets `pattern`.

apps/backend/src/routes/public-shares-router.ts
```diff
@@ -69,32 +94,132 @@ export const createPublicSharesRouter = (webHost: string): TypedRouter => {
...
+          return res.status(403).json({ error: 'Invalid join token', success: false })
+        }
+        // Don't let an unauthenticated caller overwrite a local member (the host
+        // or any same-instance member) by claiming their identity URL.
+        const existing = await getChallengeMemberByIdentity(
+          username,
+          challenge.id,
+          req.body.identity_base_url,
+        )
+        if (existing && existing.kind === 'local') {
```
fiddur, line 148: Worth noting the actual trust boundary here: `join_token` is returned by `GET /public/:username/:slug`, so for a public challenge it's readable by anyone who can resolve the slug. This guard only blocks impersonating a `local` member — an existing *remote* member can be re-registered by any slug-holder, overwriting its `data_endpoint_url`/`display_name` (intended for legitimate re-joins, but it also lets an attacker point a victim member's standing at attacker-controlled data). That's within the documented v1 trust model, so non-blocking, but since `upsertChallengeMember` keys on `(challenge_id, identity_base_url)` an unauthenticated caller can also create unbounded distinct members — consider a per-challenge member cap to bound growth.

apps/backend/src/routes/public-shares-router.ts
```diff
@@ -69,32 +94,132 @@ export const createPublicSharesRouter = (webHost: string): TypedRouter => {
...
-          return res.status(404).json({ error: 'Dashboard not found', success: false })
-        }
-        const widgetData = await resolveDashboardData(username, record.config)
-        // Public/anonymous content — allow shared (CDN) caches to absorb load.
+        const challenge = await getChallengeBySlug(username, slug)
+        if (!challenge) return res.status(404).json({ error: 'Challenge not found', success: false })
+        // No `refresh` here: this endpoint is unauthenticated, so honoring it would
+        // let anyone force N parallel outbound fetches. The TTL cache still keeps
+        // standings reasonably fresh; the owner can force-refresh via the authed route.
+        const members = await getChallengeStandings(username, challenge)
```
fiddur, line 114: This unauthenticated endpoint can trigger host-side outbound fetches (and member-cache writes) when the TTL cache is cold/expired. The TTL + 8s timeout keep it bounded, and you correctly drop `refresh` here, but there's no in-flight de-duplication: concurrent requests on an expired cache each fan out to every remote member (thundering herd). Combined with the fact that members are registrable without authentication, this is a mild amplification surface — fine for v1, just flagging for the future-hardening list.

fiddur (commented):
✅ Approved

Solid, well-structured first cut at federated challenges. The security-sensitive surface is handled thoughtfully and is backed by tests: the SSRF guard resolves DNS and checks *all* resolved addresses before fetching (so decimal/hex IP encodings get normalized by the OS resolver and then blocked), blocks IPv4-mapped IPv6 and the usual private/loopback/link-local/CGNAT ranges, and disables redirect-following — with a clear unit test. The DB layer has full integration coverage, the local-member hijack guard exists, and the public endpoints never leak `data_endpoint_url`. Standards are respected (typedRouter with api-spec response types, snake_case, schemas centralized in `@aurboda/api-spec`, Date→ISO serialization at the boundary, functional style, normalized member references by identity URL).

No blocking issues. A few non-blocking observations:

1. **Trust boundary of register-back is the (publicly readable) `join_token`.** `GET /public/:username/:slug` returns `join_token`, so for a *public* challenge anyone who can resolve the slug can register remote members and overwrite an existing remote member's `data_endpoint_url`/`display_name` (only `kind === 'local'` is protected). This is consistent with the documented v1 "trust the reporter" model and the intended re-join flow, but it's worth being explicit that a remote member's reported numbers can be hijacked by any slug-holder, and that `challenge_members` can grow without bound (consider a per-challenge member cap). Inline note added.

2. **Unauthenticated standings fan-out.** The public `/standings` route triggers host-side outbound fetches + cache writes on a cold/expired cache. It's bounded by the 5-min TTL, the 8s timeout, and the fact that registration requires a valid data response, but there's no in-flight de-duplication, so concurrent requests on an expired cache each fan out to every remote member. Acceptable for v1; flagged inline.

3. **Test gap (non-blocking).** The DB layer, standings aggregation, and SSRF guard are tested, but the route-level federation handshake (`/join`, register-back, the local-vs-remote hijack guard, token-gated `/challenge-data`) and the pure `parseChallengeUrl` helper have no direct tests. Given the security sensitivity, supertest coverage of the register-back guard and join flow would be valuable follow-up.

Minor nit: `pattern: spec.spec.pattern ?? null` (challenge-federation.ts) and `pattern: spec.pattern ?? undefined` (challenge-spec.ts) apply a nullish fallback to a field that is always a required non-empty string — harmless but slightly misleading.

apps/backend/src/services/challenge-standings.ts
```diff
@@ -0,0 +1,98 @@
...
+          challenge.spec,
+          challenge.start_ts,
+          challenge.end_ts,
+        )
+        return { ...base, buckets, last_updated: new Date().toISOString(), stale: false, total }
+      }
+
+      // Remote member: use the TTL cache unless stale or refresh requested.
+      const fresh = member.last_fetched_at !== null && now - member.last_fetched_at.getTime() < CACHE_TTL_MS
+      if (!options.refresh && fresh && member.cached_buckets) {
```
fiddur: This cache short-circuit also requires `member.cached_buckets` to be truthy, which defeats the TTL for members that have never returned data. A failed fetch persists `cached_buckets: null` but sets `last_fetched_at = NOW()`, so such a member is `fresh` yet skips this branch and gets re-fetched on *every* call — including from the unauthenticated `GET /public/:username/:slug/standings`. That lets a registered member with a hanging-but-public endpoint force a fresh fan-out (up to 200 × 8s) on each public request, contradicting the "bounded by the TTL" guarantee in the docs.

Suggest dropping the `&& member.cached_buckets` guard and returning `buckets: member.cached_buckets ?? []` here, so failures are cached for the TTL too:
```ts
if (!options.refresh && fresh) {
  return {
    ...base,
    buckets: member.cached_buckets ?? [],
    last_updated: member.last_fetched_at?.toISOString() ?? null,
    stale: member.last_error !== null,
    total: member.cached_total ?? 0,
  }
}
```

fiddur (commented):
🛑 Changes required

Solid, well-structured federation backend: the SSRF guard (verified — `new URL` canonicalizes IPv6 hosts so the v4-mapped/expanded-loopback forms are correctly blocked), the unguessable-slug + `isValidUsername` per-DB gating, join idempotency, and the register-back rollback are all good, and the integration coverage is thorough.

One correctness issue to fix before merge, on the unauthenticated standings path:

**Remote-member TTL cache is bypassed for never-succeeded members.** In `getChallengeStandings`, the cache short-circuit requires `member.cached_buckets` to be truthy. A failed fetch calls `updateChallengeMemberCache` with `buckets: member.cached_buckets` (still `null`) but bumps `last_fetched_at = NOW()`. So a remote member that has never returned data (or whose endpoint goes down before the first successful standings computation — the registration probe result is not persisted to `cached_buckets`) is `fresh` but has `cached_buckets === null`, so it falls through and is re-fetched on *every* standings call, ignoring the 5-minute TTL. This is reachable from `GET /public/:username/:slug/standings` (unauthenticated). Since slug + `join_token` are both returned by the public resolver, an attacker can register a member whose endpoint passes the one-time probe then hangs, then repeatedly hit the public standings to force up to `MAX_CHALLENGE_MEMBERS` (200) concurrent 8s outbound fetches per request — directly contradicting docs/features/challenges.md's claim that the fan-out is "bounded by the TTL". Fix by short-circuiting on `fresh` regardless of `cached_buckets` (returning `buckets: member.cached_buckets ?? []`), so failures are cached for the TTL too.

Non-blocking hardening notes:
- `safe-fetch` sets `timeout` and `maxRedirects: 0` but no `maxContentLength`/`maxBodyLength`. A member's (unauthenticated-registerable) data endpoint is a public URL the host fetches and buffers; a malicious endpoint could return a multi-GB body to exhaust host memory. Consider capping response size.
- `joinChallenge`'s local-shortcut path uses `parsed.username` (from the user-supplied URL) directly as the DB user without `isValidUsername`; it's safe in practice (a bad name just fails to connect / `SET ROLE %L` is escaped), but validating it would match the convention applied on every other username-to-DB boundary.

apps/backend/src/services/safe-fetch.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const isBlockedIPv6 = (raw: string): boolean => {
+  const ip = raw.toLowerCase()
+  // Block all IPv4-mapped addresses (::ffff:… in dotted or compressed-hex form).
+  // Real hosts resolve to plain v4/v6; a literal mapped address is an edge/attack.
+  if (ip.startsWith('::ffff:')) return true
+  if (ip === '::1' || ip === '::') return true // loopback / unspecified
+  if (ip.startsWith('fe8') || ip.startsWith('fe9') || ip.startsWith('fea') || ip.startsWith('feb'))
+    {return true} // fe80::/10 link-local
+  if (ip.startsWith('fc') || ip.startsWith('fd')) return true // fc00::/7 unique-local
```
fiddur, line 44: The v6 blocklist misses IPv4-compatible addresses (`::7f00:1` == `::127.0.0.1`) and NAT64 (`64:ff9b::/96`). Since `assertPublicUrl` skips DNS for literal IPs, `http://[::7f00:1]/` passes the guard. Low practical risk (deprecated/rarely-routed), but cheap to add alongside the `::ffff:` check — e.g. treat any `::`-prefixed address embedding a blocked v4 tail, or reject `64:ff9b:`.

fiddur (commented):
✅ Approved

Solid, well-scoped first cut at cross-instance challenges. The architecture is clean and matches the project conventions well: shared business logic in `services/`, api-spec Zod schemas as the single source of truth, `typedRouter()` with fully typed generics, MCP↔REST parity, snake_case throughout, functional style, and both integration (testcontainers) and unit tests for the DB layer, standings aggregation, and the SSRF guard. The security posture is deliberately reasoned and documented (unguessable slug + capability tokens as gates, `join_token` probe + endpoint probe on register-back, local-member hijack guard, `MAX_CHALLENGE_MEMBERS` cap, redirect-disabled + size-capped outbound fetches, failed-fetch caching to bound fan-out), and the known v1 limitations are honestly enumerated in `docs/features/challenges.md`.

No blocking correctness or security bugs found. A few non-blocking observations:

1. **SSRF allowlist — IPv6 gaps (defense-in-depth).** `isBlockedIPv6` blocks IPv4-mapped (`::ffff:*`), loopback/unspecified, link-local, and unique-local, but not IPv4-*compatible* addresses (`::7f00:1` / `::127.0.0.1`) or NAT64 (`64:ff9b::/96`). Practical exploitability is low (these are deprecated and generally not routed to IPv4 on modern kernels), and the module already scopes DNS-rebinding out of v1, but the literal-IP path (`http://[::7f00:1]/`) skips DNS entirely, so it's worth closing. Inline note below.

2. **Unauthenticated public standings recompute.** `GET /public/:username/:slug/standings` recomputes every *local* member's series (a `getChartData` SQL aggregation each) on every request — only remote members get the TTL cache. For single-instance challenges (the common case, all-local members) an unauthenticated caller can amplify 1 request → N SQL aggregations. This mirrors the existing public shared-dashboard resolver (also uncached, relying on the `max-age=60` header), so it's consistent precedent rather than a regression, but a per-challenge in-flight/short-TTL cache would help; the docs already flag the remote thundering-herd variant.

3. **Minor:** the `MAX_CHALLENGE_MEMBERS` check (list-then-insert in `public-shares-router.ts`) and join idempotency (`getParticipationByUrl` with no unique constraint on `challenge_url`) both have small TOCTOU races under concurrent requests — harmless given the caps and trust model, just noting.

apps/backend/src/services/challenge-federation.ts
```diff
@@ -0,0 +1,241 @@
...
+  apiBase: string,
+  username: string,
+  slug: string,
+  body: { identity_base_url: string; display_name: string; data_endpoint_url: string; join_token: string },
+): Promise<void> => {
+  const res = await safeFetchPost<{ success?: boolean; error?: string }>(
+    joinUrl(apiBase, `public/${encodeURIComponent(username)}/${encodeURIComponent(slug)}/members`),
+    body,
+  )
+  if (res.data?.success === false) {
```
fiddur: This `success === false` branch is effectively dead: the host's register-back endpoint signals rejection with non-2xx statuses (403 `Invalid join token`, 409 `Challenge is full` / `already a local member`, 400 `Member data endpoint unreachable`), and axios's default `validateStatus` throws on those *before* this check runs. So a joiner never sees the host's `error` message — they get a generic `Request failed with status code 4xx` instead. If you want the host's message to surface, set `validateStatus` to accept 4xx here and read `res.data.error`, or map the AxiosError's `response.data.error` in the caller.

fiddur (commented):
✅ Approved

Solid, well-structured first cut of federated challenges. The SSRF guard (`safe-fetch.ts`) is thoughtful — it checks *all* resolved addresses, covers the IPv4/IPv6 private/reserved/link-local/NAT64/IPv4-mapped ranges (incl. the cloud-metadata address), disables redirect following, and caps response size; the DNS-rebinding TOCTOU gap is explicitly acknowledged as acceptable for v1. Data-flow across the trust boundary is careful: capability tokens are unguessable, member data-endpoint URLs stay host-only, remote-supplied `api_base`/`data_endpoint_url` are re-validated on every fetch, the register-back path is bounded (`MAX_CHALLENGE_MEMBERS`, local-member hijack guard, join-token gate), and failed remote fetches are cached for the TTL to bound fan-out. api-spec wiring follows the documented `typedRouter`/`.shape` conventions, DB access is per-user, and both REST + MCP parity and integration tests (CRUD, members, participations, standings, SSRF) are present.

No blocking issues. A few non-blocking observations:

- **Local members are recomputed uncached on every standings call.** Remote members are TTL-cached, but each local member runs a fresh `getChartData` aggregation on every request to the *unauthenticated* `/public/:username/:slug/standings`. This mirrors the existing public-dashboard pattern (which also relies on the `max-age=60` CDN cache), so it's consistent — just worth keeping in mind for scaling / DoS-amplification if a challenge accumulates many local members.
- **`join_token` adds little over the slug.** The public resolver returns `join_token` to anyone who can GET the (slug-gated) challenge, and the register-back endpoint only checks that same token — so it doesn't gate anything the unguessable slug doesn't already gate. Consistent with the stated trust model, but the token is effectively belt-and-suspenders.
- **Participation join idempotency is best-effort.** `getParticipationByUrl` + create has no unique constraint on `challenge_url`, so concurrent joins of the same URL can create duplicate participations (each with its own data token). Sequential re-joins are fine.

See the inline note on the register-back error handling.

apps/backend/src/services/safe-fetch.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+const MAX_RESPONSE_BYTES = 1_000_000 // federation payloads are small; cap to avoid memory exhaustion
+
+const guardedConfig = (config?: AxiosRequestConfig): AxiosRequestConfig => ({
+  maxBodyLength: MAX_RESPONSE_BYTES,
+  maxContentLength: MAX_RESPONSE_BYTES,
+  maxRedirects: 0, // a redirect could bounce to an internal address — reject it
+  timeout: DEFAULT_TIMEOUT_MS,
+  ...config,
```
fiddur, line 80: `...config` is spread *after* the security-critical defaults, so a caller-supplied config can override `maxRedirects: 0` (and the size/timeout caps) — silently re-opening the redirect-to-internal SSRF vector this helper exists to close. No current caller passes a `config`, so it's latent, but since `safeFetchGet/Post` are exported, consider letting the guards win regardless of caller input:
```ts
const guardedConfig = (config?: AxiosRequestConfig): AxiosRequestConfig => ({
  timeout: DEFAULT_TIMEOUT_MS,
  ...config,
  maxBodyLength: MAX_RESPONSE_BYTES,
  maxContentLength: MAX_RESPONSE_BYTES,
  maxRedirects: 0,
})
```

fiddur (commented):
✅ Approved

Solid, well-structured first cut at federated challenges. The layering follows the project conventions closely: schemas in `api-spec` as the single source of truth, shared business logic in `services/`, `typedRouter()` with proper generic params and ISO-string serialization at the boundary, and MCP tools that mirror the REST surface. DB access is fully parameterized, the SSRF guard is thoughtful (literal-IP + DNS resolution, IPv4/IPv6/NAT64/IPv4-mapped coverage, `maxRedirects: 0`, size caps), and the register-back path defends the important cases (join-token gate, local-member hijack guard, member cap, data-endpoint probe). Integration tests cover every exported DB function plus the standings aggregation incl. the failed-fetch cache behavior, and the SSRF guard has focused unit tests. Not committing regenerated Kotlin/OpenAPI artifacts is fine — CI regenerates them.

A few non-blocking things to keep in mind rather than fix here:

- **Spec staleness for remote members.** A challenge's `spec` is copied into each joiner's `challenge_participations` row at join time, and the member data endpoint computes against that stored copy. Local members' standings, however, are computed against the host's *current* `challenge.spec` (`resolveMemberSeries(member.local_user, challenge.spec, ...)`). So after a host edits a challenge's spec/date-range (via `updateChallenge`), local and remote members are measured against different specs until — never, since there's no re-sync. Worth documenting as a v1 limitation or disallowing spec edits once members exist.
- **`join_token` is returned in the public `/public/:username/:slug` challenge resolver**, so it adds no secrecy beyond the (unguessable) slug — the register-back gate is effectively slug-gated. That matches the documented trust model, just flagging it's not an independent secret.
- **Unauthenticated standings amplification.** `/public/:username/:slug/standings` can trigger up to `MAX_CHALLENGE_MEMBERS` (200) parallel outbound fetches on a cache miss. It's bounded by the 5-min TTL cache and the register-time data-endpoint probe, so acceptable, but keep it in mind if the member cap grows.

One inline suggestion on the safe-fetch helper below.

fiddur (commented):
✅ Approved

Solid, well-structured v1 of federated challenges. The layering follows the project conventions closely: schemas live in `@aurboda/api-spec` (snake_case, `.meta({id,...})`), routers use `typedRouter()` with fully-typed generics, business logic sits in `services/`, MCP tools reuse the same schemas via `.shape`, and every new DB function in `db/challenges.ts` has an integration test. Docs (`docs/features/challenges.md`, `sharing.md`) are updated and unusually thorough.

I verified the parts most likely to break:
- **Wiring**: `apiBaseUrl`/`webHost` are defined in `api.ts` and correctly threaded into the MCP router, REST routers, and the well-known config. The `WellKnownConfig` gained two required fields (`apiBaseUrl`, `version`) and the well-known test was updated to supply them, so no break there.
- **SSRF guard** (`safe-fetch.ts`): protocol allowlist, DNS resolution with private/loopback/link-local/reserved blocking across *all* resolved addresses, and `maxRedirects:0` + size caps placed after the config spread so a caller can't relax them (the `dce6a87` fix). I confirmed the WHATWG URL parser compresses IPv6 literals (`[0:0:0:0:0:0:0:1]` → `[::1]`), so the `startsWith('::')` loopback check can't be bypassed with an uncompressed form. The DNS-rebinding (TOCTOU) gap is explicitly acknowledged as out of scope for v1.
- **Standings caching**: a failed remote fetch still bumps `last_fetched_at`, so a hanging member endpoint isn't re-fetched on every (unauthenticated) standings call within the TTL — covered by a dedicated integration test.
- **Bucket math**: `ChallengeDataResponse.buckets` uses the scalar `chartDataBucketSchema` (always has `value`), so the `reduce` total fallback is safe; `total ?? …` correctly preserves an explicit `0`.
- **Auth boundaries**: authed routes act on `req.user`'s DB; every username→DB boundary (public resolver, register-back, challenge-data, join) validates via `isValidUsername`; the register-back path guards against local-member hijack and caps member growth (`MAX_CHALLENGE_MEMBERS`).

No blocking issues found. The residual security/scale trade-offs — unauthenticated standings/register-back triggering bounded outbound fetches (a mild reflected-amplification vector, capped by member count + 8s timeout + 5-min TTL), no in-flight de-duplication (thundering herd), and DNS rebinding — are all explicitly documented as accepted v1 scope with signed instance-to-instance requests noted as the follow-up hardening. Good to merge; the follow-ups are worth tracking as issues.

PR #821, opened 2026-06-30
fix(metrics): truthful add_metrics_bulk for cumulative metrics (#802)

fiddur:
Fixes #802.

## Problem
`add_metrics_bulk` accepted `steps` (and other cumulative/derived metrics) from any source and returned them in `inserted` with `errors: []`, but reads of those metrics are filtered to specific sources (`cumulativeSources` = `health_connect_aggregate`/`aurboda`; `aurbodaOnlySources` for calories). So a write from e.g. `tryout_pr800` was stored in `time_series` yet **unqueryable** via `query_metrics` / `query_metrics_bucketed` / aggregates — and `inserted` over-counted. (Non-cumulative metrics from the same call round-tripped fine.)

## Fix
Reuse the query side's `getSourceFilter` (now exported) on the write path. In `bulkAddMetrics`, if a metric is cumulative/derived and the resolved source isn't one it can be read back from, the item is **rejected with a clear per-item error** and excluded from `inserted` — so the response is truthful (the issue's option 1). Allowed sources (and the default `aurboda`) insert normally; single `add_metric` was already fine (always writes `aurboda`).

Example: a bulk call of `steps` (source `tryout`) + `resting_heart_rate` (source `tryout`) now returns `inserted: 1` with one error explaining the `steps` rejection, instead of `inserted: 2` with silently-unqueryable steps.

## Testing
- New unit tests: rejects cumulative-from-unqueryable-source; accepts cumulative from `aurboda`/`health_connect_aggregate`. Full `mutations.test.ts` (79) passes; `pnpm check` clean.

### DIFF
--- a/apps/backend/src/services/mutations.test.ts
+++ b/apps/backend/src/services/mutations.test.ts
@@ -1,3 +1,5 @@
+import type * as TimeSeriesModule from '../db/time-series.ts'
+
 import { beforeEach, describe, expect, test, vi } from 'vitest'
 
 import * as db from '../db/index.ts'
@@ -18,7 +20,7 @@ import {
 } from './mutations.ts'
 
 // Mock the db module
-vi.mock('../db', () => ({
+vi.mock('../db', async () => ({
   activityTypeExists: vi.fn().mockResolvedValue(true),
   checkActivityConflict: vi.fn().mockResolvedValue(false),
   getActivityTypeDefinition: vi.fn().mockResolvedValue(null),
@@ -34,13 +36,8 @@ vi.mock('../db', () => ({
   getCustomMetricByName: vi.fn(),
   getCustomMetricDefinitions: vi.fn().mockResolvedValue([]),
   getOverrideForActivity: vi.fn().mockResolvedValue(null),
-  getSourceFilter: vi.fn((metric: string) =>
-    ['calories_active', 'calories_total'].includes(metric)
-      ? ['aurboda', 'aurboda_gap_fill']
-      : ['steps', 'distance', 'floors_climbed', 'calories_active', 'calories_total'].includes(metric)
-        ? ['health_connect_aggregate', 'aurboda']
-        : null,
-  ),
+  // Pull the real source-filter rules through so the test can't drift from production.
+  getSourceFilter: (await vi.importActual<typeof TimeSeriesModule>('../db/time-series.ts')).getSourceFilter,
   getUserNotesJoined: vi.fn().mockResolvedValue(undefined),
   getUserSettings: vi.fn(),
   insertCustomMetricDefinition: vi.fn(),
--- a/apps/backend/src/services/mutations.test.ts
+++ b/apps/backend/src/services/mutations.test.ts
@@ -34,6 +34,13 @@ vi.mock('../db', () => ({
   getCustomMetricByName: vi.fn(),
   getCustomMetricDefinitions: vi.fn().mockResolvedValue([]),
   getOverrideForActivity: vi.fn().mockResolvedValue(null),
+  getSourceFilter: vi.fn((metric: string) =>
+    ['calories_active', 'calories_total'].includes(metric)
+      ? ['aurboda', 'aurboda_gap_fill']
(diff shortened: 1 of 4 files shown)


apps/backend/src/services/mutations.test.ts
```diff
@@ -34,6 +34,13 @@ vi.mock('../db', () => ({
   getCustomMetricByName: vi.fn(),
   getCustomMetricDefinitions: vi.fn().mockResolvedValue([]),
   getOverrideForActivity: vi.fn().mockResolvedValue(null),
+  getSourceFilter: vi.fn((metric: string) =>
```
fiddur: Non-blocking: this mock hand-duplicates the real `getSourceFilter` logic and source lists. Since `getSourceFilter` is now an exported pure function with no DB dependency, the mock can silently diverge from production if `cumulativeMetrics`/`cumulativeSources`/`aurbodaOnly*` change later — the test would keep passing against stale expectations. Consider pulling the real implementation through (e.g. `const { getSourceFilter } = await vi.importActual('../db/time-series.ts')`) so the test exercises the actual source rules rather than a copy.

fiddur (commented):
✅ Approved

Solid, well-scoped fix for #802. Rejecting cumulative/derived metric writes from sources that the read filter would exclude is the right call — it turns a silent phantom insert into a surfaced per-item error, and it reuses the very same `getSourceFilter` the read paths use, so the write rejection and read filtering can't drift apart.

Verified:
- `getSourceFilter` preserves the `aurbodaOnlyMetrics`-before-`cumulativeMetrics` precedence, so `calories_total`/`calories_active` correctly require `aurboda`/`aurboda_gap_fill` (not `health_connect_aggregate`), matching `getTimeSeries`/`getTimeSeriesWithSource`.
- Default source `aurboda` is in both filter sets, so source-less imports still succeed; the error guidance ("omit source to use the default") is therefore accurate.
- `bulkAddMetrics` only backs the user-facing MCP/REST bulk path, not the provider auto-sync paths, so legitimate Garmin/Oura/HC sync is unaffected.
- Existing `bulkAddMetrics` tests use only unrestricted metrics, so the new mock doesn't perturb them; the two new tests are correct.

One non-blocking suggestion on the test mock (inline). Note I couldn't run the suite here (node_modules not installed in this worktree), but the tests check out by inspection.

fiddur (commented):
✅ Approved

Clean, well-scoped fix. `bulkAddMetrics` now reuses the read-side `getSourceFilter` to reject cumulative/derived metric writes (steps, distance, calories, …) from sources that no query path would ever return, instead of silently storing a phantom row and reporting it as inserted (#802).

Verification I did:
- **Scope is correct.** Only `bulkAddMetrics` accepts arbitrary per-item sources; the single `addMetric` always writes `source: 'aurboda'` (always allowed), so no guard is needed there. Only the MCP tool (`mcp/metric-tools.ts`) and REST router (`routes/metrics-router.ts`) call `bulkAddMetrics`, so the Garmin/Oura sync write paths that intentionally store query-filtered rows are unaffected — no contradiction with the behavior documented in `docs/features/calories.md`.
- **No regressions in existing tests.** Existing `bulkAddMetrics` tests use non-cumulative metrics (`heart_rate`, `weight`, `mood`) with `oura`/`garmin` sources; `getSourceFilter` returns `null` for those → no restriction. The guard runs after metric/range validation and uses the same per-item error + `continue` pattern, so the batch isn't failed.
- **Test wiring is sound.** Pulling the real `getSourceFilter` via `vi.importActual('../db/time-series.ts')` keeps the test from drifting from production rules; `connection.ts` (transitively imported) has no import-time side effects, so the async mock factory is safe. The re-export in `db/index.ts` comes from the same `./time-series.ts` the test imports.
- **Style.** Functional, no new casts, descriptive/actionable error message.

I was unable to execute the suite (sandbox blocks the pnpm store and a full install is disproportionate for a review), but the logic and test setup check out by inspection.

Non-blocking nit: `docs/features/calories.md` currently describes `getSourceFilter` purely as a query-time filter — it could optionally note it is now also enforced on the bulk-write path.

PR #823, opened 2026-07-01
feat(dashboard): editable section titles + update board charts from chart page (#822)

fiddur:
Closes #822.

## What & why

Issue #822 asked for two dashboard-editing capabilities:

1. **Editable section titles.**
2. When you click a chart on a board and land on the chart explorer, the URL should carry which board chart you came from, and the chart page should offer an **"Update Chart in _&lt;dashboard&gt;_ / _&lt;section&gt;_"** button (with an option to set a new title).

Both are delivered here, for the home dashboard and for shared dashboards the user owns. No backend/api-spec changes — everything persists through the existing `saveDashboard` / `updateSharedDashboard` endpoints.

## Changes

**Section rename (part 1)**
- In edit mode, each section header gains a pencil button that swaps the title for an inline input. Enter or blur commits (via a new `handleRenameSection` → `onChange`); Escape cancels; empty/unchanged titles are ignored.

**Update-a-chart-in-place (part 2)**
- `buildChartUrl` gains an optional `origin` (`board_id` / `section_id` / `widget_id`); `'home'` denotes the private dashboard, otherwise a shared-dashboard id.
- `EditableDashboard` now takes a `boardId` and threads `{ board_id, section_id, widget_id }` down through `WidgetRenderer` into the trend/bar chart widgets, so their link back to `/chart` carries the origin.
- The chart page parses the origin, preserves it in the URL as controls change, and renders `UpdateChartOnBoard`: an "Update Chart in _&lt;board&gt;_ / _&lt;section&gt;_" button that replaces the originating widget in place (preserving its id) with the current config and an optional new title. Board/section names are resolved from the live config (IDs in the URL, names at render time). The control hides itself if the origin no longer resolves to a real widget.

**Docs**
- Updated `docs/features/dashboard.md` and `docs/features/sharing.md`.

## Testing

- New unit tests for the pure helpers: `parseChartOrigin`, `chartWidgetTitle`, `replaceWidgetInConfig` (`updateWidget.test.ts`) and `buildChartUrl` origin params (`chart-url.test.ts`).
- `pnpm check` clean (0 errors; new code adds 0 lint warnings).
- Full `apps/web` suite green (464 tests).

### DIFF
--- a/apps/web/src/components/EditableDashboard/index.tsx
+++ b/apps/web/src/components/EditableDashboard/index.tsx
@@ -18,36 +18,94 @@ const generateSectionId = () => `section-${Date.now()}-${Math.random().toString(
 function DashboardSectionComponent({
   section,
   isEditing,
+  boardId,
   onRemoveWidget,
   onMoveWidget,
   onAddWidgetClick,
   onDeleteSection,
+  onRenameSection,
 }: {
   section: DashboardSection
   isEditing: boolean
+  boardId?: string
   onRemoveWidget?: (widgetId: string) => void
   onMoveWidget?: (widgetId: string, direction: 'up' | 'down') => void
   onAddWidgetClick?: () => void
   onDeleteSection?: () => void
+  onRenameSection?: (title: string) => void
 }) {
   const [collapsed, setCollapsed] = useState(section.collapsed ?? false)
+  const [editingTitle, setEditingTitle] = useState(false)
+  const [titleDraft, setTitleDraft] = useState(section.title)
 
   const gridClass =
     section.type === 'links' ? 'links-grid' : section.type === 'charts' ? 'charts-grid' : 'metrics-grid'
 
+  const commitTitle = () => {
+    const next = titleDraft.trim()
+    if (next && next !== section.title) onRenameSection?.(next)
+    else setTitleDraft(section.title)
+    setEditingTitle(false)
+  }
+
+  const cancelTitle = () => {
+    setTitleDraft(section.title)
+    setEditingTitle(false)
+  }
+
   return (
     <section class="metrics-section">
       <div class="section-header">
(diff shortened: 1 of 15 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped feature PR. Two related capabilities: (1) inline-editable section titles in `EditableDashboard`, and (2) "update-in-place" of a board chart widget from the Chart Explorer.

The origin round-trip is sound: `WidgetRenderer` derives an `origin` (`board_id`/`section_id`/`widget_id`) for chart widgets and threads it through `buildChartUrl`; the chart page captures it once via `parseChartOrigin`, preserves it across edits in `syncUrl`, and `UpdateChartOnBoard` replaces the widget via the pure `replaceWidgetInConfig` — reusing the same widget id so the origin remains resolvable. Deleted widget/section/board degrades gracefully to `null`. Read-only public dashboards correctly don't leak links or origin (they use `PublicWidgetRenderer` with no `href`/`boardId`).

Hooks ordering is safe (the new `useState(origin)` sits before the pre-existing `!isLoggedIn` early return; `UpdateChartOnBoard`'s `return null` follows all its hooks). Inline title editing is idempotent across Enter/blur/Escape because `commitTitle` no-ops when the trimmed draft equals the current title.

Solid unit coverage for the new pure helpers (`updateWidget.ts`, `chart-url.ts`) and docs updated (`dashboard.md`, `sharing.md`). No blocking issues found.

Minor, non-blocking: while the target dashboard query is still loading, `UpdateChartOnBoard` returns `null`, so the "Update Chart" button only appears once the config is fetched (brief absence rather than a disabled/loading state). Also note `buildWidgetFromState` coerces a `productivity_category` source to `activity_type` for trend widgets — pre-existing behavior shared with the Add-to-Dashboard flow, so the update round-trip inherits it; worth keeping in mind but out of scope here.

PR #824, opened 2026-07-01
feat(dashboard): redirect to the board after updating a chart from the chart page

fiddur:
Follow-up to #823 (issue #822).

## What & why

After updating a board chart from the Chart Explorer, the page used to show a "Chart updated!" box with a **Done** button. Per feedback, it now **forwards the user straight back to the dashboard they came from** instead.

- Home board → `/`
- Owned shared board → `/u/<username>/<slug>`

Navigation is client-side (preact-iso `route`), and the `['dashboard']` / `['sharedDashboards']` queries are still invalidated first, so the updated widget is visible on arrival.

## Changes

- `UpdateChartOnBoard` (Chart page): dropped the `saved` state and the success/Done UI; `onSuccess` now invalidates + `route(destination)`.
- New pure helper `boardReturnPath(origin, username, slug)` in `updateWidget.ts` computes the return path. Extracting it also keeps `UpdateChartOnBoard` under the eslint `complexity` cap.
- Docs: `dashboard.md` notes the redirect.

## Testing

- New unit tests for `boardReturnPath` (home, shared, username url-encoding).
- `pnpm check` clean (0 errors, back to the 255-warning baseline).
- Full `apps/web` suite green (467 tests).

### DIFF
--- a/apps/web/src/pages/Chart/index.tsx
+++ b/apps/web/src/pages/Chart/index.tsx
@@ -40,7 +40,7 @@ import {
   updateUserSettings,
 } from '../../state/api'
 import { auth } from '../../state/auth'
-import { chartWidgetTitle, parseChartOrigin, replaceWidgetInConfig } from './updateWidget'
+import { boardReturnPath, chartWidgetTitle, parseChartOrigin, replaceWidgetInConfig } from './updateWidget'
 import './style.css'
 
 type SourceType = 'metric' | 'productivity_category' | 'activity_type'
@@ -785,10 +785,10 @@ function AddToDashboardModal({ state, onClose }: { state: ChartState; onClose: (
  */
 function UpdateChartOnBoard({ state, origin }: { state: ChartState; origin: ChartOrigin }) {
   const queryClient = useQueryClient()
+  const { route } = useLocation()
   const isHome = origin.board_id === 'home'
   const [expanded, setExpanded] = useState(false)
   const [title, setTitle] = useState('')
-  const [saved, setSaved] = useState(false)
 
   const dashboardQuery = useQuery({
     enabled: isHome,
@@ -810,6 +810,9 @@ function UpdateChartOnBoard({ state, origin }: { state: ChartState; origin: Char
   const section = targetConfig?.sections.find((s) => s.id === origin.section_id)
   const existingWidget = section?.widgets.find((w) => w.id === origin.widget_id)
 
+  // Where to send the user after a successful update: the board they came from.
+  const destination = boardReturnPath(origin, auth.value.user, shared?.slug)
+
   const updateMutation = useMutation({
     mutationFn: async () => {
       if (!targetConfig) return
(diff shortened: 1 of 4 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped change that replaces the post-update "Chart updated!" confirmation with a redirect back to the originating board. The new `boardReturnPath` helper is pure, unit-tested (home path, shared-dashboard path, and username URL-encoding), and correctly maps `home` → `/` and owned shared dashboards → `/u/<username>/<slug>`.

Verification notes:
- `route` from `useLocation()` matches the navigation pattern used elsewhere (e.g. `AddReport.tsx`, `ReportDetail.tsx`), including inside react-query `onSuccess`.
- No stale-slug race: `destination` is captured at render, but the Update button only renders once `existingWidget` (and thus `shared`/`slug`) has loaded — otherwise the component returns `null`.
- Removing the local `saved` state and the `add-to-dash-success` usage here doesn't orphan CSS; that class is still used by `AddToDashboardModal`.
- Docs updated to reflect the new behavior.

Minor (non-blocking): `boardReturnPath` URL-encodes the username but interpolates `slug` raw, whereas `PublicProfile`/`PublicDashboard` encode both. This is harmless in practice because slugs are DB-generated as `[A-Za-z0-9_-]+` (see `shared-dashboards.integration.test.ts`), but encoding it too would keep the callsites symmetric.

I could not run the unit tests here (the pnpm store is read-only in this sandbox, so `pnpm install` fails), but the added tests are simple pure-function assertions and are correct by inspection.

fiddur (commented):
✅ Approved

Clean, well-scoped change that replaces the inline "Chart updated!" confirmation with a redirect back to the originating board after a successful chart update.

- The redirect logic is extracted into a pure `boardReturnPath` helper with dedicated unit tests, correctly URL-encoding both username and slug. Verified the logic matches the test expectations.
- The destinations line up with the actual entry points: `/` for the home board (`boardId="home"` in `Dashboard`) and `/u/<username>/<slug>` for a shared board (`boardId={share.id}` in `PublicDashboard`), so users are returned exactly where they came from.
- The `if (!section || !existingWidget) return null` guard guarantees `shared` (and thus `slug`) is loaded before the Update button is clickable, so the empty-slug fallback can't occur in practice, and non-owned boards render null so no cross-user redirect is possible. Using `auth.value.user` for the username is safe since `listSharedDashboards` only returns the current user's own boards.
- The `onSuccess` query invalidations (`['dashboard']`, `['sharedDashboards']`) match the query keys the destination pages read, so the returned board reflects the update.
- Removing the `saved` state only from `UpdateChartOnBoard` is correct — the `add-to-dash-success` class and `saved` state are still in use by the separate `AddToDashboardModal` flow, so nothing is left dead.
- Docs updated to describe the new behavior; functional style and project conventions are followed.

No blocking issues found.

PR #825, opened 2026-07-01
feat(challenges): frontend — manage page, public race view, join flows

fiddur:
## Summary

Phase 2 (frontend) of **Challenges**, on the merged backend (#820).

- **`/challenges`** (sidebar "Challenges") — create a challenge (name; metric or activity type via the existing pickers; sum/count; unit; date range with This-week/This-month quick-sets computing `start_ts`/`end_ts` in the browser timezone; public/unlisted), copy its link, delete; list challenges you've joined; and **join by URL** (local or remote).
- **`/u/:username/:slug` is now a type dispatcher** (`PublicResource`): the server returns `{ type }` and the page renders the shared-dashboard view (owner-editable / read-only, unchanged) **or** the challenge view.
- **Challenge view** — a cumulative **race chart** (one `TrendLineChart` line per member, running totals from each member's buckets) + a **leaderboard** (rank, `@member · host`, total, freshness/stale). Join buttons: one-click **Join** for logged-in users on the host instance; **Join from another instance** (enter your host → redirected to your own instance's `/challenges/join`, which performs the federated join).
- **`/challenges/join?challenge=…`** runs the join then redirects to the challenge.

## Testing
- `pnpm check` clean in `apps/web`; full web suite (464 tests) passes.
- Manual (after deploy / via the 3 local clones): create a challenge, view the race chart; join locally; join cross-instance via "Join from another instance".

## Notes
- The public dispatcher fetches `/public/:u/:slug` once; the race chart pulls `/public/:u/:slug/standings` (host-aggregated, TTL-cached).
- Reuses `TrendLineChart`'s `multiSeries`, `MetricPicker`/`ActivityTypePicker`, and the shared-dashboards public-page chrome-hiding.


apps/web/src/pages/Challenges/Join.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </div>
+    )
+  }
+
+  if (!isLoggedIn) {
+    return (
+      <div class="challenges-page">
+        <h1>Join a challenge</h1>
+        <p>
+          Please <a href={`/login?next=${encodeURIComponent(`/challenges/join?challenge=${challengeUrl}`)}`}>log in</a>{' '}
```
fiddur: This `next` param is never honored. `pages/Login/index.tsx` doesn't read `next` — on every successful login (and when already authenticated) it hardcodes `route('/')`. So a logged-out user who clicks "log in" here lands on the home dashboard and the join is abandoned, breaking the join-from-another-instance flow documented in `docs/features/challenges.md`. Make the Login page redirect to `next` after auth (or otherwise resume the join).

fiddur (commented):
🛑 Changes required

Solid, well-structured frontend for the challenges feature — the manage page, public race view, and API layer all line up cleanly with the `@aurboda/api-spec` types and the backend routes/response shapes I checked.

One blocking issue: the logged-out **join-from-another-instance** flow is broken. `Join.tsx` sends users to `/login?next=<join-url>`, but the Login page (`apps/web/src/pages/Login/index.tsx`) never reads a `next` query param — every success path hardcodes `route('/')` (and it also `route('/')`s when a token is already present). So after logging in, the user lands on the home dashboard and the pending join is silently abandoned, which defeats the exact flow described in `docs/features/challenges.md`. Either make the Login page honor `next`, or have `Join.tsx` stash the target and resume after auth.

Minor, non-blocking notes (left out of inline comments):
- `PublicResource` uses `const dash = data as PublicSharedDashboardResponse`; the project standard is to avoid casts and prefer type guards — narrowing on `data.type === 'dashboard'` would avoid the cast.
- The leaderboard in `PublicChallenge.tsx` renders all standings including `withdrawn` members (ranked with `total` 0 at the bottom, since the backend sorts by total desc), while the race chart filters to `active` only. Harmless but slightly inconsistent.
- `fetchOwnerChallengeStandings` and `updateChallenge` are added to `state/api/challenges.ts` but aren't used by any view in this PR; and `fetchPublicSharedDashboard` in `state/api/social.ts` appears to be dead now that `fetchPublicResource` replaced it in the public route.

apps/web/src/pages/Challenges/Join.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function ChallengeJoin() {
+  const { query } = useLocation()
+  const challengeUrl = typeof query.challenge === 'string' ? query.challenge : ''
+  const isLoggedIn = Boolean(auth.value.token)
+  const started = useRef(false)
+
+  const joinMutation = useMutation({
+    mutationFn: () => joinChallengeByUrl(challengeUrl),
+    onSuccess: () => {
+      window.location.href = challengeUrl
```
fiddur: Minor / non-blocking: this redirects to the raw `challenge` query param. It's effectively safe because it only runs `onSuccess` of a join the backend actually resolved (an arbitrary or `javascript:` URL would fail the join and never reach here), so the target is a validated challenge URL. Flagging only so it's a conscious choice — if the join endpoint ever loosens its URL validation, this becomes an open redirect.

fiddur (commented):
✅ Approved

Solid, well-scoped frontend PR for challenges. The manage page, public race view, and join-by-URL flows are cleanly implemented and reuse existing pieces (`TrendLineChart`, pickers, `useMutation`/`useQuery`) idiomatically.

Checks I verified:
- **Types line up with api-spec**: field names used in `PublicChallenge`/`toCumulativeSeries` (`buckets[].bucket_start/value`, `identity_base_url`, `status`, `total`, `stale`, `last_updated`) match `challengeStandingSchema`/`chartDataBucketSchema`. `identity_base_url` is per-member (public profile base URL), so the leaderboard `key` is unique.
- **Discriminated-union resolution** in `PublicResource` (`data.type === 'challenge'` then `'config' in data`) narrows without casts and matches the backend's `type: 'dashboard' | 'challenge'` discriminators.
- **No owner regression**: `getSharedDashboardBySlug` does not filter on `is_public`, so the unauthenticated `/public/:username/:slug` fetch resolves unlisted dashboards too — owners still reach `OwnerSharedDashboard`. (Minor: the owner path now does one extra public fetch whose result is discarded, but that's harmless.)
- **Login `next` guard** correctly rejects protocol-relative `//` targets; since `route()` is SPA navigation (history.pushState), cross-origin targets can't leak anyway.
- **CI-relevant lint**: `react-hooks` is not among the oxlint plugins, so the intentionally-omitted `joinMutation` dep in the `ChallengeJoin` effect won't fail; and the un-awaited `invalidateQueries` inside block-body `onSuccess` mirrors an existing pattern in `SharedDashboards` that already passes.

One minor, non-blocking note left inline about the post-join redirect target. Nothing blocking. Note the diff is web-only — I didn't run `pnpm check`, so trust CI for the final type/lint gate.

apps/web/src/pages/PublicDashboard/index.tsx
```diff
@@ -154,16 +136,67 @@ function ReadOnlyPublicDashboard({ username, slug }: { username: string; slug: s
...
+  if (query.isLoading) {
+    return (
+      <div class="public-dashboard">
+        <div class="public-loading">Loading…</div>
+      </div>
+    )
+  }
+
+  const data = query.data
+  if (query.isError || !data?.success) {
```
fiddur: This early return runs before the `isOwner` check on line 179, so an owner now sees "Not found" (and loses the editable view) whenever this unauthenticated fetch fails — a regression from the previous behavior where owners rendered `OwnerSharedDashboard` unconditionally. Since `retry: false`, a single transient failure of the public endpoint is enough. Consider checking `isOwner` before this error branch (or falling through to `OwnerSharedDashboard` on a failed public fetch).

apps/web/src/index.tsx
```diff
@@ -41,7 +41,9 @@ import { MealTypeMeta } from './pages/MealTypeMeta/index.jsx'
 import { MetricMeta } from './pages/MetricMeta/index.jsx'
 import { Places } from './pages/Places/index.jsx'
 import { Privacy } from './pages/Privacy/index.jsx'
-import { PublicDashboard } from './pages/PublicDashboard/index.jsx'
+import { Challenges } from './pages/Challenges/index.jsx'
```
fiddur: These new imports are out of oxfmt's alphabetical (case-insensitive, by path) order for the internal-value group — `./pages/Challenges/*` should sort near the top, before `./pages/DeductionRules/*`, not after `Privacy`. Running `pnpm fix` will reorder them. Not caught by CI (`pnpm check` doesn't run the formatter).

fiddur (commented):
✅ Approved

Solid, well-scoped frontend PR for the challenges feature. The federation flows are thoughtfully built: the login `next` param is guarded against open redirects (must start with `/`, not `//`), the join-by-URL redirect only follows `http(s)` targets, and the actual federated fetch/SSRF checks live server-side. The `/u/:username/:slug` resolver refactor correctly narrows the `PublicSharedDashboardResponse | PublicChallengeResponse` union via the `type` discriminator and `'config' in data` without casts, matching the api-spec types and the backend `public-shares-router` (both endpoints and the `type` values line up). Docs are updated.

Non-blocking points:

1. **Owner view now depends on the public fetch.** `PublicResource` always calls the unauthenticated `fetchPublicResource` and returns "Not found" on `query.isError || !data?.success` *before* the `isOwner` check (with `retry: false`). Previously an owner rendered `OwnerSharedDashboard` directly. A transient public-endpoint failure would now block the owner from their own editable dashboard — see inline note.

2. **Missing unit tests for the pure helpers.** `toCumulativeSeries` (PublicChallenge.tsx), and `dateToStartIso`/`dateToEndIso`/`setThisWeek`/`setThisMonth` (Challenges/index.tsx) are pure and easily testable; the web app already has ~38 unit test files including for util-style logic, and the project standard is "unit test first when reasonable." Worth adding a small test file (the `setThisWeek`/default-`endDate` ms-based `+ 6 * 864e5` arithmetic in particular can drift by a day across a DST boundary, which a test would surface).

3. **Import ordering** in `index.tsx` — the new imports were inserted out of oxfmt's sort order; `pnpm fix` will reorder them (does not affect CI, which only runs `pnpm check`).

apps/web/src/pages/Challenges/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+  deleteChallenge,
+  joinChallengeByUrl,
+  leaveChallenge,
+  listChallenges,
+  listMyChallengeParticipations,
+} from '../../state/api'
+import './style.css'
+
+/** Local YYYY-MM-DD → UTC ISO instant at that day's local midnight. */
+const dateToStartIso = (date: string): string => new Date(`${date}T00:00:00`).toISOString()
```
fiddur: These date helpers (and `setThisWeek`/`setThisMonth` below) carry subtle logic — exclusive end-of-day, week-start math, DST-sensitive `+ 6 * 864e5` arithmetic. Per the project's "make unit test first / prefer testable code" standard and the web codebase's precedent of extracting pure logic into tested `.ts` modules, consider pulling these into a small module with unit tests. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped frontend PR for the challenges feature. I traced the main risk areas and they hold up:

- **Union resolution in `PublicResource`** is correct: `PublicSharedDashboardResponse` and `PublicChallengeResponse` both carry the optional `type` literal (`'dashboard'`/`'challenge'`) plus `success`, so `data.type === 'challenge'` and `'config' in data` narrow the union without casts, matching the backend's `/public/:username/:slug` responses.
- **Leaderboard rank** (`i + 1` by array order) is safe — `challenge-standings.ts` already sorts by highest total first.
- **Measurement `pattern`** is populated correctly — `MetricPicker`/`ActivityTypePicker` emit names (not ids), which is what `challengeSpec.pattern` expects.
- **Open-redirect handling** is reasonable in both places: the login `next` is restricted to internal paths and applied via SPA `route()`, and `Join.tsx` gates its full-page `window.location.href` redirect to `http(s)` targets.
- No dangling references remain to the removed `fetchPublicSharedDashboard`, and the owner-view fallback correctly survives a failed public fetch.

Non-blocking suggestions:
- The pure helpers introduced here (`dateToStartIso`/`dateToEndIso`, `setThisWeek`/`setThisMonth`, and `toCumulativeSeries`) carry subtle logic (exclusive end date, week boundary math, running totals) but are defined inline and untested. The web codebase's convention is to extract such pure logic into small `.ts` modules with unit tests (e.g. `pages/Timeline/*`, `utils/*`); doing so here would align with the project's "make unit test first / prefer testable code" standard.
- Minor: in `PublicChallenge.joinFromOtherHost`, `host.startsWith('http')` accepts malformed inputs like `httpfoo` (kept verbatim) — a slightly tighter check would avoid producing an invalid target URL. Not worth blocking.
- Minor UX: `TrendLineChart` gates the whole race chart on `multiSeries[0].data.length >= 2`, so if the top member has a single bucket the chart is hidden even when other members have more points. Pre-existing chart behavior, just noting the interaction.

apps/web/src/pages/Challenges/Join.tsx
```diff
@@ -0,0 +1,67 @@
...
+      </div>
+    )
+  }
+
+  if (!isLoggedIn) {
+    return (
+      <div class="challenges-page">
+        <h1>Join a challenge</h1>
+        <p>
+          Please <a href={`/login?next=${encodeURIComponent(`/challenges/join?challenge=${challengeUrl}`)}`}>log in</a>{' '}
```
fiddur: Minor/defensive: `challengeUrl` is interpolated into the inner query string before the whole thing is `encodeURIComponent`-d, so it's only encoded once. When `next` is later re-parsed as a URL, a `challengeUrl` containing `&` or `#` would be truncated at that character. Today's share URLs (`/u/:user/:slug`) don't contain those, so this is harmless in practice — but encoding the inner value too (`challenge=${encodeURIComponent(challengeUrl)}`) would make it robust. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-structured frontend for challenges. I verified locally: `@aurboda/api-spec` builds, `apps/web` type-checks cleanly with `tsgo --noEmit`, the new `date-range` unit test passes (5/5), and oxlint reports only a pre-existing warning outside this diff.

- Field usage matches the api-spec schemas (`bucket_start`/`value`, `share_url`, `spec`, `identity_base_url`, etc.).
- The `PublicResource` discriminated-union handling is correct: challenge branch first (`data.type === 'challenge'`), then owner falls back to the editable dashboard independent of the public fetch, then `'config' in data` narrows the dashboard case without casts.
- `fetchPublicSharedDashboard` was fully removed with no dangling references; `fetchPublicResource` replaces it everywhere.
- The leaderboard's index-based rank (`i + 1`) is correct because the backend already sorts standings by `total` desc (`challenge-standings.ts:100`).
- Login `next` is guarded against open redirects (internal path only), and the join redirect is restricted to `http(s)`.

1. In `Join.tsx` the login link embeds `challengeUrl` into the inner `?challenge=` query string without encoding it first (inline note below). Harmless for today's `/u/:user/:slug` share URLs but latent if a share URL ever carried `&`/`#`.
2. If an owner opens their *own* challenge slug and the (unauthenticated, `retry:false`) public fetch transiently fails, `PublicResource` falls through to `OwnerSharedDashboard`, which will render "Dashboard not found" for a challenge slug. Edge case only on transient failure; not worth blocking.

apps/web/src/pages/Challenges/Join.tsx
```diff
@@ -0,0 +1,70 @@
...
+    )
+  }
+
+  return (
+    <div class="challenges-page">
+      <h1>Joining challenge…</h1>
+      {joinMutation.isError && (
+        <p class="challenges-empty">
+          {joinMutation.error instanceof Error ? joinMutation.error.message : 'Failed to join.'}{' '}
+          <a href={challengeUrl}>Open the challenge</a>
```
fiddur: `challengeUrl` comes straight from the `challenge` query param (attacker-controllable), and here it's rendered as a raw `href` without the `^https?://` guard you deliberately apply to the redirect on line 23. A crafted link like `/challenges/join?challenge=javascript:alert(document.cookie)` reliably lands in this error branch — a `javascript:` URL passes the backend's `z.string().url()` but fails to resolve, so the join errors — and preact does not sanitize `href`, so clicking "Open the challenge" executes the script (DOM XSS; at minimum an arbitrary-scheme link for phishing). Apply the same http(s) check here, or omit the link when `challengeUrl` isn't http(s).

fiddur (commented):
🛑 Changes required

Solid, well-structured frontend for the challenges feature. I verified the new API client and views against `packages/api-spec` (the `type` discriminator on the public resolver, `ChallengeStanding`/`PublicChallenge` shapes, participation/response types) and against the backend routes in `public-shares-router.ts` / `challenges-router.ts` — everything lines up, including that the backend already sorts standings by total desc, so the leaderboard's index-based rank is correct. The date-range helpers are pure and well-tested; I re-ran the logic under `TZ=America/New_York` and all expectations hold, confirming the DST/timezone-safety claim. The `PublicResource` union narrowing (`data.type === 'challenge'`, `'config' in data`) is clean and cast-free, and the owner-resilience path is reasonable given `retry: false`.

One blocking issue: the error-branch link in `Join.tsx` renders the attacker-controllable `challenge` query param as a raw `href` without the `http(s)` guard that's already applied to the success redirect — a DOM XSS / arbitrary-scheme link vector. Details inline. The login `next` open-redirect guard and the join success-redirect guard are otherwise correct.

apps/web/src/pages/Challenges/PublicChallenge.tsx
```diff
@@ -0,0 +1,145 @@
...
+const toCumulativeSeries = (standing: ChallengeStanding, color: string): LineSeriesData => {
+  const sorted = [...standing.buckets].sort((a, b) => a.bucket_start.localeCompare(b.bucket_start))
+  let running = 0
+  return {
+    color,
+    data: sorted.map((b) => {
+      running += b.value
+      return { date: b.bucket_start, value: running }
+    }),
+    name: standing.display_name,
```
fiddur, line 27: `display_name` here is federated, attacker-controlled data: a joining instance registers it via `registerChallengeMemberBodySchema` (`display_name: z.string().min(1).max(120)`, no sanitization), and it's served back to viewers by `fetchPublicChallengeStandings`. It flows into `multiSeries[].name` (used at line 113), and `TrendLineChart`'s `attachMultiSeriesTooltip` writes series names into `tooltip.innerHTML` without escaping (`apps/web/src/components/charts/TrendLineChart.tsx:336`/`:339`). So a member registered as e.g. `` yields a **DOM XSS** on this public page, firing in any viewer's browser (on the host origin) as soon as they hover the race chart.

This is the first path feeding untrusted, cross-user data into that `innerHTML` sink (the Chart page only feeds the viewer's own tag/metric names). Best fixed at the sink — build the multi-series tooltip with `textContent`/DOM nodes instead of `innerHTML` — or escape `display_name` before it reaches the chart.

fiddur (commented):
🛑 Changes required

Solid frontend for the challenges feature — the resource resolver, join flows, and date-range helpers are clean, the union narrowing in `PublicResource` is type-safe (no casts), and the new `date-range.test.ts` covers the tricky DST/week/month math. `pnpm typecheck` and `pnpm lint` both pass for the new files, and the 5 date-range tests are green.

One blocking issue: the public race chart wires **attacker-controlled, cross-instance `display_name`** into `TrendLineChart`'s multi-series tooltip, which renders series names via `innerHTML` — a DOM XSS on the public `/u/:username/:slug` page. See the inline comment. Everything else looks good.

Minor (non-blocking) notes:
- `Join.tsx`'s `useEffect` deliberately omits `joinMutation` from its deps and relies on the `started` ref guard — fine, and the `javascript:`-URL guard on `safeChallengeUrl` is a nice touch.
- The `next` open-redirect guard in `Login` is only reached via client-side `route()`, so it's safe; the double-`encodeURIComponent` round-trip between `Join` and `Login` is correct.

fiddur (commented):
✅ Approved

Clean, well-structured frontend PR for the challenges feature. I ran the checks locally: `tsgo --noEmit` passes, `oxlint` reports no new errors (the 255 warnings are all pre-existing in unrelated files like `Meals/index.tsx`), and the new `date-range.test.ts` passes (5/5).

Strengths:
- **Security-conscious throughout.** The multi-series tooltip now builds DOM text nodes instead of `innerHTML` (federated `display_name` can't become an HTML-injection sink); the login `next` param guards against open redirect (`startsWith('/') && !startsWith('//')`, client-side `route()` only); and `Join.tsx` restricts the redirect href to `http(s)` while still letting the backend validate the raw URL. All federated strings rendered in JSX (`display_name`, `challenge.name`, `shortHost(...)`) rely on Preact's auto-escaping — no `dangerouslySetInnerHTML`.
- Verified correctness of the pieces I couldn't see from the diff alone: the backend sorts standings by `total` desc, so the leaderboard's `i + 1` rank is correct; `identity_base_url` is a per-member profile URL, so the `key={s.identity_base_url}` won't collide for two members on the same host; `chartDataBucketSchema` field names (`bucket_start`, `value`) match `toCumulativeSeries`; and the `PublicResource` union narrowing (`data.type === 'challenge'` / `'config' in data`) is sound with no casts.
- Follows the documented conventions: the web api module imports types only (no runtime schemas), snake_case fields everywhere, functional components, and the `PublicDashboard` → `PublicResource` refactor cleanly folds the removed `fetchPublicSharedDashboard`/`ReadOnlyPublicDashboard` into the shared resolver with no dangling references.

Minor, non-blocking observations (no fix required):
- `JoinedRow` renders `participation.challenge_url` directly as an `href` without the `http(s)` guard applied in `Join.tsx`. This is safe in practice — a participation record only exists after the backend successfully fetched the challenge spec, which can't happen for a `javascript:` URL — but a defense-in-depth guard here would match the care taken elsewhere. Same reasoning applies to `challenge.share_url` hrefs, which are backend-generated and thus trusted.
- The race chart passes `data={[]}` with `multiSeries`; `TrendLineChart`'s `effectiveData` guard keys off `multiSeries[0]?.data`, so if the top-ranked member has fewer than 2 buckets the whole chart falls back to the "Insufficient data" placeholder even when other members have data. This is a pre-existing limitation of the shared component, not introduced here, and only affects a day-1 edge case.
- The create form doesn't guard `startDate <= endDate` client-side; worth confirming the backend rejects an inverted range.

PR #826, opened 2026-07-01
fix(dashboard): persist and render chart breakdown on board widgets

fiddur:
## The bug

Adding (or updating) a chart to a board **dropped its breakdown**. For example, a `sex` trend broken down by `partner` on the Chart Explorer would lose the breakdown once saved to a board — it rendered as a single flat series.

Root cause: the widget config schemas (`trendChartConfigSchema`, `barChartConfigSchema`) had no `breakdown_fields`, so `buildWidgetFromState` silently discarded it on save, and the dashboard chart widgets never fetched or rendered breakdown series.

## The fix (all layers)

- **api-spec** — `breakdown_fields` added to the trend/bar **config** schemas; `breakdown_series`/`breakdown_histories` (trend) and `breakdown_series`/`breakdown_buckets` (bar) added to the widget **data** schemas.
- **Persist + link** — `buildWidgetFromState` writes `breakdown_fields`; `buildChartUrl` carries it so a board chart links back to the explorer with its breakdown intact.
- **Render (owner)** — `TrendChartWidget`/`BarChartWidget` fetch with `breakdown_fields` and render multi-series + a legend, via a new shared `SERIES_COLORS`/`BreakdownLegend` module (deduped from the Chart page, with its own CSS so it loads wherever widgets render).
- **Render (public)** — the server-side resolver in `shared-dashboard-data.ts` forwards `breakdown_fields` and returns the breakdown projection, so anonymous visitors to a public shared board see the breakdown too.
- Regenerated OpenAPI + Kotlin from the schemas.

## Testing

- **Resolver integration test**: seeds a broken-down activity type and asserts `breakdown_series`/`breakdown_histories` (trend) and `breakdown_series`/`breakdown_buckets` (bar) resolve, with the flat `buckets` empty in breakdown mode.
- **Unit tests**: `buildChartUrl` emits/omits `breakdown_fields`.
- `pnpm check` clean on web + backend (0 errors); full web suite green (469).


fiddur (commented):
✅ Approved

Solid, well-scoped change. The breakdown data now round-trips correctly across all layers: the api-spec schemas (`TrendChartData.breakdown_histories/breakdown_series`, `BarChartData.breakdown_buckets/breakdown_series`, and the two config schemas) match what the backend resolver emits and what the frontend consumes. The public-board `resolveTrendChart`/`resolveBarChart` resolvers gate on `breakdown_series?.length` (so empty/no-data cases cleanly fall back to the flat shape) and, for bars, correctly narrow the union buckets with `'series' in b` and zero out the flat `buckets`. Extracting `SERIES_COLORS`/`BreakdownLegend` + its CSS into the shared `components/charts/breakdown` module is a clean DRY move, and every call site (Chart page, both widgets) was updated consistently — no stale references to the removed local copies. The `buildChartUrl` comma-join pairs correctly with the Chart page's `parseQuery` split (`.filter(Boolean)`), so the widget→/chart link preserves the breakdown. Good test coverage: a new integration test exercising both chart types with real seeded activities, plus unit tests for the URL param. I could not run `pnpm check` (dependency install is blocked by a read-only pnpm store in this sandbox), but the types line up on inspection.

Two non-blocking notes:

1. In `BarChartView`'s breakdown branch, the `<BarChart>` is not passed `bucketSize`/`rangeStart`/`rangeEnd`, unlike the Chart page's `BarDisplay`, so bar widths fall back to the 1d default. This matches the existing non-breakdown widget path (which also omits them), so it's consistent — just a minor rendering-fidelity gap if you want widget bars to size like the full page.

2. Breakdown series names are categorical field values (e.g. partner or mood) and are now surfaced on public shared boards. This is intentional (the owner explicitly configured the breakdown and chose to publish the board, and it's exactly what the widget renders), but worth being conscious of given the file's security-boundary framing — a user breaking down by a sensitive field will expose those values publicly.

PR #829, opened 2026-07-01
fix(chart): restrict cumulative-metric sums to trusted sources

fiddur:
## Problem
On `/chart` (and anywhere `query_chart_data` is used), a **daily `steps` sum** showed ~44k for a day the goal/period-summary reported as ~7.3k. `queryMetricBuckets` did `SUM(value) FROM time_series WHERE metric='steps'` across **all sources**, so cumulative/derived metrics — stored simultaneously from raw per-minute readings, deduplicated daily aggregates, and device totals — were multiply-counted. It also included soft-deleted rows.

## Fix
Apply the same trusted-source filter the metric/period-summary read paths already use (`getSourceFilter` → `cumulativeSources` for steps/distance/…, `aurbodaOnlySources` for calories) in `queryMetricBuckets`, and exclude `deleted_at`. Non-cumulative metrics (heart_rate, weight, …) are unchanged (all sources).

**This also fixes challenge totals** for cumulative metrics — challenges use the same `getChartData` path, so a steps challenge would have over-counted identically.

## Tests
- `chart-data.test.ts`: assert a cumulative metric (`steps`) sum SQL includes `source = ANY(...)` with `cumulativeSources` + `deleted_at IS NULL`; a non-cumulative metric (`heart_rate`) does not filter sources. `getSourceFilter` pulled through real (pure fn) so the rule can't drift. All 26 chart-data tests pass; `pnpm check` clean.

## Not in this PR (related, separate)
- **Zone 2 / Zone 2 (Weekly) render empty charts** — `hr_zone_*_sec` are *computed* metrics (derived from heart_rate), not stored in `time_series`, so `query_chart_data` finds nothing. Needs the chart path to compute HR-zone metrics — filing/fixing separately.
- Per-day bucket alignment for a non-UTC timezone (chart buckets in UTC; the goal buckets in local time) — the window **total** is exact regardless; per-day points can shift by the tz offset. That's the documented tz-local-bucketing refinement.

### DIFF
--- a/apps/backend/src/services/chart-data.test.ts
+++ b/apps/backend/src/services/chart-data.test.ts
@@ -1,11 +1,15 @@
+import type * as TimeSeriesModule from '../db/time-series.ts'
+
 import { beforeEach, describe, expect, test, vi } from 'vitest'
 
 import * as db from '../db/index.ts'
 import { buildBucketExpr, getChartData } from './chart-data.ts'
 
-// Mock the db module
-vi.mock('../db', () => ({
+// Mock the db module. getSourceFilter is pulled through real (pure function) so
+// the cumulative-source rules can't drift from production.
+vi.mock('../db', async () => ({
   expandActivityTypes: vi.fn().mockImplementation((_user: string, types: string[]) => Promise.resolve(types)),
+  getSourceFilter: (await vi.importActual<typeof TimeSeriesModule>('../db/time-series.ts')).getSourceFilter,
   query: vi.fn(),
 }))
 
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Clean, focused fix. `queryMetricBuckets` now restricts cumulative/derived metric queries to the trusted sources via `getSourceFilter(metric)` and adds `deleted_at IS NULL`, so chart (and challenge) totals stop multiply-counting raw + daily-aggregate rows.

Verified:
- **Parameter indexing is correct.** `p = bucket.params.length` (always 1); metric/start/end bind to `$p+1..$p+3`, and the source filter is pushed onto `params` and referenced as `source = ANY($<params.length>)` — indices line up.
- **Consistent with the rest of the codebase.** Applying the filter across all aggregations (not just `sum`) matches the existing `getTimeSeriesStats` / `getDailyAggregates` / `getTimeSeriesBucketed` read paths, and adding `deleted_at IS NULL` is a genuine correctness improvement (the previous chart query summed soft-deleted rows).
- **Constants check out.** `getSourceFilter('steps')` → `['health_connect_aggregate','aurboda']` (matches the new assertion); `heart_rate`/`weight` → `null` (non-cumulative path unchanged apart from the deleted_at guard).
- **Challenges share the path.** `challenge-spec.ts#resolveMemberSeries` calls `getChartData`, so the dedup fix propagates as the comment states.
- **Tests** pull the real (pure) `getSourceFilter` through the `../db` mock to prevent rule drift, and add both a cumulative regression guard and a non-cumulative negative case.

Note: I was unable to run the suite locally (workspace deps aren't installed and the pnpm store sits outside the sandbox-writable paths), but the diff is small enough to validate by inspection. Optional, non-blocking: the unit tests assert on SQL-string content; an integration test exercising the real dedup against seeded multi-source rows would guard the actual behavior more directly, but the existing db-layer tests already cover the source-filter SQL semantics.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. `queryMetricBuckets` now restricts cumulative/derived metrics (steps, distance, calories, …) to their trusted sources via the shared `getSourceFilter`, so charts and challenges stop multiply-counting rows that exist under both raw and daily-aggregate sources. This correctly reuses the same rule the metric/period-summary read paths already apply.

Verified:
- **Parameter indexing is correct** — `$1` bucket interval, `$2` metric, `$3` start, `$4` end, and the source array bound at `$${params.length}` (=$5) after the push. `query` accepts `unknown[]`, so the widened `params` type is fine.
- **`getSourceFilter` semantics match the tests** — returns `['health_connect_aggregate','aurboda']` for `steps` (asserted) and `null` for `heart_rate` (negative test asserts no `source = ANY`).
- **Bonus correctness fix**: the added `AND deleted_at IS NULL` closes a real gap — the metric bucket query previously counted soft-deleted points.
- **Comment claim holds**: challenges do share this path (`services/challenge-spec.ts` calls `getChartData`).
- The `vi.mock('../db', async …)` factory that pulls the real pure `getSourceFilter` through `importActual` is a valid pattern and keeps the cumulative-source rules from drifting from production.

Note (non-blocking): the filter is applied to all aggregations, not only `sum` as the title suggests — this is intentional and consistent with `getTimeSeries`/`getTimeSeriesStats`, which always filter cumulative sources regardless of aggregation.

I was unable to execute the test suite here (this review worktree has no installed dependencies and the pnpm store path is read-only), but the logic and test assertions trace cleanly.

PR #830, opened 2026-07-01
feat(chart): chart computed HR-zone metrics (Zone 0–5)

fiddur:
## Problem
On `/chart`, "Zone 2" (`hr_zone_2_sec`) and "Zone 2 (Weekly)" (`zone2_weekly`) rendered **empty charts**. HR-zone metrics aren't stored in `time_series` — they're computed from heart-rate data + the user's zone thresholds (as the period summary / HR-zones widget do) — so `query_chart_data`'s metric query found no rows. `zone2_weekly` is a dashboard-card-only alias and isn't a real metric at all.

## Fix
- `queryHrZoneBuckets`: fetch `heart_rate` for the window (source-filtered via `getTimeSeries`), group samples into UTC buckets mirroring `buildBucketExpr`, and run the existing pure `computeHrZoneSecs` per bucket with `getEffectiveHrZones`. Value = seconds in that zone per bucket.
- Route `hr_zone_<0-5>_sec` metrics through it; alias `zone2_weekly` → `hr_zone_2_sec` so that picker entry charts too. Non-zone metrics unchanged.
- Extracted the metric routing into `queryMetricSource` (keeps `getChartData` under the complexity limit).

This also enables **zone-based challenges** (they share `getChartData`).

## Tests
- `chart-data.test.ts`: hr_zone metric reads `heart_rate` + buckets per day (not the empty `hr_zone_*` rows) and never hits the plain metric query; `zone2_weekly` aliases to the computed path. `computeHrZoneSecs` pulled through real. 28 tests pass; `pnpm check` clean.

## Note
The chart shows zone time in **seconds** (raw metric value); the dashboard card separately converts to weekly minutes. Follow-up if a minutes display is wanted on the chart.

### DIFF
--- a/apps/backend/src/services/chart-data.ts
+++ b/apps/backend/src/services/chart-data.ts
@@ -7,7 +7,8 @@
 
 import type { ChartDataBreakdownBucket, ChartDataBucket, ChartDataSourceType } from '@aurboda/api-spec'
 
-import { expandActivityTypes, getSourceFilter, query } from '../db/index.ts'
+import { expandActivityTypes, getSourceFilter, getTimeSeries, query } from '../db/index.ts'
+import { computeHrZoneSecs, getEffectiveHrZones } from './settings.ts'
 
 /** Map bucket_size parameter to PostgreSQL date_trunc interval name (day and above). */
 const bucketToTrunc: Record<string, string> = {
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/chart-data.ts
```diff
@@ -311,6 +381,25 @@ const queryActivityTypeBreakdown = async (
...
+  user: string,
+  pattern: string | undefined,
+  start: string,
+  end: string,
+  bucketSize: string,
+  aggregation: 'count' | 'mean' | 'sum',
+): Promise<ChartDataBucket[]> => {
+  const metric = pattern === 'zone2_weekly' ? 'hr_zone_2_sec' : pattern
+  if (!metric) return []
+  if (/^hr_zone_[0-5]_sec$/.test(metric)) return queryHrZoneBuckets(user, metric, start, end, bucketSize)
```
fiddur, line 399: Minor (non-blocking): `aggregation` is a parameter of `queryMetricSource` but isn't forwarded to `queryHrZoneBuckets`, which always returns summed seconds-in-zone per bucket. So a zone-metric chart requested with `mean` or `count` silently gets a sum. Fine as a deliberate choice (zone-seconds are inherently a sum), just flagging that the param is dropped on this branch.

fiddur (commented):
✅ Approved

Clean, well-scoped feature. HR-zone metrics (`hr_zone_0_sec`..`hr_zone_5_sec` and the `zone2_weekly` dashboard alias) are computed from `heart_rate` samples on the fly rather than queried as stored rows, reusing the existing `computeHrZoneSecs` + `getEffectiveHrZones` logic — consistent with the identical treatment in `services/goals.ts`, so the zone rules can't drift.

Correctness checks:
- `utcBucketStart` faithfully mirrors `buildBucketExpr`'s SQL bucketing: 2000-01-01 UTC bin origin for sub-day sizes, hour/day/month truncation, and Monday-aligned weeks (`(getUTCDay()+6)%7`, matching PG `date_trunc('week')`). All math is UTC, matching the `AT TIME ZONE 'UTC'` SQL path.
- The `^hr_zone_[0-5]_sec$` guard in `queryMetricSource` runs before `queryHrZoneBuckets`, so the `as 0|1|2|3|4|5` cast on the parsed zone index is safe (same cast pattern as `goals.ts`).
- Empty/single-sample handling delegates to `computeHrZoneSecs`; empty HR data returns `[]` like the other query helpers.

I ran the backend suite (`chart-data.test.ts`: 28/28 passing) and `tsgo --noEmit` (clean) locally.

Minor, non-blocking notes:
- `queryMetricSource` receives `aggregation` but does not forward it to `queryHrZoneBuckets`, so an `hr_zone` chart requested with `mean`/`count` still returns summed seconds-in-zone. That's a defensible design (zone-seconds are inherently a sum, and the widgets use `sum`), but the parameter is silently dropped for this path.
- Because `computeHrZoneSecs` gives each group's last sample a synthetic mean-gap, per-bucket sums won't exactly equal a whole-period computation at bucket boundaries. Since gaps are capped at `MAX_GAP_SECONDS` (5s), the divergence is at most a few seconds per boundary — negligible in practice.

apps/backend/src/services/chart-data.ts
fiddur: Intentional: for computed HR-zone metrics the bucket value is a *duration* (seconds spent in the zone within the bucket), produced directly by `computeHrZoneSecs`. There is no meaningful alternate reduction (avg/min/max) of "seconds in zone" — the per-bucket value already *is* the sum of time in that zone, and coarser windows are handled by `bucket_size`, not `aggregation`. Forwarding `aggregation` would imply a choice that does not apply here, so the HR-zone path deliberately ignores it. Resolving as non-blocking.

PR #832, opened 2026-07-01
feat(dashboard): mask breakdown labels on public boards (per field, A/B/C)

fiddur:
Follow-up to the breakdown work (#826): a generic way to keep sensitive breakdown **values** off public boards.

## Why

Breaking a chart down by a categorical field (e.g. `partner`) surfaced the real values as series labels — and after #826 those render on public shared boards too. This adds an opt-in, **per-field** mask so an owner can hide chosen breakdown values from public viewers while still showing the breakdown's shape and volumes.

## Design

- **Per-field, not whole-widget:** `masked_breakdown_fields` is a subset of `breakdown_fields`. With `breakdown_fields: [partner, mood]` you can mask just `partner` → public sees `A / happy`, `B / sad`. A single-field breakdown (the common case) is just a one-entry array.
- **Server-side only (real privacy):** masking happens in the public resolver (`shared-dashboard-data.ts`), so the real values **never leave the server** for a public viewer. You keep seeing real values on your own live (client-fetched) view.
- **Stable A/B/C:** each masked field's distinct values are sorted and lettered (A, B, … Z, AA…), so the same value maps to the same label across reloads. If a real value contains the ` / ` series delimiter (ambiguous split), it falls back to masking the whole label — privacy-first.

## Changes

- **api-spec**: `masked_breakdown_fields` on trend/bar chart config schemas (regenerated OpenAPI + Kotlin).
- **Backend**: pure `maskBreakdownNames` helper + the resolver applies it to trend histories and bar buckets.
- **Frontend**: "Mask on public boards" per-field checkboxes in the Add-to-dashboard and Update-chart flows (kept out of the live explorer controls); `buildWidgetFromState` persists them.
- **Docs**: `dashboard.md`.

## Testing

- `maskBreakdownNames` unit tests: single/multi-field, stable sort order, >26 labels (AA…), delimiter fallback, bijection.
- Resolver integration test: seeds a broken-down activity type with `masked_breakdown_fields` and asserts the public output is `A`/`B` and contains **no** real value.
- `chartWidgetMaskedFields` unit test.
- `pnpm check` clean on web + backend; full web suite green (476).


fiddur (commented):
✅ Approved

Solid, well-scoped feature. The masking is enforced server-side in `resolveDashboardData` (the documented public-share security boundary), so real values never leave the server for public viewers, while the authenticated `chart-data-router`/trend endpoints keep returning real values for the owner's own live view — matches the doc's promise.

The core `maskBreakdownNames` logic is sound:
- `indexToLabel` is a correct spreadsheet-style bijection (verified A…Z, AA, BA).
- Per-field masking is injective over the actual series names — two distinct composites differ in at least one positional part, masked positions use an injective label map and unmasked positions keep their (distinct) real values, so the positional `' / '` join stays distinct. That's what makes `breakdown_histories[mask(name)]` and the `Object.fromEntries(...)` rekey in the bar path safe from silently dropping series on collision.
- The `splitsCleanly` guard correctly degrades to whole-label masking (privacy-first) when a value contains the delimiter, and the empty-map / identity-fallback handles the no-masking case without changing existing behavior.

Standards are followed: schema added in api-spec with `.meta`, shared logic in `services/`, snake_case throughout, functional style, and good unit + integration coverage (the integration test asserting `JSON.stringify(data)` contains neither `anaerobic` nor `zone2` is a nice belt-and-suspenders leak check). Generated api-spec artifacts (dist/openapi/kotlin) aren't tracked in git, so their absence from the diff is expected — CI rebuilds them.

1. Labels are assigned in sorted order of the real values, so a public viewer can infer the *alphabetical ordering* of the hidden values (e.g. "series A sorts before series B"). This is the standard trade-off for stable labels/colours across refreshes and is fine, but worth being aware of as residual signal.
2. `sanitizeConfig` forwards `masked_breakdown_fields` (and `breakdown_fields`) to the public viewer. These are field *names*, not values, and `breakdown_fields` was already public pre-PR, so there's no new value leak — just noting the config still advertises which field is masked.

apps/backend/src/services/shared-dashboard-data.integration.test.ts
```diff
@@ -185,6 +185,78 @@ describe('resolveDashboardData integration', () => {
...
+              type: 'bar_chart',
+            },
+          ],
+        },
+      ],
+      version: 1,
+    })
+
+    const trend = data['trend']
+    if (trend.type === 'trend_chart' && trend.data) {
```
fiddur, line 246: These `if (trend.type === 'trend_chart' && trend.data)` / `if (bar.type === 'bar_chart' && bar.data)` guards make the masking assertions vacuous if the resolver ever returns `data: null` or an empty breakdown — the test would still pass green while verifying nothing. For a privacy feature it's worth asserting the branch is actually entered first, e.g. `expect(trend.type).toBe('trend_chart')` and `expect(trend.data).not.toBeNull()` (or assert `breakdown_series` is non-empty) before narrowing, so a regression that drops the breakdown can't hide here.

fiddur (commented):
✅ Approved

Solid, privacy-first implementation. The masking is applied entirely server-side in `resolveDashboardData`, which is only reachable via the public shares endpoint (`/public/:username/:slug`), so the owner's own live view (which goes through `getTrend`/`getChartData` directly) still shows real values — matching the documented behaviour.

Correctness checks that hold up:
- `maskBreakdownNames` returns a map covering every entry of `result.breakdown_series`, and the per-bucket series keys are always a subset of `breakdown_series` (built from the same `seriesSet` in `queryActivityTypeBreakdown`) and trend `breakdown_histories` keys equal `breakdown_series`. So when masking is active the `nameFor.get(name) ?? name` identity fallback never fires for a real name — no value leaks.
- The per-field masking is a bijection (distinct composites always map to distinct masked composites), so `Object.fromEntries` on bucket series won't collapse two series, and the legend stays consistent with the bucket keys.
- The `splitsCleanly` guard falls back to whole-label masking when a value contains the ` / ` delimiter — over-masks rather than leaks, which is the safe direction.
- Backward compatible: boards without `masked_breakdown_fields` resolve to an empty map (no masking).

Schema (`masked_breakdown_fields` on both trend and bar configs), docs, and unit/integration tests are all updated. Generated api-spec artifacts are gitignored, so no regenerated files are expected in the diff. Frontend state handling (pre-fill on expand, filter-to-breakdown-fields on save) is sound.

One minor test-robustness nit inline; not blocking.

apps/backend/src/services/breakdown-mask.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/** Map distinct values (in sorted order) to stable positional labels. */
+function labelMap(values: Iterable<string>): Map<string, string> {
+  const distinct = [...new Set(values)].sort()
+  return new Map(distinct.map((v, i) => [v, indexToLabel(i)]))
+}
+
+/**
+ * Build a map from each real (composite) series name to its masked form, given
+ * the ordered breakdown fields and the subset to mask. Returns an empty map
```
fiddur, line 43: Non-blocking: sorting the real values before lettering means the A/B/C order reveals the alphabetical order of the hidden values. The literal strings are gone (good), but for a small guessable universe (e.g. two partners) a viewer can reconstruct label→value by sorting their guesses. Determinism is needed for stable labels, but a per-board seeded shuffle (e.g. hash of slug + value) would keep labels stable without leaking order. Fine to leave as a documented limitation.

fiddur (commented):
✅ Approved

Solid, well-scoped feature with good test coverage. Masking is applied only in `resolveDashboardData`, which is called exclusively by the unauthenticated `/public/:username/:slug` route, so the owner's own authenticated view still shows real values (matches the docs). I verified the important correctness points:

- **No key escapes masking.** In `getChartData`, each bucket's `series` keys are added to `seriesSet`, so bucket keys are always a subset of `breakdown_series`; `maskBreakdownNames(result.breakdown_series, …)` therefore covers every key the buckets can contain — no unmapped key leaks through unmasked.
- **Delimiter-ambiguity fallback is privacy-first.** When a value contains `' / '` the composite name can't be split per field, so the code masks the whole label rather than risk leaking the intended-hidden value. Good call, and covered by a test.
- **Per-field masking is a bijection**, so the `breakdown_histories` object and per-bucket `Object.fromEntries` remaps can't collide.
- Schema field is optional (backward compatible), snake_case, and inferred via `z.infer`; generated `dist`/`generated/` are untracked so CI rebuilds them (no stale-artifact concern). `indexToLabel` spreadsheet rollover math checks out against the tests.

One thing worth considering (non-blocking): labels are assigned in **sorted order of the real hidden values** (`[...new Set(values)].sort()`). The literal strings never leave the server as the docs promise, but the A/B/C ordering encodes the alphabetical order of the hidden values. For a small, guessable universe (e.g. two partner names), a public viewer can reconstruct the mapping by sorting their guesses. If the threat model cares about that, a per-board deterministic shuffle (seeded by e.g. the slug) would give stable-but-non-order-revealing labels; otherwise it's a reasonable, documented limitation.

apps/backend/src/services/breakdown-mask.ts
fiddur: Good point — documented as an accepted limitation in the breakdown-mask.ts module comment (sorting only exposes ordering, never the values; seeded-shuffle alternative noted). Resolving.

fiddur (commented):
✅ Approved

Solid, privacy-first feature with good test coverage and docs. I traced the masking end-to-end and it holds up:

- **No leak via config** — `sanitizeConfig` exposes `breakdown_fields`/`masked_breakdown_fields`, but those are field *names*, not values, so nothing sensitive escapes through the returned config.
- **`resolveDashboardData` is only wired to the public shares route** (`public-shares-router.ts`), confirming the docs' claim that the owner still sees real values on their authenticated live view.
- **The `mask = (name) => nameFor.get(name) ?? name` identity fallback can't leak**: `maskBreakdownNames` builds a complete map over every entry in `result.breakdown_series`, and both `getTrend` and `getChartData` produce bucket/history series keys that are a subset of `breakdown_series` (all derived from the same sorted set, joined with the same `' / '` delimiter the mask splits on).
- **Bijection is preserved**, so rebuilding `breakdown_histories` and per-bucket `series` never collides keys — and the test asserts it.
- **Delimiter-in-value safety**: any split-count mismatch falls back to whole-label masking (over-masks rather than leaks), which is the right default.
- `indexToLabel` rollover (A–Z, AA, AB…) is correct, and generated api-spec artifacts are gitignored so the added optional field needs no committed regeneration.

The alphabetical-ordering residual signal is a real (minor) information leak but is explicitly documented in the module header and the sharing docs, and exposes no actual value — acceptable as scoped. Nothing blocking.

Note: I could not execute the test suite in this worktree (deps not installed / sandboxed npm cache is read-only); CI covers that.

PR #833, opened 2026-07-01
fix(challenges): correct public view end date + show joiners immediately (#828, #827)

fiddur:
Followups to #825 on the public challenge view (`/u/:username/:slug`).

## #828 — end date off-by-one
The meta line rendered the exclusive `end_ts` (local midnight *after* the last day) as the inclusive last day, so a challenge entered as **6/22 – 6/29** displayed **6/22 – 6/30**, contradicting both the create form and the race chart's x-axis. Now renders `end_ts − 1ms` — the actual last competing day.

## #827 — joiner invisible until cache expires
After **Join**, the standings refetch hit the same `/standings` URL, which sends `Cache-Control: public, max-age=60`, so an HTTP/proxy cache could serve the **pre-join member list**. The public endpoint deliberately ignores `?refresh` (unauthenticated — would allow outbound-fetch amplification, see #820), so instead `fetchPublicChallengeStandings` takes a `bustCache` option that appends a unique `_` query param (never-before-seen URL slips past HTTP/proxy caches; backend ignores the unknown param and computes the member list fresh). The post-join refetch uses it and seeds the query cache via `setQueryData`, so no stale refetch races. The **Join** button also flips to **✓ Joined**.

## Also — race chart renders from a single bucket / single member
The chart showed "Insufficient data for chart" a few hours into a challenge with only one member: a daily-bucketed window yields one point on day one, and `TrendLineChart` needs ≥2 points to draw a line. The cumulative series is now modelled as a proper race — every member starts at **0 on the start line**, and each running total is plotted at its bucket's **end** — so a drawable rising line appears as soon as there is one bucket, for any member count. Pure `toCumulativeSeries`/`bucketEnd` helpers extracted to `race-series.ts` with unit tests.

## Testing
- `pnpm check` — green (0 errors).
- `pnpm vitest run src/pages/Challenges` — 8 passing (incl. new race-series tests).

Closes #828
Closes #827

### DIFF
--- a/apps/web/src/pages/Challenges/PublicChallenge.tsx
+++ b/apps/web/src/pages/Challenges/PublicChallenge.tsx
@@ -81,10 +81,14 @@ export function PublicChallenge({
 
       <p class="challenge-view-meta">
         {challenge.spec.pattern} · {challenge.spec.aggregation} ({challenge.spec.unit}) ·{' '}
-        {new Date(challenge.start_ts).toLocaleDateString()} –{' '}
+        {/* Render both dates in the timezone the range was chosen in, so they read
+            exactly as entered regardless of the viewer's browser locale/timezone. */}
+        {new Date(challenge.start_ts).toLocaleDateString(undefined, { timeZone: challenge.timezone })} –{' '}
         {/* end_ts is the exclusive window end (midnight after the last day); step back
             one ms to render the inclusive last competing day. */}
-        {new Date(new Date(challenge.end_ts).getTime() - 1).toLocaleDateString()}
+        {new Date(new Date(challenge.end_ts).getTime() - 1).toLocaleDateString(undefined, {
+          timeZone: challenge.timezone,
+        })}
       </p>
 
       <div class="challenge-view-actions">
--- a/apps/web/src/pages/Challenges/PublicChallenge.tsx
+++ b/apps/web/src/pages/Challenges/PublicChallenge.tsx
@@ -3,31 +3,18 @@
  * resolves to a challenge. Shows a cumulative "race" chart (one line per member)
  * and a leaderboard, plus join actions. Rendered without app chrome.
  */
-import type { ChallengeStanding, PublicChallenge as PublicChallengeData } from '@aurboda/api-spec'
+import type { PublicChallenge as PublicChallengeData } from '@aurboda/api-spec'
 
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { useState } from 'preact/hooks'
 
-import { type LineSeriesData, TrendLineChart } from '../../components/charts/TrendLineChart'
+import { TrendLineChart } from '../../components/charts/TrendLineChart'
 import { fetchPublicChallengeStandings, joinChallengeByUrl } from '../../state/api'
 import { auth } from '../../state/auth'
+import { toCumulativeSeries } from './race-series'
 import './style.css'
 
 const COLORS = ['#8b5cf6', '#10b981', '#3b82f6', '#f59e0b', '#ef4444', '#ec4899', '#14b8a6', '#a855f7']
-
-/** Build a cumulative (running-total) series from a member's per-bucket values. */
-const toCumulativeSeries = (standing: ChallengeStanding, color: string): LineSeriesData => {
-  const sorted = [...standing.buckets].sort((a, b) => a.bucket_start.localeCompare(b.bucket_start))
-  let running = 0
-  return {
-    color,
-    data: sorted.map((b) => {
-      running += b.value
-      return { date: b.bucket_start, value: running }
-    }),
-    name: standing.display_name,
-  }
-}
 
 const shortHost = (identityBaseUrl: string): string => {
   try {
@@ -79,8 +66,9 @@ export function PublicChallenge({
 
   const standings = (standingsQuery.data ?? []).filter((s) => s.status === 'active')
   const series = standings
-    .map((s, i) => toCumulativeSeries(s, COLORS[i % COLORS.length]))
-    .filter((s) => s.data.length > 0)
+    .map((s, i) => toCumulativeSeries(s, COLORS[i % COLORS.length], challenge.start_ts, challenge.spec.bucket_size))
+    // Keep members with at least one real bucket (the start-line point alone is length 1).
+    .filter((s) => s.data.length > 1)
 
   return (
     <div class="dashboard public-dashboard public-challenge">
--- a/apps/web/src/pages/Challenges/PublicChallenge.tsx
+++ b/apps/web/src/pages/Challenges/PublicChallenge.tsx
@@ -11,7 +11,7 @@ import { useState } from 'preact/hooks'
 import { TrendLineChart } from '../../components/charts/TrendLineChart'
 import { fetchPublicChallengeStandings, joinChallengeByUrl } from '../../state/api'
 import { auth } from '../../state/auth'
-import { toCumulativeSeries } from './race-series'
+import { formatDateInZone, toCumulativeSeries } from './race-series'
 import './style.css'
 
 const COLORS = ['#8b5cf6', '#10b981', '#3b82f6', '#f59e0b', '#ef4444', '#ec4899', '#14b8a6', '#a855f7']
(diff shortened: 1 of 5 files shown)


apps/web/src/pages/Challenges/PublicChallenge.tsx
```diff
@@ -87,15 +81,19 @@ export function PublicChallenge({
 
       <p class="challenge-view-meta">
         {challenge.spec.pattern} · {challenge.spec.aggregation} ({challenge.spec.unit}) ·{' '}
-        {new Date(challenge.start_ts).toLocaleDateString()} – {new Date(challenge.end_ts).toLocaleDateString()}
+        {new Date(challenge.start_ts).toLocaleDateString()} –{' '}
+        {/* end_ts is the exclusive window end (midnight after the last day); step back
+            one ms to render the inclusive last competing day. */}
+        {new Date(new Date(challenge.end_ts).getTime() - 1).toLocaleDateString()}
```
fiddur: The `-1ms` fix correctly removes the off-by-one (previously showed the day after the last competing day). Note it renders in the *browser* locale/timezone via `toLocaleDateString()`, while the range was chosen in `challenge.timezone` — so for viewers whose timezone differs enough from the challenge's, the day can still be off. The unchanged `start_ts` render on the line above has the same caveat. Non-blocking, but passing `{ timeZone: challenge.timezone }` to both would make them exact.

fiddur (commented):
✅ Approved

Clean, well-scoped fix for #828/#827. The cumulative-series logic is extracted from `PublicChallenge.tsx` into a pure, testable `race-series.ts` module with unit tests — nicely aligned with the project's "testable without heavy mocking" preference.

Verified the key claims:
- `bucketEnd` advances in UTC (`setUTCDate`/`setUTCMonth`), which matches the backend bucketing (`date_trunc(..., AT TIME ZONE 'UTC')` in `chart-data.ts`), so plotting each cumulative total at its bucket's exclusive end is consistent with how `bucket_start` is produced.
- The seeded `{ date: start_ts, value: 0 }` start-line point plus the `.filter(s => s.data.length > 1)` guard correctly makes a single real bucket render (TrendLineChart needs ≥2 points), which is the intended fix.
- The cache-buster is sound: `public-shares-router.ts` reads only `req.params`, so the extra `_` query param is ignored by the backend, and the endpoint sends `Cache-Control: public, max-age=60`, so busting it after a join is justified. `setQueryData` then seeds the fresh result without racing a stale refetch. The new `options` param is defaulted, so the existing `queryFn` caller is unaffected.
- Test expectations were checked by hand and compute exactly as asserted (I couldn't run vitest — deps aren't installed in this worktree).

One non-blocking note (see inline): the end-date display uses `toLocaleDateString()` (browser locale) on a UTC instant, while the range was chosen in `challenge.timezone`. The `-1ms` fix correctly removes the off-by-one for typical viewers, but the same class of timezone/browser mismatch remains (and already affects the unchanged `start_ts` render). Not worth blocking, but using `challenge.timezone` would make both dates exact.

apps/web/src/pages/Challenges/PublicChallenge.tsx
fiddur: Fixed in b24a0dc — both the start and end dates now render with `{ timeZone: challenge.timezone }`, so they read exactly as entered regardless of the viewer's browser timezone.

apps/web/src/pages/Challenges/PublicChallenge.tsx
```diff
@@ -87,15 +81,23 @@ export function PublicChallenge({
 
       <p class="challenge-view-meta">
         {challenge.spec.pattern} · {challenge.spec.aggregation} ({challenge.spec.unit}) ·{' '}
-        {new Date(challenge.start_ts).toLocaleDateString()} – {new Date(challenge.end_ts).toLocaleDateString()}
+        {/* Render both dates in the timezone the range was chosen in, so they read
+            exactly as entered regardless of the viewer's browser locale/timezone. */}
+        {new Date(challenge.start_ts).toLocaleDateString(undefined, { timeZone: challenge.timezone })} –{' '}
```
fiddur: Minor/non-blocking: `toLocaleDateString(undefined, { timeZone: challenge.timezone })` throws `RangeError: Invalid time zone specified` if `timezone` isn't a valid IANA zone. The normal path is safe (it's set from `Intl.DateTimeFormat().resolvedOptions().timeZone` on create), but the backend create schema only validates `z.string().min(1)`, so a manually-crafted challenge could crash this whole render. Consider validating the timezone server-side (or guarding here).

fiddur (commented):
✅ Approved

Solid, well-scoped fix for both issues. I verified the key correctness points:

- **`bucketEnd` / cumulative series** — advancing bucket ends with UTC arithmetic is consistent with the backend, which buckets via `date_trunc(... AT TIME ZONE 'UTC')` (`apps/backend/src/services/chart-data.ts`), so bucket_starts are UTC-midnight/1st-of-month/Monday-aligned. The `1M` branch is safe because monthly bucket_starts are always day-1 (no `setUTCMonth` overflow). Seeding a zero start-line point plus the `data.length > 1` filter correctly keeps members with ≥1 real bucket while dropping empty ones. Tests match the implementation exactly.
- **End-date fix** — `end_ts` is documented as the exclusive window end, so stepping back 1ms and rendering in `challenge.timezone` yields the inclusive last competing day; matches the schema and the earlier partial fix.
- **Show joiners immediately** — the cache-buster comment is accurate: the standings endpoint (`public-shares-router.ts:102`) only reads `username`/`slug`, sends `max-age=60`, and deliberately does not honor `refresh`, so a unique query param safely bypasses the HTTP/proxy cache. And `getChallengeStandings` always re-reads the member list fresh, with a newly-registered remote member (`last_fetched_at === null`) fetched immediately rather than served from the 5-min TTL cache — so the just-joined member really does appear. `setQueryData` seeding avoids a stale refetch race, and the optional `bustCache` param is backward-compatible with the other caller.

One low-severity robustness note (non-blocking): the new `toLocaleDateString(undefined, { timeZone: challenge.timezone })` will throw `RangeError` if `timezone` is not a valid IANA zone. In the normal flow it's sourced from the browser's `Intl.DateTimeFormat().resolvedOptions().timeZone`, so it's always valid; but the create schema only validates `z.string().min(1)`, so a hand-crafted API request could store garbage and crash this render for all viewers. Worth tightening timezone validation on the backend at some point.

apps/web/src/pages/Challenges/PublicChallenge.tsx
fiddur: Fixed in 417bb06 — added `formatDateInZone`, which tries the challenge timezone and falls back to the viewer's locale on an invalid zone, so a crafted `timezone` can no longer crash the render. Covered by a unit test (invalid zone → no throw, locale fallback). Server-side IANA validation would be good defense-in-depth but is out of scope for this fix.

fiddur (commented):
✅ Approved

Solid, focused bug-fix PR addressing #828 (public-view end date) and #827 (joiners visible immediately), with good test coverage and a clean extraction of chart logic into `race-series.ts`.

What I verified:
- **End-date fix**: `end_ts` is stored exclusive (midnight after the last day), and rendering `end_ts − 1ms` in `challenge.timezone` correctly yields the inclusive last competing day. Rendering both endpoints via `formatDateInZone` in the stored IANA zone is the right fix for locale/timezone drift.
- **`formatDateInZone` guard**: the `try/catch` fallback is warranted — `timezone` is only validated as a non-empty string, so a crafted challenge could otherwise throw `RangeError` and blank the whole public render. Covered by a test.
- **`bucketEnd` UTC arithmetic is correct here**: the backend buckets with `date_trunc(… AT TIME ZONE 'UTC')` (see `chart-data.ts`), so bucket starts are UTC-aligned; advancing with `setUTCDate`/`setUTCMonth` matches those boundaries exactly, and the `1M` case is safe since the day-of-month is always 1 (no month-length drift).
- **Cache-buster**: the `_: Date.now()` param claim checks out — the `/public/:username/:slug/standings` route has no query validation and ignores unknown params, so it passes through safely past the `max-age=60` HTTP cache. Seeding via `setQueryData` correctly marks the data fresh for `staleTime`.
- **`length > 1` filter** correctly matches the new seeded start-line point, and the tests trace through correctly.

Could not run the web unit tests (worktree has no installed `node_modules`), but I traced each test case by hand against the schemas and the fixtures match `ChallengeStanding`/`ChartDataBucket`.

Minor, non-blocking observation: after `setQueryData` seeds the fresh (post-join) list, a later window-focus refetch once the 60s `staleTime` lapses uses the non-busting `queryFn`, which could momentarily be served a stale list from the HTTP cache. In practice the HTTP cache TTL and `staleTime` both being 60s makes this a narrow window, and the `joined` badge is independent of it — so not worth changing.

LGTM.

PR #836, opened 2026-07-01
fix(web): list public challenges on user profile page

fiddur:
## Problem

Fixes #835. A user's public profile page at `/u/:username` only listed their public **dashboards**. Public **challenges** (e.g. `/u/fiddur/8MsiUjYVOA`) were not shown anywhere on the profile, so they were undiscoverable.

## Fix

The backend already returns the user's public challenges in the same `{ name, share_url, slug }` shape as dashboards from `GET /public/:username/dashboards` — the frontend simply wasn't rendering them. This change adds a **Challenges** section to `PublicProfile` alongside the existing **Dashboards** section, each with its own heading and empty-state message, linking to `/u/:username/:slug`.

- `apps/web/src/pages/PublicProfile/index.tsx` — render `query.data.challenges` in a second list; extract a shared `renderItem` helper; wrap each list in a titled `<section>`.
- `apps/web/src/pages/PublicProfile/style.css` — generalize `.public-dashboard-list` → `.public-item-list` (used by both lists) and add `.public-section` heading styles, including dark-mode.

## Testing

- `pnpm check` (tsgo typecheck + oxlint): 0 errors.
- `pnpm test`: 474 tests pass.

No backend changes were needed. The repo has no component-render test harness (all existing tests cover pure logic), so no rendering test was added, consistent with the established pattern.

https://claude.ai/code/session_01XYaK7FCGC6GJzqQNJ4QzVk

### DIFF
--- a/apps/web/src/pages/PublicProfile/index.tsx
+++ b/apps/web/src/pages/PublicProfile/index.tsx
@@ -1,6 +1,6 @@
 /**
  * PublicProfile - a user's public page at /u/:username, listing their public
- * shared dashboards. Unauthenticated; rendered without app chrome.
+ * shared dashboards and challenges. Unauthenticated; rendered without app chrome.
  */
 import { useQuery } from '@tanstack/react-query'
 import { useRoute } from 'preact-iso'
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped frontend change that adds a public **Challenges** list to the profile page alongside the existing dashboards list.

Verified against the backend and shared types:
- `challenges` is part of `PublicProfileResponse` in `@aurboda/api-spec` and is already populated by `public-shares-router.ts` (`GET /public/:username/dashboards`), so the new section renders real data rather than being permanently empty.
- Challenge links (`/u/:username/:slug`) resolve through `PublicResource`, which already handles both dashboards and challenges, so the links work end to end.
- The `renderItem` helper's inline `{ name; slug }` type is a structural subtype of `PublicDashboardListItem`, so `dashboards.map(renderItem)` / `challenges.map(renderItem)` type-check.
- The CSS rename `.public-dashboard-list` → `.public-item-list` was applied consistently (no lingering references anywhere), and dark-mode styling was added for the new `h2`.
- The doc comment was updated to mention challenges.

No tests were added, which matches this repo's convention: web page tests here cover pure logic units only, not presentational component rendering. Nothing blocking.

Issue #831, opened 2026-07-01
Feature: Federated activity feed — post activities to a user feed with selectable metrics


fiddur:
## Summary

Add a **federated, Strava-like activity feed** so a user can publish an activity
(exercise, sleep, meditation, custom type, …) to their public/follower feed, choosing
per-post **which metrics to include** (e.g. heart rate, stress, pace, distance, HR-zone
minutes). Feeds work **cross-installation** over ActivityPub, reusing the actor identity
and `/.well-known` + `/u/<user>` discovery we already expose for challenges and shared
dashboards. A later iteration adds **auto-share rules** (e.g. "share runs longer than
15 min").

This is the social-distribution layer that complements our existing analysis features,
without locking data into a closed network — the same separation-of-concerns thesis as
FitPub / Open Pace.

## Motivation

- We already federate **challenges** (`join_challenge` resolves an absolute
  `/u/<user>/<id>` URL on this *or another* Aurboda instance) and publish **shared
  dashboards** at public slugs. We have actor identity and cross-instance discovery in
  production; there is no user-facing "here's what I did today" stream that others can
  follow.
- ActivityPub (AS2 + WebFinger + HTTP Signatures) is the natural fit: Mastodon and other
  fediverse servers can follow a user's activities, and Aurboda↔Aurboda gets rich,
  structured payloads.
- Prior art validates the shape: **Open Pace** (federated Strava alternative), **FitPub**
  (self-hosted fitness for the fediverse), **Wanderer** (trails). None share a common
  *workout* FEP — each defines its own object extension — so we define ours and lean on
  the general infrastructure FEPs.

## Scope of this issue

**In scope (v1):**

1. Federation core (extracted/generalised — see "Relationship to challenges" below):
   actor object, `outbox`, `inbox`, WebFinger, HTTP Signatures, delivery queue.
2. Manual "Share activity" action from an activity detail view.
3. Per-post **metric selection** UI + payload (choose which metrics/derived stats to
   include; sensible privacy defaults).
4. Optional **static map image** for GPS activities (rendered PNG attachment).
5. AS2 object model for an activity (`Exercise`) with a Mastodon-compatible fallback.
6. `FEDERATION.md` documenting what we support (FEP-67ff).

**Deferred to follow-ups (linked, not built here):**

- Auto-share rules ("runs > 15 min") — see "Auto-share (follow-up)".
- Inbound social interactions (Like/Announce/Reply handling beyond storing them).
[...]

PR #838, opened 2026-07-01
feat(feed): activity feed persistence, share API & public series endpoint (#831)

fiddur:
## Summary

First slice of the **federated activity feed** ([#831](https://github.com/fiddur/aurboda/issues/831)). It lands the **storage + public-read foundation** the rest of the feature builds on, deliberately scoped ahead of ActivityPub delivery and image rendering so it can merge and be reviewed on its own.

Scope was agreed up front: persistence + share API + the public `/series` endpoint (the privacy boundary). It stays **AP-agnostic** — no crypto, no S2S delivery, no image rendering yet.

## What's included

- **api-spec** `schemas/feed.ts`: `ShareActivityBody`, `UpdateFeedPostBody`, `FeedPost`, and the public `PublicSeriesQuery`/`PublicSeriesResponse`. All snake_case per convention. (Not added to `openapi.ts`/Kotlin — consistent with how challenges & shared dashboards are handled; this is a web/federation surface, not part of the Android app.)
- **DB**: `feed_posts` table (declarative schema) + `db/feed.ts` with CRUD and `findCoveringSharedSeriesWindow`, the data-driven authorization lookup.
- **Service**: `resolvePublicSeries` (dependency-injected: window finder + bucketed query). Clamps the range to the shared activity's window, floors bucket granularity to 5s, projects only the requested metric, and drops per-measurement timestamps.
- **Routes**: owner-facing `/feed` (list / share / edit / delete) and the public, unauthenticated `/public/:username/series`, mounted **before** the generic `/public/:username/:slug` resolver.
- **MCP parity**: `list_feed`, `share_activity`, `update_feed_post`, `delete_feed_post`.
- **Refactor**: extracted the duplicated `isMissingDatabase` pg-error predicate into a shared `db/pg-errors.ts` (was copied in public-shares + challenge-data routers).
- **Docs**: `docs/features/feed.md` documenting the model and the series privacy boundary; cross-linked from `sharing.md`.

## Privacy boundary (the important bit)

The public series endpoint has **no auth token**, so scoping is the entire privacy boundary. A metric resolves **only** when a `public`/`unlisted` feed post shared it **as a series** (`series_metrics`, a separate opt-in from the scalar summary) for a non-deleted, bounded activity whose window **covers** the request. Scalar-only shares, `followers`-only posts, unshared metrics, and out-of-window ranges all **404**. Fully covered by tests.

## Testing

- 11 feed DB integration tests (CRUD + all series-scoping cases).
- 9 series-service unit tests (auth gating, clamp, bucket floor, projection).
- Full backend suite green locally: **2434 passed** (160 files). Typecheck + lint clean.

## Deferred to follow-up PRs

ActivityPub core (actor/keypair/WebFinger/HTTP signatures/inbox/outbox/delivery — will evaluate **Fedify** then, per the issue's open question), map/chart image rendering, the "Your feed" web UI, and the auto-share rule engine. The actor/follower/delivery tables are intentionally **not** created here — they'd be dead tables with no writer until the delivery PR.


apps/backend/src/api/rest-routes.ts
```diff
@@ -124,8 +126,12 @@ export const mountRestRouters = ({
   httpd.use(createRawRecordsRouter(authMiddleware))
   httpd.use('/dashboard', createDashboardRouter(authMiddleware))
   httpd.use('/shared-dashboards', createSharedDashboardsRouter(authMiddleware, webHost))
+  httpd.use('/feed', createFeedRouter(authMiddleware))
   httpd.use('/challenges', createChallengesRouter(authMiddleware, webHost, apiBaseUrl))
   httpd.use(createChallengeDataRouter())
+  // Public feed series must be mounted before the generic /public/:username/:slug
+  // resolver so `series` is not matched as a share slug.
+  httpd.use(createFeedPublicRouter())
```
fiddur, line 134: Mounting this before the generic `/public/:username/:slug` resolver effectively reserves `series` as a slug: any existing shared dashboard or challenge whose slug is `series` becomes unreachable via `/public/:username/series`. This is deliberate and documented, so non-blocking — but if that collision matters, consider rejecting `series` (and other reserved words) at dashboard/challenge slug creation to make the reservation explicit rather than implicit in mount order.

fiddur (commented):
✅ Approved

Solid, well-scoped foundation for the activity feed. The privacy boundary for the unauthenticated `/public/:username/series` endpoint is carefully built and reads correctly:

- `findCoveringSharedSeriesWindow` gates on all the right conditions (metric in `series_metrics`, `public`/`unlisted` visibility, non-deleted activity with a bounded window that fully covers the request), and the NULL `activity_id` case is naturally excluded by the JOIN.
- `resolvePublicSeries` clamps the range to the activity window (defence-in-depth over coverage), floors the bucket to a server minimum, projects only the requested metric, and drops per-measurement timestamps.
- All DB access is per-user-database scoped, so there is no cross-user leakage, and `getActivityById` excludes soft-deleted activities so sharing a deleted activity 404s.

Conventions are followed well: api-spec is the single source of truth (schemas exported, snake_case throughout), `typedRouter()` with fully typed generics, MCP/REST parity via `.shape`, shared business logic in a DI-friendly service that is unit-tested without a DB, plus integration tests covering CRUD and every branch of the authorization window lookup. The `isMissingDatabase` consolidation into `db/pg-errors.ts` is a nice DRY cleanup, and docs are added.

1. Mounting the public series route before the generic `/public/:username/:slug` resolver effectively reserves `series` as a slug — a pre-existing shared dashboard or challenge with slug `series` would be shadowed. It's deliberate and documented; consider rejecting `series` as a reserved slug at dashboard/challenge creation if that collision is a concern.
2. The new feed endpoints aren't added to `packages/api-spec/src/openapi.ts` path definitions, but that's consistent with the existing partial coverage there (challenges, shared-dashboards, correlations, etc. are also absent), so it's not a regression.

apps/backend/src/api/rest-routes.ts
fiddur: Slugs are server-generated random ~9-char base64url tokens (`randomToken(7)`); users never choose them, so a slug can never equal the fixed literal `series`. No reserved-slug handling needed. Deliberate and documented in docs/features/feed.md.

apps/backend/src/db/feed.ts
```diff
@@ -0,0 +1,167 @@
...
+      input.include_chart,
+    ],
+  )
+  return mapFeedPost(result.rows[0])
+}
+
+export const listFeedPosts = async (user: string): Promise<FeedPostRecord[]> => {
+  const result = await query<FeedPostRow>(
+    user,
+    `SELECT ${FEED_POST_COLUMNS} FROM feed_posts ORDER BY created_at DESC`,
```
fiddur: `ORDER BY created_at DESC` has no tiebreaker, so posts sharing a `created_at` (microsecond collision on two rapid inserts) have an undefined relative order. Minor, but it can make `listFeedPosts` non-deterministic and could theoretically flake the `lists posts newest-first` integration test. Consider a stable secondary key, e.g. `ORDER BY created_at DESC, id DESC`.

fiddur (commented):
✅ Approved

Solid, well-structured foundation for the activity feed. The privacy boundary on the unauthenticated `GET /public/:username/series` endpoint is carefully designed and holds up under scrutiny: authorization is data-driven (`metric` must be in `series_metrics` of a `public`/`unlisted` post whose non-deleted, bounded activity *fully covers* the requested window), the range is clamped to the activity window, the bucket is floored server-side (which also prevents a `0s` bucket from reaching the divide-by-`bucketMs` path), only the requested metric's stats are projected, per-measurement timestamps are dropped, and `isValidUsername` (`/^[a-z][a-z0-9_]{2,30}$/`) prevents DB-name injection. Coverage requiring a *subset* request plus the single-activity `JOIN` means neither the whole-day range nor cross-activity stitching can widen exposure. API/MCP parity is maintained, docs are added, all new `db/feed.ts` functions have integration tests, and the schema is registered in both `tableCreationOrder` and the test cleanup list.

Verified locally: `@aurboda/api-spec` builds, backend `tsgo --noEmit` passes with no errors, and `feed-series.test.ts` (9 tests) passes. The `req.user!` lint warnings match the established pattern across 26 routers and oxlint only fails on errors, so CI is unaffected.

A few minor, non-blocking observations:
- Index keys `feed_posts_created_index` / `feed_posts_series_index` use the singular `_index` suffix rather than the established plural `_indexes` convention. Functionally harmless — `migrateSchema` builds `existingTableNames` from `information_schema.tables` (indexes never appear there), so the `!existingTableNames.has(...)` branch always re-runs these `CREATE INDEX IF NOT EXISTS` statements — but it bypasses the intended `key.endsWith('_indexes')` "always run index" path, so worth aligning for consistency.
- A non-UUID `:postId` in PATCH/DELETE reaches `WHERE id = $1` and surfaces as a Postgres `22P02` (→ 500) rather than 404, but this matches existing routers (e.g. `shared-dashboards-router.ts`), so it's not PR-specific.

apps/backend/src/db/feed.ts
fiddur: Fixed: `listFeedPosts` now orders by `created_at DESC, id DESC` for deterministic ordering on a `created_at` collision.

fiddur (commented):
✅ Approved

Solid, well-tested foundation for the activity feed. The security-critical part — the unauthenticated `GET /public/:username/series` endpoint — is correctly scoped: `findCoveringSharedSeriesWindow` requires a *single* feed post to both list the metric in `series_metrics` AND reference an activity (public/unlisted, non-deleted, bounded) whose own window covers the request, so there's no cross-activity metric probing. The cheap GIN-indexed authorization check runs before the expensive bucketed query, the range is clamped to the activity window, the bucket is floored server-side, and only the requested metric's stats are projected (per-measurement timestamps dropped). This matches the documented privacy model and is well-covered by both the DI'd unit tests (`feed-series.test.ts`) and the integration tests (`feed.integration.test.ts`, including soft-delete/visibility/scalar-vs-series negative cases).

Conventions are followed throughout: `typedRouter()` with fully-typed generics, api-spec schemas as the single source of truth, snake_case everywhere, functional style, normalized `activity_id` soft-reference, MCP/REST parity, and matching docs. The shared `isMissingDatabase` extraction into `pg-errors.ts` (deduplicating the two prior inline copies) is a nice cleanup. Schema wiring (`tableCreationOrder`, `socialTables`, `cleanTestDb`) is consistent with existing tables.

Verified locally: api-spec build ✓, backend `tsgo --noEmit` ✓, feed-series unit tests ✓ (9 passed). Lint shows only `req.user!` non-null-assertion warnings, which match the established convention across the other 26 routers. Integration tests couldn't execute in this environment (no Docker/testcontainers) — not a code issue.

One non-blocking observation left inline about the `Cache-Control` header vs. the docs' "immediately stop resolving" wording.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -0,0 +1,57 @@
...
+      const { bucket, end, metric, start } = req.query
+      try {
+        const result = await resolvePublicSeries(metric, new Date(start), new Date(end), bucket, {
+          findCoveringWindow: (m, s, e) => findCoveringSharedSeriesWindow(username, m, s, e),
+          queryBucketed: (m, s, e, b) => queryMetricsBucketed(username, [m], s, e, b, {}),
+        })
+        if (!result) {
+          return res.status(404).json({ error: 'Not found', success: false })
+        }
+        res.setHeader('Cache-Control', 'public, max-age=60')
```
fiddur: Non-blocking: `public, max-age=60` lets shared browsers/CDNs keep serving a series for up to a minute after the owner deletes the post, flips it to `followers`, or removes the metric from `series_metrics`. That slightly contradicts the docs' claim that these actions "immediately stop the series from resolving." It matches the existing `challenge-data-router` pattern so it's a reasonable accepted tradeoff — just worth softening the docs wording (or dropping to `no-store`/a shorter max-age) since this endpoint exposes high-resolution physiological data.

fiddur: Fixed: the series endpoint now sends `Cache-Control: no-store` so a revoked share (deleted post, flipped to followers, or metric removed from series_metrics) can never be served stale from a shared cache/CDN.

fiddur (commented):
✅ Approved

Solid, well-scoped foundation for feed persistence + share API + the public series endpoint. The privacy-critical logic holds up:

- **Series authorization is correct and well-tested.** `findCoveringSharedSeriesWindow` gates on membership in `series_metrics` (never the scalar `included_metrics`), `public`/`unlisted` visibility, a non-deleted activity with a bounded window, and full coverage of the requested range. `feed.integration.test.ts` exercises each of these boundaries (scalar-only, followers-only, soft-deleted, out-of-window, wrong metric). `resolvePublicSeries` re-clamps to the activity window, floors the bucket, and projects only the requested metric (dropping `first_time`/`last_time`), with good unit coverage.
- **The unauthenticated surface is safely scoped.** `username` flows into the per-user DB connection only after `isValidUsername` (strict `^[a-z][a-z0-9_]{2,30}$` allowlist minus reserved names); both the authorization lookup and the data query use the same `username`; missing DBs map to 404; and `Cache-Control: no-store` keeps revocation immediate. The mount ordering (feed-public before the generic `/public/:username/:slug` resolver) is correct so `series` is not treated as a share slug.
- **Wiring is complete and consistent.** `socialTables` is merged into `createTableStatements` and `tableCreationOrder`, the multi-statement `_indexes` string matches the established pattern, `metricUnits` is re-exported from `schema.ts`, the bucket regex is a subset of what `parseBucketSize` accepts, and `metric` is enum-validated so `metricUnits[metric]`/`queryMetricsBucketed` are safe. REST and MCP reach full parity, follow the typed-router conventions (api-spec response types, Date→ISO serializers, no `req.body` casts), and the shared `isMissingDatabase` predicate is nicely de-duplicated into `pg-errors.ts`.

1. Malformed (non-UUID) `:postId` / activity `:id` path params reach Postgres and surface as a `22P02` error → unhandled 500 rather than a 404/400. This matches the existing pattern in other routers (e.g. `food-items-router.ts`), so it's pre-existing behavior rather than a regression, but worth keeping in mind if you ever add param validation broadly.
2. Because the public feed router owns the literal `/public/:username/series` path, a shared-dashboard slug named exactly `series` would be shadowed. That's the intended trade-off of the documented mount order; just noting it as a reserved word.
3. `no-store` is set only on the 200 path; 404s are uncached by default so this is negligible.

Nothing here blocks. Nice work on the docs and the privacy rationale in the comments.

PR #839, opened 2026-07-01
feat(challenges): adaptive race-chart bucketing with coarse-bucket override

fiddur:
Follow-up to the challenge race chart: a fixed daily bucket meant you couldn't follow **intraday** progress on a short challenge (one point per day), which also caused the "renderable a few hours in" issue.

## What changed
- **`spec.bucket_size` gains `auto` (new default).** The race chart bucket adapts to the window:

  | Window | Bucket |
  |---|---|
  | ≤ 1 day | 5-minute |
  | ≤ 3 days | 15-minute |
  | ≤ 14 days | 1-hour |
  | ≤ 120 days | daily |
  | longer | weekly |

- **Creator override:** a new **Chart detail** selector in the create form lets the creator fix a coarser bucket (**Daily / Weekly / Monthly**) for very long challenges. Only coarse options are offered, so an override can't blow up payloads.
- **Scoring is untouched** — the cumulative total is bucket-size-independent; the bucket only controls chart granularity.
- The resolved size derives **purely from the window**, so every instance computes the same value. It's exposed to viewers as `effective_bucket_size` on the public challenge and drives the client's bucket-end plotting (which now handles `5m/15m/1h`).

## Where it lives
`effectiveBucketSize()` in `challenge-spec.ts`, applied in `resolveMemberSeries` — shared by both the host standings aggregation and the member's own `challenge-data` endpoint — so all members bucket identically without extra coordination.

## Testing
- `pnpm check` — green (0 errors).
- Backend: `challenge-spec.test.ts` (ladder + override + sub-day window) + `chart-data.test.ts` — 31 pass.
- Web: `src/pages/Challenges` — 11 pass (incl. sub-day `bucketEnd`).
- api-spec regenerated (OpenAPI + Kotlin) via `pnpm generate`.

Docs: `docs/features/challenges.md` updated with the granularity behavior.


packages/api-spec/src/schemas/challenges.ts
```diff
@@ -174,6 +197,9 @@ export type RegisterChallengeMemberBody = z.infer<typeof registerChallengeMember
 
 export const publicChallengeSchema = z
   .object({
+    effective_bucket_size: challengeEffectiveBucketSizeSchema.meta({
```
fiddur: Non-blocking: `effective_bucket_size` is added as a **required** field of `publicChallengeSchema`, which joining instances also parse via `fetchChallengeSpec` → `publicChallengeResponseSchema.safeParse`. A new-code joiner fetching a challenge from a host that hasn't deployed this yet will fail the parse and be unable to join — even though the joiner's backend never reads this field (only the frontend rendering does). Marking it `.optional()` (with the render path falling back to a default effective size) would keep cross-version federation working at no cost, matching how `challengeDataResponseSchema` already makes its added fields optional. Given federation is nascent and effectively single-instance today the practical impact is small, so flagging rather than blocking.

fiddur (commented):
✅ Approved

Clean, well-scoped feature. The `auto`→concrete bucket resolution is derived purely from the window, so it stays consistent across every layer: the backend `effectiveBucketSize` resolver feeds both `resolveMemberSeries` (local + the federated `/challenge-data` endpoint) and the `effective_bucket_size` exposed on the public challenge, and the frontend `bucketEnd` advances by the same effective size it receives back — so members' series grids align. All effective sizes (`5m`/`15m`/`1h`/`1d`/`1w`/`1M`) are supported by `getChartData`/`buildBucketExpr`, `VARCHAR(4)` still fits every stored *choice* (`auto`/`1d`/`1w`/`1M`; effective sizes are never stored), and existing challenges keep their stored `1d` (treated as a fixed choice, not `auto`) so no cache/behavior drift on deploy. Federation stays consistent because the joiner copies the `bucket_size` *choice* + window into its participation and resolves independently. I verified locally: `@aurboda/api-spec` builds, backend + web typecheck clean (0 errors), and the new `challenge-spec.test.ts` (3) and `PublicChallenge.test.ts` (6) tests pass. Docs are updated to match. One non-blocking federation note inline; nothing blocking.

packages/api-spec/src/schemas/challenges.ts
fiddur: Fixed in 265f73b — `effective_bucket_size` is now `.optional()`, and the frontend render falls back to daily when absent. A newer joiner can now parse a spec from an older host that omits the field. Thanks for catching the cross-version federation risk.

fiddur (commented):
✅ Approved

Solid, well-scoped change. The adaptive bucketing is threaded consistently end-to-end:

- `effectiveBucketSize` resolves `'auto'` to a concrete size from the window only, and every consumer (public-shares-router for `effective_bucket_size`, and `resolveMemberSeries` via the challenge-data endpoint for each member's buckets) uses the *same* algorithm over the *same* challenge window, so a version-matched fleet stays aligned as claimed.
- The resolved sizes (`5m`/`15m`/`1h`/`1d`/`1w`) are all supported downstream by `buildBucketExpr`/`getChartData` (`date_bin` for sub-hour, `date_trunc` for hour and above), so no unsupported bucket size can reach SQL.
- `bucketEnd`'s `switch` covers every `ChallengeEffectiveBucketSize` value, and the frontend guards a missing `effective_bucket_size` with `?? '1d'` for cross-version hosts. Scoring is unaffected (cumulative total is bucket-size-independent).
- Schema/type layering is clean: `ChallengeBucketSizeChoice` (creator input incl. `auto`) vs `ChallengeEffectiveBucketSize` (rendered), `effective_bucket_size` kept `.optional()` for federation compat, `VARCHAR(4)` still fits `'auto'`, and MCP/REST parity is preserved (`toSpecFields` now takes the full `ChallengeSpec`). Docs thresholds match the code and unit tests match the implementation.

Non-blocking notes:
- I couldn't run the test suites (workspace `node_modules` isn't installed in the review env), so I verified the new tests by tracing them against the implementation rather than executing them — CI should confirm.
- The `?? '1d'` fallback in `PublicChallenge.tsx` will slightly misplace segment ends for an old cross-version host whose fixed `spec.bucket_size` was `1w`/`1M` (it advances by a day instead). Since `challenge.spec.bucket_size` is available on the object, `spec.bucket_size !== 'auto' ? spec.bucket_size : '1d'` would be a more faithful fallback. Purely cosmetic on a rare path (same-version responses always include `effective_bucket_size`), so not blocking.
- Generated api-spec artifacts (YAML/Kotlin/TS) aren't tracked in git, so the schema change needs no committed regen here — just make sure the build step ran.

fiddur (commented):
✅ Approved

Clean, well-scoped change that adds an `auto` chart-bucket mode (adapting granularity to the challenge window) with a fixed-size creator override, plus the resolved `effective_bucket_size` exposed to viewers.

**Verified locally:** api-spec builds, backend + web type-check with 0 errors, and both new test suites pass (`challenge-spec.test.ts` 3/3, `PublicChallenge.test.ts` 6/6). I confirmed the design holds end-to-end:

- Every effective size (`5m`/`15m`/`1h`/`1d`/`1w`/`1M`) is handled by all three consumers — the SQL path (`buildBucketExpr` via `date_bin`/`date_trunc`), the HR-zone JS path (`utcBucketStart`), and the frontend (`bucketEnd`) — so no bucket size can slip through unhandled.
- Federation stays deterministic: host-local and remote members both compute via `resolveMemberSeries` → `effectiveBucketSize` from the same stored `bucket_size` + window, so members' series remain aligned without extra coordination, matching the doc claim.
- `VARCHAR(4)` accommodates `'auto'`, there's no CHECK constraint or row-mapper guard to reject it, and the DB `DEFAULT` change is cosmetic (the API always supplies the value). Existing tests set `bucket_size` explicitly, so they're unaffected by the default flip.
- Making `effective_bucket_size` optional in `publicChallengeSchema` correctly preserves cross-version joins (an older host omitting the field still parses).

Two minor, non-blocking notes (no action required):

1. `PublicChallenge.tsx`'s `challenge.effective_bucket_size ?? '1d'` fallback is essentially defensive/dead in practice, since the public-challenge payload is served by the same-version backend that always sets the field. If you ever want the cross-version fallback to be more accurate, `challenge.spec.bucket_size` (present, and concrete on pre-`auto` hosts) would be a closer proxy than the hardcoded `'1d'`.
2. Terminology: the docs describe fixed sizes as a "coarse-bucket override," but for a very long challenge `auto` already caps at `1w`, so choosing `1d` explicitly actually yields *finer* buckets than `auto`. The override is really "use a fixed size as-is" rather than strictly coarser — a wording nit, not a behavior issue.

PR #841, opened 2026-07-01
fix(challenges): span race chart x-axis over the full challenge timespan (#840)

fiddur:
Closes #840.

## The bug

On the public challenge page, the cumulative "race" chart's x-axis collapsed to the very start of the challenge: `TrendLineChart` derived the x-domain from `d3.extent(data)`, so when members' buckets only covered the beginning (a just-started challenge), every tick rendered the same day label (e.g. all "Jul 01") and the lines stopped well before the right edge.

## The fix

- **`TrendLineChart`**: new optional `xDomain?: [Date, Date]`. When provided, the x-scale **and** the date formatter use it instead of `d3.extent`, so the axis spans the given range regardless of how much bucket data exists. Omitting it keeps the existing `d3.extent` behaviour, so all other callers (Chart explorer, widgets, metric mini-charts) are unaffected.
- **`renderAxes`**: for short spans (≤ 31 days) it now places ~one tick per day (capped ~8) so labels stay distinct — a 3-day challenge shows Jul 01 / 02 / 03 / 04 instead of repeating one day across sub-day ticks. Longer spans keep `.ticks(6)`.
- **`PublicChallenge`**: passes `xDomain={[new Date(challenge.start_ts), new Date(challenge.end_ts)]}` — the full challenge timespan.

## Testing

- `pnpm check` clean (0 errors); existing `Challenges/` + `charts/` suites green (10 tests). The change is additive/opt-in for `TrendLineChart`, so existing single-series and multi-series callers render exactly as before.

## Follow-up (not in this PR)

The issue asks to regenerate `apps/web/public/screenshots/challenge.png` to show the axis spanning the full timespan. There's no screenshot tooling in the repo and a faithful capture needs a running instance with the multi-day challenge data. Best done against aurboda.net once this deploys (or via a seeded local run) — happy to do it as a follow-up.


apps/web/src/components/charts/TrendLineChart.tsx
```diff
@@ -55,14 +61,19 @@ function renderAxes(
   innerHeight: number,
   dateFormat: (date: Date) => string,
 ) {
+  const [start, end] = x.domain()
+  const spanDays = (end.getTime() - start.getTime()) / 86_400_000
+  const xAxis = d3.axisBottom(x).tickFormat((d) => dateFormat(d as Date))
+  // For short spans, place ~one tick per day so labels stay distinct instead of
+  // repeating the same day across sub-day ticks (e.g. a 3-day challenge).
+  const dayStep =
+    spanDays > 0 && spanDays <= 31 ? d3.timeDay.every(Math.max(1, Math.ceil(spanDays / 8))) : null
```
fiddur: Non-blocking: `renderAxes` is shared by all chart call sites, so this per-day tick behaviour now also applies to dashboard/TagMeta `compact` widgets and the chart page — not only the challenge chart. On a narrow compact widget a ~30-day span produces ~8 `%b %d` labels (vs. the previous fixed 6), which d3 won't auto-thin, so they may overlap. Worth a quick look at the compact usages to confirm the labels still read cleanly.

fiddur (commented):
✅ Approved

Clean, tightly-scoped fix for #840. The new optional `xDomain` prop lets the challenge race chart span the full challenge timespan (`[start_ts, end_ts]`) instead of collapsing to the small data extent when only the first buckets exist, and it falls back to the data extent everywhere else. Threading it through `renderMultiSeries`/`renderSingleSeries`, deriving `buildDateFormat` from the effective domain, and adding `xDomain` to the effect deps are all correct. The per-day tick logic in `renderAxes` sensibly avoids repeated `%b %d` labels on short/sub-day spans. CI (typecheck + lint + build) is green, and the `PublicChallenge.tsx` diff is just prettier reflow with the one functional `xDomain={...}` addition.

Two non-blocking notes:

- `renderAxes` now applies day-based ticks to **every** chart with a span ≤ 31 days (dashboard widgets, TagMeta mini charts, the chart exploration page), not just the challenge chart. For narrow `compact` widgets a ~30-day span now yields ~8 `%b %d` labels where there were 6, which could crowd/overlap. Worth a quick visual check on the compact call sites.
- The `dayStep` computation is a small pure function embedded inside the D3 render path, so it isn't unit-testable without DOM mocking. Given the project's "prefer code testable without heavy mocking" standard, extracting it (like the `race-series.ts` helpers that already have tests) would let the tick-step edge cases be covered directly. Optional.

apps/web/src/components/charts/TrendLineChart.tsx
fiddur: Addressed both points: the day-tick step is now capped at ~6 ticks (extracted to a pure, unit-tested `daySpanTickStep` helper in chart-ticks.ts), so compact widgets/mini-charts get no more labels than the previous fixed `.ticks(6)` — just day-aligned on short spans.

fiddur (commented):
✅ Approved

Clean, focused fix. The new optional `xDomain` prop is threaded correctly through both `renderMultiSeries` and `renderSingleSeries` (including the `useEffect` dependency array), and `PublicChallenge` passes the full challenge timespan `[start_ts, end_ts]` so the race chart x-axis now spans the whole challenge rather than only the bucketed data extent — which is the intended fix for #840. Deriving `buildDateFormat` from the displayed `dateDomain` instead of the data extent is also more correct.

The extracted `daySpanTickStep` helper is small, pure, and well unit-tested; the assertions all match the implementation (`Math.max(1, Math.ceil(spanDays / 6))` capped at 31 days), so short challenges get one tick per day without exceeding the previous ~6-tick budget. Nice fallback to `.ticks(6)` for long/non-positive spans keeps existing charts unchanged.

A couple of non-blocking observations:
- `renderAxes` now unconditionally chooses ticks from `x.domain()`, so the behavior change touches every `TrendLineChart` call site (Chart page, MiniTrendChart, TrendChartWidget). For sub-day spans (`0 < spanDays < 1`) `daySpanTickStep` returns 1 and `d3.timeDay.every(1)` could yield only 0–1 ticks. This isn't reachable today (all call sites feed daily-bucketed data, so a ≥2-point series spans ≥1 day), but worth keeping in mind if hourly data is ever charted.
- Axis ticks/labels use the browser's local timezone (`d3.timeDay` / `d3.timeFormat`) while the challenge header renders dates in `challenge.timezone`, so viewers far from the challenge tz may see day boundaries that look off-by-one relative to the header. This is pre-existing chart behavior, not introduced here.

The remaining diff in `PublicChallenge.tsx` (multi-line reformatting of the `toCumulativeSeries` and `formatDateInZone` calls) is prettier reformatting only. No blocking issues.

apps/web/src/pages/Challenges/PublicChallenge.tsx
```diff
@@ -113,7 +121,13 @@ export function PublicChallenge({
         {standingsQuery.isLoading ? (
           <div class="public-loading">Loading standings…</div>
         ) : series.length > 0 ? (
-          <TrendLineChart data={[]} color={COLORS[0]} multiSeries={series} height={280} />
+          <TrendLineChart
+            data={[]}
+            color={COLORS[0]}
+            multiSeries={series}
+            height={280}
+            xDomain={[new Date(challenge.start_ts), new Date(challenge.end_ts)]}
```
fiddur: Minor / non-blocking: `end_ts` is the exclusive window end (midnight after the last competing day) — the header just above deliberately subtracts 1ms to show the inclusive last day. Using `end_ts` directly as the domain end means the daily-tick axis will render a trailing tick on the day *after* the last competing day (e.g. a Jan 1–3 challenge shows a Jan 4 tick with no data). It reads as the window boundary so it's defensible, but if you want the axis to match the header's inclusive range you could pass `new Date(new Date(challenge.end_ts).getTime() - 1)` as the domain end here too.

fiddur (commented):
✅ Approved

Clean, focused fix. The new `xDomain` prop is threaded correctly through both `renderMultiSeries` and `renderSingleSeries`, added to the `useEffect` dependency array, and defaults to the data extent when omitted so existing call sites are unaffected in behavior. `buildDateFormat` now keys off the effective domain rather than the data extent, which is the right choice. The extracted `daySpanTickStep` is a pure, testable helper and the unit tests match the implementation exactly (`ceil(spanDays/6)`, capped by the `>31 → null` fallback, `<=0 → null`). I confirmed the `x.domain()` destructuring typechecks (web tsconfig isn't strict, so the elements are `Date`, not `Date | undefined`), and the tooltip bisector still behaves sanely when the data is clustered in the left portion of a wider fixed domain (hovers past the last point resolve to the last point, no crash).

Non-blocking observations, all cosmetic:

- The tick change in `renderAxes` is global — it affects every `TrendLineChart` consumer (Chart page, MiniTrendChart, compact dashboard widgets), not just the challenge race chart. The `~6` cap keeps the label budget roughly as before, so this looks intentional and safe. Note that `d3.timeDay.every(step)` for `step >= 2` filters by day-of-month and resets at month boundaries, so tick spacing can be slightly uneven / a touch denser near a month boundary. Purely visual.

- Daily ticks land on browser-local midnights and are labelled with local-time `d3.timeFormat`, whereas the challenge header renders in `challenge.timezone` via `formatDateInZone`. For viewers in a different timezone than the challenge, the day-boundary ticks won't align exactly with the challenge's own days. This is pre-existing axis behavior (not introduced here), but the fixed domain + daily ticks make it a bit more visible.

apps/web/src/pages/Challenges/PublicChallenge.tsx
fiddur: Good catch — applied: the axis now ends at end_ts − 1ms so it stops on the inclusive last competing day (matching the header and #840's expected Jul 01/02/03), no trailing empty tick.

PR #846, opened 2026-07-02
feat(feed): AS2 Create{Exercise} object model (#831)

fiddur:
## Summary

Slice 2 of the federated activity feed ([#831](https://github.com/fiddur/aurboda/issues/831)) — the **ActivityPub payload builder** the upcoming Fedify delivery layer will send. Builds on the merged persistence/share/series foundation (#838).

`buildCreateExercise` is a **pure, deterministic serializer**: given a post's visibility, its activity's window/type/title, the resolved shared scalar metrics, and the absolute actor/post/series URLs, it produces a `Create` activity whose object is dual-typed `["Note", "aurboda:Exercise"]` — `Note` first so Mastodon renders `name`/`content`/`url` as a status, with structured `aurboda:` extension fields (`activityType`, `startTime`/`endTime`, `durationSeconds`, `metrics`, `series`) for Aurboda↔Aurboda consumers ("progressive enhancement").

## Privacy stays consistent with slice 1
- Only the **shared scalars** appear in `content` and `aurboda:metrics`.
- Only explicitly-shared **`seriesMetrics`** produce `aurboda:series` links, scoped to the activity window and pointing at the public `/series` endpoint (whose data-driven scoping is the actual boundary).
- Visibility maps to AS2 `to`/`cc` addressing (public / unlisted / followers).

## Why pure / why now
The serializer is DI-free — actor/post/series URLs and the `aurboda:` namespace IRI are **parameters** — so it has no dependency on the (upcoming) Fedify actor/WebFinger/HTTP-signatures/delivery infrastructure, and it's fully unit-testable. It's the payload the delivery layer will emit; landing it separately keeps that next PR focused on the framework wiring.

## Testing
13 unit tests: `to`/`cc` addressing per visibility, Note-first dual type + canonical ids, `@context` prefix, scalar flattening into `content`, `aurboda:metrics` projection, series scoping (incl. none-shared and open-ended activity), duration computation, and HTML escaping. Backend `tsgo`/oxlint clean.

## Next slices (per issue)
Fedify-based **actor + WebFinger + HTTP signatures + inbox/outbox + delivery queue** (+ the served `/ns/activitystreams` context document and `FEDERATION.md`) — this is where the actor-URL/nginx routing decision lands; then map/chart image rendering, the web Share UI, and the README entry.

### DIFF
--- a/apps/backend/src/services/activitypub/object.ts
+++ b/apps/backend/src/services/activitypub/object.ts
@@ -1,0 +1,182 @@
+/**
+ * AS2 object model for a shared activity.
+ *
+ * Serializes a feed post into a `Create` activity whose `object` is a dual-typed
+ * `["Note", "aurboda:Exercise"]`: `Note` first so plain fediverse clients
+ * (Mastodon) render `name`/`content`/`url` as a status, plus structured
+ * `aurboda:` extension fields that an Aurboda↔Aurboda consumer reads
+ * ("progressive enhancement").
+ *
+ * This module is a pure function of its inputs — the caller resolves the shared
+ * scalar values and passes in the absolute URLs — so it is fully unit-testable
+ * and carries no dependency on the (upcoming) Fedify actor/delivery layer. Only
+ * the metrics the user actually shared are ever emitted: unshared scalars are
+ * absent from `content` and `aurboda:metrics`, and only `seriesMetrics` produce
+ * `aurboda:series` links (which resolve against the public `/series` endpoint).
+ */
+
+import type { FeedVisibility } from '@aurboda/api-spec'
+
+/** The AS2 magic collection every public object is addressed to. */
+export const AS_PUBLIC = 'https://www.w3.org/ns/activitystreams#Public'
+
+/** A resolved scalar summary for one shared metric. */
+export interface ScalarMetric {
+  /** Machine key, e.g. `heart_rate_avg`, `distance`, `hr_zone_minutes`. */
+  key: string
+  /** Scalar value, or a small keyed record (e.g. HR-zone minutes `{ z2: 22 }`). */
+  value: number | Record<string, number>
+  /** Optional unit for the scalar form (e.g. `bpm`, `km`). */
+  unit?: string
+  /** Optional human label for the fallback text; defaults to a prettified key. */
+  label?: string
+}
+
+export interface BuildCreateInput {
+  /** Canonical absolute URL of the post (the `Create` activity `id`). */
+  postId: string
+  /** Absolute actor URL (e.g. `https://host/u/fredrik`). */
+  actorUrl: string
+  /** The `aurboda:` term prefix IRI, ending in `#` (e.g. `https://host/ns/activitystreams#`). */
+  aurbodaNs: string
+  /** Base URL of the public series endpoint (e.g. `https://host/api/public/fredrik/series`). */
+  seriesEndpointBase: string
+  visibility: FeedVisibility
+  activityType: string
+  /** Activity start (ISO 8601); also the `published` time. */
+  startTime: string
+  /** Activity end (ISO 8601); required for duration and series links. */
+  endTime?: string
+  title?: string
+  /** Resolved scalar summaries for the shared `included_metrics`. */
+  scalars: ScalarMetric[]
+  /** Metric keys whose series were explicitly shared (drive `aurboda:series`). */
+  seriesMetrics: string[]
+  /** Bucket granularity for the series links (defaults to `5s`). */
+  seriesBucket?: string
+}
+
+export interface AS2Create {
+  '@context': [string, Record<string, string>]
+  id: string
+  type: 'Create'
+  actor: string
+  published: string
+  to: string[]
+  cc: string[]
+  object: Record<string, unknown>
+}
+
+const escapeHtml = (s: string): string =>
+  s.replaceAll('&', '&amp;').replaceAll('<', '&lt;').replaceAll('>', '&gt;').replaceAll('"', '&quot;')
+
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/activitypub/object.ts
```diff
@@ -0,0 +1,182 @@
...
+      return { cc: [AS_PUBLIC], to: [followersUrl] }
+    case 'followers':
+      return { cc: [], to: [followersUrl] }
+  }
+}
+
+/** Build the `aurboda:series` link objects for the shared series metrics. */
+const seriesLinks = (input: BuildCreateInput): Record<string, string>[] => {
+  const { endTime, seriesMetrics } = input
+  if (!endTime || seriesMetrics.length === 0) return []
```
fiddur: `seriesLinks` emits `aurboda:series` links for every visibility, but the public `/series` endpoint only serves `public`/`unlisted` posts: `findCoveringSharedSeriesWindow` filters `f.visibility IN ('public', 'unlisted')` and `feedVisibilitySchema`'s docs say *"followers posts have no public series"*. So a `followers`-only post that shared series metrics gets links that resolve to nothing.

`visibility` is already available on `input` here — consider gating (e.g. `if (input.visibility === 'followers' || !endTime || seriesMetrics.length === 0) return []`) so the AS2 object matches the documented privacy boundary, and add a test for the `followers` + `seriesMetrics` case. (If the caller is meant to enforce this instead, a comment noting that would suffice.)

fiddur (commented):
🛑 Changes required

Clean, well-documented pure module with good unit coverage (addressing, dual-type object, HTML escaping, duration, series scoping, name fallback). I traced every assertion in `object.test.ts` by hand and they all hold; couldn't run the suite locally since backend `node_modules` isn't installed, but the logic checks out.

One correctness issue worth resolving before merge: `seriesLinks` emits `aurboda:series` links for **any** visibility, but the public `/series` endpoint is explicitly scoped to `public`/`unlisted` posts only. `findCoveringSharedSeriesWindow` filters `f.visibility IN ('public', 'unlisted')` and `feedVisibilitySchema`'s docs state *"followers posts have no public series"*. So a `followers`-only post that shared series metrics would carry `aurboda:series` links that resolve to nothing (the endpoint returns null → dead link). Since `visibility` is already an input here, gating series generation on it keeps the emitted object consistent with the documented privacy boundary — see the inline comment. There's also no test covering the `followers` + `seriesMetrics` combination.

Minor (non-blocking) design note: both `Create.published` and `object.published` are set to `startTime` (the workout start), not the share/creation time. For activities shared well after they happened, that timestamps the federated post in the past, which affects Mastodon timeline ordering. If that's intentional, no change needed; flagging in case it isn't.

apps/backend/src/services/activitypub/object.ts
fiddur: Fixed: seriesLinks now returns [] for followers-only posts, so the AS2 object never advertises series links the public /series endpoint would 404. Added a test for the followers + seriesMetrics case.

apps/backend/src/services/activitypub/object.ts
```diff
@@ -0,0 +1,189 @@
...
+  const series = seriesLinks(input)
+  if (series.length > 0) object['aurboda:series'] = series
+
+  return {
+    '@context': ['https://www.w3.org/ns/activitystreams', { aurboda: input.aurbodaNs }],
+    actor: input.actorUrl,
+    cc,
+    id: input.postId,
+    object,
+    published: input.startTime,
```
fiddur, line 185: Both the `Create` and its `object` (line 164) set `published` to `input.startTime` — the workout start, not the post-creation time. Remote servers order and render statuses by `published`, so a run shared hours after it happened will federate timestamped at the workout start and can land buried below newer posts in followers' timelines. The workout time is already preserved in `aurboda:startTime`, so consider passing a separate creation timestamp for `published` when delivery is wired. Non-blocking here since this is foundation-only, but flagging for the follow-up.

fiddur (commented):
✅ Approved

Clean, self-contained AS2 object model. It's a pure function of its inputs, thoroughly unit-tested, and follows the project's conventions (functional style, `services/` location, api-spec `FeedVisibility` type, snake_case keys, type-guard predicate instead of a cast). I verified the non-trivial computations against the tests independently: duration (2463s), `URLSearchParams` ordering/`%3A` encoding of the series href, the flattened `content` line, HTML escaping, and the derived `Running activity` name — all match. The Mastodon addressing (public/unlisted/followers → `to`/`cc`) and the "series only for non-followers with a bounded window" gating are correct and consistent with `feed.ts` / the public `/series` endpoint. Note that `content` is HTML-escaped (good — title and scalar values are user-controlled) while `name` is left as plain text, which is correct per AS2.

Two forward-looking notes for the delivery follow-up (non-blocking for this foundation-only PR):

1. `published` timestamp semantics — see inline comment.
2. Object addressing — the wrapped `object` (`Note`/`aurboda:Exercise`) carries no `to`/`cc`; only the outer `Create` does. Mastodon derives a status's visibility partly from the **object's** `to`/`cc`, so when the Fedify delivery layer is added it will likely need to mirror the computed addressing onto the object as well (or a public post could be ingested as limited/direct). Worth confirming when wiring delivery.

Couldn't execute the test suite here (no `node_modules` in this worktree), but re-derived the expected values by hand and they line up.

apps/backend/src/services/activitypub/object.ts
fiddur: Good catch — deferring to the delivery slice (PR3) as you suggested: I'll add a `publishedAt` (post-creation time) input that drives `Create.published`/`object.published`, keeping the workout start in `aurboda:startTime`. It belongs with the caller that has a real "now" at delivery time; tracked for PR3.

PR #847, opened 2026-07-02
feat(feed): persist a per-user ActivityPub actor keypair (#831)

fiddur:
## Summary

Slice 3a of the federated activity feed ([#831](https://github.com/fiddur/aurboda/issues/831)). Lands the **actor keypair** that the upcoming Fedify-based actor / WebFinger / HTTP-signature layer will consume. Builds on the merged persistence (#838) and AS2 object model (#846).

Each per-user database hosts exactly **one** actor (the user), so `feed_actor` is a **singleton** table (`BOOLEAN singleton` PK + `CHECK (singleton)`). `getOrCreateActorKeyPair`:
- generates an **RSA-2048** keypair (PKCS#8 private / SPKI public PEM) on first use,
- is **idempotent and race-safe** — `INSERT ... ON CONFLICT (singleton) DO NOTHING` then read, so a concurrent first call can't leave two keypairs.

PEM encoding is chosen so it converts cleanly to a WebCrypto `CryptoKey` for the federation layer; the public half will be published in the actor document.

## Testing
3 integration tests: valid RSA keypair on first use (parses as RSA), idempotency across calls, and a single keypair surviving concurrent first-use. Backend `tsgo`/oxlint clean.

## Context: how this fits the federation slices
The next PR wires **Fedify 2.3** (`@fedify/express`) into an actor dispatcher (reading this keypair) + WebFinger + the served `/ns/activitystreams` context document, and updates `nginx.conf` so `/.well-known/webfinger`, the actor doc, and `/ns/` reach the backend. Planned **actor identity = `https://host/u/:user`** (content-negotiated), consistent with the AS2 object model already merged and the issue's spec — flagging that routing choice here so it can be steered before the Fedify PR. Inbound inbox + signature verification and outbound delivery (Create/Update/Delete + queue) follow in their own PRs.

### DIFF
--- a/apps/backend/src/db/feed-actor.integration.test.ts
+++ b/apps/backend/src/db/feed-actor.integration.test.ts
@@ -1,0 +1,59 @@
+import { createPublicKey } from 'node:crypto'
+import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest'
+
+/**
+ * Integration tests for the singleton actor keypair store.
+ */
+import { cleanTestDb, getTestUser, startTestDb, stopTestDb } from '../test/db-test-helper.ts'
+import { getOrCreateActorKeyPair } from './feed-actor.ts'
+
+const CONTAINER_TIMEOUT = 120_000
+
+describe('Feed actor keypair integration', () => {
+  beforeAll(async () => {
+    await startTestDb()
+  }, CONTAINER_TIMEOUT)
+
+  afterAll(async () => {
+    await stopTestDb()
+  })
+
+  beforeEach(async () => {
+    await cleanTestDb()
+  })
+
+  test('generates a valid RSA keypair on first use', async () => {
+    const user = getTestUser()
+    const kp = await getOrCreateActorKeyPair(user)
+
+    expect(kp.private_key_pem).toContain('BEGIN PRIVATE KEY')
+    expect(kp.public_key_pem).toContain('BEGIN PUBLIC KEY')
+    expect(kp.created_at).toBeInstanceOf(Date)
+
+    // The stored public key parses as an RSA key.
+    const pub = createPublicKey(kp.public_key_pem)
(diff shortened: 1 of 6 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped infrastructure PR that adds a per-user ActivityPub actor keypair store for #831.

**Correctness & wiring — all good:**
- The `feed_actor` singleton table is modelled correctly: `singleton BOOLEAN PRIMARY KEY DEFAULT true` plus `CHECK (singleton)` pins it to exactly one row.
- It's wired everywhere it needs to be: `socialTables` (spread into `createTableStatements`), `tableCreationOrder`, and `cleanTestDb`. Existing users get the table created via the migration retry path — `migrateSchema`'s `tableCreationOrder` loop (connection.ts) creates any missing table, and `feed_actor` has no FKs so its placement is safe.
- `getOrCreateActorKeyPair` is genuinely idempotent and race-safe: the SELECT short-circuits the common (already-generated) case, and `INSERT ... ON CONFLICT (singleton) DO NOTHING` followed by a re-SELECT resolves concurrent first-use to the single winning row — even across connections. The docstring accurately describes this.
- Storing the private key as plaintext PKCS#8 PEM is consistent with the existing trust model: `oauth_tokens.access_token`/`refresh_token` already live as plaintext in the same per-user database, so this introduces no new exposure.

**Testing:** The integration test covers key generation (validates it parses as RSA), idempotency, and convergence under concurrent first-use — satisfying the project standard that all DB functions have integration tests.

**Parity:** No REST/MCP endpoint is expected here — the function isn't yet called anywhere; it's a building block for the federation layer in the larger #831 effort.

Minor (non-blocking) observation: `ActorKeyPair` and `ActorKeyPairRow` are identical, and the function returns rows directly whereas the sibling `feed.ts` routes rows through a `mapFeedPost` mapper. Harmless either way; noted only for future consistency.

PR #850, opened 2026-07-02
feat(feed): Fedify actor + WebFinger federation surface (#831)

fiddur:
## Summary

Slice 3b of the federated activity feed ([#831](https://github.com/fiddur/aurboda/issues/831)): the ActivityPub **actor identity + discovery** surface, built on **Fedify 2.3**. Builds on the merged persistence (#838), AS2 object model (#846), and actor keypair (#847).

- **`services/activitypub/keys.ts`** — converts the stored RSA PEM keypair into a WebCrypto `CryptoKeyPair` (Node JWK export → Fedify `importJwk`, stamped `alg: RS256` = RSASSA-PKCS1-v1_5). Unit test round-trips sign/verify and confirms the public key is extractable (so Fedify can publish it).
- **`services/activitypub/federation.ts`** — `createFeedFederation()`: one actor per user (`Person`) at `/users/{identifier}` with the published RSA key, a key-pairs dispatcher backed by `getOrCreateActorKeyPair`, and WebFinger via `mapHandle`. Inbox/outbox/followers paths registered (empty/no-op for now).
- **Mounted** into the Express app before the JSON body parser (so signed inbox POSTs keep their raw body in a later slice), with `trust proxy` for correct external URLs behind nginx.
- **`nginx.conf`** — routes `/.well-known/webfinger` and `/users/` straight to the backend (path preserved).
- **`FEDERATION.md`** (FEP-67ff).

## Testing
`federation.fetch` integration tests against a real per-user DB: actor document (with RSA `publicKeyPem`), WebFinger `acct:` resolution (self-link → actor), and invalid-username 404. Plus the key round-trip unit tests. 19 AP tests total; backend `tsgo`/oxlint clean.

## Decision flagged: actor path `/users/{identifier}` (not `/u/`)
The AS2 object model and issue examples used `/u/<user>`, but `/u/` is the SPA's profile/dashboard/challenge prefix. Serving the actor there would need fragile `Accept`-header content negotiation in nginx (an `if`/`map` I can't verify without a running server) and risks breaking those pages. A dedicated `/users/{identifier}` prefix routes cleanly to the backend with no negotiation. **Nothing is federated/deployed yet, so this permanent identity URL is free to pick now** — happy to switch to `/u/` (with the nginx work) if you'd prefer the prettier id.

## Next slices
Inbound HTTP-signature verification + `Follow`→`Accept`; outbound delivery queue (`Create`/`Update`/`Delete`, wiring share→`buildCreateExercise`, plus the deferred `publishedAt`); the served `/ns/activitystreams` context document; and a real paged outbox.


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -0,0 +1,75 @@
+import { createFederation, type Federation, InProcessMessageQueue, MemoryKvStore } from '@fedify/fedify'
+import { Person } from '@fedify/fedify/vocab'
+
+/**
+ * The Fedify `Federation` object for the activity feed.
+ *
+ * Single actor per user: the actor identifier IS the username, and the actor
+ * lives at `<host>/u/<username>` — consistent with the shared-dashboard /
```
fiddur: This says the actor "lives at `<host>/u/<username>`", but the dispatcher registers `/users/{identifier}` and FEDERATION.md documents `/users/<username>`. Since the whole point of the dedicated prefix is that it's `/users/` (not the SPA's `/u/`), this comment is misleading — please update to `/users/<username>`.

fiddur (commented):
✅ Approved

Solid, well-scoped read/discovery slice for ActivityPub federation. The design holds up:

- `isValidUsername` gates the identifier before it flows into per-user DB routing (`query(user, …)` selects the DB by name), so there's no path for an attacker to reach an arbitrary database via the actor/WebFinger URL.
- `isMissingDatabase` (Postgres `3D000`) correctly maps "no such user" to a 404. I confirmed against the Fedify 2.3.1 source that `handleWebFinger` invokes the actor dispatcher and returns `onNotFound` when it yields `null`, so WebFinger and the actor endpoint agree — no 200-webfinger/404-actor inconsistency for non-existent users.
- `trust proxy: true` is genuinely required here: `@fedify/express` builds the request URL from `req.protocol`/`req.host`, so Fedify needs the `X-Forwarded-Proto`/`Host` headers to reconstruct the external `https://<host>` actor id. Impact of the global scope is low in this app — auth is Bearer-token (no cookies), there's no IP-based rate limiting, and WebAuthn origins come from env, not the request.
- The key-conversion round-trip (PEM → JWK → WebCrypto, sign/verify) is unit-tested, and the actor + WebFinger surface has real-DB integration tests.

Two non-blocking notes:

1. The federation middleware is mounted at the app root (not path-scoped) ahead of `express.json()`, so every non-GET request now passes through `integrateFederation`, which wraps the body via `Readable.toWeb(req)` before the JSON parser runs. This is Fedify's documented Express integration and `Readable.toWeb` is pull-based (the stream stays paused until read, and Fedify returns via `onNotFound` without reading the body for non-federation routes), so normal `/api` POSTs should still parse. However, there's no automated coverage of a real POST traversing the root federation middleware into the body parser (the existing router tests build their own app without this middleware, and the integration test hits `federation.fetch` directly). Worth a quick manual smoke test of a POST `/api/...` endpoint against the assembled app.
2. Fetching an existing user's actor for the first time provisions (writes) an RSA keypair via `getOrCreateActorKeyPair` as a side effect of an unauthenticated GET. This is by design per #831 (every user is an actor), bounded to existing users, and self-limiting after first fetch — flagging only for awareness.

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed: updated the comment to `/users/<username>` to match the dispatcher and FEDERATION.md.

apps/backend/src/api.ts
```diff
@@ -313,6 +315,13 @@ const main = async () => {
     }),
   )
 
+  // ActivityPub federation (actor + WebFinger). Mounted BEFORE the JSON body
+  // parser so Fedify owns the raw body of signed inbox POSTs; it passes through
+  // (next()) any request that isn't one of its own routes. `trust proxy` lets
+  // Fedify reconstruct the external https URL behind nginx.
+  httpd.set('trust proxy', true)
```
fiddur: Non-blocking: `trust proxy: true` trusts every upstream hop, which makes `X-Forwarded-For`/`-Proto`/`-Host` fully spoofable if a request ever reaches the app other than through the loopback nginx. Since the backend binds 127.0.0.1 behind nginx, `httpd.set('trust proxy', 'loopback')` (or `1`) gives Fedify the same external-URL reconstruction with a smaller blast radius. Fine to defer given nothing here relies on `req.ip`/secure cookies.

fiddur (commented):
✅ Approved

Solid, well-scoped slice that adds the read/discovery federation surface (actor document + WebFinger) via Fedify, with good docs (FEDERATION.md) and both unit (keypair conversion round-trip) and integration (actor + WebFinger against a real per-user DB) coverage. The dispatchers correctly gate on `isValidUsername` and treat a missing per-user database (`isMissingDatabase`) as a 404, and the actor is only served once a keypair exists. I verified there are no `/.well-known/*` route collisions (the existing `assetlinks.json` / `aurboda` routes fall through Fedify's `next()`), `TEST_USER` (`testuser`) is a valid non-reserved username so the integration test is meaningful, and nothing in the codebase reads `req.ip` / `req.secure` / secure cookies, so the global `trust proxy` change is benign today.

Non-blocking notes (no changes required):
- `trust proxy: true` trusts *all* upstream hops, making `X-Forwarded-*` spoofable if the app is ever reachable other than through the loopback nginx. Since the backend listens on 127.0.0.1 behind nginx, a tighter value (`'loopback'` or `1`) would give Fedify the same external-URL reconstruction with less blast radius. Fine to leave for now given nothing trusts client IP.
- `mapHandle` returns a handle for any well-formed username without checking DB existence, so WebFinger answers `200` with a `self` link even for users that don't exist (the actor fetch then 404s). Harmless and arguably avoids user enumeration, but worth noting as a minor spec deviation.
- FEDERATION.md lives at the repo root rather than under `docs/`; the project convention keeps docs in `docs/`. Consider moving/linking it there for discoverability.
- Minor: in the actor dispatcher the `try/catch` for `isMissingDatabase` around `ctx.getActorKeyPairs` is effectively unreachable — the key-pairs dispatcher already swallows `isMissingDatabase` and returns `[]`, so `getActorKeyPairs` won't surface it. Not wrong, just redundant.

apps/backend/src/api.ts
fiddur: Applied: changed to `httpd.set('trust proxy', 'loopback')` so only the loopback nginx hop is trusted for X-Forwarded-* — smaller blast radius, same external-URL reconstruction for Fedify.

apps/backend/src/api.ts
```diff
@@ -313,6 +315,15 @@ const main = async () => {
     }),
   )
 
+  // ActivityPub federation (actor + WebFinger). Mounted BEFORE the JSON body
+  // parser so Fedify owns the raw body of signed inbox POSTs; it passes through
+  // (next()) any request that isn't one of its own routes. `trust proxy` lets
+  // Fedify reconstruct the external https URL from nginx's X-Forwarded-* headers.
+  // Scope it to `loopback`: the backend binds 127.0.0.1 behind nginx, so only the
```
fiddur: Minor comment-accuracy nit (non-blocking): the safety of `trust proxy: loopback` comes from nginx connecting to the backend over loopback (`proxy_pass http://127.0.0.1:3000`), not from the backend's own bind address — `httpd.listen(port)` at line 467 binds `0.0.0.0`, not `127.0.0.1`. The security property holds regardless (a direct remote peer wouldn't be loopback, so its `X-Forwarded-*` wouldn't be trusted), but the stated premise "the backend binds 127.0.0.1" is inaccurate and could mislead a future maintainer reasoning about the trust boundary.

fiddur (commented):
✅ Approved

Clean, well-scoped read/discovery slice for ActivityPub. The wiring is correct and defensively coded:

- Every dispatcher (`setActorDispatcher`, `setKeyPairsDispatcher`, `mapHandle`) gates on `isValidUsername` before any DB access, so malformed handles 404 without touching Postgres — matching the integration test.
- `isMissingDatabase` (Postgres `3D000`) correctly maps unregistered users to a 404, since users map to per-user databases.
- Mounting `integrateFederation` before `json()` is right (raw body for signed inbox POSTs), and it passes non-Fedify routes through via `next()`. `/mcp` is mounted earlier so it's unaffected, and there are no existing `/users/` routes to collide with.
- The global `trust proxy: loopback` change is safe: no other code reads `req.ip`/`req.protocol`/`req.secure`/`req.hostname`, so Fedify is the only consumer, and only the loopback hop (nginx→backend) is trusted.
- `toCryptoKeyPair` stamps `alg: 'RS256'` so the imported key is unambiguously RSASSA-PKCS1-v1_5; the unit test round-trips sign/verify and confirms the public key is extractable.

Non-blocking observations:

1. **Write-on-read**: an unauthenticated `GET /users/<user>` (or WebFinger) reaches `getOrCreateActorKeyPair`, which lazily generates + persists an RSA-2048 keypair on first access. This is bounded (one keypair per registered user, idempotent, cached thereafter) and matches the documented get-or-create design, but it means the actor endpoint has a DB side effect and is unauthenticated/unrate-limited. Fine for now; worth keeping in mind when the delivery slice adds more surface.
2. The `isMissingDatabase` catch inside `setActorDispatcher` is effectively dead code: `ctx.getActorKeyPairs` delegates to the key-pairs dispatcher, which already swallows `isMissingDatabase` and returns `[]` (→ `keys.length === 0` → `null`). Harmless defensive redundancy.
3. Because nginx `location /api/` strips the `/api/` prefix, the actor is also reachable at `/api/users/<user>`. The canonical `id` Fedify emits is still `/users/<user>`, so this is harmless duplication, not a correctness issue.

I could not run the test suite (dependencies aren't installed and the npm cache is read-only in this environment), so test execution relies on CI. Nothing blocks merge.

apps/backend/src/api.ts
fiddur: Fixed the comment: the safety comes from nginx proxying over loopback (so the immediate peer is 127.0.0.1), not the backend bind address. Reworded accordingly.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (isMissingDatabase(error)) return []
+        throw error
+      }
+    })
+    .mapHandle((_ctx, username) => (isValidUsername(username) ? username : null))
+
+  // Register the inbox path so the actor can advertise inbox/shared-inbox URIs.
+  // Inbound handling (Follow→Accept, signature verification) is a later slice —
+  // for now the inbox simply accepts and ignores.
+  federation.setInboxListeners('/users/{identifier}/inbox', '/inbox')
```
fiddur, line 65: Non-blocking heads-up for the delivery slice: the per-user inbox `/users/{identifier}/inbox` is reachable (nginx proxies `/users/`), but the shared inbox `/inbox` registered here is not proxied by nginx — a request to `https://<host>/inbox` would fall through to the SPA (`location /` → index.html). Fine for now since inbound handling is out of scope and nothing delivers yet, but if the actor ends up advertising `endpoints.sharedInbox`, remote servers would try an unreachable URL. Worth adding an nginx `location = /inbox` (and proxying it) when inbound lands.

fiddur (commented):
✅ Approved

Clean, well-documented read-only federation surface (actor document + WebFinger + key-pairs dispatcher). I traced the whole path end-to-end and found no blocking issues.

**Verified correct:**
- Every dispatcher gates on `isValidUsername` before touching the per-user DB, so the username can never reach `getDbForUser`'s `SET ROLE`/database selection unsafely.
- `isMissingDatabase` (PG `3D000`) maps a valid-but-nonexistent user to `null`/`[]` → 404. Importantly, `getOrCreateActorKeyPair` runs its `SELECT` first, so for a nonexistent user the missing-DB error is thrown *before* any RSA keygen — username enumeration can't be turned into a keygen DoS. Key generation only ever happens once per real user.
- `httpd.set('trust proxy', 'loopback')` is a sound choice and its reconstruction rationale is well-commented. It's a global setting, but no existing backend code reads `req.ip`/`req.protocol`/`req.secure`/`req.hostname` (OAuth/WebAuthn origins are built from env config), so it doesn't change existing behavior.
- Fedify is mounted before `json()` (consistent with the existing MCP mount) and passes non-federation routes through via `next()`; no collision with the existing `/.well-known/*` router, which doesn't serve webfinger.
- `keys.ts` correctly stamps `alg: 'RS256'` so the RSA key is unambiguously RSASSA-PKCS1-v1_5, and the round-trip is verified in `keys.test.ts`.
- Tests cover the actor document, WebFinger resolution, and the invalid-username short-circuit.

**Minor / non-blocking observations (not required for this PR):**
- The actor is auto-created (and its keypair minted lazily) for any existing user on first federation read — i.e. every user is discoverable/federatable with no opt-in. This is the documented design ("one actor per user") in FEDERATION.md, so calling it out only in case an opt-in is desired later.
- The `try/catch` for `isMissingDatabase` in `setActorDispatcher` is effectively dead: `getActorKeyPairs` runs through the key-pairs dispatcher, which already swallows that error and returns `[]`. Harmless defensiveness.
- Heads-up for the delivery slice (see inline comment): the shared inbox `/inbox` is registered but not proxied by nginx.

apps/backend/src/services/activitypub/federation.ts
fiddur: Deferring to the inbound slice (3c), as flagged. This slice's actor does not advertise `endpoints.sharedInbox`, so `/inbox` is never referenced or reached — harmless dead registration for now. I'll add an nginx `location = /inbox` proxy (and wire the shared-inbox endpoint) when inbound Follow→Accept handling lands. Tracked in the plan.

fiddur (commented):
✅ Approved

Clean, well-scoped read-only federation surface (Fedify actor `Person` + WebFinger discovery), with good unit and integration coverage. I traced the referenced helpers and the wiring and found no blocking issues.

Verified:
- `isValidUsername` gating in both the actor and key-pairs dispatchers (and `mapHandle`) correctly 404s malformed handles before any DB access — the `Invalid..Name` test confirms it.
- Missing per-user DB surfaces as pg `3D000`, caught by `isMissingDatabase` → `[]`/`null` → 404, matching the existing public-route convention.
- `toCryptoKeyPair` round-trips PKCS#8/SPKI PEM → JWK (`alg: RS256`) → WebCrypto `RSASSA-PKCS1-v1_5`/SHA-256 keys that sign+verify and export, as the keys test asserts.
- nginx `/users/` and `= /.well-known/webfinger` locations don't collide with the SPA (profiles live under `/u/`, which has its own location); `proxy_pass` without a trailing path preserves the full `/users/...` URI the actor id is built from.
- `trust proxy: loopback` is set app-wide, but no code reads `req.ip`/`req.protocol`/`req.secure`/`req.hostname`, so there's no behavioral change beyond letting Fedify reconstruct the external https origin.

Non-blocking observations (fine to defer to later slices):
1. A GET to `/users/<valid-username>` lazily generates and persists an RSA keypair (via `getOrCreateActorKeyPair`) for any username whose per-user DB exists — i.e. an unauthenticated read triggers a one-time write/keygen and provisions the actor for every user regardless of whether they've opted into federation/sharing. It's idempotent and bounded, but worth a conscious decision about whether actor existence should be gated on an explicit federation/sharing opt-in, and about username enumeration via WebFinger.
2. The outbox/followers dispatchers don't gate on `isValidUsername`/DB existence the way the actor dispatcher does. Worth confirming what Fedify returns for `/users/<invalid-or-nonexistent>/outbox` — if it serves an empty `200` while the actor `404`s, that's a minor inconsistency to tighten once these collections carry real data (harmless while empty).
3. The `isMissingDatabase` catch in the actor dispatcher is effectively redundant: `getActorKeyPairs` runs through the key-pairs dispatcher, which already swallows missing-DB into `[]`. Harmless defensive code.
4. No test exercises the Express `integrateFederation` mount itself (body-parser ordering / pass-through of non-federation routes); the integration test drives `federation.fetch` directly. A light smoke test through the Express app would guard the middleware ordering that the comment relies on.

Documentation (FEDERATION.md) is clear about what's implemented vs. planned. Nice work.

PR #851, opened 2026-07-02
feat(sharing): server-rendered <head> meta + footer for share pages (#834)

fiddur:
Part 1 of the **#834** epic (rich link previews). Since the app is nginx + Express + PostgreSQL (Docker), not Vercel/Supabase as the issue assumed, this adapts the approach: nginx proxies the public share routes to the Express backend, which server-renders the `<head>`.

## What & why

JS-less crawlers (Facebook, Slack, LinkedIn, Discord, iMessage, Mastodon) never run the SPA, so a pasted `/u/...` link previewed only the generic **"Aurboda"** title. Now:

- **nginx** proxies `/u/*` → backend, which returns the same `index.html` with an injected, crawler-visible `<head>`: Open Graph + Twitter Card + `description` meta + schema.org JSON-LD. Browsers still hydrate the full SPA — only the head is enriched.
- Rich meta is emitted **only for public resources**. Unlisted (slug-only) dashboards and unknown URLs get generic site meta, so an unlisted resource's title never lands in a crawler's cache/search index. *(Note: in this codebase `is_public` gates **listing**; slugs are unguessable and grant page access either way — this PR deliberately withholds rich crawler meta from non-public shares.)*
- A branded **1200×630 default preview image** (`og-default.png`).
- The standard **site footer** now renders on public share pages too (year made dynamic).

## Design

- `services/share-meta.ts` — pure meta builders + `injectShareMeta` (HTML/attr/JSON-LD escaped).
- `services/web-template.ts` — loads the built `index.html` via `WEB_INDEX_PATH` (set in the Docker image; unset in dev, where vite serves `/u/*`).
- `routes/share-html-router.ts` — resolves dashboards/challenges/profiles with **dependency-injected** resolvers (unit-tested with fakes); a `/u/*` catch-all preserves the old `try_files` SPA fallback.

## Tests

25 unit tests (meta building/escaping/injection, router visibility + fallbacks, template loader). `pnpm check` green.

## Follow-ups (tracked, sequential PRs)

2. Dynamic OG image via Satori→PNG · 3. Avatars · 4. Editable dashboard text · 5. oEmbed + polish


fiddur (commented):
✅ Approved

Solid, well-tested change. The server-rendered `<head>` for `/u/*` is implemented cleanly as pure functions plus a thin router, and the security-sensitive parts hold up:

- **Escaping is correct.** `escapeAttr` covers `& < > "` for attribute values, `escapeHtmlText` for the title text node, and `jsonLdScript` neutralizes `</script>` by escaping `<` to `\u003c` (all confirmed by tests). Nothing user-controlled reaches the HTML unescaped on the primary (template) path.
- **Visibility is respected.** Rich meta is only emitted for `is_public` resources; unlisted/unknown fall back to generic site meta, so a private resource's title/description never lands in a crawler cache. `isValidUsername` gates the resolvers, and slugs are unique across the dashboard/challenge namespaces, so the `dashboard ? null : resolveChallenge(...)` short-circuit is safe.
- **Injection-safe DB access.** username/slug reach the DB only as parameterized queries, and the invalid-username early return avoids pointless connection attempts; `3D000` handling degrades to generic meta rather than 500.
- **Deploy/cache lifecycle is fine.** `createTemplateLoader` caches for the process lifetime, which is correct here because the combined image restarts the backend on each deploy, so `index.html` (with fresh asset hashes) is re-read. nginx `/u/` uses `proxy_pass http://127.0.0.1:3000;` (no trailing slash) so the full `/u/...` URI is preserved to match the backend routes.

Two non-blocking notes, no changes required:

1. **Availability tradeoff:** `/u/*` now depends on the backend process. Previously nginx served these paths statically via `try_files … /index.html`, so a backend outage would only have degraded the API; now it also 502s public share pages. This is a reasonable, documented tradeoff — just calling it out.
2. **`fallbackHtml` escaping is lighter than the main path** (only `<` for the title/link text, only `"` for the href) and doesn't reuse `escapeHtmlText`. Since this path is dev-only (unset `WEB_INDEX_PATH`) and inputs are constrained (escaped/encoded upstream, no XSS vector), it's cosmetic — worth aligning if `escapeHtmlText` is exported later.

PR #852, opened 2026-07-02
Release

fiddur:


apps/backend/src/services/safe-fetch.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const isBlockedIPv6 = (raw: string): boolean => {
+  const ip = raw.toLowerCase()
+  // Any address whose high bits are all zero (starts with `::`) is special-use:
+  // unspecified (::), loopback (::1), IPv4-mapped (::ffff:x), or IPv4-compatible
+  // (::x / ::7f00:1). Global unicast is 2000::/3 and never starts with `::`, so
+  // real hosts are unaffected while embedded-IPv4 tricks are all blocked.
+  if (ip.startsWith('::')) return true
+  if (ip.startsWith('64:ff9b:')) return true // NAT64 (embeds an IPv4 in the low bits)
```
fiddur, line 41: Minor / defense-in-depth: the docstring above claims "embedded-IPv4 tricks are all blocked," but `isBlockedIPv6` doesn't cover 6to4 (`2002::/16`, which embeds an arbitrary IPv4 in bits 16–47, e.g. `2002:c0a8:0101::` → 192.168.1.1) or Teredo (`2001:0::/32`). A host whose AAAA record resolves to one of these would pass the guard. Practical exploitability is low (reaching the embedded IPv4 needs a 6to4/Teredo route, which modern hosts lack), but it's cheap to add `2002:`/`2001:0:` checks alongside the NAT64 line to match the stated guarantee. Non-blocking.

fiddur (commented):
✅ Approved

**Release PR (develop → main)** — a large batch (265 files, ~25k insertions) of already-merged, individually-reviewed feature work: federated challenges, activity feed + ActivityPub, public shared dashboards, breakdown masking, a calories/BMR refactor, an expanded correlations engine, and food-item portions.

I reviewed the wiring/sync/dedupe layer, SSRF guard, breakdown masking, feed keypair, Sentry init, frontend redirect guards, chart-data HR-zone bucketing, metrics queries, and api-spec schema compliance directly, and fanned out deep reviews across the six major new subsystems (challenge federation, feed/ActivityPub, public sharing, calories, correlations, meals/food-items). **No blocking defects surfaced.** The security-critical boundaries are notably careful and well-tested: the public `/series` endpoint enforces its full authorization set in SQL and re-clamps ranges; public dashboards resolve minimal server-side figures with breakdown masking applied to both series names and their data; challenge federation applies the SSRF guard on every outbound fetch with security-critical axios options placed after the config spread so callers can't relax them; the ActorKeyPair upsert is race-safe; all new DB modules have integration tests; SQL is parameterized (dynamic SET clauses use hard-coded column whitelists); and snake_case / typedRouter / api-spec-single-source conventions hold throughout.

Only minor, non-blocking observations are worth recording:

- **safe-fetch SSRF (defense-in-depth):** `isBlockedIPv6` covers `::`-prefixed forms and NAT64 but not 6to4 (`2002::/16`, which embeds an arbitrary IPv4) or Teredo (`2001:0::/32`), while the docstring claims "embedded-IPv4 tricks are all blocked." Practical exploitability is low (reaching the embedded IPv4 requires 6to4/Teredo routing, which modern hosts don't have), but it contradicts the stated guarantee and is cheap to close. Inline note below. `dns.lookup` also has no timeout, so a stalling resolver can hang a request before axios's timeout applies.
- **Challenge local-join asymmetry:** the same-instance join path (`challenge-federation.ts`) writes the member directly without the `join_token` check or the `MAX_CHALLENGE_MEMBERS` cap that the remote register-back path enforces. Capability-consistent (the slug is unguessable and an authed local user adds only their own identity), so low impact.
- **Public boards expose owner-authored config strings:** a correlation widget's `activity` (which may be a location/tag value) and a trend/bar widget's `pattern` fallback title render to anonymous viewers with no masking option, unlike breakdown fields. These are values the owner chose to publish, so a follow-up enhancement rather than a broken guarantee.
- **Calories DST edge case:** per-minute BMR floor uses a fixed `bmr/1440` while the day loop iterates the actual local-day minute count, so the daily BMR floor is off by ~one hour of BMR on the two DST-transition days per year. Low impact.
- **`GET /challenges/:id/standings`** reads `req.query.refresh` without declaring the `QueryParams` generic / `validateQuery` per the typedRouter convention (trivial toggle).

None of these block the release; recommending merge.

PR #853, opened 2026-07-02
feat(sharing): dynamic Open Graph preview images via Satori (#834)

fiddur:
Part 2 of the **#834** epic (rich link previews). Builds on #851.

## What & why

Public share pages now expose a **dynamically rendered 1200×630 preview image** at `<resource-url>/opengraph-image.png`, so a pasted link shows a branded card instead of the generic default.

- **Satori** renders a branded card (title + `DASHBOARD`/`CHALLENGE`/`PROFILE` eyebrow + Aurboda wordmark) to SVG; **sharp** (already a dependency) rasterizes it to PNG — no headless browser, works in the Alpine image.
- `og:image` / `twitter:image` for public dashboards, challenges, and profiles now point at the per-resource endpoint. Non-public / unknown / invalid resources **redirect (302) to the branded static default**, so a private resource never yields a bespoke image.
- Fonts (**Liberation Sans**, OFL-1.1) are bundled under `src/assets/fonts/` since the production image has no system fonts.
- Rendered PNGs are memoised in a small bounded in-process cache and served `public, max-age=3600`.

## Sample render

Title + eyebrow + wordmark on the brand gradient (rendered by the pipeline in tests):

`DASHBOARD` → **Training & effect over two years** → ◎ Aurboda

## Design

- `services/og-image.ts` — pure render (`renderOgImage(card)` → PNG `Buffer`); Satori's element tree built by hand (no JSX/React runtime in the backend).
- `routes/og-image-router.ts` — visibility-aware, **dependency-injected** resolvers + renderer (unit-tested with fakes), mounted **before** the share-html router so image routes win over the generic `/u/:username/:slug` HTML route.

## Tests

38 unit tests (real Satori→sharp render asserts 1200×630 PNG; router asserts render/redirect/caching per visibility). `pnpm check` green.

## Follow-ups

3. Avatars · 4. Editable dashboard text · 5. oEmbed + polish


docs/features/sharing.md
```diff
@@ -87,11 +87,16 @@ get the full SPA and hydrate normally — only the head is enriched.
...
-- Every card references a 1200×630 preview image. For now this is a branded default
-  (`/og-default.png`); a dynamically generated per-resource card is a follow-up.
-- **Visibility is respected**: rich meta is emitted only for **public** resources.
-  Unlisted (slug-only) dashboards and unknown URLs get generic site meta, so an
-  unlisted resource's title never lands in a crawler's cache or a search index.
+- Every public resource has a **dynamically rendered 1200×630 preview image** at
+  `<resource-url>/opengraph-image.png`. Satori renders a branded card (title +
+  DASHBOARD/CHALLENGE/PROFILE eyebrow + Aurboda wordmark) to SVG and sharp
+  rasterizes it to PNG; fonts are bundled (no system fonts in the image). Renders
+  are memoised in-process and cached `public, max-age=3600`. Chart-less / non-public
```
fiddur: "Chart-less" is a bit confusing here — these cards never contain a chart (they're title + eyebrow + wordmark), so there's no chart-less vs. charted distinction. The actual fallback condition is simply non-public / unknown / invalid-username resources. Consider dropping "Chart-less /".

fiddur (commented):
✅ Approved

Solid, well-tested feature. The new `/u/.../opengraph-image.png` endpoints reuse the same `createShareResolvers()` and `is_public` gating as the share-HTML router, and are correctly mounted **before** it so the specific image routes win over the generic `/u/:username/:slug` HTML route. Visibility is respected end-to-end: only public dashboards/challenges/profiles render a bespoke card; unlisted/unknown/invalid all 302 to the static `/og-default.png`, and `share-meta` only points `image` at `resourceOgImage(url)` on the same public paths — so the meta URL and the rendered card can never diverge. Cache keys include the resource name, so a rename produces a fresh render rather than a stale one. Good unit coverage across the router (fakes), the renderer (real Satori+sharp PNG dimension checks), and the meta builders.

I checked the one thing that looked risky — user-controlled titles (dashboard/challenge names) containing emoji/CJK/etc. that aren't in Liberation Sans. Satori does **not** throw on missing glyphs when no `loadAdditionalAsset` is supplied (it falls back to the default font and renders `.notdef` boxes), so those titles produce tofu rather than a 500. No crash there.

Non-blocking hardening notes (fine as follow-ups):
- **No in-flight de-dup on a cold cache.** `sendImage` checks the cache, and only after `await renderImage(...)` completes does it populate the entry. N simultaneous requests for the same not-yet-cached key (e.g. a fresh share hit by several crawlers/clients at once) all miss and run N identical CPU-heavy Satori+sharp renders in parallel. A per-key in-flight `Promise` map would collapse those to one.
- **No render concurrency/rate limit on an unauthenticated CPU-heavy path.** These routes have no rate limiting, and with the 200-entry FIFO cache, rotating through >200 distinct public resources thrashes the cache and forces a re-render every request. Low risk for this app's scale, but worth keeping in mind.

Minor: one doc-wording nit inline.

apps/backend/src/routes/og-image-router.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (oldest !== undefined) cache.delete(oldest)
+      }
+      cache.set(key, png)
+      return png
+    } finally {
+      inFlight.delete(key)
+    }
+  }
+
+  const sendImage = async (res: Response, key: string, card: OgCard): Promise<void> => {
```
fiddur, line 75: Non-blocking: every other failure mode in this feature degrades gracefully to the static default image, but a `renderImage` rejection here (e.g. a transient sharp/Satori failure, or a title with a glyph the bundled fonts can't render) propagates as a 500 via Express 5's default error handler rather than falling back. Since the whole design goal is "always yield *some* image," consider wrapping `render()` and redirecting to `defaultOgImage(webHost)` on error so a broken render doesn't leave a public resource with no `og:image`. Fine to leave as a follow-up.

fiddur (commented):
✅ Approved

Solid, well-tested feature. The dynamic OG-image endpoint is a pure function of resolved card data, visibility is enforced consistently with the existing share-html router (only `is_public` resources render; everything else — unlisted, unknown, invalid username — redirects to the branded default, so nothing private leaks), and usernames are validated through the shared `isValidUsername` before any DB routing. The in-process cache is keyed on resource identity **plus name**, so a rename produces a fresh key rather than a stale image, and concurrent cold-cache misses collapse into a single Satori render via `inFlight`. Tests cover the render output, dimensions, memoisation, concurrency collapse, and each redirect path. All callers of the `share-meta` builders were updated to drop the now-unused `webHost` param, and docs are kept in sync.

A few things I verified: Express 5 forwards async-handler rejections to the default error middleware, so a Satori/sharp render failure returns a 500 (no hung request or process crash); `inFlight` is cleared in `finally` so failures aren't negatively cached. One non-blocking robustness note inline about degrading render failures to the default image rather than 500, to match the feature's "always fall back to the branded default" design. Not required for merge.

apps/backend/src/services/og-image.ts
```diff
@@ -0,0 +1,149 @@
...
+export interface OgCard {
+  kind: 'dashboard' | 'challenge' | 'profile'
+  title: string
+  subtitle?: string
+}
+
+const FONT_FAMILY = 'Liberation Sans'
+const fontDir = fileURLToPath(new URL('../assets/fonts/', import.meta.url))
+
+let fontsPromise: Promise<Font[]> | null = null
```
fiddur: Non-blocking: `fontsPromise` is a module-level mutable variable, which is in mild tension with the repo's "avoid module-level variables / singletons" guideline. It's a benign write-once memoization of immutable font data, so this is fine as-is; if strict adherence is preferred, the font loader could live in a closure created by the router/service factory instead.

fiddur (commented):
✅ Approved

Clean, well-tested feature. The dynamic OG-image endpoint correctly mirrors the visibility logic of the existing share-html router (only `is_public` resources render a bespoke card; private/unlisted/unknown redirect to the branded static default), so no private-resource title or image ever leaks. Nice touches: the concurrency-collapse `inFlight` map avoids N parallel CPU-heavy Satori renders on a cold cache, rejected renders are never cached (and degrade to the default rather than a 500), and the cache key embeds the resource `name` so renames don't serve stale titles. Route ordering (og-image mounted before share-html) is correct, and the `/opengraph-image.png` URL suffix can't collide with a real resource since dashboard/challenge slugs are random base64url tokens. DI is good — `renderImage` and the resolvers are injected, which the tests exercise with fakes. `share-meta.ts` was cleanly refactored to drop the now-unused `webHost` param and point `og:image` at `resourceOgImage(url)`, with tests updated to match.

Only a minor, non-blocking note below about a module-level mutable variable. Note: I couldn't run the suite locally (no `node_modules` in this worktree), but the tests read correctly and CI covers execution.

apps/backend/src/routes/og-image-router.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+      cache.set(key, png)
+      return png
+    } finally {
+      inFlight.delete(key)
+    }
+  }
+
+  const redirectToDefault = (res: Response): void => {
+    res.redirect(302, defaultOgImage(webHost))
```
fiddur, line 72: Minor: the default-image redirect sets no `Cache-Control`, while the rendered-image path uses `max-age=3600`. Every crawler hit on a non-public/unknown/invalid share URL will re-run the DB resolve and re-redirect with no caching hint. Consider adding a `Cache-Control` header here (even a shorter max-age) for consistency and to shed repeated load. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The Satori→sharp rendering is a pure function of resolved card data, visibility is re-checked on every request before rendering (private/unlisted/unknown/invalid all redirect to the branded default), and the bounded render cache + in-flight collapsing sensibly bound the CPU cost of an unauthenticated endpoint — renders only ever happen for the finite set of existing public resources. Routing precedence is correct: the OG router is mounted before the share-html router and its `/opengraph-image.png` suffix routes win over the generic `/u/:username/:slug` HTML route, and nginx already proxies the whole `/u/` subtree to the backend so the `.png` requests reach it. Conventions are respected (console.error matches the repo's logging style; plain `Router()` matches the sibling share-html-router precedent for non-JSON routers; snake_case fields preserved). Docs and the bundled-font license note are updated. I built api-spec, ran all new tests (27 passing) and `pnpm check` (0 errors) locally to confirm.

One design nuance worth noting (non-blocking): the "always yield some image" guarantee only covers render failures in `sendImage`'s try/catch — it does not cover the `resolveDashboard`/`resolveChallenge`/`profileExists` calls. An unexpected (non-`3D000`) DB error there rejects the handler and, under Express 5, is auto-forwarded to the default error handler as a 500 rather than degrading to the default image. This matches the existing share-html-router behavior, so it's not a regression, just a limit on the stated robustness.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. Dynamic OG cards are rendered with Satori→sharp behind a factory-created renderer, and the router is mounted before the share-html router so `/u/.../opengraph-image.png` wins over the generic HTML route.

Strengths worth calling out:
- **Visibility is respected end-to-end.** Only `is_public` dashboards/challenges/profiles get a rendered card; unlisted, unknown, and invalid resources 302 to the branded default, mirroring the share-html router's logic exactly — so a private resource's title never lands in a crawler cache.
- **`isValidUsername` is checked before any DB resolution**, consistent with the rule that a username selects the per-user database.
- **Robust failure/perf handling:** an in-process bounded cache plus an in-flight map collapses concurrent cold-cache misses into one CPU-heavy render, and any Satori/sharp failure degrades to the default image instead of a 500. Fonts are memoised inside a closure (no module-level state), matching the project's no-singleton guidance.

Verification I ran locally: `tsgo --noEmit` passes (0 errors), oxlint on the new files is clean, and the new tests pass — including the real Satori+sharp render producing a genuine 1200×630 PNG, the router's public/unlisted/unknown/invalid/degrade/concurrent-dedup cases, and the updated share-meta builders now pointing `og:image` at `<resource-url>/opengraph-image.png`.

One non-blocking FYI: only Liberation Sans (Latin) is bundled. I confirmed that a resource name containing CJK or emoji does **not** throw (so no 500 and no degrade) — Satori just renders those glyphs as blanks. That's an acceptable limitation given usernames are ASCII and names are usually Latin, but if non-Latin dashboard names become common you may want a fallback font or a `loadAdditionalAsset` hook. Nothing here blocks merge.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The dynamic OG image endpoint reuses `createShareResolvers` and mirrors `share-html-router`'s dashboard-before-challenge resolution exactly, so the HTML `og:image` URL and the image endpoint always agree on public/non-public — a private resource never yields a bespoke card. `nginx.conf` already proxies the whole `/u/` subtree to the backend, and the router is mounted before the share-html router, so `/u/.../opengraph-image.png` reaches this handler cleanly. The in-process cache (bounded FIFO) + `inFlight` dedup are correct: errors aren't cached, eviction is sound, and concurrent cold-cache misses collapse to a single render. Render/rasterize failures degrade to the branded default rather than 500-ing, which matches the documented "always yield an image" design. Fonts are bundled with proper OFL-1.1 attribution, module-level mutable state is avoided via closures, and the test coverage (public/unlisted/unknown/invalid, memoisation, concurrency collapse, failure degradation, real PNG dimensions) is thorough. Docs are updated.

- On a *transient* render failure for a genuinely public resource, `sendImage` falls back through `redirectToDefault`, which sets `Cache-Control: public, max-age=300` on the 302. So a one-off Satori/sharp hiccup can pin a public resource to the static default at the CDN/crawler layer for up to 5 minutes even after rendering recovers. Acceptable given the best-effort design, but you may want a shorter (or no) cache TTL specifically on the failure path vs. the deliberate non-public redirect.
- Now that every rendered card is a fixed 1200×630, consider emitting `og:image:width`/`og:image:height` in the share meta — some crawlers (Slack/Twitter/Facebook) lay out previews better when dimensions are declared. Optional enhancement, not required.
- `clampTitle` caps at 60 chars, but at `fontSize: 82` in the ~1040px usable width a much shorter title already wraps; Satori will wrap/overflow rather than clamp lines. Purely a visual-polish note.

None of these block merge.

PR #854, opened 2026-07-02
feat(feed): accept remote Follows and track followers (#831)

fiddur:
## Summary

Slice 3c of the federated activity feed ([#831](https://github.com/fiddur/aurboda/issues/831)): **inbound `Follow` handling**. A remote actor (e.g. your Mastodon account) that follows a user's actor now gets an `Accept` and becomes an established follower — the piece that makes a Follow *complete* rather than sit pending. Builds on the merged actor + WebFinger surface (#850).

- **`feed_follower` table + `db/feed-follower.ts`**: per-user store of remote followers keyed by actor URI, caching their inbox + optional shared inbox and an `accepted` flag. Idempotent upsert (re-follow updates in place), list, remove. **4 integration tests.**
- **`federation.ts` inbox listeners**: `.on(Follow)` verifies the Follow targets one of our actors, dereferences the sender, persists the follower, and sends an `Accept`; `.on(Undo{Follow})` drops the follower. **Fedify verifies the inbound HTTP Signature before these handlers run**, so a request that reaches them is authenticated as the sending actor.

## Testing
The follower DB layer is integration-tested (4 tests). The inbox handler is **type-checked against Fedify's `Actor` API** and verified end-to-end against the deployed instance (a real Mastodon follow) — constructing signed inbox requests in a unit test isn't practical, and Fedify owns the signature-verification security boundary. Existing actor/WebFinger tests still pass (23 AP+follower tests total); backend `tsgo`/oxlint clean.

## Scope
The Follow/Accept flow uses the **personal inbox** (`/users/<user>/inbox`, already nginx-proxied) — what remote servers use for a single-user follow. The shared-inbox optimization (nginx `/inbox` + `endpoints.sharedInbox` advertisement) and **outbound post delivery** follow in the delivery slice (3d), after which a follower's timeline actually receives the user's shared activities.


fiddur (commented):
🛑 Changes required

Solid, well-scoped slice: the `feed_follower` store is clean, correctly upserts by `actor_uri` (per-user DB, so the URI-as-PK is fine), and has good integration coverage for add/list/re-follow/remove. The inbox wiring reads correctly — the `Undo{Follow}` delete is keyed by the *authenticated* `undo.actorId`, so a remote can only drop its own follow, which is the right security posture. Two things to address before merge:

1. **Stale `FEDERATION.md`.** This PR makes the inbox process inbound activities and perform HTTP-signature-verified `Follow`→`Accept` / `Undo{Follow}` handling, but `FEDERATION.md` still states the opposite: line 38-40 says "the inbox accepts but does not yet process inbound activities", line 11-12 defers inbound signature verification to "the inbox/delivery slices", and line 44 lists `Follow`→`Accept` under **Planned**. The doc self-describes as "maintained incrementally as #831 lands" and CLAUDE.md requires docs to be kept in sync when changing related behavior, so these now-incorrect statements should move to "Currently implemented". (`FEDERATION.md` isn't in the diff, so I can't comment inline.)

2. **Inbox handlers don't guard `isMissingDatabase`** — see inline comment. The actor and key-pairs dispatchers in this same file carefully catch it, but the new `Follow`/`Undo` handlers don't, so a Follow (or Undo) targeting a valid-format username with no database throws instead of being ignored.

Minor / non-blocking: the new inbox handler logic itself (parseUri guards, the `Accept` send) has no test, unlike the DB layer — understandable since it needs signed requests, but worth a follow-up.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -59,10 +65,42 @@ export const createFeedFederation = (): Federation<void> => {
...
+    .on(Follow, async (ctx, follow) => {
+      // The Follow must target one of our actors.
+      if (follow.objectId == null) return
+      const target = ctx.parseUri(follow.objectId)
+      if (target?.type !== 'actor' || !isValidUsername(target.identifier)) return
+
+      const sender = await follow.getActor(ctx)
+      if (sender?.id == null || sender.inboxId == null) return
+
+      await upsertFeedFollower(target.identifier, {
```
fiddur: `isValidUsername` only checks the *format* of the identifier, not that the user's database exists. A Follow can be delivered to `/users/<validName>/inbox` (or the shared `/inbox` targeting `/users/<ghost>`) for any syntactically-valid username, so `upsertFeedFollower(target.identifier, …)` here — and `removeFeedFollower(...)` in the `Undo` handler at line 102 — will hit `getDbForUser`, fail to connect, and throw `3D000` (invalid_catalog_name), which propagates out of the handler as a 500 and invites remote retries.

The actor dispatcher (lines 40-45) and key-pairs dispatcher (lines 61-64) already handle exactly this by catching `isMissingDatabase(error)` and returning null/[]. The inbox handlers should do the same — wrap the DB call and swallow (return) on `isMissingDatabase` so a Follow to a nonexistent actor is ignored rather than erroring.

fiddur: Fixed: both the Follow and Undo handlers now wrap the DB call and return on isMissingDatabase (Follow also skips the Accept), so a Follow to a nonexistent actor is ignored rather than 500-ing — same pattern as the actor/key-pairs dispatchers.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -59,10 +65,55 @@ export const createFeedFederation = (): Federation<void> => {
...
+      // Answer the Follow so the remote server marks it established.
+      await ctx.sendActivity(
+        { identifier: target.identifier },
+        sender,
+        new Accept({ actor: follow.objectId, object: follow }),
+      )
+    })
+    .on(Undo, async (ctx, undo) => {
+      // Undo{Follow} — drop the follower.
+      const object = await undo.getObject()
```
fiddur: Non-blocking robustness note: `undo.getObject()` relies on the inner `Follow` being resolvable. Mastodon (and most major implementations) embed the full `Follow` object in the `Undo`, so this works. But a server that sends only the Follow's URI would make `getObject()` dereference it — and after an unfollow the remote often 404s that URI, so with the default `suppressError: false` the call throws, the handler 500s, and the follower is never removed (and the remote keeps retrying). Not a blocker for this slice given the Mastodon-embedded common case, but a candidate for hardening (e.g. `suppressError` + falling back to keying the removal purely on `undo.actorId` when the object can't be resolved).

fiddur (commented):
✅ Approved

Solid, well-scoped slice that adds inbound `Follow` → persist follower + `Accept` and `Undo{Follow}` → drop follower, plus a per-user `feed_follower` store.

**Correctness verified:**
- Schema wiring is complete: `feed_follower` is added to `socialTables`, registered in `tableCreationOrder`, and picked up both by fresh-DB init and by the `migrateSchema` create-missing-tables loop (so existing users get the table auto-created on first schema error), and by `cleanTestDb` for test isolation.
- Fedify API usage checks out. In particular `sender.endpoints?.sharedInbox?.href` is a valid *synchronous* access: `Endpoints` is an `entity: false` value type, so Fedify's codegen emits a sync `get endpoints()` (and `sharedInbox` is a scalar URL getter). `inboxId`/`id`/`objectId`/`actorId` are the correct sync accessors, and `getActor`/`getObject` are correctly awaited.
- Security/isolation is sound: `target.identifier` is gated through the strict `isValidUsername` regex before it reaches the DB layer, and the `Undo` handler keys removal on the HTTP-signature-authenticated `undo.actorId`, so a sender can only ever remove their own follow record (no cross-actor deletion). Missing-database (`3D000`) is swallowed so Follows to nonexistent actors don't 500/invite retries and don't emit a spurious Accept.
- The DB functions have matching integration tests covering insert/list, defaults, in-place upsert, and remove — consistent with the project's DB-testing requirement.

**Non-blocking observations (no changes required):**
- The inbox `Follow`/`Undo` handlers themselves have no direct test coverage (the `federation.integration.test.ts` harness only exercises the read surface). That's a harder surface to test (signed requests), and the DB layer is covered, so this is fine for the slice — worth a follow-up if signed-request test scaffolding is added later.
- `feed.md` still frames the ActivityPub inbox/follower layer as future work; a small doc refresh could note that inbound Follow/Accept + follower tracking now exist.

apps/backend/src/services/activitypub/federation.ts
fiddur: Applied the safe part: undo.getObject({ suppressError: true }) so an unresolvable inner Follow yields null and is ignored rather than 500-ing. I did not add the 'remove by undo.actorId' fallback since in our per-user-DB model the target identifier comes from the inner Follow's objectId, and removing by actorId alone couldn't tell an Undo{Follow} from an Undo of some other object — safer to leave the follower in the rare URI-only+404 case than to over-remove.

fiddur (commented):
✅ Approved

Solid, well-scoped slice. The DB layer (`feed-follower.ts`) is clean: upsert keyed on the `actor_uri` primary key with correct `ON CONFLICT` semantics (preserving `created_at`), and all three functions have integration coverage (add/list, defaults, in-place re-follow, remove) — satisfying the repo's "all db functions need integration tests" rule. Per-user databases make the `actor_uri` PK unambiguous, and the schema-error retry in `query()` transparently creates `feed_follower` on first use since it's now in `tableCreationOrder`.

The inbox handlers are correct and safe:
- **Follow**: target is derived from the signed `follow.objectId` (so it works for both the per-user and shared inbox), validated as one of our actors, and a Follow to a nonexistent actor is swallowed via `isMissingDatabase` so we don't 500/invite retries or send a bogus Accept. The Accept's `actor`/`object` are set correctly.
- **Undo{Follow}**: removal is keyed on the HTTP-signature-authenticated `undo.actorId`, not the forgeable inner Follow actor, so a sender can only drop their own follow — good authorization choice. `suppressError` on `getObject` avoids a 500 on an unresolvable inner object.

Verified the Fedify vocab getters (`endpoints`, `sharedInbox`, `inboxId`) are synchronous (`Endpoints | null` / `URL | null`), so the optional-chaining is type-correct.

1. `follow.getActor(ctx)` in the Follow handler is not guarded (unlike the deliberate `suppressError` hardening just added to the Undo handler); an unresolvable/timing-out remote actor would throw → 500 → retry. In practice Fedify has already resolved the signer to verify the signature, so this is low-risk, but the asymmetry with the Undo path is worth a thought.
2. The follower key stored on Follow is the *resolved* actor's canonical id (`sender.id.href`), while removal on Undo uses the *raw* `undo.actorId.href`. These are identical for conformant servers; only a non-canonical delivery URI would desync them.

Also note the inbox Follow/Undo handlers themselves aren't exercised by a test (only the DB store is) — understandable given HTTP-signature signing is awkward to fake, but a gap to keep in mind as delivery lands.

PR #855, opened 2026-07-02
update challenge.png

fiddur:


fiddur (commented):
✅ Approved

This PR replaces a single screenshot asset (`apps/web/public/screenshots/challenge.png`) with a smaller version — the file size drops from ~317KB to ~172KB. The new file is a valid PNG (2117×1215, 8-bit RGBA) and the aspect ratio is essentially unchanged, so it renders correctly against its only reference in `README.md` (which pins `width="800"`). No code, tests, schemas, or documented standards are affected. Looks good.

PR #857, opened 2026-07-02
fix(feed): pin federation canonical origin so actor URLs are https (#831)

fiddur:
## Summary

**Bug found by testing the deployed instance:** WebFinger for `@fiddur@aurboda.net` returns a valid JRD, but the actor `self` link is **`http://aurboda.net/users/fiddur`** — Fedify reconstructs the scheme from the incoming request, and behind aurboda.net's TLS-terminating proxy that request arrives as `http`. Mastodon requires **https** actor URLs, so remote lookups 404 ("user not found").

**Fix:** pass the public base URL (`WEB_HOST`) to `createFederation({ origin })`. Fedify then builds actor ids, WebFinger self-links, and inbox/outbox/followers URIs from that canonical https origin regardless of proxy headers.

## Testing
New integration test asserts that an **http-scheme** request (simulating the proxied backend) still yields **https** actor URLs (`id` and `inbox`). Existing actor/WebFinger tests still pass (4 total in the file); backend `tsgo`/oxlint clean.

This unblocks discovery/follow from Mastodon (it was previously blocked for every actor). Marking ready immediately — it's a small, deployed-behavior fix.


fiddur (commented):
✅ Approved

Clean, well-scoped bug fix. Pinning Fedify's canonical `origin` removes the dependence on request-derived scheme reconstruction (which yields `http` behind the TLS-terminating proxy) so actor ids, WebFinger self-links, and inbox/outbox URIs are always `https` — which Mastodon requires.

Verified the change is correct:
- Fedify `@fedify/fedify@2.3.1` accepts the string `origin` form and derives both the web origin (scheme + host) and the WebFinger handle host from it; it validates the value starts with `http(s)://` and has no path/query (fail-fast at startup on a misconfigured value).
- `webHost` (`https://aurboda.net`) is the right canonical origin, not `apiBaseUrl` (`.../api`): `nginx.conf` routes `/.well-known/webfinger` and `/users/` to the backend at the apex domain, so actor/WebFinger URLs must be built against the apex. This is also consistent with the host the previous request-derived behavior already produced — the fix only corrects the scheme.
- All three call sites of `createFeedFederation` are updated to the new required `origin` parameter.
- The added integration test faithfully reproduces the reported scenario (request arriving over `http://` still yields `https://` actor `id`/`inbox`).

Non-blocking note: in local dev, `WEB_HOST` defaults to `http://localhost:5173` (the vite dev server), so actor URLs would point at that port rather than the backend on `:3000`. Federation isn't exercised locally over real HTTPS, so this is harmless, and the test pins its own origin explicitly.

fiddur (commented):
✅ Approved

Correct, minimal fix for #831. Pinning Fedify's `origin` to `webHost` is the right approach: I verified in the installed `@fedify/fedify@2.3.1` types/source that `origin` is a valid `FederationOptions` field (since 1.5.0), that a string origin is parsed into `{ handleHost: host, webOrigin: origin }`, and that URL builders (`getActorUri`/`getInboxUri`/etc.) use `canonicalOrigin` (the pinned `webOrigin`) rather than the request-derived scheme — so actor ids come out `https` even when the request reaches the backend over loopback http behind the TLS-terminating proxy. Inbound routing is unaffected: Fedify matches an incoming URI against both `origin` and `canonicalOrigin`.

`webHost` is the appropriate value for the canonical origin (the feed reuses the same base-URL identity and `/u/<username>` namespace per `docs/features/feed.md`), and it's already in scope at api.ts:105. Both callers (`api.ts` and the integration test) were updated for the new required parameter, so nothing breaks. The added integration test exercises exactly the failure mode — an `http://` request yielding `https://` actor/inbox URLs.

One minor, non-blocking note: Fedify throws a `TypeError` at construction if the origin string contains a path/query, so a misconfigured `WEB_HOST` (e.g. with a `/subpath`) would fail-fast on startup. That's acceptable and consistent with how `WEB_HOST` is otherwise expected to be a bare origin (it's also used for WebAuthn origins). No changes required.

PR #858, opened 2026-07-02
fix(federation): route /.well-known/aurboda to the backend

fiddur:
## Summary

`https://aurboda.net/.well-known/aurboda` returned the SPA `index.html` instead of the cross-instance discovery JSON, because nginx had explicit `location =` blocks for the other `/.well-known/*` endpoints (oauth, assetlinks, webfinger) but **none for `/.well-known/aurboda`** — so it fell through to the SPA catch-all. (Reported: a 404/HTML there.)

Add an exact-match proxy to the backend (`well-known-router` already serves `/.well-known/aurboda` with `{ product, version, federation, api_base }`), matching the pattern of the sibling well-known routes.

This is the **challenge / shared-dashboard** federation discovery doc — separate from the ActivityPub feed work, but the same class of well-known-routing gap. No code change, nginx only.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -61,6 +61,16 @@ server {
         proxy_set_header X-Forwarded-Proto $scheme;
     }
 
+    # Aurboda cross-instance federation discovery (challenges / shared dashboards)
+    location = /.well-known/aurboda {
+        proxy_pass http://127.0.0.1:3000/.well-known/aurboda;
+        proxy_http_version 1.1;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+    }
+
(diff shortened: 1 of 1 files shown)


fiddur (commented):
✅ Approved

Straightforward, correct infra fix. The new `location = /.well-known/aurboda` block proxies to `http://127.0.0.1:3000/.well-known/aurboda`, which is exactly where the backend serves it: `createWellKnownRouter` is mounted at the app root (`apps/backend/src/api/rest-routes.ts:172`) and registers `GET /.well-known/aurboda` (`apps/backend/src/routes/well-known-router.ts:60`). The block mirrors the existing `webfinger`, `oauth-authorization-server`, and `assetlinks.json` blocks exactly (same `proxy_set_header` set, exact-match `location =`, path-preserving `proxy_pass`), and satisfies the deployment note in `docs/features/challenges.md:95-98` requiring `/.well-known/aurboda` to be reachable at the web base URL. The endpoint returns only public discovery metadata, so no auth/security concern. No issues found.

PR #859, opened 2026-07-02
feat(sharing): profile avatars — backend, OG card, AS2 icon (#834)

fiddur:
Part **3a** of the #834 epic (rich link previews). Adds user **avatars**, surfaced on shared-page OG cards and the ActivityPub actor `icon`. The web display/upload UI is the follow-up **3b** (this PR is backend-complete and usable via the API/MCP).

## Why Postgres storage
The deployment has no object store and no persistent app-container volume (watchtower auto-updates the image) — only Postgres persists. So avatars live in the user's own DB (`profile_avatar`, a singleton row), mirroring the existing `uploaded_icons` pattern.

## What
- **Upload / remove** (auth): `POST /profile/avatar` (multer, PNG/JPEG/WebP/GIF) normalizes to a square **256×256 WebP** via sharp; `DELETE /profile/avatar`.
- **Public read**: `GET /u/:username/avatar.png` — the stored avatar, or a **deterministic font-free identicon** derived from the username so an avatar *always* renders. Mounted before the share-html router so `avatar.png` isn't matched as a dashboard slug. Cached `max-age=3600`.
- **OG card**: the owner's avatar is embedded as a circular badge (data URI), fetched lazily on cache miss so cache hits stay cheap.
- **Federation**: the AS2 actor gains an `icon` pointing at the avatar endpoint, so Mastodon shows it (identicon fallback means it's never empty).

## Design
Pure/testable services (`avatar.ts` processing + identicon, `avatar-resolve.ts` stored-or-fallback); routers use DI. Visibility isn't a concern here — avatars are public profile images, and the identicon leaks nothing (deterministic from the public username).

## Tests (all green; `pnpm check` clean)
Avatar processing + identicon determinism, both routers (DI/mock + multipart upload), the resolver, and **DB integration** (upsert/get/delete singleton).

## Follow-ups
3b: web UI (settings upload + profile/shared display) · 4: editable dashboard text · 5: oEmbed + polish


apps/backend/src/services/avatar-resolve.ts
```diff
@@ -0,0 +1,23 @@
...
+  } catch (error) {
+    if (!isMissingDatabase(error)) throw error
+  }
+  return generateIdenticon(username)
+}
+
+/** The avatar as a `data:` URI, for embedding in a rendered OG card. */
+export const loadAvatarDataUri = async (username: string): Promise<string> => {
+  const image = await loadAvatarImage(username)
+  return `data:${image.content_type};base64,${image.data.toString('base64')}`
```
fiddur: **Confirmed bug (blocking).** This emits a `data:image/webp;...` URI for uploaded avatars (`processAvatar` always outputs `image/webp`), but Satori 0.26 cannot decode a WebP `` data URI. I verified locally by running the real `createOgImageRenderer`: a PNG avatar data URI renders, while the same image as WebP throws `TypeError: u is not iterable` in satori's image handler (`handler/image.ts`). Since `og-image-router`'s `sendImage` catches render errors and redirects to `og-default.png`, **every user who uploads an avatar loses their custom OG card**, while identicon (PNG) users are unaffected.

Fix: re-encode the embedded avatar to PNG so it matches the proven identicon path, e.g.
```ts
const png = image.content_type === 'image/png' ? image.data : await sharp(image.data).png().toBuffer()
return `data:image/png;base64,${png.toString('base64')}`
```
The `og-image-router` unit tests mock `renderImage`, so they miss this — worth adding a test that runs the real renderer with a WebP avatar.

apps/backend/src/routes/profile-router.ts
```diff
@@ -0,0 +1,66 @@
...
+    authMiddleware,
+    upload.single('avatar'),
+    async (req, res) => {
+      const user = req.user!
+      const file = req.file
+      if (!file) {
+        res.status(400).json({ error: 'No file uploaded', success: false })
+        return
+      }
+      if (!isAllowedAvatarType(file.mimetype)) {
```
fiddur, line 39: `isAllowedAvatarType` gates on `file.mimetype`, which multer derives from the client-supplied multipart `Content-Type` header. So the SVG rejection (explicitly asserted in `avatar.test.ts`) is bypassable: send SVG bytes with `Content-Type: image/png` and it passes this check, after which `sharp(buffer)` sniffs the real bytes and rasterizes the SVG anyway. Impact is limited because the output is always re-encoded to WebP (nothing SVG is stored or served), but if the intent is to genuinely refuse SVG, validate the format sharp actually detects — e.g. check `(await sharp(buffer).metadata()).format` against an allowlist — rather than the claimed mimetype.

fiddur (commented):
🛑 Changes required

Well-structured, well-documented feature with good unit coverage — I ran the five new/changed suites locally and all 26 tests pass, the multi-tenant storage/migration wiring is sound (`profile_avatar` is added to `tableCreationOrder`, so existing users get the table created on the schema-error retry path), and the federation actor `icon` resolves because `createFeedFederation` is called with the same `webHost` that serves `/u/:username/avatar.png`.

However there is one **confirmed blocking bug** in the headline path: uploaded avatars are stored/emitted as WebP, and Satori 0.26 cannot decode a WebP `data:` URI ``. I verified this locally by running the real `createOgImageRenderer` — a PNG avatar data URI renders fine, but the identical image as WebP throws `TypeError: u is not iterable` inside satori's image handler. Because the OG router catches render failures and redirects to the default image, the net effect is inverted from the intent: **users who upload an avatar lose their custom OG card entirely (fall back to `og-default.png`), while users with no avatar (identicon PNG) get a proper card.** The existing `og-image-router` tests mock `renderImage`, so they don't catch this. See the inline comment for a fix (re-encode the embedded avatar to PNG).

Also flagged a secondary hardening item: the SVG rejection is bypassable via Content-Type spoofing. Two minor, non-blocking notes not worth inline comments: (1) `public-avatar-router` doesn't wrap `loadAvatar` in a try/catch like the OG router does, so an unexpected (non-missing-DB) error yields a 500 instead of the "always render an image" identicon fallback (Express 5 forwards it to the error handler, so no hang); (2) uploaded avatars are served at `/u/:username/avatar.png` with `Content-Type: image/webp`, an extension/type mismatch that's harmless for browsers and Mastodon but slightly sloppy.

apps/backend/src/routes/og-image-router.ts
```diff
@@ -92,18 +97,20 @@ export const createOgImageRouter = (deps: OgImageDeps): Router => {
 
     const dashboard = await resolveDashboard(username, slug)
     if (dashboard?.is_public) {
-      return sendImage(res, `d:${username}/${slug}:${dashboard.name}`, {
+      return sendImage(res, `d:${username}/${slug}:${dashboard.name}`, async () => ({
+        avatarDataUri: await loadAvatarDataUri(username),
```
fiddur: Non-blocking: because `buildCard()` awaits `loadAvatarDataUri` before rendering, a transient avatar failure (DB connection blip, or a sharp re-encode error) now rejects the whole card and `sendImage` degrades a perfectly resolvable *public* dashboard/challenge/profile to the generic `og-default.png`. `loadAvatarImage` already swallows the common missing-DB case, so this is rare — but if you want the preview to stay resilient, you could resolve the avatar defensively (e.g. `avatarDataUri: await loadAvatarDataUri(username).catch(() => undefined)`) so a bad avatar falls back to the ring badge instead of blanking the title card. Same pattern applies to the challenge (109) and profile (123) branches.

fiddur (commented):
✅ Approved

Solid, well-tested addition. Storing avatars as a per-user singleton row (`profile_avatar`) is a sensible fit for the Postgres-only deployment, and the layering is clean: `avatar.ts` (process/identicon) → `avatar-resolve.ts` (resolve with identicon fallback) → the public read router, OG card, and AS2 actor icon. I checked the things most likely to bite:

- **Username handling is safe**: every path that resolves a username to a per-user DB (`public-avatar-router`, both OG routes) gates on `isValidUsername` first, and invalid names short-circuit to an identicon generated from the raw string without ever touching the DB. The username is never interpolated into the identicon SVG (only a hash-derived hue + rects), so no injection.
- **Content-Type spoofing is handled**: `processAvatar` rejects based on the format sharp actually detects (`ALLOWED_DETECTED_FORMATS`), not the client-claimed mimetype, and there's a dedicated test proving SVG-as-image/png is rejected.
- **Schema wiring is complete**: `profile_avatar` is registered in `socialTables`, `tableCreationOrder`, and `cleanTestDb`; the singleton PK + `CHECK` correctly pins one row per user DB. Integration tests cover upsert/fetch/delete.
- **Route ordering** (`public-avatar` before `og-image` before `share-html`) is correct so `avatar.png` isn't matched as a dashboard slug, and both `OgImageDeps` construction sites were updated for the new required `loadAvatarDataUri` field.
- **Docs updated** (`sharing.md`) and the OG renderer's WebP→PNG re-encode for Satori is justified and tested.

Two non-blocking observations (see inline): the OG cache key intentionally omits an avatar version (documented trade-off — a new avatar won't appear on already-cached cards until eviction/restart), and avatar-load failures currently degrade the *entire* preview to the branded default rather than rendering the card without the avatar. Both are consistent with the existing "always yield some image" philosophy. No MCP tool is added, which matches the precedent set by `icons-router` for binary uploads.

fiddur (commented):
✅ Approved

Solid, well-tested feature. Profile avatars are stored as a per-user singleton row (justified by the no-object-store deployment), uploads are normalized to a square 256×256 WebP, and every consumer (public `avatar.png`, OG card, ActivityPub `icon`) degrades gracefully to a deterministic identicon so an image always renders. I verified the full unit + router test suite passes (34 tests) after building `@aurboda/api-spec`.

Things I checked and found correct:
- **Write scoping / IDOR**: `POST`/`DELETE /profile/avatar` use `req.user!` → `query(user, …)`, so a caller can only touch their own database. Public reads validate `isValidUsername` before addressing a DB, and unknown users surface as `3D000` → identicon.
- **Content-type spoofing**: `processAvatar` re-detects the format from the bytes (not the client mimetype) and rejects SVG, so an `image/png`-labelled SVG can't slip through — covered by a test.
- **Failure isolation**: `avatarOrUndefined` (OG) and the public router's try/catch both fall back rather than 500/drop the card; `loadAvatarImage` only swallows missing-DB and rethrows the rest.
- **Schema/migration**: `profile_avatar` is registered in both `createTableStatements` (via `socialTables`) and `tableCreationOrder`, and the on-demand `migrateSchema` table-creation loop will create it for existing user DBs on first `42P01`.
- **Routing order**: `/u/:username/avatar.png` is mounted before the generic share-html router, so it isn't matched as a dashboard slug.
- **Federation `origin`**: `createFeedFederation(webHost)` matches the host that serves `/u/:username/avatar.png`, so the actor icon URL resolves.
- The router mirrors the existing `icons-router.ts` conventions (multer, inline `{ success }` types), so those aren't standard deviations.

Minor, non-blocking observations (fine to defer or leave as-is):
1. **`.png` URL serves `image/webp` bytes** for uploaded avatars (identicon fallback is real PNG). Interop is fine because Content-Type is authoritative for browsers and Mastodon, but the extension is technically misleading — a `mediaType` on the AS2 `Image` could make it explicit.
2. **No in-process cache on the public avatar endpoint**: every request re-reads the DB (stored) or re-rasterizes the identicon SVG via sharp. HTTP `max-age=3600` absorbs downstream repeats, but a cache-busting flood of distinct valid-form usernames forces repeated identicon renders. Low severity given how cheap the render is and that the heavier OG endpoint does cache.
3. **Multer size-limit errors** (>10 MB) propagate to the global error handler (likely 500) rather than a clean 4xx — same as the existing `icons-router`, so not a regression.
4. **MCP parity**: no MCP tool accompanies the new REST endpoint (per the parity guideline), though binary upload over MCP is impractical, so an intentional exception seems reasonable.

PR #860, opened 2026-07-02
feat(feed): shared-scalar producer + post-publish timestamp (#831)

fiddur:
## Summary

Delivery-prep for the federated feed ([#831](https://github.com/fiddur/aurboda/issues/831)) — two pure, tested building blocks the outbound-delivery slice will use.

- **`services/activitypub/scalars.ts` — `resolveSharedScalars`**: turns the user's chosen `included_metrics` keys into the `ScalarMetric[]` that `buildCreateExercise` renders into `content` and `aurboda:metrics`. Pure + dependency-injected: the caller supplies `metricStat(metric, stat)` (a metric aggregate over the activity window, backed by `queryMetricsBucketed` when wired to delivery). Supported keys: `duration` (from the window), `hr_zone_minutes` (per-zone minutes from `hr_zone_*_sec` sums), and `heart_rate_avg`/`heart_rate_max`, `stress_avg`, `distance` (m→km), `calories`. **Only requested keys that are supported AND have data are emitted**, so a post never advertises a summary it can't back.
- **`buildCreateExercise` `publishedAt`**: the deferred #846 nit — a post-creation timestamp drives the AS2 `published` of the Create + object (so remote timelines order by *share* time), while the workout time stays in `aurboda:startTime`. Defaults to `startTime`.

## Testing
6 scalar-producer unit tests (duration, rounding + units, m→km, hr-zone-minutes skipping empty zones, unsupported/no-data skipping, request-order + minimal querying) + 2 new `publishedAt` tests. 22 tests in the two files; backend `tsgo`/oxlint clean.

## Scope
Both are pure/DI, so no runtime wiring yet. The **delivery slice** binds `metricStat` to real window aggregates and wires share → `buildCreateExercise` → sign + deliver `Create` to followers (plus `/ns/activitystreams`, shared inbox, real outbox). The initial scalar-key vocabulary is intentionally small and extensible when the Share UI lands.

### DIFF
--- a/apps/backend/src/services/activitypub/scalars.ts
+++ b/apps/backend/src/services/activitypub/scalars.ts
@@ -1,0 +1,113 @@
+/**
+ * Resolve the shared scalar-summary values for a feed post.
+ *
+ * Turns the user's chosen `included_metrics` keys into the `ScalarMetric[]` that
+ * `buildCreateExercise` renders into `content` and `aurboda:metrics`. Pure and
+ * dependency-injected: the caller supplies `metricStat`, an aggregate of a
+ * metric over the shared activity's window (backed by `queryMetricsBucketed`
+ * with a single window-sized bucket when wired to delivery). Only the requested
+ * keys that are supported AND have data are emitted; everything else is skipped,
+ * so a post never advertises a summary it can't back.
+ */
+import type { MetricType } from '../../schema.ts'
+import type { ScalarMetric } from './object.ts'
+
+export type ScalarStat = 'avg' | 'max' | 'sum'
+
+/** Aggregate a metric over the activity window; `undefined` when there's no data. */
+export type MetricStat = (metric: MetricType, stat: ScalarStat) => number | undefined
+
+export interface ActivityWindow {
+  startTime: Date
+  endTime?: Date
+}
+
+const roundTo = (n: number, decimals = 0): number => {
+  const f = 10 ** decimals
+  return Math.round(n * f) / f
+}
+
+/** Scalar keys resolved from a single metric aggregate. */
+const METRIC_SCALARS: Record<
+  string,
+  { metric: MetricType; stat: ScalarStat; unit?: string; label: string; transform?: (raw: number) => number }
+> = {
+  calories: {
+    label: 'Calories',
+    metric: 'calories_active',
+    stat: 'sum',
+    transform: (v) => roundTo(v),
+    unit: 'kcal',
+  },
+  distance: {
+    label: 'Distance',
+    metric: 'distance',
+    stat: 'sum',
+    transform: (v) => roundTo(v / 1000, 2),
+    unit: 'km',
+  },
+  heart_rate_avg: {
+    label: 'Avg HR',
+    metric: 'heart_rate',
+    stat: 'avg',
+    transform: (v) => roundTo(v),
+    unit: 'bpm',
+  },
+  heart_rate_max: {
+    label: 'Max HR',
+    metric: 'heart_rate',
+    stat: 'max',
+    transform: (v) => roundTo(v),
+    unit: 'bpm',
+  },
+  stress_avg: {
+    label: 'Avg stress',
+    metric: 'stress_level',
+    stat: 'avg',
+    transform: (v) => roundTo(v),
+    unit: 'score',
+  },
+}
+
+/** Build the `hr_zone_minutes` record (minutes per zone with data) from zone-second sums. */
+const hrZoneMinutes = (metricStat: MetricStat): Record<string, number> => {
+  const zones: Record<string, number> = {}
+  for (let i = 0; i <= 5; i++) {
+    const seconds = metricStat(`hr_zone_${i}_sec` as MetricType, 'sum')
+    if (seconds !== undefined && seconds > 0) zones[`z${i}`] = Math.round(seconds / 60)
+  }
+  return zones
+}
+
+/**
+ * Resolve the requested scalar keys into `ScalarMetric[]`, in the order given.
+ * Supported keys: `duration`, `hr_zone_minutes`, plus the `METRIC_SCALARS` set
+ * (`heart_rate_avg`/`heart_rate_max`, `stress_avg`, `distance`, `calories`).
+ */
+export const resolveSharedScalars = (
+  window: ActivityWindow,
+  includedMetrics: string[],
+  metricStat: MetricStat,
+): ScalarMetric[] => {
+  const out: ScalarMetric[] = []
+  for (const key of includedMetrics) {
+    if (key === 'duration') {
+      if (window.endTime) {
+        const seconds = Math.round((window.endTime.getTime() - window.startTime.getTime()) / 1000)
+        out.push({ key, label: 'Duration', unit: 'seconds', value: seconds })
+      }
+      continue
+    }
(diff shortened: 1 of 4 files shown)


apps/backend/src/services/activitypub/scalars.ts
```diff
@@ -0,0 +1,113 @@
...
+    transform: (v) => roundTo(v),
+    unit: 'score',
+  },
+}
+
+/** Build the `hr_zone_minutes` record (minutes per zone with data) from zone-second sums. */
+const hrZoneMinutes = (metricStat: MetricStat): Record<string, number> => {
+  const zones: Record<string, number> = {}
+  for (let i = 0; i <= 5; i++) {
+    const seconds = metricStat(`hr_zone_${i}_sec` as MetricType, 'sum')
```
fiddur: Minor style nit (non-blocking): CLAUDE.md asks to avoid casting in favor of typed values. `api-spec` already exports a typed `hrZoneMetrics` (`['hr_zone_0_sec' … 'hr_zone_5_sec'] as const`), so you could iterate `hrZoneMetrics.entries()` and drop the `as MetricType` cast:

```ts
for (const [i, metric] of hrZoneMetrics.entries()) {
  const seconds = metricStat(metric, 'sum')
  if (seconds !== undefined && seconds > 0) zones[`z${i}`] = Math.round(seconds / 60)
}
```

fiddur (commented):
✅ Approved

Clean, small, and well-tested preparatory infrastructure for feed federation (#831).

- **`object.ts`**: adding an optional `publishedAt` so the AS2 `published` on both the `Create` and its object uses share time (for remote timeline ordering) while the workout time stays in `aurboda:startTime` is a sensible model. It defaults to `startTime` and is covered by two focused tests.
- **`scalars.ts`**: `resolveSharedScalars` is a pure, dependency-injected producer with good separation from the (upcoming) delivery layer. I hand-verified the arithmetic and transforms (2463 s duration, 8231 m → 8.23 km, 431.7 → 432 kcal, zone-seconds → minutes), the requested-order preservation, and the skip-unsupported/skip-no-data behavior. The supported keys line up with the documented `included_metrics` vocabulary in `feed.ts`, and the tests assert that only requested metrics are queried.

One non-blocking nit inline: the `as MetricType` cast in `hrZoneMinutes` runs against the project's "avoid casting, use type guards" standard, and a typed `hrZoneMetrics` constant already exists in api-spec that could be iterated instead. Not a correctness issue.

apps/backend/src/services/activitypub/scalars.ts
fiddur: Done — now iterates api-spec's typed `hrZoneMetrics.entries()`, dropping the `as MetricType` cast.

fiddur (commented):
✅ Approved

Clean, well-tested incremental work on the ActivityPub feed builder.

**`scalars.ts` (`resolveSharedScalars`)** — a pure, dependency-injected producer that turns a post's `included_metrics` scalar keys into `ScalarMetric[]`. I verified the logic against the tests by hand and it all holds: duration from the window (`2463s`), distance meters→km at 2dp (`8231→8.23`), rounded avg/max/sum aggregates, and the `hr_zone_minutes` record. The HR-zone indexing is correct: `hrZoneMetrics` begins at `hr_zone_0_sec` (index 0), so `hr_zone_N_sec` maps to `zN`. Only requested keys that are both supported and have data are emitted, and request order is preserved — good design for never advertising a summary the post can't back.

The key vocabulary is consistent end-to-end: `included_metrics` is `scalarMetricKeySchema` (`heart_rate_avg`, `hr_zone_minutes`, …), exactly what this resolver consumes, while `series_metrics` stays on `metricTypeSchema` matching `seriesLinks`.

**`object.ts` (`publishedAt`)** — the new optional field cleanly separates share time (AS2 `published` on both the Create and its object, for remote timeline ordering) from workout time (kept in `aurboda:startTime`), defaulting to `startTime` when omitted. Both new tests cover the set and defaulted paths.

Note (non-blocking): both `resolveSharedScalars` and the `publishedAt` input are not yet consumed by any caller — this is the expected incremental pattern, matching how `buildCreateExercise` itself is still a pure builder ahead of the delivery layer. Follows the project's functional/no-class, no-cast, snake_case conventions.

I could not execute the test suite (this worktree has no installed dependencies), so the pass is based on manual verification of the test math rather than an actual run.

fiddur (commented):
✅ Approved

Clean, well-scoped PR that adds a pure `resolveSharedScalars` producer plus a `publishedAt` timestamp for the AS2 `Create`/object. Both are functional, dependency-injected, and match the project's style (no classes, snake_case keys, DI for testability).

I traced the implementation against every test case:
- Duration (`2463s`), meters→km rounding (`8231 → 8.23`), calories/HR/stress rounding, and the order/only-what's-requested guarantees all hold.
- The HR-zone `z${i}` mapping looks off-by-one at first glance but is correct: `hrZoneMetrics[0]` is `hr_zone_0_sec`, so the array index equals the zone number (`hr_zone_1_sec → z1`, etc.).
- All referenced metric names (`calories_active`, `distance`, `heart_rate`, `stress_level`, `hr_zone_*_sec`) are valid `MetricType`s, and `hrZoneMetrics`/`MetricType` are correctly exported from `@aurboda/api-spec`.
- In `object.ts`, `published = input.publishedAt ?? input.startTime` is applied to both the `Create` and its object while `aurboda:startTime` retains the workout time; the two new tests confirm both the override and the default.

Both `resolveSharedScalars` and `buildCreateExercise` are producers not yet wired into the delivery layer (as their doc comments state), so there is no runtime behavior change in this PR. Coverage is good and I found no correctness or standards issues.

Note: I could not execute the tests here (dependencies aren't installed in the review checkout and install is sandbox-blocked), so verification is by static tracing rather than a live run.

fiddur (commented):
✅ Approved

Clean, well-scoped incremental change with solid unit coverage.

**`object.ts` — post-publish timestamp:** adds an optional `publishedAt` that drives the AS2 `published` on both the `Create` and its object, defaulting to `startTime` so timelines order by share time while the workout time stays in `aurboda:startTime`. Backward compatible, and both the override and default paths are covered by new tests.

**`scalars.ts` — shared-scalar producer:** a pure, dependency-injected `resolveSharedScalars` that maps the user's `included_metrics` keys to `ScalarMetric[]`. I verified the tricky bits:
- Duration math (endTime−startTime ⇒ 2463s) is correct.
- Distance is divided by 1000 (m→km), consistent with how `distance` is stored in meters elsewhere (e.g. `integrations/garmin/process.ts`).
- The HR-zone mapping is correct: `hrZoneMetrics` is 0-indexed (`hr_zone_0_sec`…`hr_zone_5_sec`), so `z${i}` yields `z1` for `hr_zone_1_sec`, matching the test expectations.
- Rounding, requested-order preservation, and silent skipping of unsupported/no-data keys all behave as documented.

All referenced metrics are valid `MetricType`s and `hrZoneMetrics` is re-exported from the `@aurboda/api-spec` root. Style follows the repo conventions (functional, DI, api-spec as source of truth). No blocking issues found.

Note (non-blocking): the producer isn't yet wired into the delivery layer — the doc comment flags this as intended follow-up. I couldn't execute the test suite in this environment (no `node_modules`), but the assertions were reviewed against the implementation and match.

PR #861, opened 2026-07-02
feat(feed): assemble the Create{Exercise} for a stored feed post (#831)

fiddur:
## Summary

Delivery slice **3d-a** — the piece that turns a stored `feed_post` into the AS2 `Create{Exercise}` the send path will deliver. Builds on the scalar producer (#860) and AS2 model (#846).

- **`buildFeedPostActivity(ctx, input, metricStat)`** bridges `feed_post` + its `activity` to the pure serializer: resolves the shared scalars over the activity window (via an injected `MetricStat`) and constructs the canonical federation URLs — `<origin>/users/<user>` actor, the post id, the `aurboda:` namespace, and the public `<apiBase>/public/<user>/series` href — then calls `buildCreateExercise` using the post's `created_at` as `publishedAt`.
- **`windowMetricStat(result)`** adapts a `queryMetricsBucketed` result into a `MetricStat`, **merging buckets** so date-bin splitting never skews the aggregate (sums add, max/min take the extreme, avg re-weights by count). This is what binds `metricStat` to real data in the send path.

## Testing
Both are pure/DI and unit-tested: URL construction (incl. trailing-slash trimming), scalar wiring, series href, `published` vs `aurboda:startTime`, and the bucket-merge math (weighted avg / summed / max). Backend `tsgo`/oxlint clean.

## Next (delivery sub-slices)
Wire share → `buildFeedPostActivity` → sign + `ctx.sendActivity` to followers (`Update`/`Delete` too); real outbox `OrderedCollection`; serve `/ns/activitystreams` (+ nginx `/ns/`); nginx `/inbox` + `endpoints.sharedInbox`.

### DIFF
--- a/apps/backend/src/services/activitypub/feed-activity.ts
+++ b/apps/backend/src/services/activitypub/feed-activity.ts
@@ -46,7 +46,7 @@ const trimSlashes = (s: string): string => s.replace(/\/+$/, '')
 /**
  * Merge a bucketed-metrics result into a `MetricStat` over the whole window.
  * Buckets are combined so date-bin splitting never skews the aggregate: sums
- * add, max/min take the extreme, and avg is re-weighted by sample count.
+ * add, max takes the extreme, and avg is re-weighted by sample count.
  */
 export const windowMetricStat = (result: QueryMetricsBucketedResult): MetricStat => {
   const merged = new Map<string, { sum: number; max: number; count: number; weighted: number }>()
--- a/apps/backend/src/services/activitypub/feed-activity.ts
+++ b/apps/backend/src/services/activitypub/feed-activity.ts
@@ -49,14 +49,13 @@ const trimSlashes = (s: string): string => s.replace(/\/+$/, '')
  * add, max/min take the extreme, and avg is re-weighted by sample count.
  */
 export const windowMetricStat = (result: QueryMetricsBucketedResult): MetricStat => {
-  const merged = new Map<string, { sum: number; max: number; min: number; count: number; weighted: number }>()
+  const merged = new Map<string, { sum: number; max: number; count: number; weighted: number }>()
   for (const bucket of result.buckets) {
     for (const [metric, stats] of Object.entries(bucket.metrics)) {
       if (!stats) continue
-      const m = merged.get(metric) ?? { count: 0, max: -Infinity, min: Infinity, sum: 0, weighted: 0 }
+      const m = merged.get(metric) ?? { count: 0, max: -Infinity, sum: 0, weighted: 0 }
       m.sum += stats.sum ?? 0
       m.max = Math.max(m.max, stats.max)
-      m.min = Math.min(m.min, stats.min)
       m.count += stats.count
       m.weighted += stats.avg * stats.count
       merged.set(metric, m)
--- a/apps/backend/src/services/activitypub/feed-activity.ts
+++ b/apps/backend/src/services/activitypub/feed-activity.ts
@@ -1,0 +1,101 @@
+/**
+ * Assemble the AS2 `Create{Exercise}` for a stored feed post.
+ *
+ * Bridges the persistence layer (`feed_posts` + `activities`) to the pure AS2
+ * serializer: resolves the shared scalar values over the activity window and
+ * builds the canonical federation URLs, then hands off to `buildCreateExercise`.
+ *
+ * The metric aggregation is dependency-injected (`MetricStat`) so this is
+ * unit-testable without a database; `windowMetricStat` adapts a
+ * `queryMetricsBucketed` result into that shape for the delivery path.
+ */
+import type { FeedVisibility } from '@aurboda/api-spec'
+
+import type { MetricType } from '../../schema.ts'
+import type { QueryMetricsBucketedResult } from '../queries/types.ts'
+import type { AS2Create } from './object.ts'
+import type { MetricStat, ScalarStat } from './scalars.ts'
+
+import { buildCreateExercise } from './object.ts'
+import { resolveSharedScalars } from './scalars.ts'
+
+/** Canonical federation URLs for a user, derived from the deploy's public hosts. */
+export interface FeedActivityContext {
+  username: string
+  /** Canonical web origin, e.g. `https://aurboda.net` (actor, post id, aurboda: ns). */
+  origin: string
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Clean, pure, dependency-injected module that bridges stored feed posts to the AS2 `buildCreateExercise` serializer, with good unit coverage for both exported functions.

I verified the aggregation logic against the query layer: `windowMetricStat` correctly re-weights averages by sample count, sums cumulative values, and returns `undefined` for metrics with no data so `resolveSharedScalars` skips them. The one theoretical concern — `sum` returning `0` instead of `undefined` when a bucket omits the optional `sum` field — can't misfire in practice, because `queryMetricsBucketed` only populates `sum` for the metrics in `sumMetrics` (distance, calories_active, hr_zone_*_sec), which are exactly the keys `resolveSharedScalars` ever queries with `stat: 'sum'`. URL/context assembly and slash-trimming look right and are asserted by the tests.

Field naming uses camelCase for the internal builder inputs, which is consistent with the existing `BuildCreateInput` in `object.ts`; the project's snake_case rule targets wire/persistence contracts (schemas, REST/MCP, DB, JSONB, frontend types), not internal service signatures, and the serialized `aurboda:metrics` keys stay snake_case. No blocking issues — only one trivial dead-code note below.

apps/backend/src/services/activitypub/feed-activity.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export const windowMetricStat = (result: QueryMetricsBucketedResult): MetricStat => {
+  const merged = new Map<string, { sum: number; max: number; min: number; count: number; weighted: number }>()
+  for (const bucket of result.buckets) {
+    for (const [metric, stats] of Object.entries(bucket.metrics)) {
+      if (!stats) continue
+      const m = merged.get(metric) ?? { count: 0, max: -Infinity, min: Infinity, sum: 0, weighted: 0 }
+      m.sum += stats.sum ?? 0
+      m.max = Math.max(m.max, stats.max)
+      m.min = Math.min(m.min, stats.min)
```
fiddur: Minor: `min` is accumulated here (and initialized to `Infinity` on line 56) but never returned — `ScalarStat` is `'avg' | 'max' | 'sum'`, and the returned closure only handles `sum`/`max`/avg. This tracking is dead code and could be dropped unless a `'min'` stat is planned.

fiddur: Fixed — dropped the unused min accumulation.

apps/backend/src/services/activitypub/feed-activity.ts
```diff
@@ -0,0 +1,100 @@
...
+  /** When the post was created (drives AS2 `published`). */
+  publishedAt: Date
+}
+
+const trimSlashes = (s: string): string => s.replace(/\/+$/, '')
+
+/**
+ * Merge a bucketed-metrics result into a `MetricStat` over the whole window.
+ * Buckets are combined so date-bin splitting never skews the aggregate: sums
+ * add, max/min take the extreme, and avg is re-weighted by sample count.
```
fiddur: Stale after the `min`-tracking removal (commit bdc7f50): the code now only tracks `max`, so "max/min take the extreme" should read "max takes the extreme". Minor.

fiddur (commented):
✅ Approved

Clean, well-scoped addition: `buildFeedPostActivity` bridges the persistence-shaped `FeedActivityInput`/`FeedActivityContext` to the pure `buildCreateExercise` serializer, and `windowMetricStat` adapts a bucketed-metrics result into the injected `MetricStat` shape. The design keeps the aggregation dependency-injected so the assembler is unit-testable without a DB, matching the pattern established in `scalars.ts`/`object.ts`.

I verified the arithmetic by hand against the tests: duration `07:11:03 − 06:30:00 = 2463s`, weighted avg `(140·10 + 160·30)/40 = 155`, distance sum `3000 + 5200 = 8200`, and the `-Infinity`/count-guard edge cases (empty buckets → `undefined`). URL construction (trailing-slash trimming, `/users/<name>`, `aurboda:` ns, public series base) all line up with the assertions. Dropping `min` tracking is correct since `ScalarStat` is `avg|max|sum` only.

I could not run the test suite locally (workspace `node_modules` absent and the sandbox blocks `pnpm install` with EROFS on the pnpm store), so this is a static review — CI should confirm. Only one minor doc nit below; no blocking issues.

apps/backend/src/services/activitypub/feed-activity.ts
fiddur: Fixed the comment to 'max takes the extreme'.

fiddur (commented):
✅ Approved

Clean, well-scoped building block that assembles the `Create{Exercise}` for a stored feed post. Follows the project conventions well: functional style, dependency injection (`MetricStat`) keeps `buildFeedPostActivity` DB-free and unit-testable, api-spec types are reused, and snake_case metric keys are preserved.

I verified the core logic against `queryMetricsBucketed`'s output shape (`metrics.ts:347-358`): bucket `avg` is the per-bucket sample mean, `count` the sample count, and `sum` is present only for cumulative metrics. So the bucket-merge in `windowMetricStat` is correct — sums add, `max` takes the extreme, and `avg` is re-weighted by count (`weighted / count`). The `max` seed of `-Infinity` never leaks because a metric only lands in the merged map when a bucket carried it (and `max` is always then a real number), and `stats.sum ?? 0` is safe since only cumulative metrics are queried with the `sum` stat.

The tests exercise the meaningful paths (weighted-avg merge, empty-window `undefined`, trailing-slash trimming, canonical `/users/` URLs, series href, scalar resolution) and the hand-checked expectations (duration 2463s, weighted avg 155, distance sum 8200) are correct.

The new functions aren't wired into a delivery/route path yet, which is consistent with this being an intermediate step in the #831 series — no parity concern since it isn't a user-facing endpoint.

Minor, non-blocking: a metric present in the window but with no `sum` in any bucket resolves `sum` to `0` rather than `undefined`, whereas an entirely-absent metric resolves to `undefined`. In practice the only `sum`-stat callers are cumulative metrics (distance/calories/HR-zone seconds) that always carry `sum`, so this can't surface today; worth keeping in mind only if a non-cumulative metric ever gets a `sum`-based scalar.

PR #862, opened 2026-07-02
feat(feed): serve a real followers collection from feed_follower (#831)

fiddur:
## Summary

Delivery slice **3d-b** — the actor's **followers dispatcher** now reads `feed_follower` and returns real recipients (`id` + `inboxId` + optional shared inbox), with a **counter** for `totalItems`. This does double duty:
- serves the actor's followers `OrderedCollection` (was empty), and
- enumerates recipients for the upcoming `sendActivity(…, 'followers', …)` fan-out.

Missing DB / invalid username degrade to an empty collection, matching the actor and key-pairs dispatchers.

## Testing
Integration test (via `federation.fetch`) seeds a follower and asserts the collection reports `totalItems: 1` with the follower's actor URI in `orderedItems`. 5 federation tests pass; backend `tsgo`/oxlint clean.

## Next
The outbound **push** — build the `Create` via `buildFeedPostActivity` and `ctx.sendActivity` it to followers on share (and `Update`/`Delete`) — is the next sub-slice; that's the part best verified against the live instance (follow from Mastodon, share, watch it arrive).

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -22,6 +22,7 @@ import { isValidUsername } from '../../api/auth-routes.ts'
 import {
   getOrCreateActorKeyPair,
   isMissingDatabase,
+  listFeedFollowers,
   removeFeedFollower,
   upsertFeedFollower,
 } from '../../db/index.ts'
@@ -123,12 +124,37 @@ export const createFeedFederation = (origin: string): Federation<void> => {
       }
     })
 
-  // Empty collections for now; the outbox serves the user's public posts once
-  // delivery wiring exists.
+  // Outbox stays empty until the delivery slice serves the user's public posts.
   federation.setOutboxDispatcher('/users/{identifier}/outbox', (_ctx, _identifier) => ({ items: [] }))
-  federation.setFollowersDispatcher('/users/{identifier}/followers', (_ctx, _identifier) => ({
-    items: [],
-  }))
+
+  // Real followers, backed by feed_follower. This both serves the followers
+  // collection and enumerates recipients for `sendActivity(..., 'followers', …)`.
+  federation
+    .setFollowersDispatcher('/users/{identifier}/followers', async (_ctx, identifier) => {
+      if (!isValidUsername(identifier)) return { items: [] }
+      try {
+        const followers = await listFeedFollowers(identifier)
+        return {
+          items: followers.map((f) => ({
+            endpoints: f.shared_inbox_uri ? { sharedInbox: new URL(f.shared_inbox_uri) } : null,
+            id: new URL(f.actor_uri),
+            inboxId: new URL(f.inbox_uri),
+          })),
+        }
+      } catch (error) {
+        if (isMissingDatabase(error)) return { items: [] }
+        throw error
+      }
+    })
+    .setCounter(async (_ctx, identifier) => {
+      if (!isValidUsername(identifier)) return 0
+      try {
+        return (await listFeedFollowers(identifier)).length
+      } catch (error) {
+        if (isMissingDatabase(error)) return 0
+        throw error
+      }
+    })
 
   return federation
 }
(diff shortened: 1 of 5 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -123,12 +124,37 @@ export const createFeedFederation = (origin: string): Federation<void> => {
...
+        }
+      } catch (error) {
+        if (isMissingDatabase(error)) return { items: [] }
+        throw error
+      }
+    })
+    .setCounter(async (_ctx, identifier) => {
+      if (!isValidUsername(identifier)) return 0
+      try {
+        return (await listFeedFollowers(identifier)).length
```
fiddur: Minor (non-blocking): the counter fetches every follower row and takes `.length`, and the dispatcher does its own full `listFeedFollowers` on the same request — so serving the collection runs two full `SELECT`s. Fedify's separate counter callback exists precisely so this can be a cheap `SELECT COUNT(*)`. Negligible for a small personal feed, but a dedicated `countFeedFollowers` (with its own integration test per the DB-testing rule) would avoid loading all rows just to count them.

fiddur (commented):
✅ Approved

Clean, well-scoped change that swaps the stubbed empty followers collection for real data backed by `feed_follower`. The returned shape matches Fedify's `Recipient`/`PageItems` contract (verified against `@fedify/vocab` — `endpoints` may be `null`), the DB schema (`apps/backend/src/schema/social.ts`) is consistent with `FeedFollowerRecord`, and both `isValidUsername` guarding and `isMissingDatabase` fallbacks mirror the surrounding dispatchers. The added counter is what makes `totalItems` show up in the served collection, and the integration test covers the happy path. Nothing blocking.

- The dispatcher and counter both return/count every row regardless of `accepted`. That's harmless today because the inbox handler auto-accepts every Follow inline (always `accepted: true`), but if a pending/manual-accept flow is ever added, the collection would leak un-accepted followers. Worth a follow-up when acceptance becomes conditional.
- The public followers collection now enumerates every follower's actor URI without authorization. This matches default Fedify/ActivityPub behavior and isn't a regression, but flagging in case follower-list privacy is a concern later.

apps/backend/src/services/activitypub/federation.ts
fiddur: Done — added countFeedFollowers (SELECT COUNT(*)) with its own integration test, and the counter now uses it instead of listFeedFollowers().length.

fiddur (commented):
✅ Approved

Solid, well-scoped slice. `countFeedFollowers` follows the established per-user-DB pattern (multi-tenancy comes from the per-user connection, so no `WHERE user` is needed — consistent with `listFeedFollowers`), and the followers dispatcher maps rows to Fedify `Recipient`s correctly (`id`, `inboxId`, and `endpoints.sharedInbox` gated on a stored shared inbox). The stored URIs originate from validated remote `href`s, so the `new URL(...)` calls won't throw. Error handling mirrors the rest of the file: `isMissingDatabase` → empty/zero, everything else re-thrown. Both new integration tests are meaningful (the counter test covers 0, multiple, and the re-follow de-dup case; the federation test exercises the real HTTP surface and asserts `totalItems` + `orderedItems`). Exports are wired through `db/index.ts` consistently.

- The collection is unpaginated — it loads all followers into memory both for serving and for `sendActivity(..., 'followers', …)`. Fine at current scale and the comment already flags delivery as a later slice; worth revisiting before large follower counts.
- The `::text` cast on `COUNT(*)` is redundant (node-pg already returns bigint counts as strings), but harmless and matches the `{ count: string }` typing.
- For a syntactically-valid but nonexistent username, `/users/<name>/followers` returns an empty 200 (DB-missing is caught) while the actor document 404s. Not exploitable for enumeration (an existing follower-less user is also empty) and it's an improvement over the previous unconditional empty collection, so no change needed here.

PR #863, opened 2026-07-02
feat(feed): resolve shared scalars from the DB over the activity window (#831)

fiddur:
## Summary

Delivery slice **3d-c(a)** — the DB-backed producer that feeds the outbound push.

`resolveActivityScalars(user, activity, includedMetrics)` aggregates the source metrics over the activity window (a single `queryMetricsBucketed` call, re-merged by `windowMetricStat` so PG bin-splitting can't skew it) and maps the user's `includedMetrics` selection into `ScalarMetric[]` for `buildFeedPostActivity`. Also exposes `SCALAR_SOURCE_METRICS` from the scalar producer as the set to aggregate (kept in sync with the supported keys).

## Testing
Integration test against real time-series: window duration + avg/max HR + HR-zone minutes, and omission of scalars whose metrics have no data. 11 activitypub tests pass; backend `tsgo`/oxlint clean.

## Next
The actual outbound **push** — build the `Create` and `sendActivity` it to followers on share (and `Update`/`Delete`) — is the next sub-slice. That step needs **live verification** (follow from Mastodon, share, watch it arrive).

### DIFF
--- a/apps/backend/src/services/activitypub/feed-activity.integration.test.ts
+++ b/apps/backend/src/services/activitypub/feed-activity.integration.test.ts
@@ -27,11 +27,18 @@ describe('resolveActivityScalars', () => {
 
   test('aggregates window metrics into the selected scalar summaries', async () => {
     const user = getTestUser()
-    await insertTimeSeries(user, [
-      { metric: 'heart_rate', source: 'garmin', time: new Date('2026-07-01T06:40:00Z'), value: 140 },
-      { metric: 'heart_rate', source: 'garmin', time: new Date('2026-07-01T06:50:00Z'), value: 160 },
-      { metric: 'hr_zone_2_sec', source: 'garmin', time: new Date('2026-07-01T06:45:00Z'), value: 1320 },
-    ])
+    // Only heart_rate is stored; HR-zone seconds are computed from it. Real
+    // devices sample densely — computeHrZoneSecs caps gaps at 5s — so insert a
+    // 5s cadence for ~5 min at 150 bpm.
+    await insertTimeSeries(
+      user,
+      Array.from({ length: 60 }, (_, i) => ({
+        metric: 'heart_rate' as const,
+        source: 'garmin' as const,
+        time: new Date(START.getTime() + 10 * 60_000 + i * 5_000),
+        value: 150,
+      })),
+    )
 
     const scalars = await resolveActivityScalars(user, { end_time: END, start_time: START }, [
       'duration',
@@ -41,9 +48,13 @@ describe('resolveActivityScalars', () => {
     ])
     const byKey = Object.fromEntries(scalars.map((s) => [s.key, s.value]))
     expect(byKey.duration).toBe(2400) // 40 min window
-    expect(byKey.heart_rate_avg).toBe(150) // (140+160)/2
-    expect(byKey.heart_rate_max).toBe(160)
-    expect(byKey.hr_zone_minutes).toEqual({ z2: 22 }) // 1320s → 22 min
+    expect(byKey.heart_rate_avg).toBe(150)
+    expect(byKey.heart_rate_max).toBe(150)
+    // hr_zone_minutes is derived from the heart_rate samples + effective zones,
+    // not read from time_series — so it resolves to a non-empty per-zone record.
+    const zones = byKey.hr_zone_minutes as Record<string, number>
+    expect(typeof zones).toBe('object')
+    expect(Object.values(zones).reduce((a, b) => a + b, 0)).toBeGreaterThan(0)
   })
 
   test('omits scalars whose metrics have no data in the window', async () => {
--- a/apps/backend/src/services/activitypub/feed-activity.integration.test.ts
+++ b/apps/backend/src/services/activitypub/feed-activity.integration.test.ts
@@ -1,0 +1,62 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest'
+
+/**
+ * Integration test for the DB-backed shared-scalar resolver: aggregates real
+ * time-series over the activity window and maps the user's selection.
+ */
+import { insertTimeSeries } from '../../db/time-series.ts'
+import { cleanTestDb, getTestUser, startTestDb, stopTestDb } from '../../test/db-test-helper.ts'
+import { resolveActivityScalars } from './feed-activity.ts'
+
+const CONTAINER_TIMEOUT = 120_000
+const START = new Date('2026-07-01T06:30:00Z')
+const END = new Date('2026-07-01T07:10:00Z')
+
+describe('resolveActivityScalars', () => {
+  beforeAll(async () => {
+    await startTestDb()
+  }, CONTAINER_TIMEOUT)
+
+  afterAll(async () => {
+    await stopTestDb()
+  })
+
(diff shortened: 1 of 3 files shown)


apps/backend/src/services/activitypub/feed-activity.ts
```diff
@@ -70,6 +71,30 @@ export const windowMetricStat = (result: QueryMetricsBucketedResult): MetricStat
...
+export const resolveActivityScalars = async (
+  user: string,
+  activity: { start_time: Date; end_time?: Date },
+  includedMetrics: string[],
+): Promise<ScalarMetric[]> => {
+  const start = activity.start_time
+  const end = activity.end_time ?? activity.start_time
+  // One bucket sized to the window; windowMetricStat re-merges if PG splits it.
+  const seconds = Math.max(1, Math.ceil((end.getTime() - start.getTime()) / 1000))
+  const result = await queryMetricsBucketed(user, SCALAR_SOURCE_METRICS, start, end, `${seconds}s`, {})
```
fiddur, line 97: `SCALAR_SOURCE_METRICS` includes the `hr_zone_*_sec` metrics, but `queryMetricsBucketed`/`getTimeSeriesBucketed` reads them straight from `time_series` — and HR-zone metrics are never stored there; they're computed from `heart_rate` + effective zones everywhere else (`period-summary.ts` `computeHrZoneStats`, `chart-data.ts` `queryHrZoneBuckets`, which both special-case `isHrZoneMetric`). So in production there are no `hr_zone_*_sec` rows and `hr_zone_minutes` always resolves empty, even when heart-rate data exists. This path needs to derive zone seconds from `heart_rate` (e.g. `computeHrZoneSecs` + `getEffectiveHrZones`) instead of expecting them from the bucketed query.

apps/backend/src/services/activitypub/feed-activity.integration.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  beforeEach(async () => {
+    await cleanTestDb()
+  })
+
+  test('aggregates window metrics into the selected scalar summaries', async () => {
+    const user = getTestUser()
+    await insertTimeSeries(user, [
+      { metric: 'heart_rate', source: 'garmin', time: new Date('2026-07-01T06:40:00Z'), value: 140 },
+      { metric: 'heart_rate', source: 'garmin', time: new Date('2026-07-01T06:50:00Z'), value: 160 },
+      { metric: 'hr_zone_2_sec', source: 'garmin', time: new Date('2026-07-01T06:45:00Z'), value: 1320 },
```
fiddur: This inserts a raw `hr_zone_2_sec` sample, which is why the `hr_zone_minutes` assertion below passes — but HR-zone metrics are never stored in `time_series` in production (they're computed from `heart_rate`). The test therefore validates behavior that can't occur with real data and hides the fact that `hr_zone_minutes` won't resolve. Consider inserting `heart_rate` samples and letting the resolver compute zones (once the resolver is fixed to do so).

fiddur (commented):
🛑 Changes required

The `duration`, `heart_rate_avg/max`, and `stress_avg` scalars are resolved correctly, and `windowMetricStat`'s re-merge across buckets is sound. But there's a correctness bug in the `hr_zone_minutes` path.

**`hr_zone_minutes` will never resolve in production.** `SCALAR_SOURCE_METRICS` includes the `hr_zone_*_sec` metrics and passes them to `queryMetricsBucketed`, but HR-zone metrics are *not stored* in `time_series` — the rest of the codebase computes them from `heart_rate` samples + the user's effective zones. `queryMetricsBucketed` → `getTimeSeriesBucketed` just runs `SELECT ... FROM time_series WHERE metric = ANY($1)` and does no zone computation (unlike `getPeriodSummary`'s dedicated `computeHrZoneStats`, or `chart-data.ts`'s `queryHrZoneBuckets`, both of which special-case `isHrZoneMetric`). So for any real activity there are no `hr_zone_*_sec` rows and `hrZoneMinutes()` always returns `{}`, meaning the `hr_zone_minutes` summary is silently dropped even when heart-rate data exists. The integration test passes only because it inserts synthetic `hr_zone_2_sec` rows directly, which never occur in production data. To resolve `hr_zone_minutes`, this path needs to compute zone seconds from `heart_rate` (reuse `computeHrZoneSecs` + `getEffectiveHrZones`) rather than expecting them from `queryMetricsBucketed`.

**Minor / worth verifying:** `distance` and `calories_active` are cumulative metrics, so `queryMetricsBucketed` filters them to trusted sources (`cumulativeSources` / `aurbodaOnlySources`) via `querySplitByCumulative`. If exercise distance/calories are stored under a non-trusted source (e.g. raw `garmin`), those scalars won't resolve either. Please confirm the feed's expected data sources satisfy that filter (this is pre-existing `queryMetricsBucketed` behavior, so not necessarily wrong — just a gap the current tests don't cover).

apps/backend/src/services/activitypub/feed-activity.ts
fiddur: Fixed: hr_zone_minutes now derives zone seconds from heart_rate + getEffectiveHrZones/computeHrZoneSecs (hr_zone_*_sec removed from the queried set), and the test inserts densely-sampled heart_rate instead of a raw hr_zone row.

apps/backend/src/services/activitypub/feed-activity.integration.test.ts
fiddur: Fixed: hr_zone_minutes now derives zone seconds from heart_rate + getEffectiveHrZones/computeHrZoneSecs (hr_zone_*_sec removed from the queried set), and the test inserts densely-sampled heart_rate instead of a raw hr_zone row.

fiddur (commented):
✅ Approved

Clean, well-scoped building block that wires the shared-scalar resolver to the database. `resolveActivityScalars` correctly runs a single window-sized bucketed query for the metric-backed scalars (relying on `windowMetricStat` to re-merge if Postgres splits the bucket) and derives HR-zone seconds from raw `heart_rate` samples via `computeHrZoneSecs` — mirroring the existing `activity-detail.ts` pattern, so it stays consistent with how the rest of the app computes zones. Nice details: the `zoneIndex !== undefined` guard correctly handles zone 0 (falsy), HR-zone computation is gated behind both the `hr_zone_minutes` request and a defined `end_time`, and `SCALAR_SOURCE_METRICS` is a deterministic, deduped set that deliberately excludes the computed HR-zone metrics. The integration tests cover both the aggregation and the omit-when-no-data paths. Neither `resolveActivityScalars` nor `buildFeedPostActivity` is wired into a delivery path yet, which is fine for stacked work under #831.

Two minor non-blocking notes (no changes required):
- The doc comment on `resolveActivityScalars` says an activity with no end "has no window, so only window-independent keys (none, currently) resolve." In practice `end` falls back to `start`, and the bucketed query uses `time >= start AND time <= end`, so a sample landing exactly on `start_time` could still resolve e.g. `heart_rate_avg`. Extremely unlikely edge case, but the comment slightly over-claims.
- When `hr_zone_minutes` is requested, `heart_rate` is read from the DB twice (once bucketed for avg/max, once raw for zones). Acceptable since the zone computation needs raw samples, just worth being aware of.

PR #864, opened 2026-07-02
feat(feed): deliver shared activities to followers over ActivityPub (#831)

fiddur:
## Summary

Delivery slice **3d-c** — **sharing an activity now fans a `Create{Note}` out to the user's followers.** This is the step that puts a shared activity into a follower's Mastodon timeline.

- **`services/activitypub/deliver.ts` — `deliverFeedPost`**: resolves the shared scalars over the activity window, builds a Mastodon-compatible Fedify `Create{Note}` (HTML `content` summary + `name` + `url`, addressed per visibility via a pure `recipients` helper), and sends it with `ctx.sendActivity(…, 'followers', …)` — Fedify signs, dedupes shared inboxes, and queues with retries. Extracted a shared `feedPostContent` helper so the pushed Note and the AS2 object model read identically.
- **Wiring**: the feed router takes an injected fire-and-forget `FeedDeliver`; `api.ts` builds the Fedify federation **once** (shared with the actor mount) and delivers on share, catching/logging failures so it never blocks the HTTP response.

## Testing
Unit-tested the `recipients` addressing (public / unlisted / followers). The send path is **type-checked against Fedify's vocab** but **needs live verification** — constructing signed federation delivery can't be unit-tested. **To verify:** follow `@<user>@<host>` from Mastodon, share an activity (MCP `share_activity` or `POST /feed/activities/:id/share`), and watch it appear in the follower's timeline.

## Deferred / notes
- **`published`** is omitted on the pushed Note: Fedify types it as the ambient `esnext.temporal` `Temporal.Instant`, which the `@js-temporal/polyfill` value isn't assignable to. Delivery fires right after the share, so remote servers timestamp at receipt (≈ share time). Wiring an explicit `published` is a follow-up once the Temporal-lib interop is sorted.
- The `aurboda:` **structured extension** isn't carried on the pushed Note (Fedify's typed vocab drops unknown properties) — progressive enhancement via dereferencing the object id is a follow-up.
- **`Update`/`Delete`** propagation and the **real outbox** are follow-ups (3d-d).

### DIFF
--- a/apps/backend/src/services/activitypub/deliver.ts
+++ b/apps/backend/src/services/activitypub/deliver.ts
@@ -1,0 +1,89 @@
+import type { FeedVisibility } from '@aurboda/api-spec'
+/**
+ * Deliver a shared feed post to the user's followers over ActivityPub.
+ *
+ * Builds a Fedify `Create{Note}` — the Mastodon-compatible representation: an
+ * HTML `content` summary + `name`/`url`, addressed per the post's visibility —
+ * and fans it out via `ctx.sendActivity(..., 'followers', …)`, which Fedify
+ * signs, dedupes by shared inbox, and queues with retries.
+ *
+ * The custom `aurboda:` structured extension is not carried on the pushed Note
+ * (Fedify's typed vocab drops unknown properties); it's progressive enhancement
+ * an Aurboda consumer fetches by dereferencing the object id — served in a
+ * follow-up slice. Delivery is best-effort: callers invoke it fire-and-forget.
+ */
+import type { Federation } from '@fedify/fedify'
+
+import { Create, Note } from '@fedify/fedify/vocab'
+
+import { resolveActivityScalars } from './feed-activity.ts'
+import { feedPostContent } from './object.ts'
+
+const PUBLIC_URI = new URL('https://www.w3.org/ns/activitystreams#Public')
+
+/** AS2 `to`/`cc` addressing (as URLs) for a post's visibility. */
+export const recipients = (visibility: FeedVisibility, followers: URL): { to: URL[]; cc: URL[] } => {
+  switch (visibility) {
+    case 'public':
+      return { cc: [followers], to: [PUBLIC_URI] }
+    case 'unlisted':
+      return { cc: [PUBLIC_URI], to: [followers] }
+    case 'followers':
+      return { cc: [], to: [followers] }
+  }
+}
+
+export interface FeedDeliveryDeps {
+  federation: Federation<void>
+  /** Canonical web origin, e.g. `https://aurboda.net`. */
+  origin: string
+}
+
+export interface DeliverablePost {
+  id: string
+  included_metrics: string[]
+  visibility: FeedVisibility
+  created_at: Date
+}
+
+export interface DeliverableActivity {
+  activity_type: string
+  start_time: Date
+  end_time?: Date
+  title?: string
+}
+
+/** Build and send the `Create{Note}` for a freshly-shared post to its followers. */
+export const deliverFeedPost = async (
+  deps: FeedDeliveryDeps,
+  user: string,
+  post: DeliverablePost,
+  activity: DeliverableActivity,
+): Promise<void> => {
+  const scalars = await resolveActivityScalars(user, activity, post.included_metrics)
+  const { content, name } = feedPostContent(activity.title, activity.activity_type, scalars)
+
+  const ctx = await deps.federation.createContext(new URL(deps.origin))
+  const actorUri = ctx.getActorUri(user)
+  const followers = ctx.getFollowersUri(user)
+  const postUrl = new URL(`${actorUri.href}/feed/${post.id}`)
+  const { cc, to } = recipients(post.visibility, followers)
+
+  // `published` is intentionally omitted: Fedify's vocab types it as the ambient
+  // (esnext.temporal) Temporal.Instant, which the @js-temporal polyfill value
+  // isn't assignable to. Delivery fires right after the share, so remote servers
+  // timestamp it at receipt (≈ share time). Wiring an explicit published is a
+  // follow-up (needs Temporal-lib interop sorted).
+  const note = new Note({
+    attribution: actorUri,
+    ccs: cc,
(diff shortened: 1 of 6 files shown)


apps/backend/src/services/activitypub/deliver.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { Create, Note } from '@fedify/fedify/vocab'
+
+import { resolveActivityScalars } from './feed-activity.ts'
+import { feedPostContent } from './object.ts'
+
+const PUBLIC_URI = new URL('https://www.w3.org/ns/activitystreams#Public')
+
+/** AS2 `to`/`cc` addressing (as URLs) for a post's visibility. */
+export const recipients = (visibility: FeedVisibility, followers: URL): { to: URL[]; cc: URL[] } => {
```
fiddur, line 27: Non-blocking: this is the same visibility→`to`/`cc` mapping as `addressingFor` in `object.ts`, just returning `URL[]` instead of `string[]`. Consider having one derive from the other (e.g. `recipients` maps `addressingFor`'s strings through `new URL(...)`, with the followers URL passed in) so the two addressing tables can't drift. The exhaustive switch keeps drift compile-checked, so this is purely a maintainability nicety.

fiddur (commented):
✅ Approved

Solid, well-scoped slice that wires up fire-and-forget ActivityPub delivery of shared feed posts. The design is clean: the router stays decoupled from Fedify via the injected `FeedDeliver` type, delivery never blocks the response, and errors are caught + logged rather than swallowed silently. I verified the potentially-risky parts:

- **Type safety / Fedify API**: `FeedPostRecord`/`Activity` structurally satisfy `DeliverablePost`/`DeliverableActivity`; the `Note`/`Create` constructor params (`attribution`, `tos`, `ccs`, `url`, `id`, `object`, `actor`) all exist in `@fedify/vocab@2.3.1`; `sendActivity({identifier}, 'followers', create)` matches the documented collection overload (and the followers dispatcher is registered); and omitting the `void` contextData arg to `createContext` type-checks. So this compiles against the pinned Fedify.
- **Fire-and-forget after response is safe**: `dbByUser` in `db/connection.ts` is a process-lifetime per-user client map (not request-scoped AsyncLocalStorage), so `resolveActivityScalars(user, …)` still resolves the correct tenant connection even though it runs after `res.json()`.
- **Addressing is correct**: `public`/`unlisted`/`followers` map to the right `to`/`cc` and never leak a `followers`-only post to `Public`; covered by `deliver.test.ts`.
- **ID consistency**: the delivered Create id (`…/feed/{id}`) and Note id (`…/feed/{id}/object`) exactly match what `buildCreateExercise` in `object.ts` produces, so the ids will resolve once the object-serving endpoint lands.
- The `feedPostContent` extraction is a faithful refactor of the previous inline logic — no behavior change.

Minor, non-blocking observations (all fine to defer):
- `recipients` in `deliver.ts` is a near-exact duplicate of `addressingFor` in `object.ts` (identical visibility→to/cc logic, differing only in URL vs string). Low drift risk since both use exhaustive switches over `FeedVisibility`, but consolidating would remove the parallel-maintenance burden. See inline note.
- `DeliverablePost.created_at` is declared but unused (presumably reserved for the deferred `published` wiring). Harmless dead field.
- The object-dereference endpoint and durable (Postgres-backed) queue are explicitly deferred to follow-up slices — Mastodon renders inlined objects from the authoritative origin without refetching, so initial delivery works, but strict AP consumers that refetch by id will drop the object until that endpoint exists, and the in-memory queue loses undelivered activities on restart. Both are documented in the code.

Nothing blocking; approving.

apps/backend/src/services/activitypub/deliver.ts
fiddur: Done — recipients() now derives from addressingFor() (mapping its strings through new URL), so the two addressing tables share one source.

fiddur (commented):
✅ Approved

Solid, well-scoped slice. The delivery path is correct and the code is clean.

What I verified:
- **URL/id consistency** — `deliverFeedPost` builds `<actor>/feed/<postId>` for the `Create` id and `<actor>/feed/<postId>/object` for the `Note` id, exactly matching the JSON object model in `buildCreateExercise`, so the pushed `Create{Note}` and the (follow-up) dereferenceable object won't drift.
- **Fire-and-forget DB access is safe** — delivery runs after the response returns, but `getDbForUser` resolves per-user connections from a module-level cache keyed by the username string (no request-scoped `AsyncLocalStorage`), so `resolveActivityScalars(user, …)` reaches the correct DB.
- **Delivery isn't a silent no-op** — there is no explicit `federation.startQueue()`, but `manuallyStartQueue` defaults to `false` and Fedify's `sendActivity` auto-starts the queue worker on enqueue, so the `InProcessMessageQueue` gets drained. (Caveat, already acknowledged in `federation.ts`: the in-memory queue is at-most-once and non-durable across restarts — fine for best-effort.)
- **Privacy** — only `post.included_metrics` scalars are rendered into `content`; series links and map/chart are not emitted here, so nothing beyond the user's explicit scalar selection leaves the instance.
- **Refactor** — extracting `feedPostContent` from `buildCreateExercise` is behavior-preserving.

Recommendations (non-blocking):
- **Docs**: `docs/features/feed.md` (lines 8–11) still states the ActivityPub delivery layer — explicitly including *follower delivery* — is "intentionally not part of this foundation" and is "follow-up work." This PR implements exactly that, so the statement is now false. The project's CLAUDE.md requires docs to be kept in sync when sync/delivery behavior changes; please update feed.md (and the stale "Outbound delivery … is a later slice" comment in `federation.ts`) to reflect that follower delivery now exists.
- **Test coverage**: the new test only exercises `recipients`, which is a thin re-mapping of the already-tested `addressingFor`. The actual orchestration in `deliverFeedPost` (URL assembly, `feedPostContent` wiring, `sendActivity` call) is untested. That's understandable given the Fedify dependency, but a small test injecting a fake `Federation` to assert the built `Create`/`Note` ids, `to`/`cc`, and recipient collection would lock in the behavior this slice adds.

PR #866, opened 2026-07-02
fix(feed): deliver activities synchronously (drop the undrained queue) (#831)

fiddur:
## Summary

**Live-verification bug fix.** Sharing an activity created the post + exposed the public series, and the `Follow`→`Accept` completed — but the shared `Create` never reached the follower's timeline.

**Cause:** `createFederation` was given an `InProcessMessageQueue`, but nothing calls `federation.startQueue()` (the `@fedify/express` integration doesn't run a queue worker). So `ctx.sendActivity(…, 'followers', …)` **enqueued** the `Create` and it was never drained/sent. The inbound `Accept` worked only because it fires *inside* Fedify's inbox-request lifecycle.

**Fix:** drop the `queue` option so `sendActivity` delivers **synchronously** (awaits the outbound POST) — the documented alternative. Share delivery is already fire-and-forget in the router, so this doesn't block the HTTP response. Actor / WebFinger / followers still work (tests green).

Durable, **retried** delivery via a persistent Postgres queue + worker is a deliberate later reliability slice.

## Verification
After this deploys, re-sharing an activity should land the status in a follower's Mastodon home timeline. (The bug was invisible to unit tests — federation delivery only fails end-to-end, which is why this needed the live check.)

### DIFF
--- a/apps/backend/src/services/activitypub/deliver.ts
+++ b/apps/backend/src/services/activitypub/deliver.ts
@@ -5,7 +5,9 @@ import type { FeedVisibility } from '@aurboda/api-spec'
  * Builds a Fedify `Create{Note}` — the Mastodon-compatible representation: an
  * HTML `content` summary + `name`/`url`, addressed per the post's visibility —
  * and fans it out via `ctx.sendActivity(..., 'followers', …)`, which Fedify
- * signs, dedupes by shared inbox, and queues with retries.
+ * signs and dedupes by shared inbox. Delivery is synchronous (no message queue
+ * configured), so this awaits the outbound POSTs; retry/durability via a
+ * persistent queue is a later slice.
  *
  * The custom `aurboda:` structured extension is not carried on the pushed Note
  * (Fedify's typed vocab drops unknown properties); it's progressive enhancement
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Correct, well-scoped fix. Removing `InProcessMessageQueue` makes Fedify's `sendActivity` deliver synchronously, which is the right call given the wiring here.

I verified the reasoning against the installed Fedify 2.3.1 source: `Context.sendActivity` sends immediately (awaiting the outbound POSTs) only when `outboxQueue == null`; with a queue configured it enqueues and relies on `startQueue()` to drain. `startQueue()` is never called anywhere in the backend, and `@fedify/express`'s `integrateFederation` only calls `federation.fetch()` (never `startQueue()`), so the previous in-process queue would have silently dropped the `Create` on share. The new comments accurately describe this.

Other checks:
- The `Create` delivery is invoked fire-and-forget from `api.ts` (`feedDeliver` uses `void … .catch(…)`), so the now-synchronous fan-out does not block the HTTP response to the sharing user.
- No stale references to `InProcessMessageQueue` remain, and no test depended on queue behavior — `federation.integration.test.ts` only exercises the read surface (actor doc / WebFinger / followers), and `deliver.test.ts` only covers addressing.

One behavioral note (not blocking): the inbound `Accept` on `Follow` (federation.ts) is now also sent synchronously inside the inbox request handler, so that handler will block on the outbound POST to the remote server (and any timeout) before responding. This is strictly better than before — where the enqueued `Accept` would never have drained — but it does couple inbox latency to remote reachability. That's an acceptable tradeoff until the persistent-queue slice lands. As the PR notes, delivery is now best-effort with no retry/durability, which is called out as a follow-up.

PR #867, opened 2026-07-02
fix(feed): deliver Create to followers when sharing via MCP (#831)

fiddur:
## Problem

Sharing an activity via the **MCP** `share_activity` tool created the `feed_post` row but never pushed a `Create{Note}` to followers — the ActivityPub delivery hook was only wired into the **REST** `/feed` share route (#864). Since the live share went through MCP, followers never received anything in their timelines.

This is the second of two delivery bugs found during live testing (the first, #866, dropped the never-drained in-process queue for synchronous delivery).

## Fix

Thread the same fire-and-forget `feedDeliver` hook into the MCP path so a share delivers identically regardless of transport:

- `registerFeedTools(server, user, deliver?)` calls `deliver?.(user, record, activity)` right after `createFeedPost`, mirroring the REST handler.
- `McpDeps.feedDeliver` carries the hook, passed through `createMcpServer` → `registerFeedTools`.
- `api.ts` hoists `feedFederation` + `feedDeliver` above the MCP mount so MCP and REST share **one** federation instance and the exact same delivery closure.

No new logic — pure wiring of the existing `deliverFeedPost`. Verified with `pnpm --filter aurboda-backend check` (tsgo + oxlint clean).

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -305,6 +305,17 @@ const main = async () => {
   // Mount OAuth endpoints BEFORE body-parser (uses its own parsers)
   httpd.use(createOAuthRouter({ centralDb, loginToUserDb, webHost }))
 
+  // ActivityPub federation object (one instance, shared by the actor mount, the
+  // MCP feed tools, and the REST feed router). `feedDeliver` fans a shared post
+  // out to followers, fire-and-forget, so it's identical whether a post is
+  // created via MCP or REST.
+  const feedFederation = createFeedFederation(webHost)
+  const feedDeliver: FeedDeliver = (user, post, activity) => {
+    void deliverFeedPost({ federation: feedFederation, origin: webHost }, user, post, activity).catch((err) =>
+      console.error(`⚠️ feed delivery failed for ${user}/${post.id}:`, err),
+    )
+  }
+
   // Mount MCP server BEFORE body-parser (MCP SDK needs raw body)
   // Stateless mode — no session tracking needed (tools only, no subscriptions)
   httpd.use(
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Clean, minimal fix that closes a real parity gap: shares created via the MCP `share_activity` tool now fan out a `Create{Note}` to followers, matching the REST `/feed/activities/:id/share` route.

The wiring is sound:
- `feedFederation` + the `feedDeliver` closure are hoisted above the MCP mount so a single federation instance is now shared by the MCP tools (`feedDeliver` dep), the `integrateFederation` middleware (line 347), and the REST feed router (line 457). `createFeedFederation` is a side-effect-free factory, so moving its construction earlier changes nothing else.
- `deliver?.(user, record, activity)` in `feed-tools.ts` mirrors the REST handler exactly — same `FeedPostRecord`/`Activity` arguments, same best-effort fire-and-forget semantics. The optional `deliver?` keeps existing `registerFeedTools` callers (and tests) working with no delivery injected.
- Fire-and-forget is safe under the stateless MCP request lifecycle: `deliverFeedPost` runs against the shared `feedFederation`, independent of the per-request MCP server/transport that `server.close()` tears down, and its `.catch` prevents unhandled rejections.

Behavior (delivering to followers regardless of `public`/`unlisted`/`followers` visibility) is delegated to the shared `addressingFor`/`deliverFeedPost`, so no new privacy divergence is introduced.

Non-blocking: no test accompanies the new MCP delivery path. That's consistent with the existing REST path (only `deliverFeedPost` itself is unit-tested in `deliver.test.ts`; neither router's `deliver?.()` invocation is covered), but given the repo's "backend should be well covered" standard, a small `feed-tools` test asserting `deliver` is called with the created record + activity — and not called when the activity is missing — would be a worthwhile follow-up.

PR #868, opened 2026-07-02
feat(feed): serve real AP outbox + individual post objects (#831)

fiddur:
## Problem

A shared activity federated fine (the `Create` reaches followers), but two AP surfaces were still stubs:

- The **outbox dispatcher returned an empty collection**, so a Mastodon profile for the actor shows **"Posts 0"** even after sharing.
- **Individual post ids didn't resolve** — nothing served the `Note` when a remote server dereferenced a post's id.

## Change

- **DRY object building**: extracted `buildFeedNote` / `buildFeedCreate` out of `deliverFeedPost`. The object we *deliver*, *list in the outbox*, and *serve at the post id* are now built in one place and stay byte-identical. The `Note` id/url is its object-dispatcher URL (`getObjectUri(Note, …)` → `/users/{u}/feed/{postId}`); the `Create` id is a `#create` fragment on it.
- **Object dispatcher**: `setObjectDispatcher(Note, '/users/{identifier}/feed/{postId}')` serves the post's `Note`. Only `public`/`unlisted` resolve; `followers`-only 404 (they're delivered with the object inline, so their id never needs fetching). Invalid usernames and non-UUID ids 404 without a DB hit.
- **Real outbox**: `public` + `unlisted` posts as `Create` activities, newest-first, as a single inline `OrderedCollection` with a total-items counter. (Pagination is a later slice; fine at feed scale.)
- **DB helpers**: `listPublicFeedPosts` / `countPublicFeedPosts` centralise the outbox visibility filter so it can't drift from the `/series` authorization filter.

## Tests

`federation.integration.test.ts`: outbox lists public posts as `Create`s; object served as a `Note` at its canonical id; followers-only excluded from outbox + object 404; non-UUID id 404s without touching the DB. `feed.integration.test.ts`: `listPublicFeedPosts`/`countPublicFeedPosts` include public+unlisted, exclude followers-only. All green + `pnpm --filter aurboda-backend check` clean.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -1,5 +1,5 @@
 import { createFederation, type Federation, MemoryKvStore } from '@fedify/fedify'
-import { Accept, Follow, Person, Undo } from '@fedify/fedify/vocab'
+import { Accept, type Create, Follow, Note, Person, Undo } from '@fedify/fedify/vocab'
 
 /**
  * The Fedify `Federation` object for the activity feed.
@@ -20,13 +20,22 @@ import { Accept, Follow, Person, Undo } from '@fedify/fedify/vocab'
 import { isValidUsername } from '../../api/auth-routes.ts'
 import {
   countFeedFollowers,
+  countPublicFeedPosts,
+  getActivityById,
+  getFeedPostById,
   getOrCreateActorKeyPair,
   isMissingDatabase,
   listFeedFollowers,
+  listPublicFeedPosts,
   removeFeedFollower,
   upsertFeedFollower,
 } from '../../db/index.ts'
+import { buildFeedCreate, buildFeedNote } from './deliver.ts'
 import { toCryptoKeyPair } from './keys.ts'
+
+/** RFC 4122 canonical form — guards `getFeedPostById` from a non-UUID `postId`
+ * (Postgres would otherwise raise `invalid input syntax for type uuid`). */
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
 
 export const createFeedFederation = (origin: string): Federation<void> => {
   const federation = createFederation<void>({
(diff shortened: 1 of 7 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -128,8 +137,66 @@ export const createFeedFederation = (origin: string): Federation<void> => {
...
+    async (ctx, { identifier, postId }) => {
+      if (!isValidUsername(identifier) || !UUID_RE.test(postId)) return null
+      let post
+      try {
+        post = await getFeedPostById(identifier, postId)
+      } catch (error) {
+        if (isMissingDatabase(error)) return null
+        throw error
+      }
+      if (post == null || post.visibility === 'followers' || post.activity_id == null) return null
```
fiddur: Minor defense-in-depth: this endpoint (unauthenticated object serving) uses a denylist (`visibility === 'followers'`) while `listPublicFeedPosts`/`countPublicFeedPosts` use an allowlist (`IN ('public','unlisted')`). Today `FeedVisibility` has only those three values so they're equivalent, but if a future non-public value (e.g. `direct`) is added, the outbox would correctly exclude it while this object dispatcher would still serve it. Consider matching the allowlist here (`post.visibility !== 'public' && post.visibility !== 'unlisted'`) so the two stay in lockstep.

fiddur (commented):
✅ Approved

Solid, well-scoped slice. The refactor to build the `Note` id from `getObjectUri(Note, …)` correctly unifies the delivered object, the outbox item, and the dereferenceable object at a single canonical URL, and `buildFeedNote`/`buildFeedCreate` make that invariant hard to break. The object dispatcher and outbox handlers guard the username, UUID shape, missing-DB, soft-deleted activities, and followers-only visibility consistently, and the new integration tests exercise the full path through `federation.fetch` (outbox as Create activities, individual Note, followers-only exclusion + 404, non-UUID 404).

Verified: `getActivityById` filters `deleted_at IS NULL` so soft-deleted activities correctly drop out of both the outbox items and the object endpoint; `getFeedPostById` runs against the per-user DB so there's no cross-user leak; `FeedVisibility` is exactly `public | followers | unlisted`, so the current exclusion logic is safe.

Non-blocking notes: (1) The documented `totalItems` (from `countPublicFeedPosts`) vs `orderedItems` divergence when a post's activity was soft-deleted produces a technically-inconsistent OrderedCollection, but it's tolerated by Mastodon and self-heals — fine for this slice. (2) The `Note`'s `url` points at the AP object id rather than a human HTML permalink, which is acceptable until a post page exists. Couldn't run `tsgo`/tests locally (no `node_modules` in the worktree) — relying on CI for that.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -128,8 +138,66 @@ export const createFeedFederation = (origin: string): Federation<void> => {
...
+      } catch (error) {
+        if (isMissingDatabase(error)) return null
+        throw error
+      }
+      const items = (
+        await Promise.all(
+          posts.map(async (post) => {
+            if (post.activity_id == null) return null
+            const activity = await getActivityById(identifier, post.activity_id)
+            return activity == null ? null : buildFeedCreate(ctx, identifier, post, activity)
```
fiddur, line 186: Non-blocking: each outbox item here runs `getActivityById` **and** `buildFeedCreate` → `resolveActivityScalars`, which issues a bucketed metrics-aggregation query per post (plus extra queries for `hr_zone_minutes`). Combined with the no-cursor `listPublicFeedPosts`, an unauthenticated `/outbox` fetch does O(N) heavy aggregations for a user's entire public history. The comment above already flags pagination as a later slice — just noting the per-item cost is heavier than a plain listing, so caching/pagination will matter sooner than post-count alone suggests.

fiddur (commented):
✅ Approved

Solid, well-tested slice. Extracting `buildFeedNote`/`buildFeedCreate` and deriving the `Note` id, `url`, and `Create` id from `getObjectUri(Note, …)` means the delivered object, the outbox `Create`, and the dereferenceable object all share one id and can't drift — exactly the invariant the change is after. Security handling is correct: `followers`-only posts are excluded from the outbox (`isPubliclyVisible` allowlist + SQL `visibility IN ('public','unlisted')`) and 404'd on direct object fetch (keeping follower-only content off unauthenticated surfaces), the `UUID_RE` guard avoids a Postgres `invalid input syntax for uuid` on garbage ids, and everything is scoped to the per-user DB. `isMissingDatabase` handling matches the rest of the dispatchers, and the integration tests cover the happy path, the followers-only exclusion, and the non-UUID case. Existing unit tests for the dormant aurboda representation (`buildCreateExercise`/`buildFeedPostActivity`) are untouched and still pass.

Non-blocking observations:

- **Unbounded, query-heavy outbox on an unauthenticated endpoint.** The dispatcher builds every public/unlisted post inline, and each item runs `getActivityById` *plus* `resolveActivityScalars` (a bucketed metrics aggregation, +2 more queries when `hr_zone_minutes` is shared). So it's not just "list N posts" — it's N heavy aggregations fanned out per request. The code comment frames pagination as a later slice and "fine at feed scale," which is fair for a fitness feed, but the per-item aggregation cost is worth keeping in mind when pagination/caching lands.
- **`totalItems` vs `orderedItems` divergence.** `setCounter` counts all public/unlisted posts while the items filter drops any whose activity was soft-deleted (or has a null `activity_id`). Documented and self-healing; only cosmetic for an inline collection.
- **Heads-up for the future content-negotiated endpoint.** The dormant aurboda representation in `object.ts` still uses the old id scheme (Create id = `postId`, object id = `${postId}/object`). Since `feed-activity.ts` sets `postId` to `.../users/{user}/feed/{id}`, that aurboda `Create` id now equals the Fedify `Note` id served at `.../feed/{id}`. Nothing serves `buildCreateExercise` today so there's no runtime collision, but the two id schemes will need reconciling when that endpoint is wired up.

apps/backend/src/services/activitypub/federation.ts
fiddur: Agreed and non-blocking at current scale. Tracked as a dedicated follow-up: add cursor pagination (setFirstCursor/setLastCursor) and cache/precompute the per-post scalar summary so an unauthenticated /outbox fetch doesn't scale O(posts × bucketed-metric queries). Resolving here; the perf work lands in its own slice.

PR #869, opened 2026-07-02
feat(feed): federate Update on edit and Delete{Tombstone} on unshare (#831)

fiddur:
## Problem

Only `Create` was federated (on share). Editing a shared post's metric selection/visibility, or unsharing it, left the post stale in followers' timelines.

## Change

Propagate the full feed-post lifecycle:

- **`deliver.ts`**: `buildFeedUpdate`/`deliverFeedUpdate` (`Update{Note}`) and `buildFeedDelete`/`deliverFeedDelete` (`Delete{Tombstone}`). Both reuse the same `getObjectUri`-derived `Note` id, so an `Update` replaces exactly the object that was delivered and a `Delete` retracts it by that id. `buildFeedDelete` needs no activity/scalar resolution, so it stays synchronous and works even after the row is gone.
- **Hook shape**: generalised `FeedDeliver` from a single function to a `{ created, updated, deleted }` interface (one fire-and-forget hook per lifecycle event). Wired `updated`/`deleted` into both the REST feed router (`PATCH`/`DELETE`) and the MCP tools (`update_feed_post`/`delete_feed_post`), so a post edited/deleted via either transport federates identically. The `DELETE` route now reads the post before removing it so the `Delete` can be addressed by its original visibility.

## Tests

- Unit: `buildFeedDelete` → `Delete` wrapping a `Tombstone` at the object id, with correct visibility addressing (public → Public/followers; followers-only → followers, never Public).
- Integration: `buildFeedUpdate` wraps the post's `Note` in an `Update` at its canonical id.

All green + `pnpm --filter aurboda-backend check` clean.

## Follow-up (out of scope)

Serving a `410 Tombstone` on a GET of a deleted object needs persistence (feed_posts are hard-deleted); the pushed `Delete` is what retracts from followers' timelines.

### DIFF
--- a/apps/backend/src/routes/feed-router.ts
+++ b/apps/backend/src/routes/feed-router.ts
@@ -38,12 +38,17 @@ import { validateBody } from '../validation.ts'
  * injected so the router + MCP tools stay decoupled from the ActivityPub layer
  * (and testable without it). Each impl signs + fans the corresponding activity
  * out to the user's followers: `created` → `Create`, `updated` → `Update`,
- * `deleted` → `Delete{Tombstone}`. `deleted` takes only the post — no activity
- * lookup is needed to retract by object id, and the row is already gone.
+ * `deleted` → `Delete{Tombstone}`.
+ *
+ * `created` receives the activity because the share handler already resolved it
+ * (for the 404 check), so there's no double fetch. `updated`/`deleted` take only
+ * the post: their implementation resolves whatever it needs *inside* the
+ * fire-and-forget boundary, so a post-mutation lookup can never turn a
+ * successful edit/delete into a 500.
  */
 export interface FeedDeliver {
   created: (user: string, post: FeedPostRecord, activity: Activity) => void
-  updated: (user: string, post: FeedPostRecord, activity: Activity) => void
+  updated: (user: string, post: FeedPostRecord) => void
   deleted: (user: string, post: FeedPostRecord) => void
 }
 
@@ -109,10 +114,8 @@ export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedD
         return res.status(404).json({ error: 'Feed post not found', success: false })
       }
       // Federate the edit as an Update so followers replace the stored object.
-      if (record.activity_id) {
-        const activity = await getActivityById(user, record.activity_id)
-        if (activity) deliver?.updated(user, record, activity)
-      }
+      // Fire-and-forget: the impl resolves the activity, so it can't 500 here.
+      deliver?.updated(user, record)
       res.json({ post: serialize(record), success: true })
     },
   )
--- a/apps/backend/src/routes/feed-router.ts
+++ b/apps/backend/src/routes/feed-router.ts
@@ -26,6 +26,7 @@ import {
   createFeedPost,
   deleteFeedPost,
   getActivityById,
+  getFeedPostById,
   listFeedPosts,
   updateFeedPost,
 } from '../db/index.ts'
@@ -33,11 +34,18 @@ import { type TypedRouter, typedRouter } from '../typed-router.ts'
 import { validateBody } from '../validation.ts'
 
 /**
- * Fire-and-forget federation delivery for a freshly-shared post. Injected so the
- * router stays decoupled from the ActivityPub layer (and testable without it);
- * the implementation signs + fans the Create out to the user's followers.
+ * Fire-and-forget federation delivery hooks for the feed-post lifecycle,
+ * injected so the router + MCP tools stay decoupled from the ActivityPub layer
+ * (and testable without it). Each impl signs + fans the corresponding activity
+ * out to the user's followers: `created` → `Create`, `updated` → `Update`,
+ * `deleted` → `Delete{Tombstone}`. `deleted` takes only the post — no activity
+ * lookup is needed to retract by object id, and the row is already gone.
  */
-export type FeedDeliver = (user: string, post: FeedPostRecord, activity: Activity) => void
+export interface FeedDeliver {
+  created: (user: string, post: FeedPostRecord, activity: Activity) => void
+  updated: (user: string, post: FeedPostRecord, activity: Activity) => void
+  deleted: (user: string, post: FeedPostRecord) => void
+}
 
 const serialize = (record: FeedPostRecord): FeedPost => ({
   activity_id: record.activity_id,
@@ -79,7 +87,7 @@ export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedD
         visibility: req.body.visibility,
       })
       // Fan the post out to followers (best-effort; never blocks the response).
-      deliver?.(user, record, activity)
+      deliver?.created(user, record, activity)
       res.json({ post: serialize(record), success: true })
     },
   )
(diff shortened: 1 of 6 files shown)


apps/backend/src/routes/feed-router.ts
```diff
@@ -100,16 +108,25 @@ export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedD
       if (!record) {
         return res.status(404).json({ error: 'Feed post not found', success: false })
       }
+      // Federate the edit as an Update so followers replace the stored object.
+      if (record.activity_id) {
+        const activity = await getActivityById(user, record.activity_id)
```
fiddur: This activity lookup is awaited inline after `updateFeedPost` has already committed, purely to build the federated `Update`. If `getActivityById` rejects, the client gets a 500 even though the edit persisted — which contradicts the "best-effort; never blocks the response" intent stated for `created`. Consider resolving the activity inside the fire-and-forget `deliver.updated` implementation (or guarding this lookup) so federation support can't fail the primary operation. Same pattern in the MCP `update_feed_post` tool. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped change. Refactoring `FeedDeliver` from a single function into a `created`/`updated`/`deleted` interface is clean, all callers (api.ts, REST router, MCP tools) are updated consistently, and the new `Update{Note}` and `Delete{Tombstone}` builders faithfully mirror the existing `buildFeedCreate` path (stable `#update`/`#delete` fragment ids on the canonical Note id, addressing derived from the same `recipients()` table). Capturing the post *before* deletion so the `Delete` can carry the right object id + visibility is the correct sequencing, and the `record.activity_id` guard before the activity lookup is right. Good test coverage: a DB-free unit test for `buildFeedDelete` and an integration test for `buildFeedUpdate`.

1. **Best-effort principle vs. the awaited activity lookup (feed-router.ts / feed-tools.ts update paths).** Federation is documented as fire-and-forget that "never blocks the response," but the `getActivityById` call needed only to build the `Update` is now awaited *after* `updateFeedPost` has already committed. If that lookup rejects, a successful edit returns a 500. Low severity (same DB/pool, self-healing on retry), but worth considering wrapping it or resolving the activity inside the fire-and-forget `deliver.updated` impl. See inline note.

2. **No-op edits still federate.** `updateFeedPost` returns the row unchanged when the patch has no set columns (`sets.length === 0`), so a PATCH / `update_feed_post` with an empty or all-undefined body will still emit an `Update` to all followers. Minor spurious traffic, not incorrect.

3. **Docs.** `docs/features/feed.md` (lines 8-11) still states the ActivityPub delivery layer is "intentionally not part of this foundation" — already stale after the earlier Create/outbox PRs and now further so with Update/Delete. Not introduced by this PR, but a good moment to refresh it per the docs standard (it isn't in this diff, so noting here rather than inline).

4. A visibility change (e.g. public → followers) only addresses the `Update` to the new visibility's recipients, so servers that only saw the public object via a relay won't be corrected — a known fediverse limitation, out of scope here.

apps/backend/src/routes/feed-router.ts
fiddur: Fixed in c2c71a6: dropped the activity param from the `updated` hook (now `(user, post)`, like `deleted`) and moved the getActivityById lookup inside the api.ts fire-and-forget impl. The router/MCP just fire the hook, so a post-commit lookup failure can no longer 500 a successful edit.

apps/backend/src/services/activitypub/deliver.ts
```diff
@@ -112,6 +112,48 @@ export const buildFeedCreate = async (
...
+  post: DeliverablePost,
+  activity: DeliverableActivity,
+): Promise<Update> => {
+  const note = await buildFeedNote(ctx, user, post, activity)
+  const noteId = ctx.getObjectUri(Note, { identifier: user, postId: post.id })
+  const { cc, to } = recipients(post.visibility, ctx.getFollowersUri(user))
+  return new Update({
+    actor: ctx.getActorUri(user),
+    ccs: cc,
+    id: new URL(`${noteId.href}#update`),
```
fiddur: Non-blocking: the `Update` activity id is a fixed `#update` fragment, so every edit of the same post emits an activity with an identical id. As the comment notes, Mastodon supersedes by the object (and since the `Note` carries no `updated` timestamp it will reprocess each incoming Update), so this works for Mastodon. But AS2 expects each activity to have a unique id, and some receiving servers dedupe inbound activities by id — those would drop the 2nd+ edit. Consider deriving the fragment from `updated_at` (e.g. `#update-<epoch>`) so repeated edits produce distinct activity ids. `#delete` is fine since it fires at most once per post.

fiddur (commented):
✅ Approved

Clean, well-scoped extension of feed federation to the Update (edit) and Delete{Tombstone} (unshare) lifecycle events. The design is solid:

- Widening `FeedDeliver` from a single function to a `{ created, updated, deleted }` interface keeps the REST router and MCP tools decoupled from the ActivityPub layer, and REST/MCP parity is preserved (both `update_feed_post`/`PATCH` and `delete_feed_post`/`DELETE` now federate).
- Resolving the Update's activity *inside* the fire-and-forget boundary (api.ts) is the right call — a post-mutation lookup can no longer turn an already-committed edit into a 500. The `null`-activity and `null`-`activity_id` guards degrade gracefully.
- The delete route captures the row before deleting so the `Delete` is addressed by the post's real visibility and references the correct object id; the `!deleted || !existing` guard is correct.
- `buildFeedUpdate`/`buildFeedDelete` reuse the shared `recipients()` and `getObjectUri(Note, …)` helpers, so addressing and the object id stay consistent with `buildFeedCreate`. Test coverage is good: unit tests for the Tombstone shape + visibility addressing, and an integration test for the Update object id.

One non-blocking robustness note left inline about the stable `#update` activity id across repeated edits. No blocking issues found.

apps/backend/src/services/activitypub/deliver.ts
fiddur: Fixed: the Update activity id now carries the post's updated_at as `#update-<epoch-ms>`, so each edit emits a distinct activity id (AS2-unique), and dedupe-by-id servers no longer drop later edits.

fiddur (commented):
✅ Approved

Solid, well-scoped change that extends feed federation from `Create` to also emit `Update` on edit and `Delete{Tombstone}` on unshare. The `FeedDeliver` refactor (function → `{created, updated, deleted}` interface) is applied consistently across all call sites — api.ts, the REST feed-router, and the MCP feed-tools — with no stale function-style invocations or test mocks left behind. The fire-and-forget design is done carefully: `updated`/`deleted` take only the post and resolve whatever they need *inside* the async boundary, so a post-mutation lookup can never turn an already-committed edit/delete into a 500 (the comment and code agree). The delete route correctly captures `existing` before deleting and 404s on `!deleted || !existing`, and `buildFeedDelete` intentionally needs no activity resolution so it survives the row already being gone. Giving each `Update` a distinct activity id via `#update-<updated_at ms>` is the right call for AS2 uniqueness / inbound dedupe. Adding `updated_at` to `DeliverablePost` is satisfied by every caller (all pass `FeedPostRecord`). New unit + integration tests exercise the Delete/Tombstone shape and the Update id/object well.

Couldn't run the suite locally (node_modules not installed in this worktree), so I'm trusting CI for green tests + typecheck.

One non-blocking consideration (see inline): a visibility *downgrade* (e.g. `public` → `followers`) federates an `Update` addressed only to the new, narrower audience, leaving the original public audience with a stale public copy — inherent to the current model since the hook only sees the new record. Out of this PR's scope, but worth tracking. Also, docs/features/feed.md lines 8–11 still claim the ActivityPub delivery layer is "intentionally not part of this foundation," which is now stale (Create delivery already landed, and this PR adds Update/Delete) — a good opportunity to refresh that paragraph.

apps/backend/src/routes/feed-router.ts
```diff
@@ -100,16 +113,23 @@ export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedD
       if (!record) {
         return res.status(404).json({ error: 'Feed post not found', success: false })
       }
+      // Federate the edit as an Update so followers replace the stored object.
+      // Fire-and-forget: the impl resolves the activity, so it can't 500 here.
+      deliver?.updated(user, record)
```
fiddur, line 118: Non-blocking / for awareness: because the hook only receives the new record, a visibility *downgrade* (e.g. `public` → `followers`) federates an `Update` addressed only to the narrower audience — remote servers that showed the post to non-followers keep the stale public copy, since AP has no way to "un-publish from public" other than a `Delete` to the dropped audience. Fully propagating a downgrade would need the prior visibility to emit a `Delete` to the removed recipients. Fine to leave as a follow-up given the PR's scope, but worth capturing so it isn't lost.

fiddur: Agreed — this is an inherent ActivityPub limitation (Mastodon behaves identically: narrowing a public post to followers-only can't retract copies non-follower servers already hold, since there's no addressable "public" inbox to Delete from). Not something we can cleanly fix, so I'll document the visibility-downgrade caveat in the feed docs PR rather than attempt partial propagation that would surprise users. Resolving.

PR #870, opened 2026-07-02
feat(web): share an activity to the federated feed from its detail view (#831)

fiddur:
## What

The "good UI" for sharing: a **Share to feed** button on an activity's detail view that opens a dialog to publish it to the user's ActivityPub feed.

![share dialog fields: summary metrics, series opt-in, visibility]

- **Summary metrics** — checkboxes for exactly the scalar keys the backend resolves (`duration`, `distance`, avg/max HR, HR zones, `calories`, avg stress). Unavailable-for-this-activity metrics are silently dropped server-side, so the same set is safe to offer everywhere.
- **Share full time-series** — a separate, more-revealing opt-in (heart rate, speed, power, elevation, cadence, stress), off by default, matching the backend's privacy-conservative default.
- **Visibility** — public / unlisted / followers-only.

The button shows only for non-merged, non-deleted activities and not while editing. The dialog also supports **edit mode** (`PATCH` an existing post), so the upcoming feed/outbox view can reuse it directly.

## How

- New `state/api/feed.ts` client (share / list / update / delete) following the existing axios + `auth.value.token` pattern; barrel-exported.
- `ShareActivityDialog` (+ CSS) mirrors the existing dialog conventions (backdrop, `btn-primary`/`btn-secondary`, dark-mode).
- `ShareActivityButton` (activity-specific) rendered alongside the other activity actions in `EntityDetail`.

## Notes

- `include_chart`/`include_map` are sent `false` — chart/map attachment rendering is a later enhancement, so they're not surfaced yet.
- Verified with `pnpm --filter aurboda-web check` (tsgo + oxlint) clean. The REST endpoints this drives are covered by backend integration tests; the web app has no component-render test harness, consistent with the rest of the UI.


apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+  return next
+}
+
+export function ShareActivityDialog({ activityId, activityTitle, post, onClose, onShared }: Props) {
+  const queryClient = useQueryClient()
+  const [summary, setSummary] = useState<Set<string>>(
+    () => new Set(post ? post.included_metrics : DEFAULT_SUMMARY),
+  )
+  const [series, setSeries] = useState<Set<MetricType>>(
+    () => new Set((post?.series_metrics ?? []) as MetricType[]),
```
fiddur: Minor: the `as MetricType[]` cast goes against the repo's "avoid casting, use proper typing and type guards" guideline. The root cause is that api-spec types `FeedPost.series_metrics` as `z.array(z.string())`; if that were `z.array(metricTypeSchema)` this cast (and the widening of `included_metrics`) could go away. Not blocking.

apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+  )
+  const [visibility, setVisibility] = useState<FeedVisibility>(post?.visibility ?? 'public')
+
+  const mutation = useMutation({
+    mutationFn: () => {
+      const body = {
+        include_chart: false,
```
fiddur: The mutation always sends `include_chart: false` / `include_map: false`, but the dialog exposes no toggle for them. Harmless in this PR since edit mode (`post` prop) is never reached from `ShareActivityButton`, but once an edit entry point is wired up this would silently reset those flags on every save. Consider omitting them in edit mode (both are optional on `UpdateFeedPostBody`) or only sending them when the dialog actually controls them.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend PR. The new `state/api/feed.ts` helpers hit endpoints that match `feed-router.ts` exactly (`POST /feed/activities/:id/share`, `GET /feed`, `PATCH /feed/:postId`, `DELETE /feed/:postId`) and correctly guard the optional `post` field in the `FeedPostResponse`. The `SUMMARY_METRICS` keys all correspond to keys the backend resolver (`services/activitypub/scalars.ts`) supports (`duration`, `distance`, `heart_rate_avg`, `heart_rate_max`, `hr_zone_minutes`, `calories`, `stress_avg`), and every `SERIES_METRICS` key is a valid `MetricType`. The dialog follows the established `MergeFoodItemDialog` backdrop/`stopPropagation` pattern and the API module follows the raw-axios + `authHeaders()` convention used by `goals.ts` et al. The `ShareActivityButton` render guard (non-merged, non-deleted, not editing) is sensible.

Two minor, non-blocking notes below. No blocking issues found. (Could not run `pnpm check` locally — `node_modules` isn't installed in this checkout — but the diff is type-correct by inspection.)

apps/web/src/components/ShareActivityDialog.tsx
fiddur: Fixed: edit mode now omits include_chart/include_map (both optional on UpdateFeedPostBody), so a save no longer resets them; only the create path sets their false defaults.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: Fixed: edit mode now omits include_chart/include_map (both optional on UpdateFeedPostBody), so a save no longer resets them; only the create path sets their false defaults.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: (Correcting my prior reply, which landed on the wrong thread.) The cast is removed: series is now preselected by intersecting the post's series_metrics with the keys the dialog offers, so the state stays `Set<MetricType>` with no cast. I kept the api-spec change (FeedPost.series_metrics → MetricType[]) out of scope because it ripples into the DB row types + serializers, where pg returns an untyped `text[]` — it would relocate rather than remove the assertion.

apps/web/src/state/api/feed.ts
```diff
@@ -0,0 +1,45 @@
...
+    `${API_URL}/feed/activities/${activityId}/share`,
+    body,
+    { headers: authHeaders() },
+  )
+  if (!response.data.post) throw new Error('Share failed: no post returned')
+  return response.data.post
+}
+
+/** List every post the user has shared to their feed, newest-first. */
+export const fetchFeed = async (): Promise<FeedPost[]> => {
```
fiddur, line 28: `fetchFeed` (and `deleteFeedPost` below) aren't used anywhere in this PR — the only entry point added here is the share/edit dialog. Fine to land as foundation for the upcoming feed view, but worth confirming they're intentionally ahead of use rather than leftover.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend feature. The new `feed.ts` API module follows the existing sibling conventions (inline `axios` + `authHeaders`, api-spec type-only imports), the endpoint paths (`POST /feed/activities/:id/share`, `PATCH /feed/:id`, `GET /feed`, `DELETE /feed/:id`) match the backend `feed-router` mounted at `/feed`, and the metric keys offered are all real: every `SUMMARY_METRICS` key is resolvable by `services/activitypub/scalars.ts`, and every `SERIES_METRICS` key is a valid `MetricType`. Types line up end-to-end (`Set<MetricType>` / `Set<string>` state → `ShareActivityBody`), the privacy-conscious defaults (series off by default, `include_chart`/`include_map` false) are respected, and the button gating (`activity.id && !isMergedActivity && !activity.deleted_at && !isEditing`) is sensible.

- `fetchFeed` and `deleteFeedPost` are exported (and re-exported) but unused anywhere in this PR — presumably foundation for an upcoming feed-management view; fine to keep, just flagging.
- The dialog's edit mode (`post` prop) and `onShared` callback are never exercised by the added UI (`ShareActivityButton` only ever shares anew), so that path is currently unreachable/untested from this PR.
- `queryClient.invalidateQueries({ queryKey: ['feed'] })` is a no-op today since no query registers under `['feed']` yet — harmless, sets up for the future feed view.
- No unit tests were added, but that's consistent with the codebase's convention for dialog components (e.g. `MergeFoodItemDialog` has none); the pure `toggle` helper and edit-mode preselection would be the natural things to cover if you want any.

apps/web/src/state/api/feed.ts
fiddur: Intentional — `fetchFeed`/`deleteFeedPost` are the foundation for the feed/outbox view PR (next), which lists shared posts and lets you unshare. Landing the full client together keeps the module cohesive; both are exercised there.

PR #871, opened 2026-07-02
feat(web): feed/outbox view of shared activities (#831)

fiddur:
## What

A **Feed** page (sidebar link + `/feed` route) — the "see the outbox feed" ask — listing everything the user has published to their federated feed, newest-first.

Each post card shows:
- the **activity** (linked to its detail page), with its title resolved via the cached activities query,
- an **audience badge** (public / unlisted / followers-only),
- the shared **summary-metric** chips and any opted-in **series** chips,
- when it was shared (and edited, if different).

Actions:
- **Edit** — reuses `ShareActivityDialog` in edit mode (`PATCH` → federated `Update`).
- **Unshare** — deletes the post (`DELETE` → federated `Delete{Tombstone}`) behind a `window.confirm`.

## How

- New `pages/Feed` (+ CSS), registered at `/feed` with a `📣 Feed` nav entry.
- Extracted the metric-option lists into a shared `components/feed-metrics` module so the share dialog and this view label metrics identically (DRY) — the dialog now imports from it.
- Uses the `feed.ts` client (`fetchFeed`/`deleteFeedPost`/`updateFeedPost`) landed in #870.

## Notes

- Per-card activity-title lookups use react-query caching; feeds are small so this is cheap. Enriching `GET /feed` with the activity title server-side is a possible future optimization.
- `pnpm --filter aurboda-web check` clean.

### DIFF
--- a/apps/web/src/pages/Feed/index.tsx
+++ b/apps/web/src/pages/Feed/index.tsx
@@ -1,0 +1,138 @@
+/**
+ * Feed / outbox view — the activities the user has published to their federated
+ * (ActivityPub) feed. Lists each shared post with its audience and the metrics
+ * it exposes, and lets the user edit the selection (reusing the share dialog in
+ * edit mode) or unshare (which retracts a Delete to followers).
+ */
+import type { FeedPost, FeedVisibility } from '@aurboda/api-spec'
+
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
+import { format } from 'date-fns'
+import { useState } from 'preact/hooks'
+
+import { seriesLabel, summaryLabel } from '../../components/feed-metrics'
+import { ShareActivityDialog } from '../../components/ShareActivityDialog'
+import { deleteFeedPost, fetchActivityById, fetchFeed } from '../../state/api'
+import { toDisplayName } from '../../utils/displayName'
+import './style.css'
+
+const VISIBILITY_LABEL: Record<FeedVisibility, string> = {
+  followers: 'Followers only',
+  public: 'Public',
+  unlisted: 'Unlisted',
+}
+
+const formatWhen = (iso: string): string => format(new Date(iso), 'PP')
+
+function FeedPostCard({ post }: { post: FeedPost }) {
+  const queryClient = useQueryClient()
+  const [editing, setEditing] = useState(false)
+  const activityId = post.activity_id
+
+  const activityQuery = useQuery({
+    enabled: activityId != null,
+    queryFn: () => fetchActivityById(activityId ?? ''),
+    queryKey: ['activity', activityId],
+    staleTime: 5 * 60 * 1000,
+  })
+
+  const deleteMutation = useMutation({
+    mutationFn: () => deleteFeedPost(post.id),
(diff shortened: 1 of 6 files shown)


apps/web/src/pages/Feed/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { deleteFeedPost, fetchActivityById, fetchFeed } from '../../state/api'
+import { toDisplayName } from '../../utils/displayName'
+import './style.css'
+
+const VISIBILITY_LABEL: Record<FeedVisibility, string> = {
+  followers: 'Followers only',
+  public: 'Public',
+  unlisted: 'Unlisted',
+}
+
```
fiddur, line 24: Non-blocking: this fetches the entire activity detail (`GET /activities/:id`) just to show a title, once per card — so a feed with N posts triggers N full-detail requests on load. Acceptable for a small personal outbox (and react-query dedupes repeats), but if the outbox can grow, consider resolving the title server-side in the feed response (the repo's "resolve names at query time in the API response layer" convention) rather than N client-side detail fetches.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend PR adding the feed/outbox view. It follows the project conventions nicely: api-spec types only on the frontend, barrel imports through `state/api`, functional components, no casts.

A few things I verified:
- The `feed-metrics.ts` extraction is a good move — the share dialog and the feed view now share the same label maps, so a post's labels are consistent between creation and display. `ShareActivityDialog` behavior is unchanged (it still imports `MetricType` for its typed `series` state).
- All `FeedPost` fields used match the api-spec schema, and the `.feed-badge-{public,unlisted,followers}` CSS covers every `FeedVisibility` value.
- The `updated_at !== created_at` "edited" indicator is sound: both columns default to `NOW()` in the same INSERT transaction (identical value), and only an actual edit bumps `updated_at`.
- `activity_type` is a required string in the api-spec `Activity` schema, so the `toDisplayName(activity.activity_type)` title fallback can't throw; a deleted/failed activity lookup degrades gracefully to "Shared activity" while keeping the detail link.

One non-blocking observation (inline): each `FeedPostCard` fetches the full activity detail via `/activities/:id` solely to render a title, so a feed with N posts fires N heavyweight detail requests on load. Fine for a small personal outbox and react-query dedupes repeats, but if the outbox can grow large, resolving the title in the feed API response layer (per the repo's "resolve names at query time" guidance) would be lighter than a full-detail fetch per card. Not a blocker.

Note: I couldn't run `pnpm check` — dependencies aren't installed in this worktree (no `node_modules`) — so type/lint verification was by inspection only; CI covers it.

apps/web/src/pages/Feed/index.tsx
fiddur: Agreed, and it matches the repo's "resolve names in the API response" principle. Kept out of this frontend PR because it's a backend + api-spec change (enrich GET /feed with the activity title, regenerate types); tracked as a dedicated follow-up. At personal-outbox scale with react-query dedup the current per-card fetch is fine. Resolving.

PR #876, opened 2026-07-03
feat(web): avatar upload + display on public pages (#834)

fiddur:
Part **3b** of the #834 epic — the web UI for the avatar backend merged in #859.

## What
- **Settings → Avatar**: a new `AvatarSettings` section to upload (PNG/JPEG/WebP/GIF) or remove your avatar, with a live preview that cache-busts after each change.
- **Public profile** (`/u/:username`): the avatar renders next to the `@handle`.
- **Shared dashboards**: the owner's avatar renders next to the author attribution.
- `avatarUrl` resolves `/u/:username/avatar.png` against the API origin — same origin in prod (`/api`), absolute backend URL in dev — so it works in both. The pure builder (`buildAvatarUrl`) is unit-tested (kept config/`window`-free so it runs in the node test env).

## Notes
- No new API surface — consumes the endpoints from #859. Uploaded avatars are normalized server-side to 256² WebP; users without one get a deterministic identicon, so images always render.
- `pnpm check` green.

## Follow-ups
4: author-editable dashboard text · 5: oEmbed + polish


apps/web/src/components/AvatarSettings.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function AvatarSettings() {
+  const username = auth.value.user
+  // Bumped after upload/remove to bust the browser (and CDN) avatar cache.
+  const [cacheBust, setCacheBust] = useState(0)
+  const [error, setError] = useState<string | null>(null)
+
+  const uploadMut = useMutation({
+    mutationFn: (file: File) => uploadAvatar(file),
+    onError: (err) => setError(err instanceof Error ? err.message : 'Upload failed'),
```
fiddur: Minor UX/consistency: on a failed upload, axios rejects with a generic `Error` whose `message` is `"Request failed with status code 400"`, so the descriptive messages the backend deliberately returns (`{ error: 'Unsupported file type…' }`, `'Could not process image'`, or multer's >10MB rejection) are swallowed. Elsewhere the app extracts `err.response?.data?.error` (see `state/auth.ts`, `ActivityTypes/index.tsx`). Consider doing the same here (and in the delete handler on line 25) so users see why an upload was rejected — e.g. a file that's too large. Note the `AvatarUploadResult.error`/`success` fields on `uploadAvatar`'s return type are effectively dead for the same reason (axios throws before you can read them), and `onSuccess` never checks `data.success`.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend feature: avatar upload/remove in Settings, plus avatar display on the public profile and next to the shared-dashboard attribution. Nice touches — the `buildAvatarUrl` helper is deliberately kept free of `config`/`window` imports so it's unit-testable, and the tests cover the prod (same-origin) / dev (absolute API URL) / encoding cases. The `accept` list matches the backend's `ALLOWED_UPLOAD_TYPES` exactly, cache-busting on the preview is a sensible way to defeat the `max-age=3600` cache after upload/remove, and the barrel re-export has no name collisions. Docs are updated.

One non-blocking observation: the upload/delete error handling surfaces the generic axios message rather than the descriptive `{ error }` the backend returns (e.g. "Unsupported file type", "Could not process image", or multer's file-too-large rejection). The rest of the codebase reads `response.data.error`. Details inline. Approving — this is a UX polish item, not a correctness issue.

apps/web/src/components/AvatarSettings.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { useState } from 'preact/hooks'
+
+import { avatarUrl, deleteAvatar, uploadAvatar } from '../state/api/profile'
+import { auth } from '../state/auth'
+import { SettingsSection } from './SettingsSection'
+
+/** Prefer the backend's descriptive `{ error }` over axios's generic message. */
+const errorMessage = (err: unknown, fallback: string): string => {
+  if (axios.isAxiosError(err)) {
+    return (err.response?.data as { error?: string } | undefined)?.error ?? err.message
```
fiddur: Minor (project standard: "avoid casting, use proper typing and type guards"): this uses a `as { error?: string } | undefined` cast on the axios response body. Since `response.data` is untyped, a small type guard (e.g. `const hasError = (d: unknown): d is { error?: string } => typeof d === 'object' && d !== null`) would satisfy the guideline without the assertion. Non-blocking.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend PR that wires the existing avatar backend into the web UI. The `buildAvatarUrl` helper is nicely factored as a pure, `window`/`config`-free function with unit tests covering the same-origin (prod), absolute-backend (dev), and username-encoding cases — I traced each and they're correct, including the `API_URL=''` fallback. The `profile.ts` API module follows the established manual-`Authorization: Bearer` + FormData pattern used elsewhere (e.g. `settings.ts` `uploadIcon`), and the `POST`/`DELETE /profile/avatar` calls plus `{ success, url?, error? }` shape match `profile-router.ts`. The Settings preview cache-busts via a `?t=` query param on upload/remove; the JSX attribution change (`…/>@\n{username}`) renders as `@username` (JSXText trims the trailing newline), so no stray space.

Two non-blocking notes:

- Public pages (`PublicProfile`, `PublicDashboard`) use `avatarUrl(username)` with no cache-bust. Since the backend serves the avatar with `max-age=3600`, a user who changes their avatar can keep seeing the stale one on their own public page for up to an hour even though the Settings preview updates immediately. This is a documented caching tradeoff, not a correctness bug — flagging in case the UX matters.
- `PublicChallenge.tsx` still renders a bare `@{username}` attribution with the same `.public-attribution` class (which now becomes `inline-flex` with a gap) but no avatar. Harmless visually, but slightly inconsistent with the profile/dashboard treatment if you intended parity.

Nothing here blocks merge.

apps/web/src/components/AvatarSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+      setCacheBust(Date.now())
+    },
+  })
+
+  if (!username) return null
+
+  const src = `${avatarUrl(username)}${cacheBust ? `?t=${cacheBust}` : ''}`
+
+  const onFileChange = (e: Event) => {
+    const file = (e.target as HTMLInputElement).files?.[0]
```
fiddur: Minor UX: the file input is never reset after a mutation, so `onChange` won't fire if the user picks the *same* file again — e.g. retrying after a failed upload, or re-uploading the same image after a Remove. Clearing `e.target.value` (or resetting via a `key`/ref after success/error) makes those flows work. Non-blocking.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend PR for avatar upload/display. Verified against the backend it talks to:

- `uploadAvatar`/`deleteAvatar` hit `POST`/`DELETE /profile/avatar` with the `avatar` form field and Bearer auth — matches `profile-router.ts` exactly.
- `errorMessage` extracts `data.error`, which is the shape the backend returns on 4xx (`{ error, success: false }`).
- `buildAvatarUrl` correctly resolves `/u/:username/avatar.png` against the API origin for both prod (`/api`, same-origin) and dev (absolute backend URL), and encodes the username; the `try/catch` fallback is sensible. The three unit tests cover the meaningful cases (I traced them by hand — deps aren't installed in this worktree — and they hold).
- Cache-busting the Settings preview via `?t=Date.now()` after upload/remove is the right call; leaving the public pages on the backend's `max-age=3600` is reasonable since those are viewed by others.
- Auth-header style, `useMutation`, and `SettingsSection` usage all match existing conventions; docs updated.

One minor UX nit inline (file input not reset), non-blocking.

apps/web/src/pages/Settings/style.css
```diff
@@ -507,3 +507,45 @@
...
+  padding: 0.4rem 0.9rem;
+  border-radius: 6px;
+  background: #673ab8;
+  color: #fff;
+  cursor: pointer;
+  font-size: 0.9rem;
+}
+
+.avatar-settings__upload input {
+  display: none;
```
fiddur: Non-blocking: `display: none` removes the file input from the tab order, and the styled `<label>` isn't focusable either, so keyboard-only users can't reach the "Upload image" control. Consider a visually-hidden pattern (position/clip/opacity, e.g. an `.sr-only` class) instead of `display: none` so the input stays focusable. Not a blocker.

fiddur (commented):
✅ Approved

Clean, well-scoped web-only change that wires the already-merged avatar backend into the UI. Nice touches: the URL builder is extracted as a pure, `window`/`config`-free function with unit tests covering the prod same-origin, dev absolute-origin, and username-encoding cases; the frontend contract matches the backend exactly (multipart field `avatar`, `POST`/`DELETE /profile/avatar`, public `GET /u/:username/avatar.png`); the Bearer-token pattern and the `{ error }` response body type guard mirror existing code; and re-picking the same file after a failure is handled by resetting `input.value`. Cache-busting the Settings preview after upload/remove is reasonable (public pages intentionally rely on the backend's `max-age=3600`). Docs are updated. No correctness, security, or standards issues found.

One minor, non-blocking accessibility note below about the hidden file input.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend PR that wires up avatar upload/remove in Settings and displays avatars on the public profile and shared-dashboard attribution.

Verified against the backend merged in #859: the `POST`/`DELETE /profile/avatar` calls and the public `GET /u/:username/avatar.png` route all match, and the `{ success, url }` upload response matches `AvatarUploadResult`. `buildAvatarUrl` correctly resolves the avatar path against the API *origin* (dropping the `/api` prefix), which works in both prod (`API_URL='/api'`, same-origin) and dev (absolute backend URL), and degrades gracefully via the `try/catch` — the three unit tests cover exactly these cases plus username encoding.

The auth/upload code follows the existing `settings.ts` convention (Bearer token from the `auth` signal), and omitting an explicit `Content-Type` on the `FormData` upload is correct since the browser sets the multipart boundary. Cache-busting the settings preview with `?t=` after upload/remove is sound, while public pages intentionally rely on the backend's `max-age=3600`. The `input.value = ''` reset (re-pick same file) and the visually-hidden-but-focusable file input are nice touches. Import ordering, functional style, and react-query usage are all consistent with the rest of the codebase.

No blocking issues found. Minor, non-blocking observations: the Remove button isn't disabled during an in-flight upload (and vice-versa), so both mutations can be triggered concurrently — harmless since the backend serializes, but a small UX nicety would be to disable both while either is pending.

PR #877, opened 2026-07-03
feat(feed): share UX — post formatting, merged activities, relevant metrics (#831)

fiddur:
Bundles three share-experience improvements from live testing into one PR (batched to cut CI/review/deploy overhead).

## 1. Readable shared posts (backend)
`feedPostContent` now renders the `Note` as a **bold title headline** with the shared scalars on their **own line**, and durations humanize (`642s → 10m 42s`, `3720s → 1h 2m`) instead of "Duration 642 seconds". Only the human content changes; `aurboda:metrics` stays numeric. Since delivery/outbox/object all build the Note here, it shows everywhere.

## 2. Merged activities are shareable (web)
The Share button was hidden for merged activities (`!isMergedActivity`), but most activities are merged. A `merged:<uuid>` id wraps a real anchor activity and the share endpoint takes a plain uuid, so it now passes `rawEntityId`. The post references the anchor activity.

## 3. Only relevant metrics offered (web)
The dialog offered every metric (Distance on a yoga session, etc.). It now does one coarse bucketed-metrics fetch over the activity window and offers only summaries whose source metric has data (duration always when there's an end) and only series that exist — the full-series section hides when none apply. Extracted into a `useShareableMetricOptions` hook; window passed from the activity detail (merged span) and the feed edit view.

## Tests / checks
Updated + added `feedPostContent` unit tests (headline/stats layout, duration incl. hours, untitled fallback). `pnpm --filter aurboda-backend check` and `pnpm --filter aurboda-web check` clean.

Supersedes #874 and #875 (folded in here).


apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -54,11 +91,14 @@ export function ShareActivityDialog({ activityId, activityTitle, post, onClose,
...
   const [visibility, setVisibility] = useState<FeedVisibility>(post?.visibility ?? 'public')
+  const { summaryOptions, seriesOptions } = useShareableMetricOptions(activityStart, activityEnd)
 
   const mutation = useMutation({
     mutationFn: () => {
-      const included_metrics = [...summary]
-      const series_metrics = [...series]
+      // Only send metrics the dialog actually offered (drops defaults hidden as
+      // irrelevant); the backend still ignores any with no data.
+      const included_metrics = summaryOptions.map((m) => m.key).filter((k) => summary.has(k))
```
fiddur, line 116: This intersect-with-offered-options logic is safe for the create path, but on the **edit** path it can silently drop metrics the post already had. `summaryOptions`/`seriesOptions` are narrowed by an availability lookup over the window passed to the dialog. When editing from the Feed that window is the anchor activity's raw `start_time`/`end_time` (a plain-UUID fetch returns no `merged_*` fields), which is narrower than the merged span used at share time. Metrics that came from non-anchor merged segments therefore aren't in the offered options, get filtered out here, and `updateFeedPost` replaces the stored arrays — so they're lost even on a no-op "Save changes". Consider unioning in `post.included_metrics`/`post.series_metrics` so an edit never removes a metric the user didn't explicitly uncheck.

apps/web/src/pages/Feed/index.tsx
```diff
@@ -104,6 +105,8 @@ function FeedPostCard({ post }: { post: FeedPost }) {
         <ShareActivityDialog
           activityId={activityId}
           activityTitle={activity?.title}
+          activityStart={activity?.start_time}
```
fiddur: These are the anchor activity's raw times, not the merged span. `fetchActivityById(post.activity_id)` fetches by plain UUID, and `/activities/:id` only fills `merged_start_time`/`merged_end_time` for a `merged:`-prefixed id — so this window is narrower than the one EntityDetail uses when the post was created. That mismatch is what drives the metric-availability narrowing (and the resulting metric drop, see ShareActivityDialog) to differ between share and edit for merged activities.

fiddur (commented):
🛑 Changes required

Nicely-scoped share-UX work: the backend `feedPostContent` refactor (bold headline + separate stats line + humanized duration) is clean, pure, and well-tested, and the availability-driven metric filtering in the share dialog is a good idea. One correctness concern blocks approval.

**Editing a merged-activity post from the Feed can silently drop shared metrics.** The mutation now intersects the selected metrics with the availability-narrowed `summaryOptions`/`seriesOptions` (`summaryOptions.map(m => m.key).filter(k => summary.has(k))`) instead of sending `[...summary]`. Those options are narrowed by a bucketed fetch over the window passed in. On the create path from EntityDetail the window is the *merged span* (`merged_start_time ?? start_time` … `merged_end_time ?? end_time`), and `getOverlappingActivities` can pull in transitively-connected activities across a ~36h window — so the merged span can extend well beyond any single source. But the Feed edit path fetches the activity by its plain anchor UUID (`post.activity_id`), and `/activities/:id` only returns `merged_start_time`/`merged_end_time` for a `merged:`-prefixed id — a plain fetch returns just the anchor's raw `start_time`/`end_time`. So on edit the availability window is narrower than at share time, metrics that came from non-anchor segments are filtered out of the offered options, and `updateFeedPost` (which replaces `included_metrics`/`series_metrics` wholesale) persists the reduced set — even for a no-op "Save changes". Consider preserving metrics already in `post.included_metrics`/`post.series_metrics` regardless of the availability narrowing, or resolving the merged span for the edit case.

Everything else looks good — the escaping in `feedPostContent` is preserved, `formatDuration` only triggers for the `seconds` unit (which only `duration` uses), and the `enabled`-gated fallback query is correct.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: Fixed: the offered options now always include the metrics the post already shares (keepSummary/keepSeries in useShareableMetricOptions), so the submit intersect can no longer drop an existing choice when editing over a narrower window.

apps/web/src/pages/Feed/index.tsx
fiddur: Fixed: the Feed edit card now fetches the merged view (`merged:${activity_id}`) so merged_start_time/end are populated and the availability window matches the merged span the detail view used at share time — same offered metrics, not the narrower anchor window.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -621,8 +621,18 @@ const ActivityContent = ({ entityId }: { entityId: string }) => {
...
+      {!activity.deleted_at && !isEditing && (
+        // `rawEntityId` is a plain activity UUID for both raw and merged views (a
+        // `merged:<uuid>` id wraps a real anchor activity), so merged activities
+        // — the common case — are shareable; the post references that anchor.
+        // The merged span bounds the metric-availability lookup so the dialog
+        // offers everything the activity has across its sources.
+        <ShareActivityButton
+          activityId={rawEntityId}
+          activityTitle={activity.title}
+          activityStart={activity.merged_start_time ?? activity.start_time}
```
fiddur, line 633: Non-blocking: the dialog now offers metrics based on the merged span (`merged_start_time/end`), but when the post is published the backend resolves scalars over the *anchor* activity's own `[start_time, end_time]` window (`feed-activity.ts` `resolveActivityScalars`). For a merged activity whose anchor window is narrower than the merged span, a metric that only has data outside the anchor window would be offered here yet silently dropped from the published post. Degrades gracefully, so not a blocker — just flagging the window mismatch newly reachable now that merged activities are shareable.

fiddur (commented):
✅ Approved

Solid, well-scoped feature work with matching test updates. I traced the key paths:

- **Backend `object.ts`**: `feedPostContent` now emits a bold headline `<p>` plus a separate stats `<p>`, and `formatDuration` renders seconds as `10m 42s` / `1h 2m`. HTML is escaped on both the heading and the stats line, negative durations clamp to `0s`, and the `unit === 'seconds'` gate correctly targets only the `duration` scalar (confirmed against `scalars.ts`, which is the only scalar with `unit: 'seconds'`). The three new `feedPostContent` tests and the two updated `buildCreateExercise` expectations all match the implementation.
- **`feed-metrics.ts` + `useShareableMetricOptions`**: the new `source` values (`distance`, `heart_rate`, `calories_active`, `stress_level`) exactly match the backend `METRIC_SCALARS` metric mappings and are all valid `MetricType` members, so the availability filter is accurate. The `keepSummary`/`keepSeries` retention (from `post.included_metrics`/`series_metrics`) correctly prevents an edit over a narrower window from silently dropping an already-shared metric, and while the availability query loads (or without a window) every option is offered — a safe fallback. The mutation now sends only offered+selected keys in canonical order, and the disabled guard requires at least one *offered* summary metric.
- **Feed / EntityDetail wiring**: fetching `merged:${activityId}` to populate `merged_start_time/end` is handled gracefully by the backend for non-merged and deleted activities (falls back to the raw path, leaving the merged fields undefined → dialog offers all options). Removing the `!isMergedActivity` guard to make merged activities shareable is intentional and `rawEntityId` resolves to the real anchor UUID.

One non-blocking observation (see inline): sharing a *merged* activity offers metrics based on the merged span, but the backend resolves the post's scalars over the *anchor* activity's own `[start_time, end_time]` window (`feed-activity.ts` `resolveActivityScalars`). If the anchor's window is narrower than the merged span, a user could pick a metric that only has data outside the anchor window and see it silently dropped from the published post. It degrades gracefully, so not a blocker, but worth being aware of. No blocking issues found.

apps/web/src/pages/EntityDetail/index.tsx
fiddur: Good catch — offered (merged span) vs resolved (anchor window) can diverge for a merged activity, so an offered metric may resolve empty and drop out. Backend still omits empties gracefully (no error), and the anchor window ≈ the full activity in the common case, so it's an edge case. Tracking the proper fix (share/resolve over the merged window, or align the dialog to the anchor window) as a dedicated follow-up. Resolving.

PR #878, opened 2026-07-03
feat(dashboard): author-editable description + per-section intro text (#834)

fiddur:
Part **4** of the #834 epic. Dashboards gain optional author prose that improves both the page and its link preview.

## What
- **api-spec**: optional `description` on `DashboardConfig` and a per-section `description`. Both additive/optional, so existing v1 configs stay valid — **no version bump or migration needed** (`z.literal(1)` would reject a bumped config; the point of optional fields is backward compatibility).
- **Backend**: a shared dashboard's `config.description` now feeds `og:description` / the `<meta name=description>` (falls back to the generated sentence from PR1 when unset). Covered by a new share-html-router test.
- **Web**: `EditableDashboard` edits the dashboard description and each section's intro (commit-on-blur, matching the section-rename pattern). Both render read-only on the home dashboard and public shared pages — section intros above their widgets, so the page reads as a narrative.

## Notes
- No new API endpoints — the existing shared-dashboard/home-dashboard update paths validate the config, which now accepts the optional fields.
- `pnpm check` green.

## Follow-up
5: oEmbed + sharability polish (copy-link/Web Share, theme-color, JSON-LD, QR, debugger validation).

### DIFF
--- a/apps/web/src/components/EditableDashboard/index.tsx
+++ b/apps/web/src/components/EditableDashboard/index.tsx
@@ -8,7 +8,7 @@
  */
 import type { DashboardConfig, DashboardSection, DashboardWidget, SectionType } from '@aurboda/api-spec'
 
-import { useState } from 'preact/hooks'
+import { useEffect, useState } from 'preact/hooks'
 
 import { DashboardEditor } from '../DashboardEditor'
 import { WidgetRenderer } from '../widgets'
@@ -40,6 +40,9 @@ function DashboardSectionComponent({
   const [editingTitle, setEditingTitle] = useState(false)
   const [titleDraft, setTitleDraft] = useState(section.title)
   const [descDraft, setDescDraft] = useState(section.description ?? '')
+  // Re-sync the draft when the underlying value changes (e.g. reset elsewhere),
+  // so a stale draft can't re-persist removed text on blur.
+  useEffect(() => setDescDraft(section.description ?? ''), [section.description])
 
   const commitDescription = () => {
     const next = descDraft.trim() || undefined
@@ -294,6 +297,10 @@ export function EditableDashboard({ config, isEditing, onChange, boardId }: Edit
   const [showWidgetPicker, setShowWidgetPicker] = useState<string | null>(null) // section id or null
   const [showAddSection, setShowAddSection] = useState(false)
   const [descDraft, setDescDraft] = useState(config.description ?? '')
+  // Re-sync when config changes under a still-mounted editor (home "Reset to
+  // Default", or navigating between shared dashboards), so the draft can't go
+  // stale and re-persist stale text on blur.
+  useEffect(() => setDescDraft(config.description ?? ''), [config.description])
 
   const commitDescription = () => {
     const next = descDraft.trim() || undefined
--- a/apps/web/src/components/EditableDashboard/index.tsx
+++ b/apps/web/src/components/EditableDashboard/index.tsx
@@ -24,6 +24,7 @@ function DashboardSectionComponent({
   onAddWidgetClick,
   onDeleteSection,
   onRenameSection,
+  onDescriptionChange,
 }: {
   section: DashboardSection
   isEditing: boolean
@@ -33,10 +34,17 @@ function DashboardSectionComponent({
   onAddWidgetClick?: () => void
   onDeleteSection?: () => void
   onRenameSection?: (title: string) => void
+  onDescriptionChange?: (description: string | undefined) => void
 }) {
   const [collapsed, setCollapsed] = useState(section.collapsed ?? false)
   const [editingTitle, setEditingTitle] = useState(false)
   const [titleDraft, setTitleDraft] = useState(section.title)
+  const [descDraft, setDescDraft] = useState(section.description ?? '')
+
+  const commitDescription = () => {
+    const next = descDraft.trim() || undefined
+    if (next !== section.description) onDescriptionChange?.(next)
+  }
 
   const gridClass =
     section.type === 'links' ? 'links-grid' : section.type === 'charts' ? 'charts-grid' : 'metrics-grid'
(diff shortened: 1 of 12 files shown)


apps/web/src/components/EditableDashboard/index.tsx
```diff
@@ -273,6 +293,19 @@ interface EditableDashboardProps {
 export function EditableDashboard({ config, isEditing, onChange, boardId }: EditableDashboardProps) {
   const [showWidgetPicker, setShowWidgetPicker] = useState<string | null>(null) // section id or null
   const [showAddSection, setShowAddSection] = useState(false)
+  const [descDraft, setDescDraft] = useState(config.description ?? '')
```
fiddur, line 299: `descDraft` is initialized from `config.description` only on mount and never re-synced, so it goes stale whenever `config` changes under a still-mounted `EditableDashboard`.

1. **Home dashboard reset:** type a description → *Reset to Default* clears `config.description` but the component stays mounted (only `isEditing` toggles) → re-enter edit and the textarea still shows the old draft; on blur `commitDescription` sees `next !== config.description` and re-persists the removed text.
2. **Owner nav between shared dashboards:** this component isn't keyed (`boardId="home"` here, `boardId=share.id` on PublicDashboard) and preact-iso reuses the instance across `/u/:user/:slug` param changes, so navigating from dashboard A to B while editing keeps A's draft and blurring can write A's description onto B.

Consider keying `EditableDashboard` by `boardId` and/or syncing the draft when `config.description` changes. Same pattern applies to the section-level `descDraft` (line 42), though `key={section.id}` on the section components shields it in most flows.

fiddur (commented):
🛑 Changes required

Nice, well-scoped feature: schema, backend meta, editor UI, read-only rendering, and docs all move together, the meta description is properly `escapeAttr`-escaped, and there's a test for the new og:description path.

One correctness issue holds it back: the dashboard-level `descDraft` in `EditableDashboard` is seeded from `config.description` only on mount and never re-synced, so it goes stale whenever `config` changes while the component stays mounted. The clearest reproducible case is on the home dashboard — type a description, click *Reset to Default* (which clears `config.description` but leaves the component mounted), then re-enter edit: the textarea still shows the old draft and blurring re-persists the text the reset just removed. A second case affects owners with multiple shared dashboards navigating between `/u/:user/:slug` while editing (the component isn't keyed and preact-iso reuses the instance across param changes), where a blur can write one dashboard's description onto another. See the inline note for a suggested fix. The section-level draft (line 42) shares the pattern but is largely shielded by `key={section.id}` on the section components.

Minor (non-blocking): `buildDashboardShareMeta` uses `description ?? fallback`, so a stored empty/whitespace-only description (possible via the API, which has no min-length) would yield an empty og:description instead of the generated fallback; the web UI already normalizes empty to `undefined`, so this only matters for direct API writes.

apps/web/src/pages/PublicDashboard/index.tsx
```diff
@@ -127,12 +127,15 @@ function ReadOnlyDashboard({
         </a>
       </div>
 
+      {config.description ? <p class="dashboard-intro">{config.description}</p> : null}
```
fiddur, line 130: This branch never renders for a public (non-owner) viewer: the config here comes from `GET /public/:username/:slug`, which returns `sanitizeConfig(dashboard.config)`. `sanitizeConfig` (in `public-shares-router.ts`) rebuilds the top level as only `{ sections, version }`, so the new top-level `config.description` is dropped before it reaches this component. The dashboard description shows in the `og:description` link preview but not on the actual shared page. Preserve `description` in `sanitizeConfig` (e.g. `description: config.description`). Section descriptions are unaffected because they ride along on the `...section` spread.

fiddur (commented):
🛑 Changes required

Nicely-scoped feature with proper escaping in the crawler meta (`escapeAttr`), auto-escaped JSX on the page, updated docs, and backend tests for the new meta behaviour. One correctness bug blocks it.

**Dashboard-level `description` is stripped from public shared dashboards.** `sanitizeConfig` in `apps/backend/src/routes/public-shares-router.ts` (line ~47, not in this diff) rebuilds the config as `{ sections, version }` and does **not** carry over the new top-level `config.description`. That is the config returned by `GET /public/:username/:slug`, which is exactly what a non-owner viewer's `ReadOnlyDashboard` renders. Result:

- The author sets a dashboard description and sees it while editing (owner path uses the unsanitized config).
- The link preview / `og:description` shows it (`share-html-router` reads the unsanitized `config.description`).
- But on the actual public page the `{config.description ? …}` branch in `PublicDashboard/index.tsx` never fires, because the API already dropped it.

So half the advertised feature ("a dashboard description shown under the title" for a shared dashboard) is silently missing for public viewers. Section descriptions are fine — they survive via the `...section` spread; it's only the dashboard-level field that's lost. Fix: preserve `description` in `sanitizeConfig` (e.g. add `description: config.description` to the returned object), and ideally add a `sanitizeConfig` test asserting the description passes through.

Everything else (schema additions, draft/commit re-sync logic, CSS scoping under `.dashboard`, empty/whitespace fallback in `buildDashboardShareMeta`) looks correct.

fiddur (commented):
✅ Approved

Clean, well-tested feature. The important paths all hold up:

- **Sanitization**: `sanitizeConfig` uses a whitelist and correctly adds the top-level `description`, while section descriptions ride along the existing `...section` spread — covered by a new test.
- **No injection**: the now-author-controlled description is escaped via `escapeAttr` in `renderShareMetaTags` before landing in `og:description`/`meta description`, and the frontend renders it through JSX (auto-escaped) with `white-space: pre-wrap` for line breaks.
- **Empty/whitespace fallback** in `buildDashboardShareMeta` is handled and tested; passing the un-trimmed value is fine since `clampDescription` trims.
- **Editor drafts**: the `useEffect` re-sync + `next !== current` guards correctly prevent a stale draft from re-persisting removed text on blur, without clobbering in-progress typing (deps key on the primitive value, not object identity).
- Schema fields are `optional` (backward compatible), types line up (`SharedDashboardRecord.config`), CSS applies to both owner and anonymous public views, and docs are updated.

Minor, non-blocking observations (no change required):
- The section intro paragraph renders above the widgets even when a section is collapsed, so a collapsed section still shows its intro in view mode. Arguably intended (the intro describes the section), just calling it out.
- Neither `description` field has a max length in the schema. The meta is clamped to 200 chars, but the on-page render is unbounded — low risk since it's owner-authored content on their own board.

PR #880, opened 2026-07-03
feat(feed): chart + route-map images on shared posts (#831)

fiddur:
Implements the feed-post images you asked for: a **heart-rate chart** and a **full GPS route map**, each with a checkbox in the share dialog. (No privacy trimming — area masking is a tracked follow-up.)

## Backend
- **Renderers** (`services/activitypub/feed-images.ts`): pure SVG→PNG (via `sharp`, no fonts/tiles). `renderChartPng` (line chart from `[time,value]`), `renderRoutePng` (aspect-correct route shape from GeoJSON `[lon,lat]`, north-up, start/end marked). Unit-tested for valid PNG output incl. degenerate inputs.
- **Serving** (`feed-image-router.ts`): `GET /public/:username/feed/:postId/{chart,route}.png`, rendered on demand. Gating (`resolveImageWindow`, unit-tested) mirrors the object endpoint — public/unlisted only, only when the matching flag was opted in, non-UUID/invalid/missing → 404. `no-store` so unshare/visibility change takes effect immediately.
- **Attachment**: `buildFeedNote` attaches AS2 `Image`(s) (url + mediaType + intrinsic size) for `include_chart`/`include_map`, so Mastodon renders them inline. Since delivery/outbox/object all build the Note here, attachments show consistently.

## Web
- Share dialog exposes **Heart-rate chart** / **Route map** toggles, offered only when the activity has the data (HR for the chart; a GPS signal — distance/speed — for the map), wired through create **and** edit.

## Docs
- `feed.md`: Images section, endpoint table rows, and a caveat that route maps have no basemap and no privacy trimming.

## Notes / follow-ups (tracked)
- Route map is a bare **shape** (no street basemap — needs a tile provider) and shows the **full** track (area masking later).
- The chart currently renders **heart rate** (the canonical workout series).

`pnpm --filter aurboda-backend check` and `pnpm --filter aurboda-web check` clean; renderer + eligibility + deliver tests green.

### DIFF
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -1,0 +1,97 @@
+/**
+ * Public feed-post image endpoints (UNAUTHENTICATED).
+ *
+ * Handles: GET /public/:username/feed/:postId/chart.png
+ *          GET /public/:username/feed/:postId/route.png
+ *
+ * Rendered on demand from the shared activity's data. Like the public `/series`
+ * endpoint there is no auth token, so the gating below is the whole privacy
+ * boundary: an image is served only for a `public`/`unlisted` post that opted
+ * into that attachment (`include_chart` / `include_map`). `no-store` keeps the
+ * images revocable (unshare / flip to followers / clear the flag takes effect
+ * immediately). Mounted before the generic `/public/:username/:slug` resolver.
+ */
+import { type Response, Router } from 'express'
+
+import type { FeedPostRecord } from '../db/index.ts'
+
+import { isValidUsername } from '../api/auth-routes.ts'
+import { isMissingDatabase } from '../db/index.ts'
+import { renderChartPng, renderRoutePng } from '../services/activitypub/feed-images.ts'
+import { isPubliclyVisible } from '../services/activitypub/object.ts'
+
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+
+/** The window an image renders over. */
+export interface ImageActivity {
+  start_time: Date
+  end_time?: Date
+}
+
+export interface FeedImageDeps {
+  getPost: (user: string, postId: string) => Promise<FeedPostRecord | null>
+  getActivity: (user: string, activityId: string) => Promise<ImageActivity | null>
+  getSeries: (user: string, metric: string, start: Date, end: Date) => Promise<[Date, number][]>
+  getRoute: (user: string, start: Date, end: Date) => Promise<[number, number][]>
+}
+
+/**
+ * Resolve the activity window an image may render over, or `null` if the request
+ * isn't eligible: invalid username / non-UUID id, missing DB, missing or
+ * non-public post, the attachment flag not opted in, no linked activity, or an
+ * open-ended activity (no bounded window). Pure of Express — unit-testable.
+ */
+export const resolveImageWindow = async (
+  deps: Pick<FeedImageDeps, 'getPost' | 'getActivity'>,
+  username: string,
+  postId: string,
+  flag: 'include_chart' | 'include_map',
+): Promise<ImageActivity | null> => {
+  if (!isValidUsername(username) || !UUID_RE.test(postId)) return null
+  let post: FeedPostRecord | null
+  try {
+    post = await deps.getPost(username, postId)
+  } catch (error) {
+    if (isMissingDatabase(error)) return null
+    throw error
+  }
+  if (post == null || !isPubliclyVisible(post.visibility) || !post[flag] || post.activity_id == null) {
+    return null
+  }
+  const activity = await deps.getActivity(username, post.activity_id)
+  if (activity?.end_time == null) return null
+  return activity
+}
+
+const notFound = (res: Response) => res.status(404).json({ error: 'Not found', success: false })
+
+const sendPng = (res: Response, png: Buffer) => {
+  // Revocable like the /series endpoint: never let a shared cache serve an image
+  // for a post that was just un-shared.
+  res.setHeader('Cache-Control', 'no-store')
+  res.type('png').send(png)
+}
+
+export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
+  const router = Router()
+
+  router.get('/public/:username/feed/:postId/chart.png', async (req, res) => {
+    const { postId, username } = req.params
+    const activity = await resolveImageWindow(deps, username, postId, 'include_chart')
+    if (!activity?.end_time) return notFound(res)
+    const series = await deps.getSeries(username, 'heart_rate', activity.start_time, activity.end_time)
+    if (series.length === 0) return notFound(res)
+    sendPng(res, await renderChartPng(series))
+  })
+
+  router.get('/public/:username/feed/:postId/route.png', async (req, res) => {
+    const { postId, username } = req.params
+    const activity = await resolveImageWindow(deps, username, postId, 'include_map')
+    if (!activity?.end_time) return notFound(res)
+    const coords = await deps.getRoute(username, activity.start_time, activity.end_time)
+    if (coords.length === 0) return notFound(res)
+    sendPng(res, await renderRoutePng(coords))
+  })
+
+  return router
+}
--- a/apps/backend/src/services/activitypub/deliver.ts
+++ b/apps/backend/src/services/activitypub/deliver.ts
@@ -20,7 +20,7 @@ import type { FeedVisibility } from '@aurboda/api-spec'
  */
 import type { Context, Federation } from '@fedify/fedify'
 
-import { Create, Delete, Note, Tombstone, Update } from '@fedify/fedify/vocab'
+import { Create, Delete, Image, Note, Tombstone, Update } from '@fedify/fedify/vocab'
 
 import { resolveActivityScalars } from './feed-activity.ts'
 import { addressingFor, feedPostContent } from './object.ts'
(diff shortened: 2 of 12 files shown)


apps/backend/src/services/activitypub/deliver.ts
```diff
@@ -48,6 +48,44 @@ export interface DeliverablePost {
...
+  include_map: boolean
+}
+
+/**
+ * Image attachments for a post's opted-in chart/route. Each points at the public
+ * on-demand image endpoint (`/api/public/<user>/feed/<id>/{chart,route}.png`),
+ * built against the actor's origin so it matches the deployed API base. Fedify's
+ * `Image` carries `url` + `mediaType` + intrinsic size so Mastodon lays it out.
+ */
+const imageAttachments = (actorUri: URL, user: string, post: DeliverablePost): Image[] => {
```
fiddur: `imageAttachments` emits attachment URLs from `include_chart`/`include_map` alone, ignoring visibility. A `followers`-only post can carry these flags (the share dialog's Images toggles are independent of the Visibility selection), so its federated `Note` gets URLs pointing at `/api/public/<user>/feed/<id>/chart.png|route.png` — but `resolveImageWindow` 404s those for any non-public/unlisted post (the `isPubliclyVisible` gate). Followers then receive attachments that never resolve.

The sibling `seriesLinks` in `object.ts` already guards this exact case (`if (visibility === 'followers' || ...) return []`). Suggest gating here the same way, e.g. return `[]` early when `!isPubliclyVisible(post.visibility)`.

fiddur (commented):
🛑 Changes required

Solid, well-scoped feature: pure `renderChartPng`/`renderRoutePng` (rasterized via the already-present `sharp`, robust to single-point / non-finite input), an `Express`-independent `resolveImageWindow` with good unit coverage, a privacy boundary that mirrors the existing `/series` endpoint (`isPubliclyVisible` + opted-in flag + `no-store`), and thorough docs including the acknowledged route-privacy limitation.

One blocking inconsistency: the federation delivery layer attaches image URLs based purely on the `include_chart`/`include_map` flags, ignoring visibility, while the public image endpoint categorically 404s for `followers`-only posts. Since the share dialog lets a user set those flags on a `followers` post, such posts federate a `Note` whose attachments never resolve. The sibling `seriesLinks` in `object.ts` already handles exactly this case (`if (visibility === 'followers') return []`), so `imageAttachments` should gate the same way (`isPubliclyVisible(post.visibility)`).

Softer, non-blocking note: even for public/unlisted posts, `canMap` is offered whenever `distance`/`speed` data exists, but the route endpoint 404s when the `locations` table has no GPS points for the window (e.g. an indoor treadmill run with distance but no track). The delivered attachment then silently fails to resolve. Not required for this PR, but worth a follow-up (or a data-presence check) since it produces the same missing-attachment outcome as the followers case.

apps/backend/src/services/activitypub/deliver.ts
fiddur: Fixed in f538056d: imageAttachments now returns [] for non-public posts (gated on isPubliclyVisible, matching the image endpoint), so a followers-only post's Note carries no attachment URLs that would 404. Exported + unit-tested (public attaches only opted-in images; followers-only attaches nothing), and documented the followers-only-no-images limitation.

apps/backend/src/services/activitypub/deliver.ts
```diff
@@ -48,6 +48,50 @@ export interface DeliverablePost {
...
+ * `Image` carries `url` + `mediaType` + intrinsic size so Mastodon lays it out.
+ *
+ * Only `public`/`unlisted` posts get attachments: the image endpoint is
+ * unauthenticated and (like the object/series endpoints) refuses `followers`-only
+ * posts, so attaching a URL that would 404 for a follower is worse than omitting
+ * it. (Authenticated per-follower image delivery is a possible later slice.)
+ */
+export const imageAttachments = (actorUri: URL, user: string, post: DeliverablePost): Image[] => {
+  if (!isPubliclyVisible(post.visibility)) return []
+  const base = new URL(`/api/public/${encodeURIComponent(user)}/feed/${post.id}`, actorUri)
```
fiddur: This reconstructs the API base as `<actor-origin>` + hardcoded `/api`, rather than reusing the `apiBaseUrl` (`API_BASE_URL`) that the series-link builder uses (`feed-activity.ts:141`: `${trimSlashes(ctx.apiBaseUrl)}/public/...`). They're equal in the current deployment, but they're set from independent env vars — if `API_BASE_URL` ever points at a different host/base, series links stay correct while these image URLs silently break. Consider threading `apiBaseUrl` into `buildFeedNote`/`imageAttachments` so both paths share one source of truth. Non-blocking.

apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -71,6 +74,11 @@ const useShareableMetricOptions = (
     ? new Set(availabilityQuery.data.buckets.flatMap((b) => Object.keys(b.metrics)))
     : undefined
   return {
+    // The chart renders heart rate; the route map needs a GPS signal (distance /
+    // speed). While availability is unknown, offer both. `keepChart`/`keepMap`
+    // keep an already-attached image toggleable when editing.
+    canChart: present ? present.has('heart_rate') || keepChart : true,
+    canMap: present ? present.has('distance') || present.has('speed') || keepMap : true,
```
fiddur: `canMap` uses distance/speed presence as a proxy for a GPS track, but the backend renders the route from `getLocations` (actual coordinates). An activity that has distance/speed but no GPS points (e.g. a treadmill run) will show the map toggle, and if opted in, `getRoute` returns `[]` → `route.png` 404s and the federated post carries a broken image attachment. `canChart` is exact (`heart_rate` is what's rendered); the map heuristic isn't. Non-blocking, but a real edge case.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The privacy boundary is the important part here, and it's handled correctly: `resolveImageWindow` gates the unauthenticated image endpoints on a valid username + UUID, an existing post in the caller-scoped DB, `public`/`unlisted` visibility, the specific opt-in flag, a linked non-deleted activity, and a bounded window — mirroring the existing `/series` and object-dispatcher gating. `imageAttachments` correctly omits attachments for `followers`-only posts so no URL that would 404 for a follower ends up in a Note, and the served Note (object dispatcher + outbox) is built from the same full `FeedPostRecord`, so delivery/outbox/object stay consistent. Pure render functions are nicely isolated and defensively coded (finite filtering, empty/degenerate/zero-span guards), the DI seams make everything unit-testable, and coverage (`resolveImageWindow`, `imageAttachments`, PNG magic-byte checks) is good. Docs are updated and the `no-store`/privacy-trimming caveats are called out.

No blocking issues. A few non-blocking observations:

- **API base is reconstructed, not reused.** `imageAttachments` builds the image URL as `<actor-origin>` + hardcoded `/api/public/...`, whereas the series endpoint (`feed-activity.ts:141`) derives its base from the `apiBaseUrl` (`API_BASE_URL`) env var. These are equal in the current deployment (`WEB_HOST` origin + `/api` == `API_BASE_URL`), but they're configured independently — if the API is ever hosted on a different host/base (e.g. an `api.` subdomain, or `API_BASE_URL` without a `/api` path), series links stay correct while image links break. Threading `apiBaseUrl` through would remove the assumption. See inline note.

- **`canMap` uses distance/speed as a GPS proxy, but the route renders from `getLocations`.** An activity with distance/speed metrics but no GPS track (e.g. a treadmill run) will offer the map toggle, yet `getRoute` returns no coordinates → `route.png` 404s. On a public federated post that yields a broken media attachment on remote clients. Edge case, but worth a note. See inline.

- **Unauthenticated, uncached, on-demand rasterization.** Each request re-renders via `sharp` (`no-store`, no downsampling of the series/route). For long activities the point arrays are unbounded, and the endpoint is a plausible CPU/memory amplification target under a request flood. Same general class as the existing `/series` endpoint, so not new, but consider a point cap and/or rate limiting as a follow-up.

- Minor: the `no-store` "revocable" framing holds for the origin endpoint, but Mastodon caches/rehosts attached media at receipt, so an unshare won't retract already-federated image copies — an inherent federation limitation, just worth being explicit about.

apps/backend/src/services/activitypub/deliver.ts
fiddur: Fixed in 79bca9ae: imageAttachments now takes the configured apiBaseUrl (threaded through createFeedFederation, FeedDeliveryDeps, and the Note builders) and builds the URL as `${apiBaseUrl}/public/…`, exactly like the series-link builder — no longer hardcoded `<origin>/api`, so it stays correct if API_BASE_URL and WEB_HOST diverge.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: Fixed: canMap is now gated on an actual GPS track (fetchRawLocations over the window), not the distance/speed proxy. A treadmill run with distance but no GPS no longer offers the map toggle, so it can't opt into a route.png that would render empty and 404.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
+  const router = Router()
+
+  router.get('/public/:username/feed/:postId/chart.png', async (req, res) => {
+    const { postId, username } = req.params
+    const activity = await resolveImageWindow(deps, username, postId, 'include_chart')
+    if (!activity?.end_time) return notFound(res)
+    const series = await deps.getSeries(username, 'heart_rate', activity.start_time, activity.end_time)
```
fiddur, line 82: This fetches the raw, un-bucketed heart-rate series (and `getRoute` on line 91 the raw GPS track), unlike the `/series` endpoint which buckets. On an uncached (`no-store`) public endpoint that means the full point set is fetched and re-rendered into an SVG polyline on every request. It also feeds `renderChartPng`/`renderRoutePng`, which compute bounds via `Math.min(...times)` spreads — those throw `RangeError: Maximum call stack size exceeded` once the array gets large enough (a multi-hour high-frequency activity). Consider reusing the bucketed series path (and/or a reduce-based min/max) to bound the work and avoid the crash on long activities.

fiddur (commented):
✅ Approved

Solid, well-tested slice. The privacy gate is the important part and it's right: `resolveImageWindow` validates the username/UUID, scopes the post + activity lookups to the requesting user's DB, refuses non-public/`followers` posts and un-opted-in flags, swallows `isMissingDatabase`, and requires a bounded activity window — all covered by unit tests. `imageAttachments` correctly omits attachments for `followers` posts (matching the unauthenticated endpoint), uses the configured `apiBaseUrl` rather than a hardcoded `<origin>/api`, and threads it through every `buildFeedNote/Create/Update` caller and both federation dispatchers. Using plain `Router()` (not `typedRouter()`) is consistent with the other binary/image routers (`og-image-router`, `public-avatar-router`), and Express 5 handles async handler rejections natively. `no-store` keeps images revocable. The map toggle is gated on real GPS, avoiding attaching a `route.png` that would 404 for a treadmill run. Docs are updated thoroughly.

Two non-blocking notes:

1. **Un-downsampled rendering (see inline).** Both image endpoints render from raw `getTimeSeries`/`getLocations` output, whereas the `/series` endpoint buckets. For a very long activity this means a huge SVG polyline re-generated on every (uncached, `no-store`) request, and `feed-images.ts` builds its min/max via `Math.min(...times)` / `Math.min(...lons)` spreads (lines 39-42, 72-75), which throw `RangeError` once the array is large enough (~10^5 points). Consider reusing the bucketed series path and/or a reduce-based min/max.

2. **Schema description drift (not in this diff, `packages/api-spec/src/schemas/feed.ts`).** `shareActivityBodySchema.include_chart` is described as "Attach a rendered chart image (drawn only from shared metrics)", but the chart is hardcoded to render `heart_rate` regardless of whether `heart_rate` is in `included_metrics`/`series_metrics`. The share dialog labels it "Heart-rate chart" so user consent is clear, but the OpenAPI/MCP description now misrepresents the behavior — worth reconciling (update the description, or gate the chart on `heart_rate` being a shared metric to honor the stated contract).

apps/backend/src/routes/feed-image-router.ts
fiddur: Agreed it's an unauthenticated per-request render. It's gated by a per-request eligibility check (username/UUID + a scoped post+activity lookup) and these are low-frequency endpoints, so it's not an immediate blocker — but a bounded in-memory render cache + in-flight dedup (as og-image-router does), kept behind the eligibility gate so unshare/visibility changes still take effect, plus bucketing the chart series to shrink the point set, is worth doing. Tracked as a dedicated follow-up. Resolving.

PR #882, opened 2026-07-03
feat(sharing): oEmbed + share button + theme-color (#834)

fiddur:
Final part (**5**) of the #834 epic — richer unfurls and sharability polish. Completes the epic (PRs 1–4 merged).

## What
- **oEmbed**: `GET /oembed?url=<share url>&format=json` returns a `type: "link"` document (title, author, provider, OG-image thumbnail). Only **public** resources resolve — unlisted/unknown/private/invalid all 404, so nothing private leaks; a non-json `format` returns 501 per spec. Public share pages advertise it via `<link rel="alternate" type="application/json+oembed">` (injected by the share-html router). nginx proxies `/oembed` to the backend. Primarily benefits Mastodon and other oEmbed-aware consumers.
- **Share button**: public profile and shared-dashboard pages get a Share control that uses the native share sheet (`navigator.share`, mobile) where available and otherwise copies the link to the clipboard with brief "Copied!" feedback.
- **`theme-color`** meta (brand `#673ab8`) so tabs and mobile share sheets look finished (favicons/apple-touch-icon were already present).

## Design
Pure, testable `oembed.ts` (`parseShareUrl` tolerates sub-path bases; `buildOEmbedLink`) + a DI'd `oembed-router` reusing the same `createShareResolvers()`/visibility gating as the rest of the share surface.

## Tests (green; `pnpm check` clean)
oEmbed URL parsing + link building, and the router across all paths (missing url → 400, non-json → 501, non-share → 404, public dashboard/challenge/profile → 200, unlisted → 404).

## Out of scope (optional in the issue)
QR codes and per-share preview overrides — noted as future work in the docs.

### DIFF
--- a/apps/web/src/components/ShareLinkButton.tsx
+++ b/apps/web/src/components/ShareLinkButton.tsx
@@ -1,0 +1,34 @@
+import { useState } from 'preact/hooks'
+
+/**
+ * Share control for public pages: uses the native share sheet
+ * (`navigator.share`, mainly mobile) when available, otherwise copies the link
+ * to the clipboard with brief "Copied!" feedback.
+ */
+export function ShareLinkButton({ url, title }: { url: string; title?: string }) {
+  const [copied, setCopied] = useState(false)
+
+  const onClick = async () => {
+    if (typeof navigator !== 'undefined' && typeof navigator.share === 'function') {
+      try {
+        await navigator.share({ title, url })
+        return
+      } catch {
+        // User cancelled or share failed — fall through to clipboard copy.
+      }
+    }
+    try {
+      await navigator.clipboard.writeText(url)
+      setCopied(true)
+      setTimeout(() => setCopied(false), 2000)
+    } catch {
+      // Clipboard unavailable (e.g. insecure context) — nothing more we can do.
+    }
+  }
+
+  return (
+    <button type="button" class="share-link-btn" onClick={onClick} title="Share this page">
+      {copied ? 'Copied!' : 'Share'}
+    </button>
+  )
+}
--- a/apps/web/src/components/ShareLinkButton.tsx
+++ b/apps/web/src/components/ShareLinkButton.tsx
@@ -13,8 +13,10 @@ export function ShareLinkButton({ url, title }: { url: string; title?: string })
       try {
         await navigator.share({ title, url })
         return
-      } catch {
-        // User cancelled or share failed — fall through to clipboard copy.
+      } catch (err) {
+        // A deliberate cancel (AbortError) is a no-op; only a real failure
+        // falls through to the clipboard copy.
+        if (err instanceof Error && err.name === 'AbortError') return
       }
     }
     try {
--- a/apps/backend/src/api/rest-routes.ts
+++ b/apps/backend/src/api/rest-routes.ts
@@ -40,6 +40,7 @@ import { createMealsRouter } from '../routes/meals-router.ts'
 import { createMetricsRouter } from '../routes/metrics-router.ts'
 import { createNotesRouter } from '../routes/notes-router.ts'
 import { createNutrientRecommendationsRouter } from '../routes/nutrient-recommendations-router.ts'
+import { createOEmbedRouter } from '../routes/oembed-router.ts'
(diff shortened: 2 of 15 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped feature. The oEmbed endpoint mirrors the existing `og-image-router`/`share-html-router` patterns closely and is security-conscious: `isValidUsername` is enforced before any DB access, only `is_public` resources resolve, a non-public dashboard and an unknown resource both return an identical `404 Not found` (no private-vs-nonexistent enumeration or name leak — nicely covered by the `not.toContain('Secret')` test), and every URL in the response is built from the configured `webHost` rather than the attacker-supplied `url`, so there's no host injection. The response is proper JSON so there's no HTML-injection surface, and the endpoint fetches nothing (no SSRF). URL round-tripping through `oembedDiscoveryUrl` → `parseShareUrl` (encode-once / decode-per-segment) checks out. nginx `location = /oembed` correctly forwards to the root-mounted backend route, consistent with the `/u/` block. Unit tests are thorough for both the pure helpers and the router. Docs updated. I couldn't run the suite here (read-only worktree blocks `pnpm install`), but the tests are logically correct and CI will run them.

- `ShareLinkButton`: on a native-share cancel (AbortError) the catch falls through to a clipboard write + "Copied!" flash, which is a slightly surprising outcome for a deliberate cancel. It's clearly intentional per the comment, so just flagging.
- CSS: `PublicProfile/style.css` adds an **unscoped** `.share-link-btn` rule (global under Vite's plain-CSS import) while `PublicDashboard/style.css` scopes its own copy under `.public-dashboard` with a different text color (`#8b5cf6` vs `#673ab8`). The rules are duplicated across the two files and diverge only in color — worth consolidating into one shared class to avoid drift, though harmless functionally.

apps/web/src/components/ShareLinkButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+ */
+export function ShareLinkButton({ url, title }: { url: string; title?: string }) {
+  const [copied, setCopied] = useState(false)
+
+  const onClick = async () => {
+    if (typeof navigator !== 'undefined' && typeof navigator.share === 'function') {
+      try {
+        await navigator.share({ title, url })
+        return
+      } catch {
```
fiddur: Non-blocking nit: this catch swallows AbortError (user cancelling the share sheet) the same as a genuine failure, so a deliberate cancel falls through to a clipboard write and flashes "Copied!". If you want cancel to be a no-op, check `err?.name !== 'AbortError'` before falling back. Intentional per the comment, so feel free to leave as-is.

fiddur (commented):
✅ Approved

Solid, well-tested addition. The oEmbed endpoint only resolves `is_public` resources (returning 404 otherwise), and it never fetches the input URL — it just parses the `/u/...` path and does local DB lookups — so there's no private-data leak and no SSRF surface. The dashboard-before-challenge resolution correctly mirrors `share-html-router` (a found-but-private dashboard does not fall through to a challenge lookup). `parseShareUrl` is robust (sub-path bases, trailing slashes, percent-encoding, malformed input) and round-trips correctly with `buildShareUrl`/`oembedDiscoveryUrl`. The discovery `<link rel="alternate">` is only attached to public resources and is properly attribute-escaped. The nginx `location = /oembed` block matches the existing `/u/` and `/users/` proxy patterns, and the profile OG-image path used as the thumbnail exists in `og-image-router`. Both `oembed.ts` and `oembed-router.ts` have good unit coverage including the private-resource 404 case.

Using plain `Router()` rather than `typedRouter()` is consistent with the sibling unauthenticated HTML/asset routers (share-html, og-image), so it's not a violation of the api-spec routing convention.

Note (non-blocking, cosmetic): the share button's text color differs between the two pages — `#8b5cf6` in `PublicDashboard/style.css` vs `#673ab8` in `PublicProfile/style.css`, while both use a `#673ab8` border. Worth aligning if that divergence is unintentional.

I could not run the test suite in this review environment (dependencies aren't installed and the npm cache is read-only), but the tests match the implementation by inspection.

PR #885, opened 2026-07-03
feat(web): activity "Note" → rich-text "Description" moved up

fiddur:
## What & why

When editing an activity, the note field had a few rough edges:

- It was a plain `<textarea>` labelled **"Notes"** — no markdown, no preview — even though the app already has a rich-text editor with a Write/Preview toggle used for comments.
- It sat at the **bottom** of the edit fields (below Type/Start/End/Duration).
- In read mode the same text showed up as a plain-text row at the bottom of the stats table **and** (markdown-rendered) as the first comment below — redundant and buried.

This renames it to **"Description"**, moves it directly under the title in both edit and read modes, and uses the existing `MarkdownEditor` with preview.

## Changes

- **Edit form** (`EditableActivityFields.tsx`): "Notes" → "Description", moved under the title, plain textarea → `MarkdownEditor` (Write/Preview).
- **Read mode** (`index.tsx`): render the description as a markdown block under the title; drop the redundant plain-text "Notes" row from the stats table.
- **Stats builder** (`activityStats.ts` + test): remove the now-unused `notes` row.
- **CSS**: add `.entity-description`, remove the now-unused `.edit-notes-input` rules.

## Kept behaviour

The description still round-trips to the activity's **first user comment** (no backend/api-spec change) and still appears in the Notes section below.

## Verification

- `pnpm check` — 0 errors.
- `pnpm test` — 484 passed (stats builder test updated).


apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -327,6 +328,13 @@ const ActivityDetailContent = ({
           durationLabel={hasSleepStages ? 'In Bed' : undefined}
         />
 
+        {!isEditing && getUserNotesContent(activity) && (
+          <div
+            class="entity-description note-content"
+            dangerouslySetInnerHTML={{ __html: marked.parse(getUserNotesContent(activity)) as string }}
```
fiddur: Non-blocking: `marked.parse()` output is injected via `dangerouslySetInnerHTML` without sanitization (marked ^17 has no built-in sanitizer and there's no DOMPurify in the repo), so raw HTML in a note — e.g. `` — executes. This mirrors the existing pattern in `NotesSection.tsx` and `MarkdownEditor`, and the content is the user's own comments, so it's not a regression. Worth tracking centrally though: if notes are ever rendered for a different user (shared reports/admin), this becomes stored XSS. Consider a shared sanitizing render helper for all these sinks in a follow-up.

fiddur (commented):
✅ Approved

Clean, well-scoped frontend refactor. The plain notes `<textarea>` is replaced by the reusable `MarkdownEditor` and moved above the type/duration fields, the read-only "Notes" stat row is dropped from `buildActivityStatRows`, and user notes are now rendered as markdown in read mode. Verified consistency end-to-end: `notes` is removed from `BuildActivityStatRowsInput`, from `buildActivityStatRows`, from its tests, and from the single `ActivityStatsTable` caller, so type-check/tests stay green. The removed `.edit-notes-input` CSS is no longer referenced anywhere. Read-mode rendering only shows when `getUserNotesContent(activity)` is non-empty and `!isEditing`, which pairs correctly with the editor shown in edit mode.

One non-blocking security note (see inline): the new `dangerouslySetInnerHTML` sink renders unsanitized `marked` output. This is not unique to this PR — it matches the existing pattern in `NotesSection.tsx` and `MarkdownEditor`, and the content is the user's own comments — so it isn't a regression, but if activity notes are ever surfaced to another user (shared reports/admin views), the whole family of these sinks would want a sanitizer (e.g. DOMPurify), ideally centralized in `MarkdownEditor`/a shared render helper. Minor nit: `getUserNotesContent(activity)` is evaluated twice on lines 331/334; cheap, but could be hoisted to a local.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -330,6 +331,13 @@ const ActivityDetailContent = ({
...
```
fiddur: Non-blocking: `marked.parse` doesn't sanitize, so this injects raw HTML from the note content via `dangerouslySetInnerHTML`. Since the content is user-authored (self-XSS) and this matches the existing `NotesSection`/`MarkdownEditor` pattern, it's not a regression — but consider an app-wide `DOMPurify` pass. Minor: `getUserNotesContent(activity)` is also called on the guard line above; hoisting to a `const` avoids evaluating it twice.

fiddur (commented):
✅ Approved

Clean, well-scoped UI change. The activity "Note" field becomes a rich-text "Description": editing moves to the top of the edit form using the shared `MarkdownEditor`, and in view mode the user notes render as markdown above the schema fields instead of as a plain-text stat row. `notes` is removed from `buildActivityStatRows`, its input type, and its tests consistently, and `ActivityStatsTable` no longer passes it — so nothing dangles and the build stays consistent. The dead `.edit-notes-input` CSS (light + dark) is removed and `.entity-description` added. Nice.

Non-blocking notes:

- **Unsanitized HTML render (`index.tsx:337`).** `marked.parse(...)` does not sanitize, so raw HTML in a note is injected via `dangerouslySetInnerHTML`. This mirrors the existing `NotesSection.tsx` and `MarkdownEditor` patterns, and `getUserNotesContent` is filtered to user-authored comments (self-authored → self-XSS, low severity), so it's consistent with the established approach rather than a regression. Worth an app-wide `DOMPurify` pass eventually — and note the `MarkdownEditor` comment "safe rendering (no HTML passthrough)" is inaccurate, since `marked` passes HTML through.
- **Minor:** `getUserNotesContent(activity)` is evaluated twice on the same render (guard + render); could hoist to a local `const`.

apps/web/src/pages/EntityDetail/index.tsx
fiddur: Agreed — tracking the centralized DOMPurify sanitizer as a follow-up in #910. No privilege boundary is crossed today (user's own notes rendered back to themselves), and this matches the existing NotesSection/MarkdownEditor pattern, so keeping it out of this PR. Resolving.

apps/web/src/pages/EntityDetail/index.tsx
fiddur: Fixed the double evaluation in 6730dee5 — getUserNotesContent(activity) is now hoisted to a single `description` const. The sanitization concern is tracked centrally in #910. Resolving.

apps/web/src/pages/EntityDetail/index.tsx
```diff
@@ -330,6 +334,13 @@ const ActivityDetailContent = ({
           durationLabel={hasSleepStages ? 'In Bed' : undefined}
         />
 
+        {!isEditing && description && (
+          <div
+            class="entity-description note-content"
+            dangerouslySetInnerHTML={{ __html: marked.parse(description) as string }}
```
fiddur, line 340: Non-blocking: `marked.parse` output goes straight into `dangerouslySetInnerHTML` with no sanitization (marked v17 doesn't strip HTML), so raw HTML in a note is rendered. This matches the existing pattern in `NotesSection.tsx`/`MarkdownEditor` and the data is the user's own, so it's fine here — flagging only in case these notes ever become shared/multi-user, at which point all three sites should run through a sanitizer.

fiddur (commented):
✅ Approved

Clean, well-scoped UI refactor. The activity "Notes" field becomes a rich-text "Description": in edit mode it now uses the shared `MarkdownEditor` (moved up under the title), and in read mode it renders as a markdown block above the stats table instead of a plain-text stat row. The obsolete `notes` handling in `buildActivityStatRows` and its tests are removed consistently — the interface, the sole call site (`ActivityStatsTable` props), and the tests all drop `notes` together, with no dangling callers. `getUserNotesContent` is reused unchanged, so save/edit semantics are preserved.

The implementation matches existing patterns in the codebase (`NotesSection.tsx`, `MarkdownEditor`), and `marked` is already a direct dependency of `apps/web`.

Non-blocking observations:

- **Unsanitized HTML render.** `marked.parse(description)` is passed to `dangerouslySetInnerHTML` without a sanitizer (marked v17 has no built-in HTML sanitization), so raw HTML/`<script>`/`onerror` in a note is injected into the DOM. This exactly mirrors the pre-existing `NotesSection.tsx` (line 107) and `MarkdownEditor` (line 62), and the content is the user's own data, so it's not a new or blocking risk here — but if any of these fields ever become shared/multi-user it would be worth routing all three through a sanitizer (e.g. DOMPurify).
- **Implicit marked config.** The read-only render in `index.tsx` relies on the global `marked.setOptions({ breaks, gfm })` side-effect executed when `MarkdownEditor`'s module loads (reached via the `EditableActivityFields` import). It works given the current import graph but is an implicit cross-module dependency; same caveat already applies to `NotesSection`.
- **Possible duplicate display.** User-typed comments (no `source`) now surface both in the new Description block and in the `NotesSection` list below. This overlap predates the PR (the old "Notes" stat row had the same overlap), so it's presumably intentional — just worth a sanity check that showing the content in both places is desired.

apps/web/src/pages/EntityDetail/index.tsx
fiddur: Same non-blocking sanitization point as the earlier threads — tracked centrally in #910 (shared DOMPurify render helper for the NotesSection/MarkdownEditor/EntityDetail sinks). Not a regression and content is the user's own, so keeping it out of this PR. Resolving.

PR #886, opened 2026-07-03
feat(web): mirror the activity chart's shown metrics into the share dialog (#831)

fiddur:
## What

Per your request: opening **Share** from an activity now defaults the dialog's checked metrics to **what's shown on the activity's chart**. Toggle a metric off on the chart → it's unchecked in the Share dialog; still fully alterable there.

## How

`ActivityChart` reports its enabled (toggled-on) metrics upward:

```
ActivityChart --onEnabledMetricsChange--> ActivityDetailContent --> ActivityContent
   --chartMetrics--> ShareActivityButton --> ShareActivityDialog
```

In **create** mode the dialog seeds:
- **summary** = `duration` (always) + each summary metric whose source is charted (e.g. charting `heart_rate` → Avg HR + Max HR + HR zones; `distance` → Distance),
- **series** = the full time-series for every charted metric (this deliberately turns high-res series on to match the chart, as you chose).

**Edit** mode still uses the post's saved selection.

`DEFAULT_SUMMARY` + `defaultsFromChart` moved to the shared `feed-metrics` module and unit-tested (HR → avg/max/zones; series-only metrics like speed; unrepresentable metrics ignored; empty fallback).

`pnpm --filter aurboda-web check` clean; new test green.

### DIFF
--- a/apps/web/src/components/ShareActivityDialog.tsx
+++ b/apps/web/src/components/ShareActivityDialog.tsx
@@ -114,6 +114,7 @@ export function ShareActivityDialog({
   const queryClient = useQueryClient()
   // Create mode mirrors the chart's shown metrics; edit mode uses the post's saved selection.
   const defaults = defaultsFromChart(chartMetrics)
+  const mirroredFromChart = !post && (chartMetrics?.length ?? 0) > 0
   const [summary, setSummary] = useState<Set<string>>(
     () => new Set(post ? post.included_metrics : defaults.summary),
   )
@@ -196,7 +197,9 @@ export function ShareActivityDialog({
           <fieldset class="share-dialog-group">
             <legend>Share full time-series</legend>
             <p class="share-dialog-note">
-              Higher resolution — more revealing than a summary. Off by default.
+              {mirroredFromChart
+                ? 'Higher resolution — more revealing than a summary. Pre-checked to match the activity chart; uncheck any you would rather not share.'
+                : 'Higher resolution — more revealing than a summary. Off by default.'}
             </p>
             <div class="share-dialog-options">
               {seriesOptions.map(({ key, label }) => (
--- a/apps/web/src/components/ShareActivityDialog.tsx
+++ b/apps/web/src/components/ShareActivityDialog.tsx
@@ -15,7 +15,7 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { useState } from 'preact/hooks'
 
 import { fetchBucketedMetrics, fetchRawLocations, shareActivity, updateFeedPost } from '../state/api'
-import { SERIES_METRICS, SUMMARY_METRICS } from './feed-metrics'
+import { defaultsFromChart, SERIES_METRICS, SUMMARY_METRICS } from './feed-metrics'
 import './ShareActivityDialog.css'
 
 interface Props {
@@ -24,6 +24,8 @@ interface Props {
   /** Activity window — when given, the dialog offers only metrics with data in it. */
   activityStart?: Date
   activityEnd?: Date
+  /** Metrics currently shown on the activity's chart; mirrored into the defaults. */
+  chartMetrics?: string[]
   /** When present, edit this existing post instead of sharing anew. */
   post?: FeedPost
   onClose: () => void
@@ -35,8 +37,6 @@ const VISIBILITIES: { value: FeedVisibility; label: string; hint: string }[] = [
   { hint: 'Anyone with the link; kept out of public timelines.', label: 'Unlisted', value: 'unlisted' },
   { hint: 'Only your followers.', label: 'Followers only', value: 'followers' },
 ]
-
-const DEFAULT_SUMMARY = ['duration', 'distance', 'heart_rate_avg', 'heart_rate_max', 'calories']
 
 const toggle = <T extends string>(set: Set<T>, key: T): Set<T> => {
   const next = new Set(set)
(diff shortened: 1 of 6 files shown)


apps/web/src/pages/EntityDetail/ActivityChart.tsx
```diff
@@ -590,6 +593,12 @@ export const ActivityChart = ({
 
   const enabledMetrics = new Set(availableMetrics.filter((m) => !disabledMetrics.has(m)))
 
+  // Surface the currently-shown metrics so a parent (the share dialog) can mirror
+  // the user's chart selection. Runs on load and each toggle (deps change ref).
+  useEffect(() => {
+    onEnabledMetricsChange?.(availableMetrics.filter((m) => !disabledMetrics.has(m)))
```
fiddur: This effect can loop. `availableMetrics` (line 559) is `chartDataQuery.data?.metrics ?? []`, so it's a **new `[]` on every render** whenever the query has no data — the initial `chartWidthPx === 0` window, while loading, and permanently on error. That unstable `[]` is in the dep array, so the effect re-runs each render and calls `onEnabledMetricsChange([])` → `setChartMetrics([])` (a fresh array, so Preact doesn't bail out) → parent re-renders → `ActivityChart` re-renders → new `[]` → effect fires again. On a happy load it's a burst of redundant full-chart re-renders; on a failed metrics fetch it never terminates.

The existing effect at line 569 shares the same unstable dep but is safe because its body is guarded and never calls a parent setter. Guard this one too — e.g. skip while data is absent:

```ts
useEffect(() => {
  if (!chartDataQuery.data) return
  onEnabledMetricsChange?.(availableMetrics.filter((m) => !disabledMetrics.has(m)))
}, [availableMetrics, disabledMetrics, onEnabledMetricsChange, chartDataQuery.data])
```

so it only notifies on real load/toggle transitions and never feeds an empty array back into the parent.

fiddur (commented):
🛑 Changes required

The feature is well factored: `defaultsFromChart` is a clean pure function extracted into `feed-metrics.ts` with good unit-test coverage, and the prop threading (`onEnabledMetricsChange` → `chartMetrics` → dialog defaults) is coherent. The dialog's edit/create split reads clearly.

One blocking issue: the new notify-effect in `ActivityChart.tsx` (lines 596–600) can drive an infinite re-render loop. `availableMetrics` is `chartDataQuery.data?.metrics ?? []`, so it is a fresh `[]` on every render whenever the query has no data — which includes the initial `chartWidthPx === 0` window, while the fetch is in flight, and permanently on query error. Because that unstable `[]` is in the effect's dependency array, the effect re-runs every render; its body calls `onEnabledMetricsChange([])` → `setChartMetrics([])` (a new array, so not bailed out by Preact) → the parent re-renders → `ActivityChart` re-renders → `availableMetrics` is a new `[]` again → the effect fires again. On the happy path this is a wasteful burst of redundant full-chart renders during load; on a failed metrics fetch (`data` stays `undefined`) it never terminates. Note the pre-existing effect at line 569 depends on the same unstable `availableMetrics` but is safe precisely because its body is guarded (`availableMetrics.length > 0` + a ref) and never calls a parent setter. The new effect should be guarded the same way — e.g. skip the notify while `chartDataQuery.data` is undefined (or gate on `availableMetrics.length > 0`) so it only fires on real load/toggle transitions and never feeds an empty array back into the parent.

Everything else looks good.

apps/web/src/pages/EntityDetail/ActivityChart.tsx
fiddur: Fixed: the effect is now keyed on the enabled set's contents (a stable `|`-joined string) instead of the availableMetrics/disabledMetrics refs, and reconstructs the list inside from that key. availableMetrics being a fresh [] each render no longer re-fires it, so the setChartMetrics loop is gone; it fires only when the shown metrics actually change.

apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -106,19 +106,27 @@ export function ShareActivityDialog({
...
+  // Preselect from the post's shared series (edit) or the charted metrics (create),
+  // intersected with keys this dialog can represent — keeps the state typed as
+  // `Set<MetricType>` without a cast.
   const [series, setSeries] = useState<Set<MetricType>>(
-    () => new Set(SERIES_METRICS.map((m) => m.key).filter((k) => post?.series_metrics.includes(k) ?? false)),
+    () =>
+      new Set(
+        post
+          ? SERIES_METRICS.map((m) => m.key).filter((k) => post.series_metrics.includes(k))
+          : defaults.series,
```
fiddur, line 129: This defaults the full time-series selection from the chart's enabled metrics. Since the chart starts with *all* available metrics enabled (`disabledMetrics` begins empty in `ActivityChart`), a user who hasn't touched the chart toggles will see the "Share full time-series" checkboxes pre-checked for every recorded series metric (heart rate, speed, power, cadence, …) on a fresh Share. That contradicts the note in that fieldset — "more revealing than a summary. Off by default." — and is a privacy-relevant change from the previous behavior (create mode defaulted series to empty). Consider keeping full series off by default (mirror only `defaults.summary`), or update the "Off by default" copy if defaulting them on is intended.

fiddur (commented):
🛑 Changes required

Clean, well-factored change: the chart→dialog plumbing is straightforward, the render-loop concern is understood and guarded (keying the effect on the enabled-set *contents* rather than the fresh-`[]` refs), the empty-key `split('|')` edge case is handled, and `defaultsFromChart` is nicely unit-tested.

One substantive concern about the **series** default (see inline). Because the chart enables *all* available metrics by default (`disabledMetrics` starts empty, so `enabledMetrics === availableMetrics`), a user who never touches the chart toggles now has *every* recorded series metric (heart rate, speed, power, cadence, …) pre-checked under "Share full time-series" the first time they open the Share dialog. That directly contradicts the on-screen note in that same fieldset — "Higher resolution — more revealing than a summary. Off by default." — and quietly opts users into sharing the more-revealing high-resolution data by default. Previously create-mode defaulted series to empty.

The summary-metric mirroring is a reasonable UX win and I'd keep it. But the series behavior should be reconciled: either keep full series off by default (mirror only summaries), or, if defaulting them on is genuinely intended, update the "Off by default" copy so the UI isn't misleading about a privacy-relevant default. Flagging as changes_required because the on-screen copy is now factually wrong regardless of which way you resolve the intent.

Minor (non-blocking): `const defaults = defaultsFromChart(chartMetrics)` runs on every render even though it's only consumed by the two lazy `useState` initializers (first render only) — harmless, just wasted work.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: This is a deliberate, informed product decision: the feature was requested specifically as 'mirror the chart toggles, and pre-check their full time-series too', chosen over a 'series stay off' option that spelled out it's more revealing. It isn't silent — the pre-checked series are visibly checked and fully alterable before the deliberate Share click, and the section note now explicitly says they're pre-checked to match the chart and to uncheck any you'd rather not share (edit mode / no-chart still shows 'off by default'). Keeping the behavior; resolving.

fiddur (commented):
✅ Approved

Clean, well-tested feature. `defaultsFromChart` is a pure, easily-testable function extracted into `feed-metrics.ts`, and the new unit tests cover the meaningful cases (HR fan-out to avg/max/zones + series, series-only metrics like speed, ignored metrics like `hrv_rmssd`, and the empty/unknown fallback to `DEFAULT_SUMMARY`). I traced each expectation against `SUMMARY_METRICS`/`SERIES_METRICS` and they match.

The render-loop concern in `ActivityChart` is handled correctly: the new effect keys on the content string `enabledKey` (not the fresh-`[]`-per-render `availableMetrics` ref), and the callback chain resolves to the stable `setChartMetrics` state setter, so there's no feedback loop. The `DEFAULT_SUMMARY` constant was cleanly relocated (only referenced by the test and `defaultsFromChart`, and only ever spread into a `Set`, so no mutation/aliasing risk).

One non-blocking product note: on create, series that appear on the chart are now pre-checked, whereas time-series sharing was previously opt-in/off-by-default ("more revealing than a summary"). This is the explicit intent of #831 and the note copy was updated to warn the user ("Pre-checked to match the activity chart; uncheck any you would rather not share"), so it's an accepted decision — just calling it out since it changes the privacy default for the full-resolution series. A minor cosmetic edge: `mirroredFromChart` shows the "pre-checked" note whenever chart metrics exist even if none of them map to an offered series option (e.g. only `hrv_rmssd` charted while `seriesOptions` lists HR from the availability query), but that mismatch is unlikely in practice since both derive from the same activity's data.

PR #889, opened 2026-07-03
fix(sharing): visible profile Share button + owner attribution avatar

fiddur:
Fixes two tryout-findings from the #834 sharing work.

## #883 — Share button invisible on public profile
The `.public-profile-header` used a `<header>` element, which matches the global `header { background-color: #673ab8 }` (mobile-nav) rule, painting a purple bar. My `#673ab8` Share button was then purple-on-purple → invisible. That purple bar was itself unintended (introduced when I wrapped the profile header in PR #876). **Fix:** use a `<div>` — removes the accidental purple bar, and the button (purple on the light page) reads fine.

## #879 — Owner's own view missing the attribution avatar
The avatar was only added to `ReadOnlyDashboard`, not `OwnerSharedDashboard`, so owners saw a bare `@handle`. **Fix:** extract a shared `AttributionLink` (avatar + handle) used by **both** views, so they can't drift again (as the finding suggested).

## Verification
`pnpm check` green. Both are UI-only; the finding repro steps (avatar upload → owner vs anonymous view; profile Share button contrast) are covered by the change.

`Fixes #883`, `Fixes #879`.

### DIFF
--- a/apps/web/src/pages/PublicDashboard/index.tsx
+++ b/apps/web/src/pages/PublicDashboard/index.tsx
@@ -26,6 +26,17 @@ import './style.css'
 const gridClass = (type: SectionType): string =>
   type === 'links' ? 'links-grid' : type === 'charts' ? 'charts-grid' : 'metrics-grid'
 
+/** Author attribution (avatar + @handle). Shared so the owner and public views
+ *  can't drift apart (#879). */
+function AttributionLink({ username }: { username: string }) {
+  return (
+    <a class="public-attribution" href={`/u/${encodeURIComponent(username)}`}>
+      <img class="public-attribution__avatar" src={avatarUrl(username)} alt="" width={28} height={28} />@
+      {username}
+    </a>
+  )
+}
+
 /** Owner view: live, editable dashboard backed by the authed shared-dashboard API. */
 function OwnerSharedDashboard({ username, slug }: { username: string; slug: string }) {
   const queryClient = useQueryClient()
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Clean, minimal frontend fix. Two well-scoped changes:

1. **`AttributionLink` extraction (PublicDashboard):** The new component's markup is byte-for-byte identical to the previous `ReadOnlyDashboard` attribution (avatar + `@handle`), and the owner view — which previously rendered only `@{username}` without an avatar — now shares it. This delivers the intended owner attribution avatar and removes the drift risk called out in #879. `avatarUrl` is already imported, and the `function` declaration matches the file's existing style.

2. **`<header>` → `<div>` (PublicProfile):** Verified the rationale — the global `header` rule in `apps/web/src/style.css` sets `background-color: #673ab8` with `display: none` (flipped to `display: block` under `max-width: 60em`), so on an anonymous, chrome-less page it either hid the header on desktop or painted a purple bar over the Share button on mobile. Since all `.public-profile-header` CSS is class-based, swapping the element preserves styling while dropping the global-`header` side effects. The inline comment documents this well.

Minor, non-blocking: swapping `<header>` for `<div>` loses the semantic banner landmark. Given the constraint (the overly-broad global `header` selector) and that a `<div>` avoids touching shared styles, this is a reasonable trade-off — no change requested.

No tests reference these components, and the change is purely presentational. Looks good.

fiddur (commented):
✅ Approved

Clean, focused fix. Both changes are correct and well-documented.

- **`PublicDashboard`**: Extracting `AttributionLink` de-duplicates the attribution markup and resolves the drift (#879) where the owner view rendered only `@username` text while the read-only view showed avatar + handle — both now render identically. `avatarUrl` is already imported and the link still uses `encodeURIComponent`.
- **`PublicProfile`**: Switching `<header>` to `<div>` is the right fix. The global `header` rule in `apps/web/src/style.css` sets `background-color: #673ab8`; `.public-profile-header` overrides `display` but not the background, so the purple bar hid the purple Share button (#883). All `.public-profile-header` CSS selectors are class-based, so styling carries over to the `<div>` unchanged, and the inline comment records the rationale.

Frontend-only, low-risk, and no existing tests reference these presentational components. Nothing blocking.

PR #907, opened 2026-07-03
fix(feed): render chart text (bundle fonts) + cache feed image renders (#887, #892)

fiddur:
Chunk 1 of the feed follow-up queue (batched).

## #887 — chart text renders as tofu boxes
The runtime image (`node:25-alpine`) ships **no fonts** (`fc-list` → 0), so the chart PNG's `<text>` (title + min/max labels) rendered as missing-glyph boxes; the route map (no text) was fine.

**Fix:** install `fontconfig` and register the already-bundled Liberation TTFs (`apps/backend/src/assets/fonts`) via `fc-cache` in the Docker runtime stage, and reference `Liberation Sans` explicitly in the chart SVG so librsvg/sharp resolves it. (Reuses the fonts already shipped for the OG-image renderer.)

## #892 — image endpoints render per request
`/public/:username/feed/:postId/{chart,route}.png` re-fetched raw series/GPS and rasterized SVG→PNG on every request.

**Fix:** a bounded `createRenderCache` (LRU + in-flight dedup, mirroring `og-image-router`), keyed per post. Past-activity data is effectively immutable so a bounded cache + restart eviction suffices, and the **eligibility gate runs first** so an unshared/visibility-changed post still 404s without touching the cache. Renderers are injected (`FeedImageDeps.renderChart/renderRoute`), so the cache behaviour is unit-tested: render-once, concurrent-dedup, per-key, null-not-cached, eviction.

Closes #887, #892. `pnpm --filter aurboda-backend check` + tests green.

Note: the font fix is a Docker runtime change (not exercised by Node CI); verified against the documented `fc-list`/`fc-match` failure and the standard fontconfig setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,9 @@ RUN pnpm --filter @aurboda/api-spec generate:openapi && pnpm --filter @aurboda/a
 # Production stage
 FROM node:25-alpine
 
-# Install nginx
-RUN apk add --no-cache nginx
+# Install nginx + fontconfig (librsvg, via sharp, needs a registered font to
+# rasterize SVG <text> in server-rendered images — the base image ships none).
+RUN apk add --no-cache nginx fontconfig
 
 # Install pnpm
 RUN npm install -g pnpm@10
@@ -46,6 +47,12 @@ COPY --from=builder /app/packages/api-spec/dist ./packages/api-spec/dist
 # Copy backend source (Node 25 runs TypeScript directly via built-in type stripping)
 COPY tsconfig.json ./
 COPY apps/backend/src ./apps/backend/src
+
+# Register the bundled Liberation fonts so librsvg/sharp can render SVG text
+# (feed chart images) instead of tofu boxes — the base image has no fonts.
+RUN mkdir -p /usr/share/fonts/liberation \
+ && cp apps/backend/src/assets/fonts/*.ttf /usr/share/fonts/liberation/ \
+ && fc-cache -f
 
 # Copy built web frontend
 COPY --from=builder /app/apps/web/dist /usr/share/nginx/html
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped fix for the two linked issues.

**Font rendering (#887):** The Dockerfile now installs `fontconfig` and registers the git-tracked Liberation fonts (`mkdir /usr/share/fonts/liberation` → `cp *.ttf` → `fc-cache -f`). The copy path is valid because `apps/backend/src` is already `COPY`d (line 49) before the font step, and `feed-images.ts` switches the SVG `font-family` to `Liberation Sans, sans-serif`, which matches the bundled font's internal family name. Correct approach to eliminate the tofu/blank text in server-rendered chart images.

**Render caching (#892):** `renderChart`/`renderRoute` are now injected through `FeedImageDeps` (a nice testability win over the previous direct import), and `createRenderCache` adds a bounded FIFO cache with in-flight de-duplication that mirrors `og-image-router`. The design is sound: the eligibility gate (`resolveImageWindow`) runs on every request *before* the cache is consulted, so un-share / visibility-flip still 404s immediately and `no-store` revocability is preserved; `null` (no-data) results are deliberately not cached; and `inFlight` is always cleaned up in `finally`, so a rejected render doesn't wedge the key. The new unit tests cover hit, concurrent dedup, per-key isolation, null-not-cached, and eviction. I could not execute the suite here (deps aren't installed in this sandbox and the Node version mismatches), but I traced the logic and tests by hand and they line up — CI will run them.

**Minor (non-blocking):** the module doc comment at the top of `apps/backend/src/services/activitypub/feed-images.ts` still describes the images as rasterized "(no fonts, no external tiles)", which this change directly contradicts — chart text now depends on a registered `Liberation Sans` font being present. Worth updating that parenthetical for accuracy (per the CLAUDE.md "keep docs/comments up to date" guidance), but it doesn't affect behavior.

PR #908, opened 2026-07-03
perf(feed): cursor-paginate the ActivityPub outbox (#890)

fiddur:
Chunk 2 of the feed follow-up queue.

## Problem (#890)
The outbox dispatcher listed **every** public post and built each `Create` inline — `getActivityById` + `resolveActivityScalars` (a bucketed metrics-aggregation query per post). An unauthenticated `/users/:u/outbox` fetch therefore scaled O(posts × queries), unbounded.

## Fix
Serve the outbox as a **cursor-paginated** `OrderedCollection` (`OUTBOX_PAGE_SIZE = 20`):
- root `/outbox` → `totalItems` + `first`/`last` page links (no inline items),
- `/outbox?cursor=<offset>` → one page of `Create`s + a `next` link.

So a fetch resolves at most one page of posts. The cursor is a simple offset (a concurrent share can shift a page boundary — fine for an occasionally-crawled outbox; noted in a comment). New `listPublicFeedPostsPage` DB helper.

## Tests
Integration: root/page split (root has `first` link, page has the `Create`), page boundary past 20 with `next` cursor, followers-only exclusion still 0. DB: `listPublicFeedPostsPage` newest-first limit/offset. `pnpm --filter aurboda-backend check` + tests green.

Closes #890.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -26,7 +26,7 @@ import {
   getOrCreateActorKeyPair,
   isMissingDatabase,
   listFeedFollowers,
-  listPublicFeedPosts,
+  listPublicFeedPostsPage,
   removeFeedFollower,
   upsertFeedFollower,
 } from '../../db/index.ts'
@@ -34,6 +34,9 @@ import { buildProfileUrl } from '../share-urls.ts'
 import { buildFeedCreate, buildFeedNote } from './deliver.ts'
 import { toCryptoKeyPair } from './keys.ts'
 import { isPubliclyVisible } from './object.ts'
+
+/** Posts per outbox page (cursor pagination). */
+const OUTBOX_PAGE_SIZE = 20
 
 /** RFC 4122 canonical form — guards `getFeedPostById` from a non-UUID `postId`
  * (Postgres would otherwise raise `invalid input syntax for type uuid`). */
@@ -167,17 +170,20 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
   )
 
   // Outbox: the user's public + unlisted posts as `Create` activities, so a
-  // Mastodon profile shows them. Served as a single inline OrderedCollection (no
-  // cursor) — fine at feed scale; pagination is a later slice. A post whose
-  // activity was soft-deleted is skipped from the items (its Create can't be
-  // built) while `setCounter` still counts it; the divergence self-heals when
-  // the stale post is removed.
+  // Mastodon profile shows them. Cursor-paginated (`OUTBOX_PAGE_SIZE` per page)
+  // so an unauthenticated fetch never resolves every post's scalars at once; the
+  // cursor is a simple offset (a concurrent share can shift a page boundary —
+  // acceptable for an occasionally-crawled outbox). A post whose activity was
+  // soft-deleted is skipped from the items while `setCounter` still counts it;
+  // the divergence self-heals when the stale post is removed.
   federation
-    .setOutboxDispatcher('/users/{identifier}/outbox', async (ctx, identifier) => {
+    .setOutboxDispatcher('/users/{identifier}/outbox', async (ctx, identifier, cursor) => {
       if (!isValidUsername(identifier)) return null
+      const offset = cursor == null ? 0 : Number.parseInt(cursor, 10)
+      if (!Number.isInteger(offset) || offset < 0) return null
       let posts
       try {
-        posts = await listPublicFeedPosts(identifier)
+        posts = await listPublicFeedPostsPage(identifier, OUTBOX_PAGE_SIZE, offset)
       } catch (error) {
         if (isMissingDatabase(error)) return null
         throw error
@@ -191,7 +197,8 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
           }),
         )
       ).filter((item): item is Create => item != null)
-      return { items }
+      const nextCursor = posts.length === OUTBOX_PAGE_SIZE ? String(offset + OUTBOX_PAGE_SIZE) : null
+      return { items, nextCursor }
     })
     .setCounter(async (_ctx, identifier) => {
       if (!isValidUsername(identifier)) return 0
@@ -199,6 +206,20 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
         return await countPublicFeedPosts(identifier)
       } catch (error) {
         if (isMissingDatabase(error)) return 0
+        throw error
+      }
+    })
+    .setFirstCursor((_ctx, identifier) => (isValidUsername(identifier) ? '0' : null))
+    .setLastCursor(async (_ctx, identifier) => {
+      if (!isValidUsername(identifier)) return null
+      try {
+        const count = await countPublicFeedPosts(identifier)
+        // Offset of the last page; '0' for an empty or single-page outbox.
+        return count <= OUTBOX_PAGE_SIZE
+          ? '0'
+          : String(Math.floor((count - 1) / OUTBOX_PAGE_SIZE) * OUTBOX_PAGE_SIZE)
+      } catch (error) {
+        if (isMissingDatabase(error)) return null
         throw error
       }
     })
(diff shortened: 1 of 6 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -167,17 +170,20 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
...
+  // cursor is a simple offset (a concurrent share can shift a page boundary —
+  // acceptable for an occasionally-crawled outbox). A post whose activity was
+  // soft-deleted is skipped from the items while `setCounter` still counts it;
+  // the divergence self-heals when the stale post is removed.
   federation
-    .setOutboxDispatcher('/users/{identifier}/outbox', async (ctx, identifier) => {
+    .setOutboxDispatcher('/users/{identifier}/outbox', async (ctx, identifier, cursor) => {
       if (!isValidUsername(identifier)) return null
+      const offset = cursor == null ? 0 : Number.parseInt(cursor, 10)
+      if (!Number.isInteger(offset) || offset < 0) return null
```
fiddur: This guard rejects negatives and non-integers but accepts arbitrarily large offsets. A crafted cursor beyond bigint range (e.g. `?cursor=99999999999999999999`) parses to `1e20`, which passes `Number.isInteger`/`>= 0` and is handed to Postgres `OFFSET`. pg serializes it as `"100000000000000000000"`, which exceeds bigint max, so Postgres raises `value out of range for type bigint` — an error that isn't `isMissingDatabase`, so it rethrows and surfaces as a 500 on the public, unauthenticated outbox. Consider also rejecting offsets above a sane bound (e.g. `> Number.MAX_SAFE_INTEGER`, or `> count`) so a bad cursor returns null/empty rather than 500ing.

fiddur (commented):
✅ Approved

Clean, well-scoped perf change: the ActivityPub outbox now serves a paginated `OrderedCollection` (root → `first`/`last` page links, `OUTBOX_PAGE_SIZE`-sized pages via offset cursors) instead of resolving every post's scalars in one unauthenticated fetch. The DB layer gets a `listPublicFeedPostsPage` that mirrors the existing visibility filter and `created_at DESC, id DESC` ordering, it's exported from `db/index.ts`, and both the DB function and the paginated HTTP surface get integration tests (page boundaries, past-page-size, empty/followers-only). The `next`/`last` cursor math checks out (based on raw `posts.length`, so soft-deleted filtering can't break pagination), and the pre-existing counter/items divergence is carried over and documented.

One robustness edge worth tightening (inline). Two minor notes, non-blocking:

- `listPublicFeedPosts` (the non-paginated variant) now has no production callers — only its own test and a doc-comment reference in `object.ts`. Consider removing it or noting why it's retained.
- The `listPublicFeedPostsPage` docstring says `limit`/`offset` are "clamped by the caller", but the caller fixes `limit` to a constant and only validates that `offset` is a non-negative integer — there's no upper bound (see inline).

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed: the guard now uses Number.isSafeInteger, so an out-of-range/crafted cursor (e.g. 1e20) 404s instead of reaching Postgres OFFSET and overflowing bigint. Added an integration test for the out-of-range cursor.

fiddur (commented):
🛑 Changes required

Solid, well-scoped change — the cursor pagination logic is correct and the added integration tests cover the meaningful edges (page boundary at 21 posts, empty `followers`-only outbox, out-of-range cursor → 404). A few things I verified as sound:

- The offset guard (`Number.isSafeInteger(offset) || offset < 0`) correctly rejects crafted cursors like `1e20` before they reach Postgres, avoiding the `bigint` `OFFSET` overflow → the `?cursor=99999999999999999999` test passes for the right reason.
- `nextCursor` keys off `posts.length === OUTBOX_PAGE_SIZE` (the fetched-row count, not the filtered `items.length`), so a page containing soft-deleted activities still advances the crawler instead of dead-ending — this is the right choice given the documented items/counter divergence.
- `setLastCursor`'s math is correct at the boundaries (`count <= OUTBOX_PAGE_SIZE → '0'`; otherwise `floor((count-1)/PAGE)*PAGE`), and `listPublicFeedPostsPage` preserves the same `visibility IN ('public','unlisted')` filter and `created_at DESC, id DESC` ordering as `listPublicFeedPosts`.

**Blocking (documented standard):** `docs/features/feed.md` is now stale and actively contradicts the shipped behavior. CLAUDE.md requires docs be kept in sync when changing related code ("When changing APIs, sync behavior... update the corresponding docs"). Specifically:
- Line ~207-208 says **"The outbox is not paginated — it serves all public posts inline... cursor pagination is a follow-up."** This PR *is* that follow-up, so the statement is now false and should be removed/rewritten.
- Line ~85 describes the outbox as "an `OrderedCollection` of the user's posts... inline"; it should now describe the paginated (cursor / `first`/`next`/`last`) collection with `OUTBOX_PAGE_SIZE` per page.

Please update those doc sections to reflect the paginated outbox; everything else looks good to merge.

fiddur (commented):
✅ Approved

Clean, well-tested performance improvement that converts the ActivityPub outbox from a single inline `OrderedCollection` to a cursor-paginated collection, so an unauthenticated fetch only resolves one page (`OUTBOX_PAGE_SIZE`) of post scalars/`Create`s at a time.

The implementation is correct on the points that matter:

- **New DB function** `listPublicFeedPostsPage` uses the exact same deterministic ordering (`created_at DESC, id DESC`) as `listPublicFeedPosts`, with parameterized `LIMIT`/`OFFSET`, and has an integration test covering the newest-first paging behavior.
- **Cursor validation** in the dispatcher (`Number.isSafeInteger(offset) || offset < 0` → `null`) correctly rejects the crafted-`1e20` case that would otherwise overflow the Postgres `bigint` `OFFSET` and 500 the request — verified by the new `out-of-range cursor` 404 test — and rejects negatives.
- **`nextCursor`** is derived from `posts.length` (raw DB rows) rather than the filtered `items.length`, so pages whose activities were soft-deleted don't prematurely terminate pagination. Good call.
- **`setLastCursor`** math (`Math.floor((count - 1) / OUTBOX_PAGE_SIZE) * OUTBOX_PAGE_SIZE`) always lands on a page-size-aligned offset that is reachable via the `next` chain, and returns `'0'` for empty/single-page outboxes.
- Fedify 2.3.1's `PageItems.nextCursor`, `setFirstCursor`, `setLastCursor`, and `setCounter` API and the `{ items, nextCursor }` return shape all check out.
- Docs (`docs/features/feed.md`) are updated: the outbox is now described as cursor-paginated and the stale "not paginated" known-limitation bullet is removed.

Only non-blocking observations (no action required): the offset scheme yields a harmless trailing empty page with a `next` link when the post count is an exact multiple of the page size (standard for offset pagination, and the code comment already notes concurrent shares can shift boundaries); and `Number.parseInt` is lenient on trailing garbage, which is moot since Fedify generates the cursor values. `listPublicFeedPosts` remains exported and still has its own integration test, so removing it from the production path leaves no dead-code concern.

LGTM.

fiddur (commented):
✅ Approved

Clean, well-scoped implementation of cursor pagination for the ActivityPub outbox, with solid test coverage and matching docs.

Correctness checks all hold up:
- **`nextCursor`** is derived from `posts.length === OUTBOX_PAGE_SIZE` (the raw DB row count), not the soft-delete-filtered `items` — so a page with some skipped stale posts still paginates forward correctly.
- **`setLastCursor`** math is right at the boundaries (`count<=20 → '0'`; `21 → 20`; `40 → 20`; `41 → 40`).
- **The offset guard** (`Number.isSafeInteger(offset) || offset < 0`) correctly turns a crafted huge/negative cursor into a 404 before it can overflow the Postgres `bigint` OFFSET — nice defensive touch, and the `404s an out-of-range cursor` test locks it in.
- The Fedify 2.3.1 API is used correctly: the `(ctx, identifier, cursor)` dispatcher signature and `{ items, nextCursor }` return match `PageItems`, and `setFirstCursor`/`setLastCursor`/`setCounter` all exist.
- Ordering `created_at DESC, id DESC` is deterministic and consistent with the existing `listPublicFeedPosts`, so offset paging is stable (concurrent-share boundary shift is documented and acceptable).

The new `listPublicFeedPostsPage` DB helper has an integration test (satisfies the CLAUDE.md rule that all DB functions be covered), the federation integration tests exercise the root/page split, past-page-size paging, and the out-of-range cursor, and `docs/features/feed.md` is updated (removing the stale "not paginated" limitation).

Minor, non-blocking: after this change `listPublicFeedPosts` is no longer called from production code (only from its own integration test and a couple of doc comments). Fine to keep as a general listing helper, but worth being aware it's now effectively unreferenced outside tests.

PR #909, opened 2026-07-03
fix(backend): avoid stack overflow on large metric result sets

fiddur:
## Symptom

`GET /activities` over a wide date range 500s with:

```
RangeError: Maximum call stack size exceeded
    at querySplitByCumulative (apps/backend/src/db/cumulative-query.ts:91)
    at getTimeSeriesMultiMetric (apps/backend/src/db/time-series.ts:227)
    at queryActivities (apps/backend/src/services/queries/activities.ts:180)
```

## Root cause

Not recursion — it's the spread. `querySplitByCumulative` appended mapped rows with:

```ts
results.push(...result.rows.map(options.mapRow))
```

`push(...bigArray)` (via `Function.apply`) passes every element as a separate argument. Once the array exceeds V8's argument-count limit (~100k+), that throws `RangeError: Maximum call stack size exceeded`. A non-cumulative metric like `heart_rate` sampled per-second over an 8-day span is hundreds of thousands of rows, so `getTimeSeriesMultiMetric` tips over.

## Fix

Append each mapped row in a plain loop (no spread), which has no argument-count limit. Applied to all three branches (cumulative / aurboda-only / non-cumulative).

## Test

Added a regression test that feeds 500k rows through `querySplitByCumulative`. It throws `RangeError` on the old spread-based code and passes with the loop.

- `pnpm check` — 0 errors
- `pnpm vitest run src/db/cumulative-query.test.ts` — 9 passed


fiddur (commented):
✅ Approved

Clean, well-targeted fix. `results.push(...rows.map(options.mapRow))` spreads every mapped row as a separate function argument, and V8 caps argument count (~65k–125k), so a large result set (e.g. multi-day per-second `heart_rate`) throws `RangeError: Maximum call stack size exceeded`. Replacing the three spreads with a per-element `for` loop (`appendMapped`) removes the limit while preserving accumulation order across the cumulative/aurboda-only/non-cumulative branches. The behavior is equivalent to the old code — the previous `.map(options.mapRow)` passed `(element, index, array)`, but `mapRow` is typed to take only `row`, so `options.mapRow(row)` is identical. The regression test with 500k rows would have thrown under the old implementation and passes now, and the comment documents the pitfall clearly.

Non-blocking FYI (outside this PR's diff): the same `push(...arr.map(...))` pattern exists at `apps/backend/src/services/correlations/hrv-activities.ts:107`. That one operates on per-record windowed slices bounded by activity duration, so it's much lower risk — worth keeping in mind but not required for this change.

PR #912, opened 2026-07-03
feat(feed): Note published + 410 Tombstone on deleted objects (#894, #895)

fiddur:
Chunk 3 of the #831 feed follow-up queue: **Delivery lifecycle** — closes #894 and #895.

## #894 — Delivered Note carries a `published` timestamp
The `Note` (and outbox `Create`) now set `published` = the post's share time (`created_at`), so remote servers order and timestamp posts correctly instead of stamping them at receipt.

Fedify types `published` as the ambient (`esnext.temporal`) `Temporal.Instant`, which our `@js-temporal/polyfill` value isn't *nominally* assignable to (their method signatures differ) — but the values are the **identical class at runtime** (Fedify imports the same polyfill). A dedicated `temporal-interop.ts` helper bridges the two with a single documented boundary cast, confined to that one spot, so every call site stays cast-free.

## #895 — Deleted object returns `410 Gone` Tombstone (not `404`)
After unshare, a `public`/`unlisted` post's object id now answers **`410 Gone` with an AS2 `Tombstone`**, so a dereferencing server learns the object is *permanently* gone rather than transiently missing.

- Deleting a post records its id in a new **`feed_tombstone`** table, atomically in the same `DELETE` statement (a CTE).
- `followers`-only posts are **never** tombstoned — their id never resolved publicly, so a 410 would leak that a post once existed.
- Fedify's `handleObject` serialises a returned `Tombstone` as **200**, so the 410 can't come from the object dispatcher. Instead a small router mounted right after the Fedify integration catches the dispatcher's `null` → `@fedify/express` `next()` fall-through and serves the 410.

## Tests
- `temporal-interop.test.ts` — epoch-ms preservation / round-trip.
- `feed-tombstone-router.test.ts` — 410 body/headers, origin trailing-slash, fall-through on no-tombstone / invalid username / non-UUID.
- `feed.integration.test.ts` — tombstone recorded for public/unlisted deletes, **not** for followers-only, none for live posts.
- `federation.integration.test.ts` — end-to-end: live object → 200 Note (now with `published`), delete → 410 Tombstone through the real Fedify + Express wiring; deleted followers-only stays 404.

Docs (`feed.md`) updated: the two former caveats ("`published` omitted", "no 410 Tombstone") are removed and the object-serving / storage sections describe the new behaviour.

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -32,6 +32,7 @@ import {
   getDeductionRulesByIds,
   getDetectedLocationById,
   getEnabledDeductionRules,
+  getFeedTombstone,
   getNamedLocations,
   insertActivities,
   insertActivity,
@@ -52,6 +53,7 @@ import { ouraClient } from './integrations/oura/client.ts'
 import { createOwnTracksRouter } from './integrations/owntracks/router.ts'
 import { stravaClient } from './integrations/strava/client.ts'
 import { createMcpRouter } from './mcp.ts'
+import { createFeedTombstoneRouter } from './routes/feed-tombstone-router.ts'
 import { createOAuthRouter } from './routes/oauth-router.ts'
 import { deliverFeedDelete, deliverFeedPost, deliverFeedUpdate } from './services/activitypub/deliver.ts'
 import { createFeedFederation } from './services/activitypub/federation.ts'
@@ -361,6 +363,11 @@ const main = async () => {
   // spoof them.
   httpd.set('trust proxy', 'loopback')
   httpd.use(integrateFederation(feedFederation, () => undefined))
+
+  // `410 Gone` Tombstone for dereferenced deleted objects. Mounted right after
+  // the federation integration: when the object dispatcher returns null for a
+  // since-deleted post, `@fedify/express` calls next(), and this catches it.
+  httpd.use(createFeedTombstoneRouter({ getTombstone: getFeedTombstone, origin: webHost }))
 
   httpd.use(json({ limit: '10mb' }))
 
--- a/apps/backend/src/db/feed.integration.test.ts
+++ b/apps/backend/src/db/feed.integration.test.ts
@@ -13,6 +13,7 @@ import {
   type FeedPostInput,
   findCoveringSharedSeriesWindow,
   getFeedPostById,
+  getFeedTombstone,
   listFeedPosts,
   listPublicFeedPosts,
   listPublicFeedPostsPage,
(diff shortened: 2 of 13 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped implementation of the two feed-federation lifecycle features (#894 `410 Tombstone`, #895 `Note.published`). I verified the correctness-critical pieces and found nothing blocking.

**What I checked and confirmed:**
- **Atomic delete + tombstone CTE** (`db/feed.ts`): the `tomb` CTE is a data-modifying `INSERT` that PostgreSQL always executes to completion even though the primary `SELECT id FROM deleted` doesn't reference it — so the tombstone is reliably recorded in the same statement. The `WHERE visibility IN ('public','unlisted')` filter and `ON CONFLICT (post_id) DO NOTHING` (which preserves the original `deleted_at`) are correct, and the integration tests exercise all four visibility/lifecycle cases. Return semantics (`result.rows.length > 0`) match the prior `rowCount`-based behavior.
- **Tombstone router** (`routes/feed-tombstone-router.ts`): guards mirror the object dispatcher exactly (`isValidUsername` + `UUID_RE`), the store is only consulted for well-formed inputs (tests assert this), `isMissingDatabase` falls through to 404, and the constructed id matches `getObjectUri(Note, …)`. Mounting after `integrateFederation` and before the JSON body parser / audit middleware is correct for an unauthenticated GET-only dereference. Missing-table on an un-migrated per-user DB self-heals via `query()`'s schema-error auto-migrate-and-retry, so there is no window where the query 500s.
- **`published` wiring** (`deliver.ts` + `temporal-interop.ts`): both `Note` and `Create` now carry `published = created_at`, consistent with the existing `object.ts` convention. The `dateToTemporalInstant` boundary cast is the one place the two structurally-different-but-runtime-identical `@js-temporal/polyfill` `Instant` types can be reconciled; it's confined to a single well-documented helper with its own unit tests, so while it's a cast (which CLAUDE.md discourages), it's a justified interop boundary rather than a papered-over type error.
- **Migration/schema**: `feed_tombstone` is added to both `socialTables` and `tableCreationOrder`; existing DBs pick it up via the migrate loop (creates missing tables). No FK, so ordering is fine.
- **Parity**: both the REST delete route and the MCP `delete_feed_post` tool go through `deleteFeedPost` (tombstone recorded) and deliver the `Delete{Tombstone}`.
- **Docs** updated to match (removed the two stale caveats, documented the 410 + `published` behavior).

**Minor, non-blocking observations:** the 410 is served with `application/activity+json` (no charset/profile parameter) regardless of `Accept`, so a browser hitting a deleted object URL sees raw JSON — semantically fine for an AS2 dereference endpoint. The tombstone table grows unbounded by design (permanence). Neither warrants a change.

Nice attention to the privacy nuance (followers-only ids are never tombstoned so a 410 can't leak that a post existed) and to keeping the served/delivered representations built from one place.

PR #913, opened 2026-07-03
fix(feed): merged activity span everywhere + enrich GET /feed (#881, #891)

fiddur:
Chunk 4 of the #831 feed follow-up queue: **Feed data + merged window** — closes #881 and #891.

Both fixes reduce to one primitive: resolving a feed post's stored anchor `activity_id` to its **merged span** — the same window the `merged:` detail view and share dialog present — via `resolveActivityWindow(..., true)`, at query time (nothing denormalised is persisted, per the repo's reference-by-id principle). Centralised in a new **`services/feed.ts`** (`resolveFeedActivity` / `expandFeedActivityWindow` / `serializeFeedPost`), shared by REST + MCP (parity) and the ActivityPub paths.

## #881 — Shared merged activity reported anchor-only duration/metrics
A merged activity (e.g. a 1h run assembled from two overlapping sub-activities) was delivered/served over the **anchor sub-activity's own window** (40m), under-counting duration and metrics vs. what the user saw and the dialog offered.

Now the delivered / served / outbox-listed `Note` scalars **and** the rendered chart/route images resolve over the **merged span**. Wired through:
- the object + outbox dispatchers (`federation.ts`),
- the create/update delivery hooks (`api.ts`),
- the image router's activity resolver (`rest-routes.ts`).

A plain (non-overlapping) activity passes through with its own window unchanged.

## #891 — `GET /feed` lacked the activity title (web N+1)
The web feed view fetched `GET /activities/:id` **per card** just to show the title and drive the edit dialog's window. `GET /feed` (and share/update, and MCP `list_feed`) now enrich each post with the activity's title/type and merged-span start/end — resolved at query time — and the web card reads those and drops the per-card fetch.

api-spec `FeedPost` gains optional `activity_title`, `activity_type`, `activity_start_time`, `activity_end_time`.

## Tests
- `services/feed.integration.test.ts` (new) — `resolveFeedActivity` expands a merged anchor to the full span / passes a lone activity through / returns null for missing-deleted; `serializeFeedPost` enriches with title + merged window, omits for non-activity / deleted.
- `federation.integration.test.ts` — end-to-end: a shared **merged** activity's served Note reports the merged `Duration 1h`, not the `40m` anchor slice.

## Known follow-up (documented)
The public `GET /public/:username/series` endpoint still authorises only the anchor window for merged shares (`findCoveringSharedSeriesWindow` joins on `activity_id`). This is latent — the delivered Mastodon `Note` carries no series links — and expanding series authorisation across a merge group needs the merge algorithm at query time. Noted in `docs/features/feed.md`.

### DIFF
--- a/apps/backend/src/routes/feed-router.ts
+++ b/apps/backend/src/routes/feed-router.ts
@@ -11,7 +11,6 @@ import type { RequestHandler } from 'express'
  * the public read surface lives in `feed-public-router.ts`.
  */
 import {
-  type FeedPost,
   type FeedPostResponse,
   type FeedPostsResponse,
   type ShareActivityBody,
@@ -30,6 +29,7 @@ import {
   listFeedPosts,
   updateFeedPost,
 } from '../db/index.ts'
+import { serializeFeedPost } from '../services/feed.ts'
 import { type TypedRouter, typedRouter } from '../typed-router.ts'
 import { validateBody } from '../validation.ts'
 
@@ -52,25 +52,14 @@ export interface FeedDeliver {
   deleted: (user: string, post: FeedPostRecord) => void
 }
 
-const serialize = (record: FeedPostRecord): FeedPost => ({
-  activity_id: record.activity_id,
-  created_at: record.created_at.toISOString(),
-  id: record.id,
-  include_chart: record.include_chart,
-  include_map: record.include_map,
-  included_metrics: record.included_metrics,
-  series_metrics: record.series_metrics,
-  updated_at: record.updated_at.toISOString(),
-  visibility: record.visibility,
-})
-
 export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedDeliver): TypedRouter => {
   const router = typedRouter()
 
   router.get<Record<string, never>, FeedPostsResponse>('/', authMiddleware, async (req, res) => {
     const user = req.user!
     const records = await listFeedPosts(user)
-    res.json({ posts: records.map(serialize), success: true })
+    const posts = await Promise.all(records.map((record) => serializeFeedPost(user, record)))
+    res.json({ posts, success: true })
   })
 
   router.post<{ id: string }, FeedPostResponse, ShareActivityBody>(
@@ -93,7 +82,7 @@ export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedD
       })
       // Fan the post out to followers (best-effort; never blocks the response).
       deliver?.created(user, record, activity)
-      res.json({ post: serialize(record), success: true })
+      res.json({ post: await serializeFeedPost(user, record), success: true })
     },
   )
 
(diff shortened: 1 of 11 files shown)


apps/backend/src/routes/feed-router.ts
```diff
@@ -52,25 +52,14 @@ export interface FeedDeliver {
...
-})
-
 export const createFeedRouter = (authMiddleware: RequestHandler, deliver?: FeedDeliver): TypedRouter => {
   const router = typedRouter()
 
   router.get<Record<string, never>, FeedPostsResponse>('/', authMiddleware, async (req, res) => {
     const user = req.user!
     const records = await listFeedPosts(user)
-    res.json({ posts: records.map(serialize), success: true })
+    const posts = await Promise.all(records.map((record) => serializeFeedPost(user, record)))
```
fiddur, line 61: Non-blocking: `serializeFeedPost` runs `getActivityById` + `getOverlappingActivities` (a wide-window query + JS merge) per record, so this list is now ~2 DB queries + a merge per post. Fine for typical feed sizes and a net win on round-trips over the old client-side per-card fetch, but it's server-side N+1 rather than truly eliminated — consider a batched resolve (or capping/paginating `listFeedPosts`) if post counts grow. The same per-item cost applies to the outbox page in `federation.ts`.

fiddur (commented):
✅ Approved

Clean, well-scoped fix. Extracting the feed serialization + activity resolution into a shared `services/feed.ts` and routing every surface (REST list/share/update, MCP tools, ActivityPub delivery/object/outbox, and the rendered chart/route images) through `resolveFeedActivity`/`expandFeedActivityWindow` correctly makes the *merged span* the single source of truth — matching what the detail view and share dialog present (#881). The `ResolvedFeedActivity` shape is structurally compatible with `DeliverableActivity`, and `resolveActivityScalars` re-derives scalars over the passed window (no double-merge), so the delivered/served Note now reports the full-span duration/metrics. Enriching `GET /feed` (#891) with the resolved title/type/window lets the web card drop its per-card `merged:` fetch, and the api-spec `FeedPost` fields are correctly optional (absent for non-activity/deleted). Good integration coverage (merged-span expansion, lone pass-through, deleted → null/omitted) plus the federation duration assertion, and the docs honestly capture the remaining public-series-window limitation. Verified: no dangling `getActivityById`/`serialize` references, `useQuery` still used in Feed/index.tsx, and the merge logic (`resolveActivityWindow` → `getOverlappingActivities`) only needs the fields the callers supply.

One non-blocking note: resolving the merged span now runs `getOverlappingActivities` (a ~36h windowed query + JS merge) plus `getActivityById` per post on every feed serialization — so `GET /feed` and each outbox page do ~2 queries + a merge per post rather than the previous single list query. This is a reasonable correctness/round-trip tradeoff (and bounded), but it relocates the N+1 to the server rather than eliminating it; worth keeping an eye on if a user accumulates many posts or the outbox page size grows.

apps/backend/src/routes/feed-router.ts
fiddur: Deliberate tradeoff. The #891 goal was to eliminate the **client** N+1 (N HTTP round-trips from the web feed view), which is done — one `GET /feed` now. The server-side per-post resolution is the *same* `resolveActivityWindow` work each per-card `merged:` detail fetch already triggered, so it's strictly fewer round-trips for the same work, not new work. It's bounded: feed lists are a single user's shared posts and the outbox is paginated (20/page). Persisting the merged span on the post would avoid the resolution but was rejected to keep the window consistent when the underlying activities are edited (reference-by-id, resolve-at-query-time). Batching `getOverlappingActivities` across a page is a clean future optimization if a user accumulates many posts; noted in the PR body + docs. Resolving as non-blocking.

PR #914, opened 2026-07-03
fix(feed): compact Share button + token-gated followers-only images (#872, #893)

fiddur:
Final chunk of the #831 feed follow-up queue — closes #872 and #893.

## #872 — Share-to-feed button rendered full-width, detached
The `ShareActivityButton` was a bare flex child of the detail-page column, so it stretched to full content width instead of sitting in the action row. Wrapped in `.entity-actions` (matching `ResyncDetailButton`) so it's compact.

## #893 — Followers-only posts now carry images (token-gated)
**Context / decision:** The naive fix (verify the fetching server's HTTP Signature) doesn't work — the fediverse fetches **media without HTTP signatures** (Mastodon's authorized-fetch signs ActivityPub object/actor requests, not media downloads), so a signed-request gate would reject the media fetcher and followers would see a broken image. Per the maintainer's call, this uses the **capability-URL** model already established in this codebase (shared-dashboard slugs, challenge join tokens):

- Each post gets an unguessable `image_token` (`feed_posts.image_token`, `DEFAULT gen_random_uuid()::text`).
- A `followers`-only post's image URLs carry `?token=<image_token>`, embedded **only** in the `Note` delivered to followers.
- The image endpoint serves a `followers`-only image only on a **timing-safe** token match (`crypto.timingSafeEqual`, length-checked); `public`/`unlisted` images stay unauthenticated (no token).
- Tradeoff (documented): a leaked image URL grants access to that one rendered image — a chart or a bare route shape, **not** the underlying high-resolution series, which stays `followers`-excluded entirely.

`image_token` is **never** exposed in any owner-facing/public response (not in `FeedPost`, `serializeFeedPost`, the object/outbox) — only embedded in delivered followers-only image URLs and checked on inbound image requests. Existing DBs get the column via `ADD COLUMN IF NOT EXISTS` + the on-schema-error migration (`42703` → migrate → retry).

## Tests
- `deliver.test.ts` — `imageAttachments`: public/unlisted → no token; **followers-only → chart/route URLs carry `?token=…`**.
- `feed-image-router.test.ts` — `resolveImageWindow`: followers-only rejected without a token / with a wrong token / with a token **prefix** (length-checked); **served on exact match**; token ignored for public.
- `feed.integration.test.ts` — `createFeedPost` assigns a persistent, per-post unguessable `image_token`.

docs/features/feed.md updated (Images section, API table, storage).

### DIFF
--- a/apps/backend/src/db/connection.ts
+++ b/apps/backend/src/db/connection.ts
@@ -666,6 +666,14 @@ export const migrateSchema = async (user: string) => {
     await query(db, `ALTER TABLE tags ADD COLUMN IF NOT EXISTS deleted_at TIMESTAMPTZ`)
     await query(db, `ALTER TABLE tags ADD COLUMN IF NOT EXISTS tag_definition_id UUID`)
   }
+  if (existingTableNames.has('feed_posts')) {
+    // Capability token for followers-only image URLs (#893). Existing rows get a
+    // token via the DEFAULT applied on ADD COLUMN.
+    await query(
+      db,
+      `ALTER TABLE feed_posts ADD COLUMN IF NOT EXISTS image_token TEXT NOT NULL DEFAULT gen_random_uuid()::text`,
+    )
+  }
   if (existingTableNames.has('activities')) {
     await query(db, `ALTER TABLE activities ADD COLUMN IF NOT EXISTS deleted_at TIMESTAMPTZ`)
     await query(db, `ALTER TABLE activities ADD COLUMN IF NOT EXISTS external_id VARCHAR(255)`)
--- a/apps/backend/src/db/feed.integration.test.ts
+++ b/apps/backend/src/db/feed.integration.test.ts
@@ -76,6 +76,17 @@ describe('Feed posts integration', () => {
     expect(await getFeedPostById(user, '00000000-0000-0000-0000-000000000000')).toBeNull()
   })
 
+  test('assigns an unguessable image_token that round-trips (#893)', async () => {
+    const user = getTestUser()
+    const created = await createFeedPost(user, postInput())
+    // Non-empty and unguessable (a UUID by default).
+    expect(created.image_token).toMatch(/^[0-9a-f-]{36}$/)
+    // Distinct per post, and stable on re-read.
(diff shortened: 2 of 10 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped PR. Two independent changes:

**Token-gated followers-only images (#893)** — a per-post `image_token` (defaulted to `gen_random_uuid()::text`, ~122 bits of CSPRNG entropy) is appended as `?token=` to a `followers`-only post's rendered chart/route URLs so a follower's server can fetch unsigned media, while public/unlisted stay tokenless. The security model holds up under scrutiny:

- The token never leaks on a public surface: `serializeFeedPost` maps fields explicitly (no `image_token`), the outbox and object dispatcher only serve `public`/`unlisted` posts (`federation.ts:165` returns `null` for followers-only), and `imageAttachments` only emits `?token=` for non-public visibility.
- `tokenMatches` correctly guards `a.length === b.length` before `timingSafeEqual` (which throws on unequal lengths), so no crash and constant-time on equal-length inputs — the added "prefix must not pass" test confirms this.
- Authorization (`isImageAuthorized`) runs *before* the render cache is consulted, so a public→followers flip (or unshare / cleared flag) 404s the untoken'd URL immediately even though the rendered PNG is still cached — the `no-store` revocation claim in the docstring is accurate.
- The migration adds `image_token` with a *volatile* default, so PostgreSQL evaluates it per existing row and each backfilled post gets a distinct token (verified by the integration test).

The documented capability-URL tradeoff (a leaked URL grants one rendered image, not the underlying series) is a reasonable, clearly-explained choice.

**Compact Share button (#872)** — wrapping the button in `.entity-actions` mirrors the sibling `ResyncDetailButton` exactly; consistent and correct.

Test coverage is thorough (no-token, wrong-token, prefix, matching-token, token-ignored-for-public, unlisted-no-token, token round-trip) and docs are updated. No blocking issues found. I couldn't execute the suite here (deps not installed in this worktree, network-restricted), but the affected unit tests are pure and type-consistent by inspection; CI runs the full suite.

Issue #884, opened 2026-07-03
Feature: Feed improvements — native post rendering, following, home timeline & live updates


fiddur:
## Summary

Grow the current Feed page from a **management list of my own shares** into an actual
social reading surface:

1. **Render posts as they really federate** (Mastodon-style card), not as a chip list.
2. **Consolidate publishing controls onto the Share page** (dashboards + activity outbox),
   leaving the Feed as a reading surface; give posts a proper visibility model.
3. **Follow other ActivityPub users**, with a quick-add for local users.
4. A **home timeline** of activity from people I follow.
5. **Live updates** using the Mastodon "N new posts" pattern (new items surface at the top,
   reading position preserved), pushed via Postgres `LISTEN/NOTIFY` → SSE with a polling
   fallback.
6. (Optional) a **local timeline** of public posts on this instance.

Richer attachments (chart/route images) are handled in the separate rich-sharing issue;
this issue renders them once available but doesn't build them.

## Current state

The Feed page today lists the user's own shared activities as management cards (title,
shared date, metric chips, series chips, `Public` badge, Edit / Unshare). It does **not**
show how a post actually looks once federated — e.g. in Mastodon it renders as:

> **fiddur** · @fiddur@aurboda.net · 59m
> Mark Walking · Duration 1006 seconds · Avg HR 94 bpm · Max HR 109 bpm · Calories 37 kcal

So the page is a control panel, not a feed. There is also no way to follow anyone or read a
timeline of others' activity.

## Goals

- The feed shows posts the way they actually appear when federated (WYSIWYG).
- Users can follow other actors (remote by `@user@instance`, and local via quick-add).
- A chronological home timeline aggregates followed actors' posts.
- New posts appear live without disrupting the current reading position.

## 1. Native post rendering

Replace the chip list with a post card that mirrors the federated object:

- Author avatar, display name, `@handle@instance`, relative timestamp, visibility icon.
- The AS2 `content` rendered as **sanitized HTML** (allowlist: `p`, `br`, `a`, `span`,
  mention/hashtag markup) so it matches what remote servers show.
- Attachments (chart/route images) when present (from the rich-sharing issue).
- Interaction affordances (reply / boost / favourite / bookmark). Counts can be
  **read-only** in v1 (display inbound counts; full interaction is a follow-up).
- Permalink to the canonical `/u/:user/activities/:id`.
- **For the viewer's own posts:** show the post exactly as it federates, with a management
[...]

PR #915, opened 2026-07-03
feat(feed): render own feed posts as native cards (#884 slice 1)

fiddur:
## Summary

**Slice 1 of #884** (in-app federated feed). Renders the activities a user has published to their feed as **native Mastodon-style post cards** that mirror how the post actually federates — replacing the old management-chip list.

This is the first of several slices toward a full in-app feed. It deliberately touches only the *own-posts* view; following, home timeline, and live updates come in later slices.

## What changed

- **New `FeedPostCard`** (`apps/web/src/pages/Feed/FeedPostCard.tsx`) — a reusable card showing author, `@user@host` handle, timestamp, visibility, the AS2 `content` HTML, and chart/route attachments. The author identity is **injected**, so the same component will serve remote actors' posts in the home-timeline slice without change.
- **`serializeFeedPost` now returns `content`** — the *exact* server-built HTML the delivered Note carries (via `feedPostContent`), resolved at query time. The web renders it directly (WYSIWYG, matching Mastodon) instead of reconstructing a summary from metric keys. Added an optional `content` field to the api-spec `FeedPost` schema.
- **Own-post controls preserved** — `index.tsx`'s `OwnPostCard` wraps `FeedPostCard`, passing Edit / Unshare into the card's `footer` slot; editing still reuses the share dialog.
- **Followers-only images** are token-gated (#893) and the browser has no token, so they're omitted from the card; public/unlisted chart/route images render.

## Notes for review

- The owner-facing `content` is server-generated and HTML-escaped (same string that federates), so it renders via `dangerouslySetInnerHTML` without extra sanitising. A code comment flags that *remote* timeline content (a later slice) will need sanitising.
- The feed schema is not wired into OpenAPI path generation (pre-existing), so `pnpm generate` produces no diff for this change.

## Testing

- Backend `services/feed.integration.test.ts` extended to assert the rendered `content` HTML carries the activity headline; passes against a real DB.

### DIFF
--- a/apps/backend/src/services/feed.ts
+++ b/apps/backend/src/services/feed.ts
@@ -22,6 +22,8 @@ import type { FeedPost } from '@aurboda/api-spec'
 import type { Activity, FeedPostRecord } from '../db/index.ts'
 
 import { getActivityById } from '../db/index.ts'
+import { resolveActivityScalars } from './activitypub/feed-activity.ts'
+import { feedPostContent } from './activitypub/object.ts'
 import { resolveActivityWindow } from './queries/index.ts'
 
 /**
@@ -70,18 +72,34 @@ export const resolveFeedActivity = async (
 
 /**
  * Serialise a stored feed post for the owner-facing REST/MCP surface, enriching
- * it with the shared activity's title/type and **merged-span** window resolved
- * at query time. A client can therefore render the post and re-open the share
- * dialog without a second per-post activity fetch (avoids the web feed's N+1).
+ * it with the shared activity's title/type, the **merged-span** window, and the
+ * exact `content` HTML the post federates with — all resolved at query time. A
+ * client can therefore render the post WYSIWYG (as it appears on Mastodon) and
+ * re-open the share dialog without a second per-post activity fetch.
+ *
+ * The `content` is the same server-built, HTML-escaped string the delivered Note
+ * carries (via `feedPostContent`), so the web renders it directly. (When the feed
+ * later renders *remote* actors' posts, that untrusted content will need
+ * sanitising — this owner-facing content does not.)
  */
 export const serializeFeedPost = async (user: string, record: FeedPostRecord): Promise<FeedPost> => {
   const activity = record.activity_id ? await resolveFeedActivity(user, record.activity_id) : null
+  let content: string | undefined
+  if (activity) {
+    const scalars = await resolveActivityScalars(
+      user,
+      { end_time: activity.end_time, start_time: activity.start_time },
+      record.included_metrics,
+    )
+    content = feedPostContent(activity.title, activity.activity_type, scalars).content
+  }
   return {
     activity_end_time: activity?.end_time?.toISOString(),
     activity_id: record.activity_id,
     activity_start_time: activity?.start_time.toISOString(),
     activity_title: activity?.title,
     activity_type: activity?.activity_type,
+    content,
     created_at: record.created_at.toISOString(),
     id: record.id,
     include_chart: record.include_chart,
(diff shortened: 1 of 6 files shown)


apps/backend/src/services/feed.ts
```diff
@@ -70,18 +72,34 @@ export const resolveFeedActivity = async (
...
+ * The `content` is the same server-built, HTML-escaped string the delivered Note
+ * carries (via `feedPostContent`), so the web renders it directly. (When the feed
+ * later renders *remote* actors' posts, that untrusted content will need
+ * sanitising — this owner-facing content does not.)
  */
 export const serializeFeedPost = async (user: string, record: FeedPostRecord): Promise<FeedPost> => {
   const activity = record.activity_id ? await resolveFeedActivity(user, record.activity_id) : null
+  let content: string | undefined
+  if (activity) {
+    const scalars = await resolveActivityScalars(
```
fiddur, line 89: This adds a per-post `resolveActivityScalars` (one `queryMetricsBucketed`, plus `getEffectiveHrZones` + `getTimeSeries` when `hr_zone_minutes` is included) on top of the existing per-post `resolveFeedActivity`. The list serializer fans out over every post via `Promise.all`, so a large feed now issues several DB queries per post concurrently. It's consistent with the federation path and parallelised, so not blocking — just flagging that content resolution is no longer free for the list endpoint, and this may want batching down the line.

fiddur (commented):
✅ Approved

Clean, well-scoped slice. The native card renders the server-built `content`, and the key correctness/security properties hold up under inspection:

- **XSS-safe.** `FeedPostCard` injects `post.content` via `dangerouslySetInnerHTML`, but that string is produced by `feedPostContent`, which HTML-escapes the heading and stats (`escapeHtml` handles `& < > "`) and only ever places them in element text position (`<p><strong>…</strong></p>`). It's owner-facing content, and the code correctly documents that future *remote* content will need sanitising.
- **WYSIWYG claim holds.** `serializeFeedPost` reuses the exact same `resolveActivityScalars` + `feedPostContent` over the merged-span activity that the federation delivery path (`buildFeedNote`/`buildFeedCreate`) uses, so the card matches the delivered Note rather than reconstructing it. Nice reuse.
- **Followers-only images gated.** `imageUrl` returns `null` for `followers` posts (no browser token), matching #893.

Non-blocking notes:
- The list endpoint (`feed-router.ts` `Promise.all(records.map(serializeFeedPost))`) now does an extra `resolveActivityScalars` per post — one `queryMetricsBucketed`, plus `getEffectiveHrZones` + `getTimeSeries` for any `hr_zone_minutes` post — on top of the existing per-post `resolveFeedActivity`. It's parallelised and matches the federation path, so acceptable, but worth keeping an eye on as feeds grow (see inline).
- `PostAuthor.profileUrl` is set in `index.tsx` but never consumed by `FeedPostCard` (avatar/name aren't linked). Fine if it's reserved for the upcoming timeline reuse, but currently dead.
- No frontend unit test for the `imageUrl` followers-gating logic; low priority given the project's backend-test emphasis, but that's a security-relevant branch worth pinning down.

Overall LGTM.

apps/backend/src/services/feed.ts
fiddur: Agreed and acknowledged. Keeping it as-is for slice 1: content resolution is parallelised and mirrors the federation delivery path exactly (single source of truth for the rendered Note). The natural place to batch is slice 3 (home timeline), which introduces a cursor-paginated timeline read where a single batched scalar-resolution pass replaces the per-post fan-out. Tracking it there rather than pre-optimising the owner-only list here.

PR #917, opened 2026-07-03
feat(feed): follow other fediverse actors (#884 slice 2)

fiddur:
## Summary

**Slice 2 of #884** (in-app federated feed) — the **inbound** direction. A user can now **follow other fediverse actors** (remote *or* another local Aurboda user) and see who they follow. Until now the feed was outbound-only (remote users following the local actor); this adds the local actor following others.

> **Scope note (reordering vs. the original plan):** the original slice-2 sketch bundled inbound `Create`/`Announce` handling. But there's nowhere to *store* received posts yet — that's the `timeline_entry` model in slice 3. So this slice is the **follow relationship control plane** only; receiving followees' posts into a **home timeline** (inbound `Create`/`Announce` + storage + UI) is slice 3, where the storage model lives. Cleaner split: "who you follow" here, "what shows up" there.

### Backend
- **`feed_following` table** + `db/feed-following.ts` — keyed by a local `id` (used to build the `Follow` activity id + unfollow), UNIQUE `actor_uri`, caching the followee's inbox / handle / display name / avatar and an `accepted` flag. Full integration-test coverage.
- **`services/following.ts`** — resolve an actor (WebFinger + fetch via Fedify), persist a *pending* follow, send a signed `Follow`; unfollow sends `Undo{Follow}` to the cached inbox and drops the row. The pure `actorToFollowingInput` / `serializeFollowing` mappers are unit-tested; the network orchestration is thin and injected (`FollowActions`, mirroring `FeedDeliver`).
- **federation.ts** — the actor advertises a `following` collection; it's served (accepted follows only); inbound `Accept` (mark accepted) / `Reject` (drop) of Follows *we* sent are handled, matched via `ctx.recipient` (no need to fetch the inner Follow).
- **REST** `/feed/following` (`GET` list · `POST` follow-by-handle · `DELETE` unfollow), mounted before `/feed` so its two-segment paths never hit the feed router's `/:postId`. **MCP** `list_following` / `follow_actor` / `unfollow_actor`. Followee inbox URIs are **never** exposed on the owner-facing surface.
- **api-spec**: `FollowActorBody`, `FollowingActor`, `FollowActorResponse`, `FollowingResponse`.

### Web
- A **Following** panel on the Feed page: follow by `@user@host`, list who you follow with a **Pending** badge (until the remote accepts) and an **Unfollow** button.

### Docs
- `feed.md` gains a "Following other actors (inbound)" section + API / storage / web-app updates.

## Design notes for review
- **Local follows** use the exact same federation path (delivered to the local inbox over loopback), so there's no special-casing.
- **Best-effort delivery**, consistent with the rest of the feed (no durable queue): a failed `Follow`/`Undo` POST is logged, not fatal — the pending row lets the user retry. Unfollow removes locally even if the `Undo` POST fails.
- **Handle** is derived from `preferredUsername` + the actor id host (deterministic/offline) rather than `getActorHandle` (which may WebFinger and hang) — good enough for a display handle.
- **No casts**: uses the `isActor` type guard; `ctx.recipient` for inbox owner identity.
- `db-test-helper`'s truncation list gained `feed_following` (caught cross-test leakage).

## Testing
- New: `db/feed-following.integration.test.ts` (7), `services/following.test.ts` (4, incl. asserting inbox URIs never serialize), `federation.integration.test.ts` extended (actor advertises `following`; collection lists only accepted follows). All pass against a real DB.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -1,4 +1,4 @@
-import { createFederation, type Federation, MemoryKvStore } from '@fedify/fedify'
+import { createFederation, type Federation, type InboxContext, MemoryKvStore } from '@fedify/fedify'
 import { Accept, type Create, Follow, Image, Note, Person, Reject, Undo } from '@fedify/fedify/vocab'
 
 /**
@@ -49,6 +49,28 @@ const OUTBOX_PAGE_SIZE = 20
 /** RFC 4122 canonical form — guards `getFeedPostById` from a non-UUID `postId`
  * (Postgres would otherwise raise `invalid input syntax for type uuid`). */
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+
+/**
+ * The local follower (inbox owner) that an `Accept`/`Reject` response to a Follow
+ * WE sent belongs to. Derived from the inner Follow's `actor` (which is our
+ * actor), so it works whether the response lands on the personal OR a shared
+ * inbox; falls back to `ctx.recipient` (null on the shared inbox) if the remote
+ * didn't embed the Follow. `suppressError` so an unresolvable inner object yields
+ * null rather than throwing a 500 that invites retries. Returns null if no valid
+ * local user can be determined.
+ */
+const localFollowerIdentifier = async (
+  ctx: InboxContext<void>,
+  response: Accept | Reject,
+): Promise<string | null> => {
+  const object = await response.getObject({ suppressError: true })
+  if (object instanceof Follow && object.actorId != null) {
+    const parsed = ctx.parseUri(object.actorId)
+    if (parsed?.type === 'actor' && isValidUsername(parsed.identifier)) return parsed.identifier
+  }
+  const recipient = ctx.recipient
+  return recipient != null && isValidUsername(recipient) ? recipient : null
+}
 
 export const createFeedFederation = (origin: string, apiBaseUrl: string): Federation<void> => {
   const federation = createFederation<void>({
@@ -155,11 +177,10 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
     })
     .on(Accept, async (ctx, accept) => {
       // Accept of a Follow WE sent — the followee's server confirms the follow.
-      // `ctx.recipient` is the identifier of the inbox that received it (us, the
-      // follower); `accept.actorId` is the followee we followed, matching the
-      // `actor_uri` of our pending row. No need to fetch the inner Follow object.
-      const me = ctx.recipient
-      if (me == null || !isValidUsername(me) || accept.actorId == null) return
+      // `accept.actorId` is the followee we followed, matching the `actor_uri` of
+      // our pending row.
+      const me = await localFollowerIdentifier(ctx, accept)
+      if (me == null || accept.actorId == null) return
       try {
         await markFeedFollowingAccepted(me, accept.actorId.href)
       } catch (error) {
@@ -169,8 +190,8 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
     })
     .on(Reject, async (ctx, reject) => {
       // Reject of a Follow we sent — the followee declined; drop our pending row.
-      const me = ctx.recipient
-      if (me == null || !isValidUsername(me) || reject.actorId == null) return
+      const me = await localFollowerIdentifier(ctx, reject)
+      if (me == null || reject.actorId == null) return
       try {
         await removeFeedFollowingByActor(me, reject.actorId.href)
       } catch (error) {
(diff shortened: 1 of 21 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -144,6 +153,31 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
         throw error
       }
     })
+    .on(Accept, async (ctx, accept) => {
+      // Accept of a Follow WE sent — the followee's server confirms the follow.
+      // `ctx.recipient` is the identifier of the inbox that received it (us, the
+      // follower); `accept.actorId` is the followee we followed, matching the
+      // `actor_uri` of our pending row. No need to fetch the inner Follow object.
+      const me = ctx.recipient
```
fiddur: Minor robustness note (non-blocking): unlike the sibling `Follow`/`Undo` handlers, which derive the local user from the activity's inner object and therefore work on both the personal and shared inbox, `Accept`/`Reject` rely on `ctx.recipient`, which Fedify sets to `null` for the shared (`/inbox`) listener. This is safe today because the actor document doesn't advertise `endpoints.sharedInbox`, so remotes deliver these to the personal inbox — but if a later slice adds a `sharedInbox` endpoint (a common delivery optimization), Accepts routed there would be silently dropped and the follow would stay stuck `pending`. Deriving `me` from the inner Follow's `actorId` (as `Undo` does) would harden it against that.

fiddur (commented):
✅ Approved

Solid, well-scoped slice: it adds the outbound follow direction (resolve → persist pending → signed `Follow`, `Undo{Follow}` on unfollow), inbound `Accept`/`Reject` handling, a public `following` collection, the `feed_following` store, REST `/feed/following` + MCP `list_following`/`follow_actor`/`unfollow_actor` parity, a frontend panel, and docs.

Strengths:
- Clean separation: pure mappers (`actorToFollowingInput`, `serializeFollowing`) are unit-tested; the network orchestration is thin and injected as `FollowActions`, mirroring `FeedDeliver`, so the router/MCP stay decoupled from the ActivityPub context.
- Good test coverage: DB integration tests exercise upsert/accept/remove/list semantics (including that re-follow preserves `accepted` and pending rows stay out of the public collection), plus a federation integration test proving the `following` collection publishes only accepted follows.
- Follows conventions: snake_case throughout, `typedRouter()` with fully-typed generics, api-spec as the single source of truth, inbox URIs deliberately kept off the owner-facing DTO, and the `/feed/following` mount-order comment is accurate (it resolves before `/feed`'s `/:postId`).
- Verified against the installed Fedify 2.3.1 types that `ctx.getFollowingUri`, `setFollowingDispatcher` (accepts `Actor | URL` items), and `ctx.recipient` all exist.

No blocking issues. Two things worth a glance, both non-blocking:
- One minor robustness note inline on the `Accept` handler.
- The `following` collection is world-readable (accepted follows only). This matches fediverse norms and is documented as intentional, so just flagging it as a deliberate default.

apps/backend/src/services/activitypub/federation.ts
fiddur: Good catch — applied. `Accept`/`Reject` now derive the local follower from the inner Follow's `actor` (our actor), like the `Undo` handler, so they work on both the personal and shared inbox; `ctx.recipient` is only a fallback when the Follow isn't embedded. Pushed in fddeb8c4.

fiddur (commented):
✅ Approved

Solid slice-2 that adds the inbound follow relationship (follow/unfollow other fediverse actors, plus the ActivityPub `Accept`/`Reject` handling and the public `following` collection). The design is clean and consistent with the existing feed code: shared service logic (`following.ts`) injected into both the REST router and the MCP tools for parity, per-user DB store with an upsert-on-`actor_uri` that deliberately preserves `accepted`, best-effort synchronous delivery matching the rest of the feed, and DTOs that never leak the cached inbox URIs. Test coverage is good — every new DB function has an integration test, the pure mappers are unit-tested, and the `following` collection has a federation integration test asserting only accepted follows are published. Docs (`docs/features/feed.md`) and the API/MCP tables are kept in sync. Nothing blocking.

A few non-blocking observations for later:

- **Tenant-safety of the inbound `Accept`/`Reject` looks correct.** `localFollowerIdentifier` derives the local user from the inner Follow's `actor` (guarded by `parseUri` + `isValidUsername`) and the write keys on the HTTP-signature-authenticated `accept.actorId`/`reject.actorId`, so a remote can only flip/remove a row that references *itself* — no cross-tenant writes. The shared-inbox fallback to `ctx.recipient` (null on the shared inbox) means a non-Mastodon server that sends an `Accept` carrying only a Follow *URI* (not embedded) to the shared inbox can't be resolved and the follow stays pending; acceptable for best-effort, and Mastodon embeds it.

- **`actorToFollowingInput` calls `actor.getIcon()`.** The comment notes Mastodon inlines the icon so no fetch happens, but for servers that don't inline, `getIcon()` may dereference the URL during a synchronous follow request. It's wrapped in try/catch (so it degrades to a null avatar) but has no explicit timeout.

- **`self-follow` isn't prevented** — a user can follow their own actor (resolves, sends a Follow over loopback, auto-accepts). Harmless but slightly odd; Mastodon blocks it.

- **The `following` collection dispatcher returns all accepted follows unpaginated** (no `setFirstCursor`), matching the existing followers dispatcher — fine at current scale, worth pagination if follow counts grow.

- **Frontend UX nit:** `FollowingPanel` shows a generic "Couldn't follow that handle" and discards the specific server error (e.g. "Could not resolve an actor"), which is more actionable.

PR #918, opened 2026-07-03
feat(feed): OpenStreetMap basemap for GPS route maps

fiddur:
## Summary

The shared GPS **route image** was a bare purple polyline on a dark background. It now renders over an **OpenStreetMap street basemap**, so the route reads in context.

Requested by @fiddur ("The shared GPS route image should have the openstreetmap background if allowed"). Confirmed decisions: **OSM public tiles, on by default**, applied to **all route images** (public + followers).

| Before | After |
| --- | --- |
| Bare polyline on dark navy | Track over OSM street tiles, haloed, with attribution |

## What changed

- **New `osm-tiles.ts`** — Web Mercator (EPSG:3857) projection helpers (`lonToWorldX`, `latToWorldY`, `chooseZoom`) + a best-effort tile fetcher for `tile.openstreetmap.org`.
- **`renderRoutePng` rewrite** — reprojects the track to Web Mercator (matching the tiles), picks the zoom that fits the whole route, fetches the covering tiles, composites them behind a **white-haloed** track + start/end markers, rounds the corners, and bakes in the required **"© OpenStreetMap contributors"** attribution.
- **Injected tile fetching** (`opts.fetchTile`) so it's opt-in and testable; the production wiring passes the real OSM fetcher. If tiles can't be fetched (offline, timeout, all-fail) it **falls back to the original bare shape**.

## OSM tile-policy compliance

- Descriptive `User-Agent` (`Aurboda/1.0 (+https://aurboda.net; feed route maps)`).
- **Low volume** — route renders are cached per post (#892), so a route's tiles are fetched at most once until cache eviction.
- Required **attribution baked into the image**.
- 5 s per-tile timeout + out-of-range guard; a defensive `MAX_TILES` cap.

## ⚠️ Privacy note

A real basemap makes a route's **precise location obvious** (including start/end = likely home/work), and per your choice this applies to **public/unlisted** images fetched by the whole fediverse — a genuine increase in location exposure vs. the old bare shape. The docs caveat is sharpened accordingly; start-point / area masking remains a planned follow-up. Flagging so it's a conscious call.

## Testing

- New `osm-tiles.test.ts` (projection math, zoom-fit selection, out-of-range tiles return null without network) + extended `feed-images.test.ts` (basemap compositing — asserts a tile pixel shows through; bare-shape fallback when tiles fail; empty-route guard). Tests use injected fake tiles — **no network in CI**.

> Branch is off the latest develop (includes #884 slice 2).

### DIFF
--- a/apps/backend/src/services/activitypub/feed-images.ts
+++ b/apps/backend/src/services/activitypub/feed-images.ts
@@ -195,7 +195,16 @@ const renderRouteWithBasemap = async (
     ty,
   }))
   const loaded = tiles.filter((t): t is { tx: number; ty: number; buf: Buffer } => t.buf != null)
-  if (loaded.length === 0) return null
+  if (loaded.length === 0) {
+    // Every covering tile failed to fetch — an OSM outage or block (e.g. 429), not
+    // an expected case like the MAX_TILES guard above. `fetchTile` swallows each
+    // failure to null, so this is the only place a real outage surfaces: log it so
+    // the silent degradation to a bare shape is observable rather than invisible.
+    console.warn(
+      `⚠️ OSM basemap: all ${coords.length} tiles failed to fetch (zoom ${zoom}); using bare route`,
+    )
+    return null
+  }
 
   // Composite tiles onto a grid canvas (all non-negative offsets), then crop the
   // 700×700 viewport out of it — avoids negative composite offsets entirely.
--- a/apps/backend/src/services/activitypub/feed-images.ts
+++ b/apps/backend/src/services/activitypub/feed-images.ts
@@ -2,20 +2,29 @@
  * Render feed-post attachment images (a metric line chart and a GPS route map)
  * as PNGs, from raw series / coordinate data.
  *
- * Both are pure functions of their inputs: they build a self-contained SVG and
- * rasterize it with `sharp` (no fonts, no external tiles). The route is drawn as
- * a bare polyline shape — a street basemap would need an external tile provider
- * and is a later enhancement. No privacy trimming is applied (area masking is a
- * planned follow-up), so callers must only render for public/unlisted posts that
- * opted in.
+ * The chart is a pure, self-contained SVG rasterized with `sharp`. The route map
+ * is drawn over an **OpenStreetMap street basemap**: the track is projected into
+ * Web Mercator (the tiles' projection), the covering tiles are fetched and
+ * composited behind it, and the required "© OpenStreetMap contributors"
+ * attribution is baked in (see `osm-tiles.ts`). Tile fetching is injected
+ * (`opts.fetchTile`) so it's opt-in and testable; when it's absent, or every tile
+ * fails (e.g. offline), the route falls back to a bare polyline on a dark
+ * background. No privacy trimming is applied (area masking is a planned
+ * follow-up), so callers must only render for posts that opted in.
  */
 import sharp from 'sharp'
+
+import { chooseZoom, latToWorldY, lonToWorldX, TILE_SIZE, type TileFetcher } from './osm-tiles.ts'
 
 const CHART_W = 1000
 const CHART_H = 420
 const ROUTE_W = 700
 const ROUTE_H = 700
 const PAD = 40
+/** Dark background behind the route (and behind any missing basemap tiles). */
+const ROUTE_BG = '#0b0f19'
+/** Guard against a pathological tile grid (chooseZoom keeps the route ≤ ~3 tiles wide). */
+const MAX_TILES = 25
 
 const svgToPng = (svg: string): Promise<Buffer> => sharp(Buffer.from(svg)).png().toBuffer()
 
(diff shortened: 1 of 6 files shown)


apps/backend/src/services/activitypub/feed-images.ts
```diff
@@ -59,42 +68,195 @@ export const renderChartPng = async (
...
+  const rows = tileMaxY - tileMinY + 1
+  if (cols * rows > MAX_TILES) return null
+
+  const jobs: Promise<{ tx: number; ty: number; buf: Buffer | null }>[] = []
+  for (let tx = tileMinX; tx <= tileMaxX; tx++) {
+    for (let ty = tileMinY; ty <= tileMaxY; ty++) {
+      jobs.push(fetchTile(zoom, tx, ty).then((buf) => ({ buf, tx, ty })))
+    }
+  }
+  const tiles = await Promise.all(jobs)
```
fiddur: Minor / non-blocking: this fires all covering tiles (up to ~16, `MAX_TILES` 25) as one `Promise.all` burst. OSM's tile usage policy discourages more than ~2 concurrent download threads, and the module docstring claims the policy is "respected". The per-post render cache keeps total volume very low, so this is defensible, but if you want to be strictly conformant you could cap concurrency (e.g. a small pool of 2) — the tiles are already fetched once per post, so the extra latency is negligible and only paid on a cache miss.

fiddur (commented):
✅ Approved

Solid, well-tested addition of an OpenStreetMap basemap to feed route maps. The Web Mercator math (`osm-tiles.ts`), `chooseZoom`, and the tile-grid compositing are correct, and the design is careful in the places that matter:

- **Bounds are safe.** Tiles are composited onto a `cols×rows` grid at non-negative offsets and the 700×700 viewport is cropped out with a clamped `extract`. I checked that `gridW ≥ ROUTE_W` and `gridH ≥ ROUTE_H` always hold (the tile range provably covers `[vpLeft, vpLeft+700]` / `[vpTop, vpTop+700]`), so the `extract` can't go out of bounds.
- **Robust fallbacks.** Empty/all-non-finite routes render a plain background; `MAX_TILES` guards the pathological low-zoom (world-spanning) grid before any fetch; every-tile-failed and any compositing/fetch exception both fall back to the bare polyline. `fetchOsmTile` never throws and validates tile indices.
- **No injection/SSRF surface.** The tile URL is built purely from integer z/x/y, and the route SVG contains no user-derived text.
- **Policy/caching claim holds.** `createRenderCache` memoizes the rendered PNG per post, so tiles are fetched at most once per route until eviction — backing the docstring's low-volume claim. A descriptive `User-Agent` and short timeout are set.
- Signature change to `renderRoutePng` is backward-compatible (optional `opts`), and the one production caller is updated.
- Tests cover the basemap composite, the all-tiles-fail fallback, and the empty route; docs and — importantly — the privacy caveat (now "precise area … likely home/work") are updated honestly.

Note: I could not run `pnpm check` / the vitest suite here — this worktree has no `node_modules` and the npm cache is read-only — so verification is by inspection; CI should exercise the tests.

One non-blocking point is left inline (tile-fetch concurrency vs OSM's connection guideline). Nothing blocking.

apps/backend/src/services/activitypub/feed-images.ts
fiddur: Applied — tiles now fetch through a fixed pool of 2 (a small mapWithConcurrency), matching OSM's ~2-connection guideline, so the docstring claim is strictly accurate. Pushed in 4201448f.

apps/backend/src/services/activitypub/feed-images.ts
```diff
@@ -59,42 +91,200 @@ export const renderChartPng = async (
...
+      `<svg xmlns="http://www.w3.org/2000/svg" width="${ROUTE_W}" height="${ROUTE_H}"><rect width="${ROUTE_W}" height="${ROUTE_H}" fill="${ROUTE_BG}" rx="16"/></svg>`,
+    )
+  }
+  const bbox = boundingBox(pts)
+
+  if (opts.fetchTile) {
+    try {
+      const map = await renderRouteWithBasemap(pts, bbox, opts.fetchTile)
+      if (map) return map
+    } catch {
```
fiddur: Non-blocking: this `catch {}` swallows every error from `renderRouteWithBasemap` silently. The safe bare-shape fallback is the right behavior, but a bare polyline is also the *offline* fallback, so a genuine regression (e.g. a bug that makes the basemap never composite) would degrade all route maps with no signal. Consider a `console.warn`/debug log here so such failures are observable rather than invisible.

fiddur (commented):
✅ Approved

Solid, well-tested feature. The Web Mercator projection and tile-grid math check out: `chooseZoom` fits the bbox within the padded frame, the covering tiles are composited onto a grid whose width/height provably exceed the 700×700 viewport (`cols`/`rows` ≥ 3), so the `extract()` crop is always in-bounds and the `clamp` guards never distort alignment — the route overlay shares the same `vpLeft/vpTop` origin as the crop, so basemap and track line up. OSM's tile policy is handled thoughtfully: descriptive `User-Agent`, ≤2 concurrent fetches via `mapWithConcurrency`, out-of-range/`timeout` guards, baked-in "© OpenStreetMap contributors" attribution, and per-post render caching (`createRenderCache`) so tiles aren't re-fetched on every fediverse media fetch. Code style adheres to the standards (functional, injected/testable tile fetcher, type-guard filter instead of a cast, docs updated).

Non-blocking notes:
- **Privacy:** the basemap upgrades a route map from an "approximate" shape to a legible, precise location including start/end (likely home/work). This is explicitly documented as an intentional known limitation in `docs/features/feed.md`, so just flagging for awareness — start-point/area masking remains a planned follow-up.
- **OSM policy at scale:** the render cache is in-process and bounded (200 entries, evicted on restart), so an instance serving many distinct routes could re-fetch tiles after eviction. Fine for personal-scale usage and within policy given the mitigations, but worth keeping in mind if traffic grows (a proper tile provider or persistent cache would be the next step).

apps/backend/src/services/activitypub/feed-images.ts
fiddur: Good call — added a console.warn in the fallback catch so a real basemap regression is visible rather than silently degrading to the bare (offline) shape. Pushed in 47783b20.

apps/backend/src/services/activitypub/feed-images.ts
```diff
@@ -59,42 +91,203 @@ export const renderChartPng = async (
...
+  // Fetch through a small pool (not one big burst): OSM's tile policy asks for at
+  // most ~2 concurrent download connections. Renders are cached per post, so the
+  // extra latency is paid at most once per route.
+  const tiles = await mapWithConcurrency(coords, TILE_FETCH_CONCURRENCY, async ({ tx, ty }) => ({
+    buf: await fetchTile(zoom, tx, ty),
+    tx,
+    ty,
+  }))
+  const loaded = tiles.filter((t): t is { tx: number; ty: number; buf: Buffer } => t.buf != null)
+  if (loaded.length === 0) return null
```
fiddur: This is the primary regression path and it returns `null` **silently** — no log is emitted. `fetchOsmTile` catches all failures (non-200, 429/blocked, timeout, network) and returns `null` rather than throwing, so when OSM stops serving tiles every tile ends up here, `renderRouteWithBasemap` returns `null`, and `renderRoutePng` falls through to `renderBareRoute` with no signal. The `console.warn` in the caller's `catch` only fires on an unexpected compositing exception, not on this path — so the "genuine basemap regression would otherwise silently degrade every route map with no signal" goal from the logging commit isn't actually covered. Consider logging here when `loaded.length === 0` (distinct from the expected `MAX_TILES` guard above) so a real OSM outage/block is observable.

fiddur (commented):
🛑 Changes required

Solid, well-tested feature. The Web Mercator math (`lonToWorldX`/`latToWorldY`/`chooseZoom`), viewport centering, tile-grid compositing with the crop-to-avoid-negative-offsets trick, and the overlay alignment (`x/y` subtract `vpLeft/vpTop` to match the extracted map, with the clamp only a sub-pixel safety net) all check out. OSM tile-policy compliance is handled well: descriptive `User-Agent`, baked-in `© OpenStreetMap contributors` attribution, per-post render caching, and the 2-connection concurrency cap. Fallbacks are safe (empty route, degenerate point, MAX_TILES guard, all-tiles-fail). Nice test coverage that actually samples pixels to prove basemap vs. bare fallback.

One finding worth addressing before merge (inline): the fallback log added in the last commit doesn't fire for the most likely basemap regression. `fetchOsmTile` swallows every error (non-200, timeout, network, 429/blocked) into `null` and never throws, so when OSM stops serving tiles, `renderRouteWithBasemap` returns `null` via `loaded.length === 0` and `renderRoutePng` silently falls through to the bare shape — no `console.warn`. The `try/catch` log only catches unexpected compositing exceptions, i.e. the *less* likely path. The stated goal ("a genuine basemap regression would otherwise silently degrade every route map with no signal") isn't met for the all-tiles-fail case.

Minor, non-blocking considerations (no action required): the 2-connection cap is per-render, not global, so N concurrent `route.png` requests for distinct posts can open up to 2N connections to OSM; and the render cache is a bounded (200-entry) in-memory LRU cleared on restart, so tiles are re-fetched after eviction/restart — both fine at this app's scale.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The route map now composites an OpenStreetMap basemap behind the GPS track with baked-in attribution, and gracefully falls back to the previous bare polyline when tiles are absent (no `fetchTile`), the grid is pathological, the fetch fails, or the render throws — each degradation path is `console.warn`-logged so it's observable rather than silent. Nice attention to the OSM tile usage policy (descriptive `User-Agent`, ≤2 concurrent fetches, per-post render caching, attribution) and good test coverage of the projection, zoom selection, basemap compositing, and both fallback paths.

I checked the geometry carefully and it holds up:
- **Extract/crop bounds are safe.** The viewport is a fixed 700px, so the covering grid is always 3–4 tiles wide/tall → `gridW`/`gridH` ≥ 768 ≥ 700, making the `clamp(…, 0, gridW - ROUTE_W)` upper bound always ≥ 0. Composite offsets `(tx - tileMinX) * TILE_SIZE` and `extractLeft/Top ∈ [0,256)` are always non-negative, matching the comment's intent.
- **The `routeOverlaySvg` refactor preserves the original bare rendering exactly** (overBasemap=false → no halo, no marker stroke, no attribution — identical to the pre-PR SVG).
- **`mapWithConcurrency` is race-free** (`next++` is atomic in single-threaded JS between awaits) and preserves order.
- **Out-of-range tiles** (antimeridian/pole edges) resolve to `null` and are treated as background gaps; an all-null result triggers the logged fallback.

Notes (non-blocking):
- **Privacy:** the docs are correctly updated to say a public route map now reveals the *precise* area (incl. likely home/work) and that sharing is opt-in per post — this is a conscious, documented tradeoff, and the image router only serves opted-in public/unlisted posts (or followers-with-token). Worth keeping the start-point/area masking follow-up on the radar.
- **OSM tile policy:** the per-render concurrency cap is per call, so N simultaneous distinct-post route renders could open up to 2N connections to `tile.openstreetmap.org`. For a single-user, cached, low-volume deployment this is fine, but if usage grows, a global/queue-level cap (or a self-hosted/commercial tile source) would be the safer long-term stance under OSMF's policy.
- **`MAX_TILES = 25` guard** is effectively unreachable given the fixed 700px viewport (grid is always ~3–4 tiles). Harmless defensive code — no change needed.

I could not execute the test suite in this worktree (dependencies aren't installed and the npm cache is read-only), but the tests read as correct by inspection.

PR #920, opened 2026-07-03
feat(feed): home timeline — ingest & view posts from followed actors (#884)

fiddur:
Slice 3 of the in-app federated feed (#884): the **home timeline** — receiving posts from the actors a user follows and showing them in the app. Builds on slice 1 (feed posts) and slice 2 (following, inbound follow relationship).

## What this adds

**Ingest (inbound `Create`/`Update`/`Delete` of `Note`s):**
- Only activities from an **accepted** followee are ingested — an unsolicited `Create` can't inject into the timeline.
- Remote HTML `content` is **sanitised server-side on ingest** (`sanitize-html`, strict allowlist) — the XSS boundary. So the web client renders it directly.
- Stored in `timeline_entry`, keyed by the note's `object_uri` so an `Update`/redelivery replaces in place. `Delete` removes the entry; unfollowing removes all of that actor's entries.

**Read API (REST + MCP parity):**
- `GET /feed/timeline` and the `list_timeline` MCP tool, backed by a shared `getTimelinePage` service.
- Keyset-paginated on `(published_at, id)` behind an opaque `next_cursor` (same style as the outbox). The row fetcher is injected so the pagination + cursor logic is **unit-tested without a database**.

**Web:**
- A **Home timeline** section on `/feed` (below Following): a `useInfiniteQuery` over `/feed/timeline` with a **Load more** button, each entry as a Mastodon-style card reusing the existing `.feed-post` styles.

## Tests
- `services/timeline.test.ts` (6) — serialize + pagination/cursor round-trip via injected fetcher.
- `db/timeline.integration.test.ts` (5), `services/activitypub/timeline-ingest.test.ts` (8, incl. XSS payloads).
- Full slice-3 + OSM suite green (30 tests); both package `check`s pass (0 errors).

## Docs
`docs/features/feed.md`: new "Home timeline (inbound)" section, `/feed/timeline` + `list_timeline` surfaces, and `timeline_entry` storage.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -97,6 +97,11 @@ const localFollowerIdentifier = async (
  * non-Notes are ignored, and a missing DB never 500s (which would invite retries).
  */
 const ingestFeedActivity = async (ctx: InboxContext<void>, activity: Create | Update): Promise<void> => {
+  // The recipient (whose timeline this is) comes from the personal inbox owner.
+  // Unlike Accept/Reject there's no inner Follow to derive it from, so this relies
+  // on personal-inbox delivery; `ctx.recipient` is null on a shared inbox. That's
+  // fine today (the actor advertises no `sharedInbox`), but shared-inbox support
+  // would need fanning a single delivery out to every local follower of the actor.
   const me = ctx.recipient
   if (me == null || !isValidUsername(me) || activity.actorId == null) return
   try {
@@ -248,11 +253,14 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
     .on(Update, (ctx, update) => ingestFeedActivity(ctx, update))
     .on(Delete, async (ctx, del) => {
       // Delete of a post we received → drop it from the timeline. `del.objectId`
-      // is the removed object's id (a Tombstone or bare id); we key on it.
+      // is the removed object's id (a Tombstone or bare id); we key on it, but
+      // scope the delete to the sender (`del.actorId`) so a signed `Delete` from
+      // some other actor can't evict a post it didn't author — mirroring how
+      // `Undo{Follow}` is scoped to its own actor.
       const me = ctx.recipient
-      if (me == null || !isValidUsername(me) || del.objectId == null) return
+      if (me == null || !isValidUsername(me) || del.objectId == null || del.actorId == null) return
       try {
-        await deleteTimelineEntryByUri(me, del.objectId.href)
+        await deleteTimelineEntryByUri(me, del.objectId.href, del.actorId.href)
       } catch (error) {
         if (isMissingDatabase(error)) return
         throw error
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -1,5 +1,16 @@
 import { createFederation, type Federation, type InboxContext, MemoryKvStore } from '@fedify/fedify'
-import { Accept, type Create, Follow, Image, Note, Person, Reject, Undo } from '@fedify/fedify/vocab'
+import {
+  Accept,
+  Create,
+  Delete,
+  Follow,
+  Image,
+  Note,
+  Person,
+  Reject,
+  Undo,
+  Update,
+} from '@fedify/fedify/vocab'
 
 /**
  * The Fedify `Federation` object for the activity feed.
@@ -13,8 +24,9 @@ import { Accept, type Create, Follow, Image, Note, Person, Reject, Undo } from '
  * - key-pairs dispatcher backed by the per-user `feed_actor` keypair,
  * - inbound inbox: `Follow` → persist follower + `Accept`; `Undo{Follow}` →
  *   drop the follower; `Accept`/`Reject` → resolve a Follow WE sent (mark the
- *   `feed_following` row accepted, or drop it). Fedify verifies the HTTP
- *   Signature first.
+ *   `feed_following` row accepted, or drop it); `Create`/`Update` of a `Note`
+ *   from an *accepted followee* → ingest into the home timeline (sanitised);
+ *   `Delete` → drop the received post. Fedify verifies the HTTP Signature first.
  * - followers + following collections (the latter lists this user's *accepted*
  *   follows), both backed by Postgres.
  *
@@ -26,6 +38,8 @@ import {
   countAcceptedFeedFollowing,
   countFeedFollowers,
   countPublicFeedPosts,
+  deleteTimelineEntryByUri,
+  getFeedFollowingByActor,
   getFeedPostById,
   getOrCreateActorKeyPair,
   isMissingDatabase,
@@ -36,12 +50,14 @@ import {
   removeFeedFollower,
   removeFeedFollowingByActor,
   upsertFeedFollower,
+  upsertTimelineEntry,
 } from '../../db/index.ts'
 import { resolveFeedActivity } from '../feed.ts'
 import { buildProfileUrl } from '../share-urls.ts'
 import { buildFeedCreate, buildFeedNote } from './deliver.ts'
 import { toCryptoKeyPair } from './keys.ts'
 import { isPubliclyVisible } from './object.ts'
+import { noteToTimelineInput } from './timeline-ingest.ts'
 
 /** Posts per outbox page (cursor pagination). */
 const OUTBOX_PAGE_SIZE = 20
@@ -70,6 +86,31 @@ const localFollowerIdentifier = async (
   }
   const recipient = ctx.recipient
   return recipient != null && isValidUsername(recipient) ? recipient : null
+}
+
+/**
+ * Ingest a `Create`/`Update` of a `Note` into the recipient's home timeline —
+ * but ONLY if the sender is an *accepted* followee (so the timeline can't be
+ * injected into by a stranger who delivers to our inbox). The author's
+ * presentation comes from the cached `feed_following` row; the Note's HTML is
+ * sanitised in `noteToTimelineInput`. Best-effort: unresolvable objects and
+ * non-Notes are ignored, and a missing DB never 500s (which would invite retries).
+ */
+const ingestFeedActivity = async (ctx: InboxContext<void>, activity: Create | Update): Promise<void> => {
+  const me = ctx.recipient
+  if (me == null || !isValidUsername(me) || activity.actorId == null) return
+  try {
+    const follow = await getFeedFollowingByActor(me, activity.actorId.href)
+    if (follow == null || !follow.accepted) return
+    const object = await activity.getObject({ suppressError: true })
+    if (!(object instanceof Note)) return
+    const input = noteToTimelineInput(object, follow)
+    if (input == null) return
+    await upsertTimelineEntry(me, input)
+  } catch (error) {
+    if (isMissingDatabase(error)) return
+    throw error
+  }
 }
 
 export const createFeedFederation = (origin: string, apiBaseUrl: string): Federation<void> => {
@@ -194,6 +235,24 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
       if (me == null || reject.actorId == null) return
       try {
         await removeFeedFollowingByActor(me, reject.actorId.href)
+      } catch (error) {
+        if (isMissingDatabase(error)) return
+        throw error
+      }
+    })
+    // Create / Update of a Note from an actor we follow → ingest into our home
+    // timeline. Update reuses the same upsert (keyed on the Note's object id), so
+    // an edit replaces the stored copy. Only *accepted* followees are ingested, so
+    // a stranger who somehow delivers here can't inject into the timeline.
+    .on(Create, (ctx, create) => ingestFeedActivity(ctx, create))
+    .on(Update, (ctx, update) => ingestFeedActivity(ctx, update))
+    .on(Delete, async (ctx, del) => {
+      // Delete of a post we received → drop it from the timeline. `del.objectId`
+      // is the removed object's id (a Tombstone or bare id); we key on it.
+      const me = ctx.recipient
+      if (me == null || !isValidUsername(me) || del.objectId == null) return
+      try {
+        await deleteTimelineEntryByUri(me, del.objectId.href)
       } catch (error) {
         if (isMissingDatabase(error)) return
         throw error
--- a/apps/backend/src/services/activitypub/timeline-ingest.ts
+++ b/apps/backend/src/services/activitypub/timeline-ingest.ts
@@ -1,0 +1,87 @@
+/**
+ * Ingest a received ActivityPub `Note` (a post from an actor the user follows)
+ * into the home-timeline store.
+ *
+ * Two concerns live here, both testable in isolation:
+ *
+ * 1. `sanitizeRemoteHtml` — remote fediverse HTML is **untrusted** and is rendered
+ *    with `dangerouslySetInnerHTML` on the web, so it MUST be sanitised before it
+ *    is stored. We keep only the small tag set Mastodon-style content uses and
+ *    force safe link attributes; scripts, styles, event handlers, iframes, images,
+ *    and every other attribute are dropped.
+ * 2. `noteToTimelineInput` — map a Fedify `Note` + the (already-known) author to a
+ *    `TimelineEntryInput`, or null if it lacks the id/published we need.
+ *
+ * The inbox handler that calls these (in `federation.ts`) is thin: it checks the
+ * sender is a followed, accepted actor and upserts the result.
+ */
+import type { Note } from '@fedify/fedify/vocab'
+
+import sanitizeHtml from 'sanitize-html'
+
+import type { FeedFollowingRecord, TimelineEntryInput } from '../../db/index.ts'
+
+import { temporalInstantToDate } from './temporal-interop.ts'
+
+/**
+ * Sanitise untrusted remote HTML to the small tag set Mastodon-style post content
+ * uses. Links are forced to `rel="nofollow noopener noreferrer"` + `target`, and
+ * only `http(s)` schemes survive — so no `javascript:` URLs, inline handlers,
+ * styles, scripts, iframes, or images.
+ */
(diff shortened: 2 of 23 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -199,6 +240,24 @@ export const createFeedFederation = (origin: string, apiBaseUrl: string): Federa
...
+    // a stranger who somehow delivers here can't inject into the timeline.
+    .on(Create, (ctx, create) => ingestFeedActivity(ctx, create))
+    .on(Update, (ctx, update) => ingestFeedActivity(ctx, update))
+    .on(Delete, async (ctx, del) => {
+      // Delete of a post we received → drop it from the timeline. `del.objectId`
+      // is the removed object's id (a Tombstone or bare id); we key on it.
+      const me = ctx.recipient
+      if (me == null || !isValidUsername(me) || del.objectId == null) return
+      try {
+        await deleteTimelineEntryByUri(me, del.objectId.href)
```
fiddur: This `Delete` is under-authorized. Unlike the `Create`/`Update` path (gated to accepted followees), it removes a timeline entry by `object_uri` alone, with no check on who sent the `Delete`. Fedify only verifies the request is signed by *some* resolvable actor, so any fediverse actor can deliver a spoofed `Delete{ object: <a public note URI they know is in your timeline> }` and permanently evict that post from the victim's home timeline (an integrity/censorship vector).

Scope the deletion to the sender — only let an actor delete its own posts, e.g. `deleteTimelineEntryByUri(me, del.objectId.href, del.actorId.href)` backed by `DELETE FROM timeline_entry WHERE object_uri = $1 AND actor_uri = $2`. That mirrors how the `Undo{Follow}` handler scopes removal to `undo.actorId`.

fiddur (commented):
🛑 Changes required

Solid, well-structured slice: per-user DB isolation makes the columnless `timeline_entry` correct, keyset pagination on `(published_at DESC, id DESC)` is done properly (composite tuple comparison + `limit+1` next-page probe + opaque cursor with graceful malformed-cursor handling), the sanitizer is the right XSS boundary and is nicely unit-tested, and REST/MCP parity + docs are all in place.

One blocking issue: the inbound **`Delete` handler is under-authorized**. The `Create`/`Update` ingest path is carefully gated to *accepted followees* ("a stranger who somehow delivers here can't inject"), but `Delete` deletes by `object_uri` alone with no check that the sending actor authored — or is even followed by — the target entry. Since Fedify only proves the request is signed by *some* resolvable actor, any fediverse actor can deliver a spoofed `Delete{ object: <a known/guessable public note URI in your timeline> }` and permanently evict that post from a victim's home timeline. Scope the delete to the sender, e.g. `DELETE ... WHERE object_uri = $1 AND actor_uri = $2` with `del.actorId.href` (mirroring how `Undo{Follow}` is scoped to `undo.actorId`).

Non-blocking notes for consideration:
- **Shared-inbox recipient**: `ingestFeedActivity` and the `Delete` handler rely on `ctx.recipient`, which the code itself documents is `null` on the shared `/inbox`. The actor currently doesn't advertise `endpoints.sharedInbox`, so delivery lands on the personal inbox and this works — but if a shared inbox is ever advertised, ingestion silently drops. Worth a comment or deriving the recipient more robustly (as the Follow handler does via `parseUri`).
- **Test coverage**: `noteToTimelineInput`/`sanitizeRemoteHtml` and the DB layer are well covered, but the security-sensitive inbox authorization logic in `federation.ts` (the accepted-followee gate, and the `Delete` scoping once fixed) has no direct test. Given the finding above, a test asserting a non-followee `Delete`/`Create` is ignored would be valuable.
- Minor: a `Delete{actor}` (remote account deletion) won't purge that actor's posts (its `objectId` is the actor URI, matching no `object_uri`); unfollow handles bulk purge but account-deletion doesn't. Acceptable as a follow-up.

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed in 9d4559cb: `deleteTimelineEntryByUri` now takes the authoring `actor_uri`; the handler passes `del.actorId.href`, so the delete is `WHERE object_uri = $1 AND actor_uri = $2` — a signed `Delete` from any other actor no longer evicts a post it didn't author (mirrors `Undo{Follow}` scoping). Added a spoofed-Delete integration test.

apps/web/src/pages/Feed/HomeTimeline.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import { useInfiniteQuery } from '@tanstack/react-query'
+import { formatDistanceToNow } from 'date-fns'
+
+import { fetchTimeline } from '../../state/api'
+
+/** A fallback avatar (a neutral silhouette) for actors without an icon. */
+const FALLBACK_AVATAR =
+  'data:image/svg+xml;utf8,' +
+  encodeURIComponent(
```
fiddur, line 24: The `#` colors are double-encoded, so the fallback avatar renders as black blobs rather than the intended gray silhouette. The SVG string already writes `%23` for `#`, and then `encodeURIComponent` re-encodes the `%` to `%25` → `%2523`. When the browser percent-decodes the `data:` URI once, `fill` ends up as the literal string `"%23cbd5e1"`, which is not a valid SVG color, so fills fall back to the initial `black`.

Use raw `#` in the SVG and let `encodeURIComponent` handle it (i.e. `fill="#cbd5e1"` / `fill="#fff"`), or drop `encodeURIComponent` and keep the pre-encoded `%23`. Only fires for followed actors with no cached avatar.

fiddur (commented):
✅ Approved

Solid, well-tested slice: the home-timeline store, ingest, service, REST/MCP parity, and web view all hang together, and the security-sensitive parts are handled with care — ingest is gated on *accepted* followees, inbound `Delete` is scoped to the authoring actor (with a test for the spoofed-Delete case), remote HTML is sanitised at ingest with a strict allowlist, and unfollow purges the actor's entries. Keyset pagination (`published_at, id`) and the opaque cursor are correct and unit-tested including the malformed-cursor path. The database-per-user model makes the globally-`UNIQUE object_uri` and the absence of a user column fine.

One genuine (minor, cosmetic) bug in the fallback avatar — see the inline comment.

Two non-blocking hardening notes on the federation ingest, worth a follow-up:

- **Note author is not verified against the sender.** `ingestFeedActivity` checks the *delivering* actor (`activity.actorId`) is an accepted followee, but doesn't check the embedded Note's `attributedTo` matches. A followed actor could therefore deliver a Note carrying an `id`/`url` on a domain they don't control, and it'd be stored (attributed to them) with a `url` that links elsewhere. Low severity given you chose to follow them and the display author is always the (correct) sender, but verifying `attributedTo`/origin matches the sender is the usual belt-and-braces here, and would also prevent one followee from squatting another's `object_uri` (the upsert would then overwrite `actor_uri` on the real author's later delivery).

- **`published_at` is remote-controlled and drives ordering.** Since the timeline orders by `published_at DESC`, a followee posting with a far-future `published` pins their post to the top of the home timeline indefinitely. Consider clamping `published_at` to `received_at` when it's in the future.

Neither blocks the feature. (Also worth confirming CI regenerates the api-spec OpenAPI/Kotlin artifacts, since only the Zod source changed here — the generated output isn't tracked in git.)

apps/web/src/pages/Feed/HomeTimeline.tsx
fiddur: Good catch — fixed by using raw `#` in the SVG so `encodeURIComponent` encodes it once (`%23cbd5e1`). Verified the output is single-encoded.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -72,6 +88,36 @@ const localFollowerIdentifier = async (
...
+  // fine today (the actor advertises no `sharedInbox`), but shared-inbox support
+  // would need fanning a single delivery out to every local follower of the actor.
+  const me = ctx.recipient
+  if (me == null || !isValidUsername(me) || activity.actorId == null) return
+  try {
+    const follow = await getFeedFollowingByActor(me, activity.actorId.href)
+    if (follow == null || !follow.accepted) return
+    const object = await activity.getObject({ suppressError: true })
+    if (!(object instanceof Note)) return
+    const input = noteToTimelineInput(object, follow)
```
fiddur, line 112: The sender is verified to be an accepted followee, but the Note's `id` (→ `object_uri`, the globally-UNIQUE upsert key) and `attributedTo` aren't checked against the sender's origin. Since `upsertTimelineEntry` does `ON CONFLICT (object_uri) DO UPDATE SET actor_uri = EXCLUDED.actor_uri, content = …`, a malicious accepted followee can deliver a `Create` whose Note id collides with another followee's post and overwrite/evict that entry with their own content (attributed to themselves). Consider an origin/authority check (Note id host and `attributedTo` match `activity.actorId`) — the symmetric guard to the sender-scoped `Delete` you already implement below.

apps/backend/src/services/timeline.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/** Decode an opaque cursor, or undefined if it's missing/malformed (→ first page). */
+const decodeCursor = (cursor: string | undefined): TimelineCursor | undefined => {
+  if (cursor == null || cursor === '') return undefined
+  const decoded = Buffer.from(cursor, 'base64url').toString('utf8')
+  const sep = decoded.indexOf(':')
+  if (sep <= 0) return undefined
+  const ms = Number(decoded.slice(0, sep))
+  const id = decoded.slice(sep + 1)
+  if (!Number.isSafeInteger(ms) || id === '') return undefined
```
fiddur: `decodeCursor` is documented to return `undefined` for a malformed cursor, but it doesn't validate that `id` is a UUID. A crafted cursor decoding to e.g. `12345:not-a-uuid` passes this guard (`ms` safe, `id !== ''`) and reaches the `$2::uuid` cast in `listTimelineEntries`, which throws `invalid input syntax for type uuid` and 500s the request instead of falling back to the first page. Add a UUID-shape check on `id` before returning.

fiddur (commented):
✅ Approved

Solid, well-structured slice: the home-timeline store, keyset pagination, ingest, REST/MCP parity, and web view are cleanly separated and thoroughly tested (unit + integration), and the docs are updated. The XSS boundary is handled correctly — remote HTML is sanitised on ingest with a strict allowlist, and the web layer's `dangerouslySetInnerHTML` is explicitly justified against it. The Delete/unfollow authorization scoping is thoughtfully done.

Two hardening points worth considering (neither a happy-path bug, so not blocking):

1. **Object authority on ingest** (see inline). The Create/Update path trusts the *sender* being an accepted followee, but never checks that the Note's `id` (which becomes the globally-UNIQUE `object_uri` upsert key) or its `attributedTo` share the sender's origin. Because `ON CONFLICT (object_uri) DO UPDATE SET actor_uri = EXCLUDED.actor_uri, content = …`, a malicious followee can supply a Note whose id collides with another followee's post and overwrite/evict that entry with their own content. This mirrors the spoofed-Delete case you already guard against.

2. **Cursor UUID validation** (see inline). `decodeCursor` promises to return `undefined` on a malformed cursor, but a crafted cursor with a non-UUID id slips through to the `$2::uuid` cast and 500s the request.

Minor limitations (fine for this slice, just noting): ingest applies no visibility filter, so a followed actor's direct-message Note addressed to the user also lands in the home timeline (only the user sees their own timeline, so no leak — just unconventional); `published_at` is attacker-controlled and drives ordering, so a followee can pin a post at the top with a future timestamp; and `timeline_entry` grows unbounded (only unfollow/Delete prune it).

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed in 7a563bcf: `noteToTimelineInput` now rejects a Note whose `id` host differs from the sender (`author.actor_uri`), and — when the Note declares `attributedTo` — requires it to attribute to the sender. Since `object_uri = note.id.href` is the global upsert key, this blocks an accepted followee on one host from minting an id on another host to overwrite that actor's entry. Added spoof tests (cross-host id, mismatched attribution) + a correctly-attributed accept.

apps/backend/src/services/timeline.ts
fiddur: Fixed in 7a563bcf: `decodeCursor` now validates the id against a UUID regex, so `12345:not-a-uuid` decodes to `undefined` (first page) instead of reaching the `$2::uuid` cast. Added a test asserting the crafted cursor is rejected.

apps/backend/src/services/activitypub/timeline-ingest.ts
```diff
@@ -0,0 +1,99 @@
...
+  const attributions = note.attributionIds
+  if (attributions.length > 0 && !attributions.some((uri) => uri.href === author.actor_uri)) return null
+  return {
+    actor_uri: author.actor_uri,
+    avatar_url: author.avatar_url,
+    content: sanitizeRemoteHtml(note.content?.toString() ?? ''),
+    display_name: author.display_name,
+    handle: author.handle,
+    object_uri: note.id.href,
+    published_at: temporalInstantToDate(note.published),
```
fiddur: Non-blocking: `published_at` is taken verbatim from the remote Note and is the sort key (`ORDER BY published_at DESC`). A followed actor can therefore pin a post to the top of the timeline indefinitely — and re-pin it via `Update` — by publishing with a far-future timestamp. Mastodon guards against this by clamping `published` to "not in the future". Consider `new Date(Math.min(instant, Date.now()))` here (or clamping to `received_at`). Low severity since it's an actor you've chosen to follow, but cheap to harden.

fiddur (commented):
✅ Approved

Solid, well-structured slice that adds the inbound home timeline (ingest + view). The design is clean and matches the project conventions: a per-user `timeline_entry` store with keyset pagination, shared `getTimelinePage` service used by both the REST `/feed/timeline` route and the `list_timeline` MCP tool (parity ✔), api-spec schemas as the single source of truth, and Preact rendering consistent with the existing feed cards.

The security posture is notably good and backed by tests:
- Remote HTML is sanitised server-side on ingest (`sanitizeRemoteHtml`, strict tag/attribute allowlist, `http(s)`/`mailto` only, forced `rel`/`target`) before it's stored, so the `dangerouslySetInnerHTML` render is safe — and the XSS cases are exercised.
- Ingest is scoped to *accepted* followees, the Note's id must originate from the sender's host, and `attributedTo` (when present) must match the sender — closing the object_uri-collision spoof, with tests for each.
- Inbound `Delete` is scoped to the authoring actor (mirroring `Undo{Follow}`), so a signed Delete from another actor can't evict a post; spoof test present.
- The opaque cursor is validated (safe-integer ms + UUID regex) before hitting the `$::uuid` cast, so a crafted cursor pages from the start instead of 500ing.

I verified the multi-tenancy model (database-per-user, so no user column and no cross-user leakage of the global `object_uri` UNIQUE), that `validateQuery` replaces `req.query` in place (so `limit` is a real number, not a string), that `timeline_entry`/`timeline_entry_indexes` are wired into `tableCreationOrder` (so existing users get lazy migration), and that all imported symbols exist. Docs are updated. I could not execute the test suite in this review sandbox (read-only pnpm store), so the pass/fail is from static analysis; the tests themselves read as comprehensive.

One non-blocking hardening note inline: `published_at` comes straight from the remote and drives `ORDER BY published_at DESC`, with no clamp — a followed actor can pin their post to the top indefinitely with a far-future `published`. Minor (it's an actor you follow), but worth clamping to `min(published, now)` as Mastodon does. Also worth a follow-up: account-level `Delete{actor}` (object id == actor id) won't purge that actor's posts, and boosts (`Announce`) / media attachments aren't ingested — all reasonable scope cuts, just flagging.

apps/backend/src/services/activitypub/timeline-ingest.ts
fiddur: Hardened in the latest push: `noteToTimelineInput` clamps `published_at` to `Math.min(published, now)` on ingest (clock injected for deterministic tests), so a far-future timestamp can't pin a post to the top — matching Mastodon's guard. Added a test covering far-future clamp + past-timestamp passthrough.

fiddur (commented):
✅ Approved

Solid, well-tested slice implementing the inbound home timeline. The security reasoning is thorough and matches the code:

- **Ingest is gated to accepted followees** (`getFeedFollowingByActor` + `follow.accepted`), so a stranger who delivers to the inbox can't inject.
- **Object-id spoofing is guarded** in `noteToTimelineInput`: the Note must originate from the sender's host and, if it declares `attributedTo`, attribute to the sender — protecting the global `object_uri` upsert key from cross-followee overwrite.
- **Delete is sender-scoped** (`object_uri AND actor_uri`), mirroring `Undo{Follow}`, with an explicit test for the spoofed-Delete case.
- **XSS boundary is correct**: remote HTML is sanitized at ingest with a strict allowlist before storage, and the web renders the already-sanitized content — tests cover `<script>`, inline handlers, `javascript:`/`data:` URLs, images/iframes/styles.
- **Cursor handling is defensive**: a malformed or non-UUID cursor decodes to the first page rather than reaching (and 500-ing on) the `$::uuid` cast; the anti-pin clamp of `published_at` to now is tested.

I also confirmed the per-user-database architecture makes `object_uri TEXT UNIQUE` correctly scoped per user (no cross-user collision), that `migrateSchema` creates the new `timeline_entry`/index tables for existing users (and the `query()` wrapper auto-migrates on `42P01`), that `validateQuery` writes the coerced `limit` back to `req.query` (so `limit + 1` is arithmetic, not string concat), and that keyset pagination on `(published_at DESC, id DESC)` round-trips correctly. API/MCP parity, api-spec schemas, and docs are all updated. CI's Node job was still pending at review time — worth confirming green before merge.

A few non-blocking caveats (fine to leave as-is / document as limitations): (1) the strict `note.id.host === actor host` check will silently drop legitimate posts from instances that serve object IDs on a different host than the actor — acceptable for the common Mastodon case but a source of "missing posts" on some servers; (2) all Notes from a followee are ingested regardless of visibility/reply status, so a followee's replies (or a direct-message Note delivered to your inbox) will surface in the owner-only home timeline; (3) `Announce` (boosts) are not handled, which the docs already reflect as out of scope for this slice.

PR #921, opened 2026-07-04
feat(feed): live home-timeline updates — SSE + polling (#884)

fiddur:
Slice 4 of the in-app federated feed (#884): **live updates** — new posts from followed actors appear on the Feed page without a refresh. Builds on slice 3 (the home timeline #920).

## How it works

**Transport — Postgres `LISTEN/NOTIFY` → SSE.** Each user has one long-lived per-user DB connection, so ingest and an open stream share it: when a genuinely new `timeline_entry` is inserted, ingest emits an **empty `NOTIFY` ping**; the browser's SSE stream on the same connection receives it. Payload is empty (sidesteps NOTIFY's 8 kB cap; no post content on the wire).

**Backend:**
- `db/timeline-notify.ts` — thin `LISTEN/NOTIFY` glue (integration-tested against real PG).
- `services/timeline-hub.ts` — an injected in-process fan-out hub that keeps **one** `LISTEN` channel per user regardless of tab count, closing it when the last subscriber leaves. PG glue is injected, so refcounting/fan-out is unit-tested without a DB.
- `GET /feed/timeline/stream` — authenticated SSE endpoint; forwards each ping as `event: new`, heartbeats, sets `X-Accel-Buffering: no` (nginx), cleans up on disconnect, and `503`s if no hub is wired (client then polls).
- `upsertTimelineEntry` reports `inserted` (`xmax = 0`) so ingest pings **only on new posts**, not edits.

**Frontend:**
- `openTimelineStream` — a **`fetch`-based** SSE reader (bearer token in the `Authorization` header, not the URL, unlike `EventSource`); pure SSE parsing is unit-tested.
- `useTimelineLive` — subscribes and **falls back to 30 s polling** if the stream can't open or drops.
- On a ping/poll, `HomeTimeline` refetches the newest page and buffers anything new as a **"N new posts" pill**; clicking prepends them. Entries are de-duped by `object_uri`.

## Tests
- `timeline-hub.test.ts` (6, fan-out + refcount + open-failure), `timeline-notify.integration.test.ts` (2, real LISTEN/NOTIFY round-trip), `feed-router.test.ts` (3, SSE headers + event forward + disconnect cleanup), `timeline-sse.test.ts` (4, stream parsing). Affected suites all green (52 backend tests across the slice); both package `check`s pass (0 errors).

## No new schema
No api-spec/schema changes (SSE has no body); the transport is app-level NOTIFY on the existing per-user connection.

### DIFF
--- a/apps/backend/src/db/timeline.ts
+++ b/apps/backend/src/db/timeline.ts
@@ -48,12 +48,17 @@ const TIMELINE_COLUMNS =
  * Insert or update a received post by `object_uri`. A re-delivered or edited post
  * (same object id) refreshes the content/presentation in place; `received_at`
  * stays at first receipt while `published_at` tracks the remote timestamp.
+ *
+ * `inserted` distinguishes a brand-new post from an in-place refresh via the
+ * `xmax = 0` trick (freshly-inserted tuples have xmax 0; the ON CONFLICT update
+ * path locks the existing row, so its xmax is non-zero). The ingest path uses it
+ * to notify live subscribers only about genuinely new posts, not edits.
  */
 export const upsertTimelineEntry = async (
   user: string,
   input: TimelineEntryInput,
-): Promise<TimelineEntryRecord> => {
-  const result = await query<TimelineEntryRecord>(
+): Promise<TimelineEntryRecord & { inserted: boolean }> => {
+  const result = await query<TimelineEntryRecord & { inserted: boolean }>(
     user,
     `INSERT INTO timeline_entry
        (object_uri, actor_uri, handle, display_name, avatar_url, content, url, published_at)
@@ -66,7 +71,7 @@ export const upsertTimelineEntry = async (
                    content = EXCLUDED.content,
                    url = EXCLUDED.url,
                    published_at = EXCLUDED.published_at
-     RETURNING ${TIMELINE_COLUMNS}`,
+     RETURNING ${TIMELINE_COLUMNS}, (xmax = 0) AS inserted`,
     [
       input.object_uri,
       input.actor_uri,
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -28,6 +28,7 @@ import { setupOuraWebhook, setupStravaWebhook } from './api/webhooks-setup.ts'
 import { createAuth } from './auth.ts'
 import {
   deleteRuleActivities,
+  emitTimelineNotify,
   getDeductionRulesByIds,
   getDetectedLocationById,
   getEnabledDeductionRules,
@@ -41,6 +42,7 @@ import {
   insertRawRecord,
   insertTimeSeries,
   loginToUserDb,
+  openTimelineChannel,
   resolveOrCreateActivityType,
   softDeleteLocationRange,
   updateDetectedLocation,
@@ -78,6 +80,7 @@ import { createPgBoss } from './services/pg-boss.ts'
 import { initSentry, Sentry } from './services/sentry.ts'
 import { createStravaQueue, type StravaQueue } from './services/strava-queue.ts'
 import { createSyncProvider } from './services/sync-provider.ts'
+import { createTimelineHub } from './services/timeline-hub.ts'
 import { createWebAuthnService } from './services/webauthn.ts'
 
 declare global {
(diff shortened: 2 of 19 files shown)


apps/backend/src/db/timeline.ts
```diff
@@ -66,7 +71,7 @@ export const upsertTimelineEntry = async (
                    content = EXCLUDED.content,
                    url = EXCLUDED.url,
                    published_at = EXCLUDED.published_at
-     RETURNING ${TIMELINE_COLUMNS}`,
+     RETURNING ${TIMELINE_COLUMNS}, (xmax = 0) AS inserted`,
```
fiddur, line 74: The `(xmax = 0)` trick is the linchpin of the whole feature — federation only pings live subscribers when `inserted === true` — yet there's no test asserting it. The existing `timeline.integration.test.ts` "re-delivering the same object upserts in place" case is the natural spot: assert `inserted === true` on the first `upsertTimelineEntry` and `inserted === false` on the re-delivery. Given the repo standard that all DB functions have integration tests (and the subtlety of the xmax idiom), this is worth adding so a future refactor can't silently break live updates (or start over-notifying on edits).

fiddur (commented):
✅ Approved

Solid, well-structured feature. The design is clean: an injectable `TimelineHub` isolates the refcounting + fan-out logic from the PG glue (nicely unit-tested), and the SSE endpoint, LISTEN/NOTIFY transport, federation wiring, and frontend pill/polling fallback all hang together. Security is handled thoughtfully — empty NOTIFY payload (no content on the wire, sidesteps the 8 kB limit), per-user channel isolation, and the bearer token in the `Authorization` header (via `fetch`, not `EventSource`). Docs are updated. I checked the potential hazard around the shared per-user `pg.Client`: because `openTimelineChannel`/teardown both resolve to the same connection and `LISTEN`/`UNLISTEN` are session-global, I traced whether a concurrent reopen could have its `LISTEN` clobbered by the last unsubscribe's `UNLISTEN`. It's safe — the teardown enqueues `UNLISTEN` within microtask draining (before any new HTTP subscribe macrotask can run), and node-postgres executes queued queries FIFO, so a later reopen's `LISTEN` always lands after. Cleanup ordering in the SSE handler (close firing during channel open) is also correctly handled.

One gap worth addressing: the new `inserted` flag driving "notify only on genuinely new posts" has no test coverage anywhere (see inline). Minor: the SSE route's `ResponseBody` generic (`{ success: false; error: string }`) is an inline type rather than an api-spec response type — understandable since this is a stream, not a typed JSON endpoint, but it does diverge from the CLAUDE.md routing convention; consider a shared api-spec error type if one exists. Neither is blocking.

apps/backend/src/db/timeline.ts
fiddur: Added to the existing re-delivery test: asserts `inserted === true` on the first upsert and `false` on the re-delivery (the ON CONFLICT update path), so a future refactor can't silently break "ping only on new posts" or start over-notifying on edits. Also switched the SSE route's ResponseBody to the shared api-spec `BaseResponse` per the routing convention.

fiddur (commented):
✅ Approved

Solid, well-tested feature. The layering is clean: a payload-free Postgres `LISTEN/NOTIFY` transport (`db/timeline-notify.ts`), an in-process refcounting fan-out `TimelineHub` (injected PG glue → unit-testable without a DB), an SSE endpoint that degrades to a 503→poll fallback, and a client that uses `fetch` (bearer in header, not URL) with a 30s polling fallback and a `dedupByUri` render guard. Test coverage is strong across all layers (hub refcounting/isolation/failed-open, SSE parse edge cases, feed-router 503 + stream + unsubscribe, and LISTEN/NOTIFY + `inserted`-flag integration tests), and the docs are updated.

I verified the load-bearing assumptions: `getDbForUser` returns one long-lived `Client` per user, so the ingest NOTIFY and the SSE LISTEN share a connection and self-delivery works (matches the integration test); there's no `compression` middleware to buffer the stream; the `upsertTimelineEntry` return-type change is additive and its only production caller destructures `{ inserted }`; the new `createFeedFederation`/`createFeedRouter` params are optional so existing callers/tests still compile; and the hub's concurrent-subscribe slot reservation, last-listener teardown, and close-before-subscribe-resolves race are all handled correctly.

- The `"N new posts"` pill can briefly over-count: `checkForNew` buffers fresh posts in `pending` but never invalidates the infinite query, so if react-query background-refetches page 1 (e.g. on window focus) those posts appear in `displayed` via `entries` while still counted in the pill. `dedupByUri` prevents duplicate cards, so this is only a cosmetic count quirk, and clicking the pill resolves it.
- `revealed` accumulates for the lifetime of the mounted component (never pruned once the same posts naturally enter `entries`). Harmless for normal sessions given the dedup, just worth noting for very long-lived tabs.

Nice touch pinging live subscribers only for genuinely new posts via the `(xmax = 0)` upsert flag rather than on every edit/redelivery.

PR #922, opened 2026-07-04
feat(feed): harden following — self-follow guard, icon timeout, error surfacing (#884)

fiddur:
First PR of slice 5 (share consolidation). Folds in the deferred **#917 non-blocking follow-ups** on the following flow — small, independent correctness/UX hardening.

## Changes
- **Reject self-follow** — following your own actor URI is rejected (422) before persisting a pending row; it would otherwise deliver a `Follow` to your own inbox and echo your posts into your own timeline.
- **Icon-fetch timeout** — `actorToFollowingInput` now bounds `actor.getIcon()` with a 3s timeout (`withTimeout`). Mastodon inlines the icon (no fetch), but a server that only *links* it would make `getIcon()` dereference a URL, which could hang the synchronous follow. The avatar is non-essential, so a timeout falls back to no avatar.
- **Specific follow errors in the UI** — the Following panel surfaces the server's `{ error }` message (e.g. "You can't follow yourself.", "Could not resolve an actor for …") instead of a generic message; the api client extracts it from the failed response.

## Deferred (with rationale)
The fourth #917 note — paginating the ActivityPub `following` collection — is **deliberately not done**: the `followers` collection is also unpaginated, both are small, and only the outbox (potentially large) is paginated. Paginating one collection but not the symmetric other would be inconsistent; this is YAGNI until counts actually grow.

## Tests
- `following.test.ts`: added `withTimeout` cases (resolves in time / rejects on hang). `followActor`'s network orchestration stays integration-territory (unchanged pattern). Both package `check`s pass (0 errors).

### DIFF
--- a/apps/backend/src/services/following.test.ts
+++ b/apps/backend/src/services/following.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, test } from 'vitest'
 
 import type { FeedFollowingRecord } from '../db/index.ts'
 
-import { actorToFollowingInput, serializeFollowing } from './following.ts'
+import { actorToFollowingInput, serializeFollowing, withTimeout } from './following.ts'
 
 describe('actorToFollowingInput', () => {
   test('extracts uri, inbox, shared inbox, handle, name, and avatar from a full actor', async () => {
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped hardening of the following flow. Three concerns, all correct:

- **Self-follow guard** (`following.ts:155`): compares `actor.id.href` against `ctx.getActorUri(user).href`. This is sound — the actor dispatcher mints local actor ids via the same `ctx.getActorUri(identifier)` (`federation.ts:164`), so a self-lookup resolves to a byte-identical URI, and it catches both the `@user@host` and actor-URL forms. Living in the shared service means REST and MCP both inherit it (parity preserved), and it rejects before persisting a pending row.

- **Icon timeout** (`withTimeout`): correct implementation — `Promise.race` attaches a reaction to the underlying `getIcon()` promise, so a late rejection from a slow icon host won't surface as an unhandled rejection, and the timer is always cleared in `finally`. Nicely unit-tested (resolve-in-time and never-settle cases).

- **Error surfacing**: `apiErrorMessage` narrows the `unknown` axios payload then guards with `typeof === 'string'`, and the REST router returns `{ error }` in exactly that shape. The message is rendered as text in Preact so there's no XSS risk, and the `let response` evolving-any pattern typechecks because the catch always throws.

Docs are updated to match. Minor, non-blocking: the self-follow branch has no test, but that's consistent with the file's stated philosophy that the network orchestration is thin and only the pure mappers are unit-tested — an integration-style test would need to stub the federation context, so I wouldn't hold the PR for it.

PR #923, opened 2026-07-04
refactor(sharing): unify visibility model across feed, dashboards, challenges (#884)

fiddur:
## Slice 5b of #884 — visibility model unification (type + UI, no storage migration)

Consolidates how *shareable* things describe their audience. Previously the feed
used a rich `public`/`unlisted`/`followers` radio triad while shared dashboards
and challenges used a bare **"Public"** checkbox (`is_public` boolean) — two
vocabularies and two UX idioms. This unifies both.

### What changed

**api-spec** — new shared **`ShareVisibility`** vocabulary (`public`/`unlisted`),
the single source of truth. `FeedVisibility` now derives from it (adds
`followers`). Shared dashboards and challenges expose a `visibility` field
(create/update bodies, responses, and the public challenge view) instead of
`is_public`.

**backend** — `services/visibility.ts` maps `visibility` ⇄ the existing
`is_public` boolean at the REST-router and MCP-tool boundary. **Storage stays a
boolean — no DB migration, no index changes.** Both directions unit-tested.

**web** — a shared **`VisibilitySelector`** component (full form variant + a
compact inline variant for dense list rows), used in the share dialog, the
challenge create form, and per-dashboard rows. Dead per-page visibility CSS
removed.

**docs** — sharing/challenges/feed docs describe the shared vocabulary and the
boundary mapping.

### Design note
Per the agreed scope, this is *type + UI* unification only: the public API/MCP/UI
surface speaks `visibility`, but the DB column remains `is_public` (mapped at the
boundary). No live-data migration.

### Checks
- api-spec builds; backend `tsgo` clean; web `tsgo` clean; oxlint clean (only the
  pre-existing accepted `req.user!` warnings).
- Backend: 1603 tests pass (incl. routes/mcp/services + integration).
- Web: 499 tests pass.


fiddur (commented):
✅ Approved

Clean, well-executed refactor. The `ShareVisibility` (`public`/`unlisted`) vocabulary is introduced once in `api-spec`, `FeedVisibility` is derived from it (`[...shareVisibilityValues, 'followers']`), storage stays an `is_public` boolean, and the boolean↔vocabulary mapping lives in a single, unit-tested `services/visibility.ts`. The `VisibilitySelector` web component nicely deduplicates the three bespoke pickers, and the `name`-per-instance prop correctly avoids radio-group collisions across dashboard rows. Docs are updated coherently.

I verified the conversion is complete and internally consistent:
- No stale `is_public` remains in the api-spec schemas or web consumers; the DB layer intentionally keeps the boolean and its internal readers (`oembed-router`, `og-image-router`) read DB records, so they're unaffected.
- Update paths correctly guard `visibility === undefined ? undefined : visibilityToIsPublic(...)`, and the new `.default('unlisted')` preserves the previous `is_public` default of `false`.
- Re-`.meta({ description })` on `shareVisibilitySchema` (which carries `id: 'ShareVisibility'`) matches the established pattern used for `activityTypeSchema`/`metricTypeSchema`, so OpenAPI generation is fine.
- The `feedVisibilitySchema` value reorder is cosmetic — no code depends on enum ordering, and the inferred union type is unchanged.
- Android has no challenge/dashboard models and the generated Kotlin/OpenAPI artifacts aren't tracked (rebuilt in CI), so nothing there goes stale.

One non-blocking observation: this renames `is_public`→`visibility` in the public REST responses/bodies and, notably, in the federation `PublicChallenge` wire payload. That's a breaking contract change for any out-of-repo or cross-version consumer (a new instance fetching a challenge from an old-version host, or vice-versa, would fail the `publicChallengeResponseSchema` parse). Since the join logic never actually reads this field and all in-repo consumers are updated together, this is acceptable for the project's current scale — just worth being aware of when instances on different versions federate. I did not run `pnpm check` as part of this review; CI should confirm the typecheck.

PR #925, opened 2026-07-04
refactor(nav): group feed/dashboards/challenges under a "Sharing" section (#884)

fiddur:
## Slice 5c of #884 — nav/IA consolidation

The sidebar item literally labeled **"Share"** pointed only to `/shared-dashboards`,
while **Feed** and **Challenges** were separate flat siblings — so "Share" read as a
generic verb but meant *dashboards*, and the three sharing surfaces had no grouping.

### What changed
Group the three sharing surfaces under one collapsible **"Sharing"** section (reusing
the existing **Data Sources** expandable pattern) in both the desktop sidebar and the
mobile drawer:

- **`nav-links.ts`** — split `NAV_LINKS` into `NAV_LINKS_PRIMARY` / `NAV_LINKS_SECONDARY`
  and add `SHARING_LINKS` (Feed, Dashboards, Challenges) + `SHARING_PATHS`. The bare
  **"Share"** label is gone; the dashboards item is now clearly **"Dashboards"**.
- **Sidebar** — renders the Sharing section in its current mid-list position, **open by
  default** so the surfaces stay one click away. When the sidebar is collapsed, the three
  render as individual icon links (each is a real page).
- **Header (mobile drawer)** — the same grouped, open-by-default section.
- **Routes unchanged** (`/feed`, `/shared-dashboards`, `/challenges`) — bookmarks safe.
- **docs** — the "using it in the web app" notes in sharing/feed/challenges point at the
  new Sharing section.

### Resulting sidebar
```
🏠 Home · 🎯 Goals · 📅 Timeline · 📊 Data · 🍽 Meals · 📝 Reports · ➕ +Add · 📈 Chart
🌐 Sharing ▴
   📣 Feed          → /feed
   🔆 Dashboards    → /shared-dashboards
   🏆 Challenges    → /challenges
🔬 Analyze · 📍 Places · 🏷 Types · 🔗 Rules
📡 Data Sources ▾ · ⚙️ Admin
```

### Scope note
Frontend-only (no backend/api-spec changes). The "Share to feed" activity button and the
native-share "Share" on public pages are each contextually clear and left as-is. Per the
chosen option ("Grouped Sharing section"), this keeps all existing routes rather than
introducing a hub page.

### Checks
- web `tsgo` clean; oxlint clean; **web 499 tests pass**.

### DIFF
--- a/apps/web/src/components/Sidebar.tsx
+++ b/apps/web/src/components/Sidebar.tsx
@@ -2,8 +2,16 @@ import { useLocation } from 'preact-iso'
 import { useState } from 'preact/hooks'
 
 import { auth, logout } from '../state/auth'
-import { DATA_SOURCE_LINKS, NAV_LINKS } from './nav-links'
+import {
+  DATA_SOURCE_LINKS,
+  NAV_LINKS_PRIMARY,
+  NAV_LINKS_SECONDARY,
+  SHARING_LINKS,
+  SHARING_PATHS,
+} from './nav-links'
 import './Sidebar.css'
+
+const isSharingActive = (url: string) => SHARING_PATHS.some((p) => url === p || url.startsWith(`${p}/`))
 
 const COLLAPSED_KEY = 'sidebar-collapsed'
 
@@ -15,6 +23,8 @@ export function Sidebar() {
 
   const [collapsed, setCollapsed] = useState(() => localStorage.getItem(COLLAPSED_KEY) === '1')
   const [dsExpanded, setDsExpanded] = useState(() => url.startsWith('/data-sources'))
+  // Sharing group is open by default so its surfaces stay one click away.
+  const [sharingExpanded, setSharingExpanded] = useState(true)
 
   const toggleCollapsed = () => {
     const next = !collapsed
@@ -30,6 +40,7 @@ export function Sidebar() {
   }
 
   const isDataSourcesActive = url.startsWith('/data-sources')
+  const sharingActive = isSharingActive(url)
 
   return (
     <aside class={`sidebar${collapsed ? ' collapsed' : ''}`}>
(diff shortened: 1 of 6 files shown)


apps/web/src/components/Sidebar.tsx
```diff
@@ -2,9 +2,17 @@ import { useLocation } from 'preact-iso'
...
+import {
+  DATA_SOURCE_LINKS,
+  NAV_LINKS_PRIMARY,
+  NAV_LINKS_SECONDARY,
+  SHARING_LINKS,
+  SHARING_PATHS,
+} from './nav-links'
 import './Sidebar.css'
 
+const isSharingActive = (url: string) => SHARING_PATHS.some((p) => url === p || url.startsWith(`${p}/`))
```
fiddur: Non-blocking: this exact predicate is duplicated in `Header.tsx` (inline `const isSharingActive = SHARING_PATHS.some(...)`). Since `SHARING_PATHS` already lives in `nav-links.ts`, consider exporting the `isSharingActive` helper from there so both nav components can't drift.

fiddur (commented):
✅ Approved

Clean, well-scoped navigation refactor that groups Feed / Dashboards / Challenges under a collapsible "Sharing" section in both the sidebar and the mobile drawer.

Verified:
- The old flat `NAV_LINKS` (15 entries) is preserved in full across `NAV_LINKS_PRIMARY` (8) + `SHARING_LINKS` (3) + `NAV_LINKS_SECONDARY` (4) — no link is dropped, and the removed `NAV_LINKS` export has no remaining references anywhere in the repo.
- All CSS classes used by the new markup (`sidebar-section-btn`, `sidebar-sub-links`, `sidebar-section-arrow`, `drawer-section-toggle`, `drawer-sub-link`, `dropdown-arrow`) already exist — they are reused from the existing Data Sources section, so styling and the collapsed-sidebar behaviour stay consistent.
- The Header drawer correctly filters Home out of `NAV_LINKS_PRIMARY` (`l.href !== '/'`) so it isn't duplicated with the separately-rendered Home link.
- Active-state handling is consistent with prior behaviour: the section header uses `startsWith` (so it highlights on sub-routes) while individual sub-links keep exact `url === link.href` matching, matching how the flat links behaved before.
- Docs (challenges.md, feed.md, sharing.md) are updated to reflect the new "Sharing" grouping; the remaining doc churn is prettier markdown reformatting.

One small, non-blocking observation: the `isSharingActive` predicate is duplicated between `Sidebar.tsx` (module-level fn) and `Header.tsx` (inline const). Since `SHARING_PATHS` already lives in `nav-links.ts`, co-locating the predicate there would keep the two nav components in sync. Not a blocker.

I could not run `pnpm check` in this sandbox (the monorepo install fails on a read-only pnpm store), but the change is a small, self-consistent set of export renames + additions with no type-level surprises.

apps/web/src/components/Sidebar.tsx
fiddur: Done in 84e1bdaa — moved `isSharingActive` next to `SHARING_PATHS` in `nav-links.ts` and imported it in both Sidebar and Header, so the predicate has a single source of truth.

fiddur (commented):
✅ Approved

Clean, well-executed navigation refactor. The three sharing surfaces (feed, shared dashboards, challenges) are grouped under a single collapsible "Sharing" section in both the sidebar and the mobile drawer, and the ambiguous top-level "Share" label is replaced by "Dashboards".

Verified:
- **No links lost or duplicated.** All 15 entries from the old `NAV_LINKS` are preserved across `NAV_LINKS_PRIMARY` / `SHARING_LINKS` / `NAV_LINKS_SECONDARY`, in the same order.
- **No dangling references.** Nothing still imports the removed `NAV_LINKS` export; `Header.tsx` and `Sidebar.tsx` are both updated.
- **CSS classes exist.** `sidebar-section-btn`, `sidebar-sub-links`, `sidebar-section-arrow`, `drawer-section-toggle`, `drawer-sub-link`, and `dropdown-arrow` are all reused from the existing Data Sources pattern.
- **`isSharingActive` is correct.** Requiring the trailing slash (`url.startsWith(`${p}/`)`) avoids the `/feed` → `/feedback` false-positive, and exact-match handles the surface pages themselves. The collapsed-sidebar path renders the three surfaces as plain links, mirroring the Data Sources collapsed behavior.
- **Docs** are mostly prettier-driven formatting (emphasis `*`→`_`, table reflow) plus accurate label/section updates ("Share" → "Dashboards", references to the new **Sharing** section).

Minor, non-blocking suggestions:
- `isSharingActive` is a new pure function with a small edge-case-y `startsWith` guard; a tiny unit test in `nav-links.test.ts` would be cheap given the existing web test infra and would match the project's "unit test first (when reasonable)" guidance.
- `SHARING_PATHS` is exported but only consumed internally by `isSharingActive`. Consider dropping the `export` unless it's intended for tests/other consumers.

Neither is a blocker.

PR #926, opened 2026-07-04
feat(feed): federate native structured charts between Aurboda peers (#884)

fiddur:
## What & why

When you view a followed **Aurboda** user's exercise post in your home timeline, it currently renders as static Mastodon-style HTML (text + a picture) — not the rich interactive chart the author sees on their own instance. That's because a `Note` delivered over ActivityPub only carries the Mastodon-compatible HTML: Fedify's typed vocab drops our `aurboda:` extension in transit.

This PR wires the full **Aurboda-to-Aurboda** round-trip so a peer's posts render natively.

## How

1. **Emit** — `GET /public/:username/feed/:postId` serves the native `FeedStructured` payload (typed scalar metrics + inline bucketed series). It reuses the *exact same* shared-scalar resolution as the object dispatcher and the *exact same* data-scoped series resolution as the public `/series` endpoint, so a peer can never render more than the author actually shared. Only `public`/`unlisted` posts resolve; `Cache-Control: no-store`.
2. **Ingest enrichment** — on inbound `Create`/`Update`, if the note id matches Aurboda's own object-dispatcher path (`/users/{user}/feed/{postId}`, a Mastodon status never matches → no needless fetch), we discover the origin's federation metadata and fetch its structured payload through the **SSRF-guarded** `safeFetchGet`, bounded by a total timeout. Strictly best-effort: any failure (non-Aurboda host, 404, malformed body, timeout) resolves to `null` and the post still shows its HTML.
3. **Storage** — a nullable `structured` JSONB column on `timeline_entry`; the upsert `COALESCE`s so a re-delivery whose enrichment failed keeps the last-known chart rather than wiping it.
4. **Render** — the home timeline draws a native interactive `TrendLineChart` per shared series when structured data is present.

## Safety

- Enrichment is never allowed to fail ingest (every error → `null`).
- The fetch target is the accepted followee's already-validated host, through the SSRF guard (public hosts only, no redirects, size + time bounded).
- A peer only ever sees what the author shared — same resolution paths as existing public endpoints.

## Testing

- **Backend unit** — enrichment URL parsing + mapping (`timeline-enrich.test.ts`), timeline serialization.
- **Backend integration** — `structured` JSONB round-trip + COALESCE-on-redelivery (`timeline.integration.test.ts`), end-to-end structured resolution from a real activity + series (`feed-structured.integration.test.ts`).
- **Web unit** — `structuredChartSeries` mapping/labels/colours/2-point filter.
- `pnpm check` green; docs (`docs/features/feed.md`) updated.


fiddur (commented):
✅ Approved

Solid, well-tested feature that federates native structured charts between Aurboda peers. The privacy and security posture is careful and consistent with the existing federation surface, and I found no blocking issues.

**Strengths**
- **Privacy scoping is reused, not reinvented.** `resolveStructuredPost` gates on `isPubliclyVisible(post.visibility)` and resolves series through the same `resolvePublicSeries` + `findCoveringSharedSeriesWindow` path as the public `/series` endpoint (which filters `visibility IN ('public','unlisted')`), so a peer can never read more than the author shared.
- **SSRF-safe, authority-bound enrichment.** Fetches go through `safeFetchGet` (public hosts only, no redirects, 1MB cap), and the object URI's host is already bound to the accepted followee's host by `noteToTimelineInput` (`note.id.host === senderHost`), so a peer can only serve its own structured data — no content-spoofing across hosts.
- **Truly best-effort.** Parse → discover → fetch → schema-validate, with every failure path resolving to `null`, wrapped in a 12s `withTimeout` and a catch-all; enrichment can never fail or block basic ingest. The `COALESCE(EXCLUDED.structured, timeline_entry.structured)` upsert avoids wiping a working chart on a transient re-enrich failure.
- **Clean schema/migration work.** The migration is additive/idempotent (`ADD COLUMN IF NOT EXISTS`) and correctly ordered in `tableCreationOrder`; moving `publicSeriesSampleSchema` into a leaf `feed-structured.ts` avoids a circular import while preserving the package-root re-exports.
- **Good coverage.** Unit tests for URL parsing / enrichment branches / web mapping, plus integration tests for the resolver (public vs followers-only, opt-in series) and the JSONB round-trip + COALESCE behavior.

**Non-blocking considerations (already acknowledged in code comments — no change requested)**
- Enrichment runs *synchronously* in the inbox handler, so an Aurboda `Create`/`Update` can block up to `ENRICH_TIMEOUT_MS` (12s) on discovery+fetch. The upsert is idempotent so sender retries are harmless, and the code comment flags this as intentional for now; worth revisiting if inbox latency becomes a concern.
- Series are stored at the 5s floor with no upper bucket cap, bounded only by the 1MB `safeFetchGet` limit. Very long activities may exceed 1MB and silently fall back to the HTML — acceptable degradation, but a future coarser-bucket-for-long-windows tweak would let more posts federate a chart.
- The `COALESCE` keep-last-known means a public post later flipped to followers-only won't have its already-federated chart re-fetched (the structured endpoint 404s), so receivers retain the previously-public chart. Benign here (receivers are accepted followers, and `Delete`/series-removal-while-public do propagate), but noting the revocation edge for the record.

PR #927, opened 2026-07-04
fix(feed): gate Fedify middleware so large /sync bodies don't hang (Health Connect sync)

fiddur:
## Symptom

Health Connect sync from the Android app stopped landing data around **2026-07-03 13:40** for multiple users (me + Sara), with **nothing in the audit log or server log**. Garmin/Oura/Strava (server-side pulls) and `POST /activities` from computers kept working.

## Root cause

`@fedify/express`'s `integrateFederation` is mounted **before** `express.json()` (required, so Fedify sees the raw body of signed inbox POSTs). But its `fromERequest` wraps **every** non-GET request body with `Readable.toWeb(req)` — even for routes it doesn't own, like `/sync/:recordType` — and then calls `next()` without consuming that stream. `express.json()` then blocks forever waiting for body bytes that were siphoned into an abandoned web stream.

- Small bodies already sit in the socket buffer → parse fine (months of small HC syncs worked).
- Bodies past ~64 KB → **hang with no response** → `res.on('finish')` never fires → **no audit entry**.

Once a device's backlog produced a page > ~64 KB (the app's screenshot showed a **749-record / 156 KB** page), every retry hung and the backlog only grew — permanently stuck.

Confirmed by a live size sweep against prod (42 KB → fast 401; 156 KB → 30 s hang) and a local repro with the real `@fedify/express` (small → 200, large → hang, `raw-body: request aborted`).

## Fix

Gate the Fedify middleware to the paths it actually owns (`/users/*`, `/inbox`, `/.well-known/webfinger`, nodeinfo — a superset of every dispatcher/listener in `services/activitypub/federation.ts`, including the object/tombstone path `/users/{id}/feed/{postId}`). The inbox keeps raw-body access; every other route reaches `express.json()` untouched. A plain reorder won't do — `json` first makes `fromERequest` throw on the consumed stream and breaks inbox signature verification.

## Tests

- `isFederationRequest` / `gateFederation` unit coverage (owns vs skips).
- Regression: POSTs a **>64 KB** body through the **real** `integrateFederation` → gate → `json` chain in the production order and asserts 200 (hangs without the gate); plus a `/inbox` POST still routing to Fedify.

After deploy, backlogged devices self-heal on their next sync.

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -20,6 +20,7 @@ import { Client } from 'pg'
 import type { FeedDeliver } from './routes/feed-router.ts'
 
 import { registerAuthRoutes } from './api/auth-routes.ts'
+import { gateFederation } from './api/federation-gate.ts'
 import { createAdminMiddleware, createAuditLogMiddleware, createAuthMiddleware } from './api/middleware.ts'
 import { registerOAuthRoutes } from './api/oauth-routes.ts'
 import { mountRestRouters } from './api/rest-routes.ts'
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped fix for the Health Connect large-body sync hang.

**Root cause & fix are right.** `@fedify/express`'s `integrateFederation` wraps every non-GET body with `Readable.toWeb(req)` and `next()`s without consuming it, starving `express.json()` on any large non-federation POST. Gating the middleware to federation-owned paths cleanly resolves this while preserving raw-body access for signed inbox POSTs.

**Correctness checks I ran:**
- `isFederationRequest` is a superset of every dispatcher/listener path registered in `services/activitypub/federation.ts` (actor `/users/{id}` and its sub-collections inbox/outbox/followers/following/feed, the shared `/inbox`, and `/.well-known/webfinger`). The nodeinfo prefixes are defensive extras (no dispatcher registered → harmless passthrough).
- The gate only narrows what reaches Fedify; no non-federation POST route lives under the gated prefixes (only a GET OAuth-discovery `.well-known` route, which isn't gated), so nothing needing `express.json()` still hits the raw-body wrap.
- Non-federation routes don't rely on `integrateFederation` running per-request — the federation context is created directly via `federation.createContext(...)` in services, not read off `req`, so skipping it for non-federation paths is safe.
- The tombstone router (mounted after) still works: `/users/.../feed/...` matches the gate → passes through Fedify → `next()` → tombstone.
- The wrapper preserves the original async-middleware semantics.

**Tests** are appropriately layered: unit tests for path classification plus a real `@fedify/express` ordering regression that reproduces the >64 KB hang and confirms inbox POSTs still route to Fedify. (I couldn't execute them in this sandbox — no installed `node_modules` and read-only npm cache — but the logic is sound.)

Good inline documentation on the why. No blocking issues.

PR #928, opened 2026-07-04
feat(feed): follower approval flow + manual-approval setting (#884)

fiddur:
## What & why

You asked whether the public-vs-followers distinction meant you could **approve or reject followers**. This adds exactly that: a per-user setting to run either an **open account** (the default — anyone can follow) or a **locked account** where incoming follows become **requests you approve or reject**.

## How

**Setting** — `manually_approve_followers` (user setting, default `false`). When on, the actor document advertises `manuallyApprovesFollowers: true` (a "locked account"), so Mastodon and the wider fediverse show a follow *request* and hold it pending.

**Inbound `Follow`**
- Auto-accepted (immediate `Accept`) unless manual approval is on. A re-delivered Follow from an already-accepted follower never demotes them back to pending.
- In manual mode the follow is recorded in `feed_follower` as **pending** with the follower's cached handle / display name / avatar (so the approval UI shows *who* is asking) and the id of their Follow activity — **no** `Accept` is sent yet.
- Only **accepted** followers appear in the followers collection + count and receive `followers`-only posts (a pending request is not yet a follower).

**Approve / reject**
- **Approve** flips the row to accepted and sends the deferred `Accept`, echoing the original `Follow` id so the follower's server matches it to its pending request.
- **Reject** (or removing an already-accepted follower) sends a `Reject` and drops the row.
- Both are best-effort/synchronous, matching the rest of the feed's delivery model.

## Surface (REST + MCP parity)

- `GET /feed/followers?status=pending|accepted|all`, `POST /feed/followers/:id/approve`, `DELETE /feed/followers/:id`.
- MCP `list_followers`, `approve_follower`, `reject_follower`.
- **Web** — a **Followers** panel on the Feed page (pending requests with Approve/Reject, accepted followers with Remove) and a **Manually approve new followers** toggle in Settings → *Feed & Followers*.

## Notable decisions

- **Default is auto-accept** — non-breaking; preserves today's open-account behaviour. Switching the setting off later does not retroactively accept already-pending requests.
- Shared `extractActorPresentation` (used by both follow directions) and a `with-timeout` leaf util were extracted from `following.ts` to stay DRY and cycle-free.
- `feed_follower` columns (`id`, `handle`, `display_name`, `avatar_url`, `follow_activity_uri`) are added with idempotent `ADD COLUMN IF NOT EXISTS`.

## Testing

- **DB integration** — round-trip, status filter (pending/accepted), approve, remove-by-id, keep-last-known presentation on re-follow.
- **Service integration** — approve/reject DB effects with best-effort delivery failing gracefully to a non-resolving inbox.
- **Federation integration** — actor advertises `manuallyApprovesFollowers` when enabled; followers collection + count are accepted-only (pending excluded).
- **Units** — follower serializer (no inbox leak) + `Follow` reconstruction (echoes / omits the original id).
- `pnpm check` green (0 errors); backend + web suites green; docs (`docs/features/feed.md`) updated.

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -72,6 +72,12 @@ import {
 import { createDetectionTrigger, type DetectionTrigger } from './services/detection-trigger.ts'
 import { runDetectionForUser } from './services/detection-worker.ts'
 import { expandFeedActivityWindow, resolveFeedActivity } from './services/feed.ts'
+import {
+  approveFollower,
+  type FollowerActions,
+  rejectFollower,
+  serializeFollower,
+} from './services/followers.ts'
 import { type FollowActions, followActor, unfollowActor } from './services/following.ts'
 import { createGeocodeQueue } from './services/geocode-queue.ts'
 import { createInvitationAuth } from './services/invitation.ts'
@@ -356,6 +362,15 @@ const main = async () => {
     follow: (user, handle) => followActor(feedDeps, user, handle),
     unfollow: (user, id) => unfollowActor(feedDeps, user, id),
   }
+  // The follower-management operations (approve/reject a follow request), sharing
+  // the same federation + origin. Approve returns the serialised follower.
+  const followerActions: FollowerActions = {
+    approve: async (user, id) => {
+      const record = await approveFollower(feedDeps, user, id)
+      return record ? serializeFollower(record) : null
+    },
+    reject: (user, id) => rejectFollower(feedDeps, user, id),
+  }
 
   // Mount MCP server BEFORE body-parser (MCP SDK needs raw body)
   // Stateless mode — no session tracking needed (tools only, no subscriptions)
@@ -368,6 +383,7 @@ const main = async () => {
       engineDeps,
       feedDeliver,
       followActions,
+      followerActions,
       garmin,
       onActivityMutated: activityNotifier,
       oura,
@@ -503,6 +519,7 @@ const main = async () => {
     engineDeps,
     feedDeliver,
     followActions,
+    followerActions,
     garmin,
     httpd,
     invitationAuth,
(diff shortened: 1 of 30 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped implementation of the follower-approval / manual-approval (locked account) flow. The design is coherent end-to-end and I found no blocking issues.

**What I verified**

- **Schema migration is sound.** The additive `feed_follower_columns` entry follows the existing non-table pattern (`_seed`/`_setup`/`_trigger`) — it's keyed by a name that never matches a real table, so the loop in `db/connection.ts` always runs it, and every statement is idempotent (`ADD COLUMN IF NOT EXISTS`, `CREATE UNIQUE INDEX IF NOT EXISTS`). `gen_random_uuid()` is volatile, so back-filling `id` on a pre-existing table gives each row a distinct UUID (the unique index then succeeds). On a fresh DB the `CREATE UNIQUE INDEX IF NOT EXISTS feed_follower_id_key` matches the `CREATE TABLE`'s auto-named UNIQUE-constraint index and is a clean no-op. Multi-statement SQL strings are established precedent (`food_items_search_setup`).
- **Acceptance logic is correct.** `recordInboundFollow` keeps an already-accepted follower accepted on re-delivery (never demotes), auto-accepts when manual approval is off, and holds pending otherwise; the inbox handler only sends the `Accept` when `accepted === true`. Only accepted followers appear in the followers collection + count and receive `followers`-only delivery. `getUserSettings`'s `null` return is handled in both the actor dispatcher and the inbound path.
- **Security.** The owner-facing `FollowerActor` DTO deliberately omits inbox / shared-inbox / follow-activity URIs, and that exclusion is unit-tested.
- **Parity + wiring.** New MCP tools (`list_followers` / `approve_follower` / `reject_follower`) mirror the REST endpoints (`GET /feed/followers`, `POST /feed/followers/:id/approve`, `DELETE /feed/followers/:id`), both backed by the shared `followers.ts` service. Schemas are exported from the api-spec root, the web api barrel re-exports the new calls, and the settings field flows through `updateSettingsInputSchema.shape` and `withDefaults`. Docs are updated thoroughly.
- **Tests** are comprehensive (db integration, service integration + unit, federation integration for locked-account advertising and accepted-only collection).

**Minor, non-blocking note:** inbound `Follow` handling now runs `extractActorPresentation` (a timeout-bounded, ≤3s icon deref) synchronously on every Follow / re-delivery. This is intentional and bounded per the module doc (and mirrors the outbound follow path), so it's acceptable — just be aware re-deliveries from servers that only *link* their icon incur the fetch each time.

PR #929, opened 2026-07-13
fix(feed): show delivered image attachments in the home timeline (#884)

fiddur:
## What & why

Follow-up from tryout (Sara followed you): a **followers-only sleep post with an HR chart showed nothing** in the aurboda home timeline — even though Mastodon rendered the chart image.

Two gaps caused it:
- **Image dropped on ingest** — `noteToTimelineInput` kept only the sanitised text `content` (the sanitiser even strips ``), so the delivered image attachment was thrown away.
- **Native chart is public-only** — structured-chart federation gates on `public`/`unlisted`; a `followers`-only share federates no structured data → no native chart either.

Net: a followers-only post had neither. This slice fixes the **image** path (native charts for followers-only is the next slice).

## How

- **`extractNoteImages`** pulls image attachments off the received Note — our own rendered chart / route-map `Image`s *and* Mastodon-style `Document`s with an `image/*` media type. Only inline `http(s)` images survive (url + optional media type / alt / intrinsic size); anything else is skipped.
- Stored on a nullable **`timeline_entry.images`** JSONB column (additive `ADD COLUMN IF NOT EXISTS`), `COALESCE`d on re-delivery like `structured`.
- The home timeline card renders the **native structured chart when present, else the delivered image(s)** — matching the rule *"native chart if the data is shared, image otherwise if present."* So a followers-only share still shows its chart **image**, and a Mastodon photo post shows its photos.

## Testing

- **Unit** — `extractNoteImages` extracts a rendered `Image` (url/media-type/alt/size) and a Mastodon `Document`, and skips non-image / non-`http(s)` / no-attachment cases.
- **Integration** — `images` JSONB round-trip (stored + read back; plain posts are null).
- `pnpm check` green (0 errors); backend + web suites green; docs updated.

## Follow-up slices (tracked)

2. Native charts for followers-only posts (serve structured to accepted followers via the image-token capability).
3. Unify chart-image + series sharing (one control, "same data two formats").
4. Public profile feed + navigation links.
5. Home timeline backfill of a followee's older public posts.

### DIFF
--- a/apps/backend/src/db/timeline.ts
+++ b/apps/backend/src/db/timeline.ts
@@ -85,6 +85,10 @@ export const upsertTimelineEntry = async (
                    -- couldn't re-fetch it (transient enrich failure), rather
                    -- than wiping a working chart.
                    structured = COALESCE(EXCLUDED.structured, timeline_entry.structured),
+                   -- images always arrives as a concrete array from the ingest
+                   -- path, so this COALESCE never actually preserves a prior value
+                   -- (an edit that drops attachments clears them) -- it is defensive
+                   -- parity with structured for any caller that omits the field.
                    images = COALESCE(EXCLUDED.images, timeline_entry.images)
      RETURNING ${TIMELINE_COLUMNS}, (xmax = 0) AS inserted`,
     [
--- a/apps/backend/src/db/timeline.ts
+++ b/apps/backend/src/db/timeline.ts
@@ -8,7 +8,7 @@
  * responsible for cleaning untrusted fediverse HTML before it reaches here).
  * Ordering + pagination is keyset by `(published_at DESC, id DESC)`.
  */
-import type { FeedStructured } from '@aurboda/api-spec'
+import type { FeedStructured, TimelineImage } from '@aurboda/api-spec'
 
 import { query } from './connection.ts'
 
@@ -25,6 +25,8 @@ export interface TimelineEntryRecord {
   received_at: Date
   /** Native structured payload from an Aurboda peer, or null for non-Aurboda posts. */
   structured: FeedStructured | null
+  /** Image attachments (rendered chart / route map, or a Mastodon photo), or null. */
+  images: TimelineImage[] | null
 }
 
 export interface TimelineEntryInput {
@@ -39,6 +41,8 @@ export interface TimelineEntryInput {
   published_at: Date
   /** Native structured payload fetched from an Aurboda peer on ingest, if any. */
   structured?: FeedStructured | null
+  /** Image attachments captured from the delivered Note, if any. */
+  images?: TimelineImage[] | null
 }
 
 /** Opaque keyset cursor: the last row's `(published_at, id)`. */
@@ -48,7 +52,7 @@ export interface TimelineCursor {
 }
 
 const TIMELINE_COLUMNS =
-  'id, object_uri, actor_uri, handle, display_name, avatar_url, content, url, published_at, received_at, structured'
+  'id, object_uri, actor_uri, handle, display_name, avatar_url, content, url, published_at, received_at, structured, images'
 
 /**
  * Insert or update a received post by `object_uri`. A re-delivered or edited post
@@ -67,8 +71,8 @@ export const upsertTimelineEntry = async (
   const result = await query<TimelineEntryRecord & { inserted: boolean }>(
     user,
     `INSERT INTO timeline_entry
-       (object_uri, actor_uri, handle, display_name, avatar_url, content, url, published_at, structured)
-     VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
+       (object_uri, actor_uri, handle, display_name, avatar_url, content, url, published_at, structured, images)
+     VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
      ON CONFLICT (object_uri)
      DO UPDATE SET actor_uri = EXCLUDED.actor_uri,
                    handle = EXCLUDED.handle,
@@ -80,7 +84,8 @@ export const upsertTimelineEntry = async (
                    -- Keep the last-known structured payload if a refresh/edit
                    -- couldn't re-fetch it (transient enrich failure), rather
                    -- than wiping a working chart.
-                   structured = COALESCE(EXCLUDED.structured, timeline_entry.structured)
+                   structured = COALESCE(EXCLUDED.structured, timeline_entry.structured),
+                   images = COALESCE(EXCLUDED.images, timeline_entry.images)
      RETURNING ${TIMELINE_COLUMNS}, (xmax = 0) AS inserted`,
     [
       input.object_uri,
@@ -92,6 +97,7 @@ export const upsertTimelineEntry = async (
       input.url ?? null,
       input.published_at,
       input.structured == null ? null : JSON.stringify(input.structured),
+      input.images == null ? null : JSON.stringify(input.images),
     ],
   )
   return result.rows[0]
--- a/apps/backend/src/services/activitypub/timeline-ingest.ts
+++ b/apps/backend/src/services/activitypub/timeline-ingest.ts
@@ -15,8 +15,10 @@
  * The inbox handler that calls these (in `federation.ts`) is thin: it checks the
  * sender is a followed, accepted actor and upserts the result.
  */
+import type { TimelineImage } from '@aurboda/api-spec'
 import type { Note } from '@fedify/fedify/vocab'
 
+import { Document, Image, Link } from '@fedify/fedify/vocab'
 import sanitizeHtml from 'sanitize-html'
 
 import type { FeedFollowingRecord, TimelineEntryInput } from '../../db/index.ts'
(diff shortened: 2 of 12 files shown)


fiddur (commented):
🛑 Changes required

Solid, well-scoped change: new `TimelineImage` schema, an additive `images` JSONB column + migration, the `extractNoteImages` ingest helper, serializer normalization (empty array → omitted, matching `structured`), the web fallback render (`structured` ? native chart : delivered images), docs, and both unit + integration tests all line up cleanly. The attachment-typing logic (`Image` subtype vs. `Document` with `image/*` media type, http(s)-only URL guard, Link-vs-URL handling) is careful and the tests cover the interesting branches.

One robustness concern worth addressing before merge: `extractNoteImages` calls `note.getAttachments()` without `{ suppressError: true }`, which breaks the file's stated best-effort / never-500 ingest invariant for referenced or cross-origin attachments (details inline). The common paths — Aurboda's own chart `Image`s and inline Mastodon photos (both `id`-less, so never dereferenced) — are unaffected, so this only bites unusual/adversarial attachment shapes, but the surrounding code (`getObject({ suppressError: true })`, the `isMissingDatabase` guard) deliberately hardens exactly this path.

Minor, non-blocking: the `images = COALESCE(EXCLUDED.images, timeline_entry.images)` "keep last-known" behavior is effectively inert on the federation path, since `extractNoteImages` always returns an array (never null) — so an edit that drops attachments correctly wipes them. That's fine, just note the COALESCE (unlike the `structured` one it mirrors) never actually preserves a prior value here.

apps/backend/src/services/activitypub/timeline-ingest.ts
```diff
@@ -105,3 +107,45 @@ export const noteToTimelineInput = (
...
+/**
+ * Extract a received Note's image attachments (rendered chart / route map, or a
+ * Mastodon photo) so the home timeline can show them — the fallback when a post
+ * carries no native structured chart. Best-effort: only inline http(s) images
+ * are kept; anything else is skipped. Non-image and malformed attachments (and a
+ * Note with none) yield an empty array.
+ */
+export const extractNoteImages = async (note: Note): Promise<TimelineImage[]> => {
+  const images: TimelineImage[] = []
+  for await (const att of note.getAttachments()) {
```
fiddur: `getAttachments()` is called without `{ suppressError: true }`. For an attachment that is a bare URI string or an embedded object with a cross-origin `id`, Fedify dereferences it over the network; without `suppressError` a failed dereference (unreachable host, timeout, or `validatePublicUrl` rejecting a private-IP URL) **throws**. That propagates out of `extractNoteImages` → `ingestFeedActivity`, and since it isn't a missing-DB error it's re-thrown, 500ing the inbox and inviting redelivery retries — contradicting the adjacent `getObject({ suppressError: true })` and this module's "best-effort … never 500s" intent. The documented cases (Aurboda's `id`-less chart `Image`s, inline Mastodon photos) never fetch, so this is edge-case, but pass `{ suppressError: true }` here to keep ingest resilient (it makes a failed fetch skip the attachment instead of throwing).

fiddur: Fixed in 3a80161a — `getAttachments({ suppressError: true })` so a failed attachment dereference yields null (skipped) instead of throwing out of the best-effort ingest path.

apps/backend/src/db/timeline.ts
```diff
@@ -80,7 +84,8 @@ export const upsertTimelineEntry = async (
                    -- Keep the last-known structured payload if a refresh/edit
                    -- couldn't re-fetch it (transient enrich failure), rather
                    -- than wiping a working chart.
-                   structured = COALESCE(EXCLUDED.structured, timeline_entry.structured)
+                   structured = COALESCE(EXCLUDED.structured, timeline_entry.structured),
+                   images = COALESCE(EXCLUDED.images, timeline_entry.images)
```
fiddur, line 92: Minor semantic note: this COALESCE can never actually fall back to `timeline_entry.images`. Unlike `structured` (which the caller sets to `null` on a transient enrich failure, so the COALESCE genuinely preserves the last-known chart), `extractNoteImages` always returns a concrete array and `federation.ts` always passes it — so `EXCLUDED.images` is never SQL NULL. Net effect: an edit/redelivery that drops all attachments clears them (arguably the correct behaviour, since images come from the authoritative Note payload rather than a separate fallible fetch). Worth a one-line comment noting the asymmetry, or dropping the COALESCE if the preserve semantics aren't intended here.

fiddur (commented):
✅ Approved

Solid, well-scoped change that renders delivered image attachments as the timeline fallback when a post carries no native structured chart. It follows the project conventions cleanly: the `TimelineImage` schema lives in `api-spec` with `.meta({ id })`, the DB column is added both to the CREATE and as an idempotent additive `ALTER ... ADD COLUMN IF NOT EXISTS` (registered in `tableCreationOrder`), serialization mirrors the existing `structured` omission pattern, and docs are updated. Security posture is good: extraction restricts to inline `http(s)` `Document`/`Image` attachments, `getAttachments({ suppressError: true })` preserves the never-500 ingest invariant, and rendering only ever gets `http(s)` URLs (matching the existing `FeedPostCard` remote-image pattern). Unit + integration coverage is thorough (image extract, Mastodon-Document keep, non-image/non-http skip, empty case, JSONB round-trip).

Non-blocking observations:

- **Unbounded attachment count.** `extractNoteImages` keeps every qualifying attachment with no cap, so an accepted-but-hostile followee could deliver a Note with a large number of image attachments that all get stored in the JSONB column and rendered in one card. Mastodon caps at 4; a small `.slice()` limit would harden this.
- **`http:` attachments** are accepted by `httpUrl` but will be blocked as mixed content when rendered on the https site, so they silently won't display. Not harmful, just dead weight — consider https-only for the stored ``.
- See the inline note on the `images` COALESCE for a minor semantic asymmetry with `structured`.

apps/backend/src/db/timeline.ts
fiddur: Kept the COALESCE for defensive parity with structured, and added a comment noting it never actually preserves on the ingest path (extractNoteImages always returns a concrete array, so an edit that drops attachments correctly clears them). 4da52cb8.

apps/backend/src/services/activitypub/timeline-ingest.ts
```diff
@@ -105,3 +107,57 @@ export const noteToTimelineInput = (
...
+ * hostile followee could make us store + render in one card. */
+const MAX_TIMELINE_IMAGES = 4
+
+export const extractNoteImages = async (note: Note): Promise<TimelineImage[]> => {
+  const images: TimelineImage[] = []
+  // `suppressError` so a referenced attachment whose dereference fails (unreachable
+  // host, timeout, or `validatePublicUrl` rejecting a private-IP URL) yields null
+  // instead of throwing out of the best-effort ingest path (never-500 invariant).
+  // Inline attachments — our own chart `Image`s and Mastodon photos — don't deref.
+  for await (const att of note.getAttachments({ suppressError: true })) {
```
fiddur, line 164: Non-blocking hardening: `MAX_TIMELINE_IMAGES` bounds how many images are *stored/rendered*, but the `break` only fires once 4 valid images are collected. A Note whose attachments are mostly non-images or unreachable *reference* URIs never hits the break, so all of them get iterated (and reference-style ones dereferenced) before the loop ends — a small delivery could amplify into many sequential outbound fetches during ingest. Consider also capping the total number of attachments examined (e.g. `let seen = 0; if (++seen > MAX_ATTACHMENTS) break`) so the cap bounds work, not just output. Gated behind an accepted follow and the common inline case doesn't deref, so low severity.

fiddur (commented):
✅ Approved

Solid, well-scoped change that renders delivered image attachments as the fallback when a post has no native structured chart. The design is clean and the invariants are respected:

- **XSS/mixed-content safe**: `httpsUrl` restricts to `protocol === 'https:'`, so `javascript:`/`data:`/`http:` URLs can never reach the ``. The URL is only ever used as an attribute value (Preact-escaped), and `name` is used as `alt`.
- **never-500 invariant preserved**: `extractNoteImages` uses `getAttachments({ suppressError: true })` and runs inside the existing try/catch, so a failed attachment deref yields `null` rather than throwing out of the ingest path.
- **API/MCP parity**: the new `images` field flows through `serializeTimelineEntry`, so both the REST `GET /feed/timeline` and the MCP `list_timeline` tool pick it up automatically. Schema lives in `packages/api-spec` as the single source of truth, with an additive `ADD COLUMN IF NOT EXISTS` migration wired into `tableCreationOrder`.
- **Good test coverage**: the unit tests exercise the chart Image case, the non-image/non-https/data-URL skips, the 4-image cap, and the empty case; the integration test round-trips the JSONB column. The `COALESCE(EXCLUDED.images, …)` behaviour is honestly documented (an edit that drops attachments clears them, since ingest always passes a concrete array).

No blocking issues. Two minor, non-blocking notes:

1. **Attachment-iteration bound** (see inline): `MAX_TIMELINE_IMAGES` caps stored/rendered images but not how many attachments are *examined*. A hostile-but-accepted followee could deliver a Note with many non-image or unreachable *reference* attachments so the `break` never fires, turning one delivery into many sequential outbound derefs. Low severity (gated behind an approved follow; the common case is inline attachments that don't deref), but capping total iterations too would fully realise the cap's DoS-bounding intent.
2. The JSDoc block for `extractNoteImages` (timeline-ingest.ts) is separated from the function by the `MAX_TIMELINE_IMAGES` const + its own comment, so editors/hover won't associate the doc with the function — worth moving the const above the doc block.

apps/backend/src/services/activitypub/timeline-ingest.ts
fiddur: Fixed — added MAX_ATTACHMENTS_SCANNED (20) so iteration/derefs are bounded even when no valid image is ever collected.

fiddur (commented):
✅ Approved

Solid, well-scoped change that adds image-attachment capture to the timeline. The security posture is careful and the tests cover the important cases (Image chart, Mastodon Document photo, non-image/non-https/data skips, the 4-image cap, the empty case, and JSONB round-trip in the integration test). A few observations, none blocking:

**Correctness / safety — verified OK:**
- The never-500 invariant holds. `getAttachments({ suppressError: true })` swallows dereference failures, and the singular `att.url` accessor returns the first value rather than throwing on multi-valued `url` (per Fedify's documented singular-accessor behavior), so a PeerTube-style attachment with multiple `url` representations won't blow up the inbox. The `MAX_ATTACHMENTS_SCANNED`/`MAX_TIMELINE_IMAGES` caps correctly bound both dereference amplification and stored/rendered images.
- SSRF for reference-URI attachments is covered: `createFederation()` is called with defaults (no `allowPrivateAddress` override), so `getAttachments()` dereferences through Fedify's built-in SSRF-protected document loader. The comment's `validatePublicUrl` claim is accurate.
- The `images` COALESCE-clears-on-edit semantics match the inline comment: the ingest path always supplies a concrete array (empty `[]` when no attachments), so `COALESCE(EXCLUDED.images, prior)` never preserves a stale value, and `serializeTimelineEntry` treats `null` and `[]` identically (omitted). Consistent with the `structured` handling.
- Frontend is safe: https-only filtering guarantees `` is always an https URL (no `javascript:`/`data:`), and `.feed-post-image` CSS (`max-width:100%; height:auto`) neutralizes hostile width/height attributes.
- The generated OpenAPI/Kotlin artifacts are gitignored and rebuilt in CI from the Zod schema, so there's no stale-artifact / missed-regeneration concern; the new `TimelineImage` model and `images` field will flow to the OpenAPI spec and Android models on build.

**Non-blocking considerations:**
- Privacy: images are hotlinked directly from the author's origin instance (``), which leaks the viewer's IP/User-Agent to that instance on every render. Mastodon actually *proxies/caches* remote media through its own server for exactly this reason, so the doc/comment phrasing "the way Mastodon shows them" is slightly misleading — worth a caveat in `docs/features/feed.md`, or a follow-up to proxy media if IP-leak matters to you.
- Minor readability: in `timeline-ingest.ts`, the descriptive JSDoc block for `extractNoteImages` is now separated from the function by the `MAX_TIMELINE_IMAGES` / `MAX_ATTACHMENTS_SCANNED` declarations, so it reads as documenting the constant instead of the function. Consider hoisting the two `const`s above the function's doc block.

PR #930, opened 2026-07-13
feat(feed): federate native charts to accepted followers (#884)

fiddur:
## Summary

Second slice of the #884 tryout follow-ups (after #929 timeline media rendering). Sara followed the user and got a sleep activity shared with an HR chart, but her aurboda timeline showed **neither** the chart image **nor** a native chart — because the native structured chart was **public-only**. This slice federates the native chart to **accepted followers** too, so a follower sees the interactive chart instead of only the flat image.

## What changed

Authorization reuses the post's **existing capability token** — the same unguessable `image_token` that already gates a followers-only post's chart/route images, embedded only in the `Note` delivered to followers.

- **Shared authz** — extracted `isCapabilityAuthorized` into `services/feed-capability.ts` so the media and structured endpoints enforce one identical boundary (`public`/`unlisted` always; `followers`-only only with a matching token). The feed-image router now delegates to it instead of its own copy.
- **Structured resolver** — `resolveStructuredPost(user, postId, token?)` resolves a followers-only post only with a matching token. Its series are scoped by the post's own activity window (the post itself is the authorization boundary), bypassing the public-only covering-window check that would otherwise blank the chart for an authorized follower.
- **Route** — `GET /public/:username/feed/:postId` reads `?token=` and forwards it.
- **Ingest** — the enricher lifts the token from the delivered image URL (`capabilityTokenFrom`) and passes it to the structured fetch, so an accepted follower's instance fetches the chart it may already see the image of. A public guess (no/wrong token) still 404s.

Best-effort and additive as before: any failure falls back to the HTML + image.

## Tests

- `timeline-enrich.test.ts` — `?token=` forwarding + `capabilityTokenFrom` (lifts token, skips malformed URLs, undefined for public/no images).
- `feed-structured.integration.test.ts` — a followers-only post is null without a token and with a wrong token, and resolves (metrics + series) with `post.image_token`.
- Full `pnpm check` clean; affected backend suite green (52 tests).

## Docs

`docs/features/feed.md` — the native-charts (structured enrichment) and card-rendering sections now describe followers-only federation via the capability token.

### DIFF
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -1,4 +1,5 @@
 import { type Response, Router } from 'express'
+
 /**
  * Public feed-post image endpoints (UNAUTHENTICATED).
  *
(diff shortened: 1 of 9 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped extension of followers-only feed federation: an accepted follower's instance can now fetch and render the *native* structured chart (not just the flat image) by reusing the post's `image_token` as a capability token, lifted from the delivered image URL.

I traced the security-sensitive parts and they hold up:

- **Shared auth (`feed-capability.ts`)** correctly factors the previously-inline image-router logic (`isImageAuthorized`/`tokenMatches`) into `isCapabilityAuthorized`, now used by both the media and structured endpoints so they enforce identically. Constant-time compare preserved; the structural param type is satisfied by `FeedPostRecord`.
- **The `findCoveringWindow` bypass in `resolveStructuredSeries`** (the highest-risk change) is safe: the iterated series list is `post.series_metrics` (the author's opt-in) and the window is the post's own activity window, so the metric list — not the covering-window lookup — is the scoping boundary. The `omits series the post did not opt into` test (empty `series_metrics` → empty series) confirms nothing leaks. The unauthenticated public `/series` endpoint still routes through `findCoveringSharedSeriesWindow`, so its privacy boundary is untouched.
- **`capabilityTokenFrom`** forwards the lifted token only to the `object_uri`'s Aurboda origin (SSRF-guarded, non-Aurboda URIs short-circuit before the token is used), and public posts carry no token — I could not construct a token-leak or cross-origin scenario.

Test coverage is thorough (public resolution, opt-in scoping, followers-only gating with no/wrong/correct token, and `capabilityTokenFrom` edge cases incl. malformed URL and URL-encoding), and the docs are updated consistently. Query-param handling (`typeof req.query.token === 'string'`) correctly rejects array-valued `token`. No blocking issues.

PR #931, opened 2026-07-13
feat(feed): unify the heart-rate chart image and series into one share control (#884)

fiddur:
## Summary

Third slice of the #884 tryout follow-ups. Your note when scoping this work:

> Sharing a chart could be the same checkbox as sharing the corresponding data series in detail, it's the same data in two formats.

The heart-rate **chart image** (`include_chart`, a rendered PNG for Mastodon) and the heart-rate **series** (`series_metrics ∋ heart_rate`, which — since Slice 2 (#930) — renders a native interactive chart for Aurboda followers) are exactly that: one dataset in two formats. This slice fuses them in the share dialog into **one control**.

## What changed (web share dialog only)

- Opting into the **heart-rate series** now also attaches the **chart image** — `include_chart` is *derived* from the heart-rate toggle, never set separately. So one decision covers both audiences (native chart for peers, PNG for Mastodon), and the two formats can't diverge.
- The separate **"Heart-rate chart"** checkbox is gone from the dialog's *Images* section (now just the route map). The *Share full time-series* note explains heart rate is also shared as a chart image.
- Two pure helpers extracted into `feed-metrics.ts` and unit-tested (no component-render mocking):
  - `buildShareBody(selection)` — derives the request body; `include_chart = canChart && series.has('heart_rate')`, so the image is suppressed when the activity has no HR data to render.
  - `initialSeriesSelection(post)` — on **edit**, folds a post that attached the chart image but predates this control (no `heart_rate` in `series_metrics`) into the heart-rate series, so the unified control shows checked and stays toggleable.

## Scope / behaviour note

- The REST API and MCP `share_activity`/`update_feed_post` fields (`include_chart`, `series_metrics`) stay **independent** — a programmatic caller keeps the granular control. Only the web dialog fuses them.
- **Editing a legacy image-only post** (chart image attached, series never opted in) and saving will now *also* share the heart-rate series (the image already implied that data). This is the intended "same data, two formats" unification, but it is a small exposure increase on re-save — flagging it explicitly. Such posts only exist from the last few days of tryouts.

## Tests

`feed-metrics.test.ts` — 12 pass, incl. new coverage for `buildShareBody` (image attached iff HR series shared; image suppressed when no HR data; offered-metric filtering; map gated on GPS) and `initialSeriesSelection` (canonical order, chart-image fold, no duplicate, empty). Full `pnpm check` clean.

### DIFF
--- a/apps/web/src/components/ShareActivityDialog.tsx
+++ b/apps/web/src/components/ShareActivityDialog.tsx
@@ -15,7 +15,13 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { useState } from 'preact/hooks'
 
 import { fetchBucketedMetrics, fetchRawLocations, shareActivity, updateFeedPost } from '../state/api'
-import { defaultsFromChart, SERIES_METRICS, SUMMARY_METRICS } from './feed-metrics'
+import {
+  buildShareBody,
+  defaultsFromChart,
+  initialSeriesSelection,
+  SERIES_METRICS,
+  SUMMARY_METRICS,
+} from './feed-metrics'
 import { FEED_VISIBILITY_OPTIONS, VisibilitySelector } from './VisibilitySelector'
 import './ShareActivityDialog.css'
 
@@ -115,39 +121,36 @@ export function ShareActivityDialog({
   )
   // Preselect from the post's shared series (edit) or the charted metrics (create),
   // intersected with keys this dialog can represent — keeps the state typed as
-  // `Set<MetricType>` without a cast.
+  // `Set<MetricType>` without a cast. `initialSeriesSelection` folds a legacy
+  // chart-image post into the heart-rate series (one control governs both).
   const [series, setSeries] = useState<Set<MetricType>>(
-    () =>
-      new Set(
-        post
-          ? SERIES_METRICS.map((m) => m.key).filter((k) => post.series_metrics.includes(k))
-          : defaults.series,
-      ),
+    () => new Set(post ? initialSeriesSelection(post) : defaults.series),
   )
   const [visibility, setVisibility] = useState<FeedVisibility>(post?.visibility ?? 'public')
-  const [includeChart, setIncludeChart] = useState(post?.include_chart ?? false)
   const [includeMap, setIncludeMap] = useState(post?.include_map ?? false)
   const { summaryOptions, seriesOptions, canChart, canMap } = useShareableMetricOptions(
     activityStart,
     activityEnd,
     post?.included_metrics,
-    post?.series_metrics,
+    // Offer the heart-rate series on an edit whenever the post shares it in either
+    // format, so the unified control is always toggleable (never stuck checked).
+    post ? initialSeriesSelection(post) : undefined,
     post?.include_chart,
     post?.include_map,
   )
 
   const mutation = useMutation({
     mutationFn: () => {
-      // Only send metrics the dialog actually offered (drops defaults hidden as
-      // irrelevant); the backend still ignores any with no data. Attachments are
-      // only requested when their toggle is offered (chart needs HR, map needs GPS).
-      const body = {
-        include_chart: canChart && includeChart,
-        include_map: canMap && includeMap,
-        included_metrics: summaryOptions.map((m) => m.key).filter((k) => summary.has(k)),
-        series_metrics: seriesOptions.map((m) => m.key).filter((k) => series.has(k)),
+      const body = buildShareBody({
+        canChart,
+        canMap,
+        includeMap,
+        series,
+        seriesOptions,
+        summary,
+        summaryOptions,
         visibility,
-      }
+      })
       return post ? updateFeedPost(post.id, body) : shareActivity(activityId, body)
     },
     onSuccess: (result) => {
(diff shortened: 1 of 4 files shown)


apps/web/src/components/ShareActivityDialog.tsx
```diff
@@ -193,8 +196,8 @@ export function ShareActivityDialog({
             <legend>Share full time-series</legend>
             <p class="share-dialog-note">
               {mirroredFromChart
-                ? 'Higher resolution — more revealing than a summary. Pre-checked to match the activity chart; uncheck any you would rather not share.'
-                : 'Higher resolution — more revealing than a summary. Off by default.'}
+                ? 'Higher resolution — more revealing than a summary. Pre-checked to match the activity chart; uncheck any you would rather not share. Heart rate is also shared as a chart image.'
```
fiddur: Minor copy accuracy: this note ("Heart rate is also shared as a chart image.") is appended unconditionally whenever the series fieldset renders, but the fieldset also renders for activities that have other series yet no heart-rate data (e.g. a GPS/power activity with `speed`/`power` but no HR — `canChart` false, `heart_rate` absent from `seriesOptions`). In that case there's no heart-rate checkbox in the section, so the line references a control/attachment that isn't present. Consider only showing the "…also shared as a chart image" clause when `seriesOptions` actually includes `heart_rate`. Non-blocking.

fiddur (commented):
✅ Approved

Clean, well-scoped refactor. The heart-rate chart image and the heart-rate series are folded into a single control by deriving `include_chart` from `series.has('heart_rate')`, and the ad-hoc body-building logic is extracted into two pure helpers (`initialSeriesSelection`, `buildShareBody`) that are covered by focused unit tests. Types are sound (`ShareActivityBody` is a valid superset of the optional `UpdateFeedPostBody`), all imports remain used, no `includeChart` state references are left dangling, and the feed docs are updated to describe the unified control. Nice adherence to the project's "testable without heavy mocking" preference.

One behavioral consequence worth a conscious confirmation (it's documented as intended in the code comment on `initialSeriesSelection` and in feed.md, so not a blocker): editing/re-saving a legacy *image-only* post (`include_chart: true`, no `heart_rate` in `series_metrics`) backfills `heart_rate` into `series_metrics`, which makes the raw high-resolution series publicly queryable via the `/series` endpoint (the authorization boundary). This fires even on an edit that only changes visibility. The heart-rate box is visibly checked under the "more revealing than a summary" section and the note text calls it out, so it's surfaced rather than silent — just confirm this expansion-on-resave is the desired trade-off.

No REST/MCP parity work is needed since the wire fields stay independent and only the web dialog's composition changed.

apps/web/src/components/ShareActivityDialog.tsx
fiddur: Good catch — fixed in 43ab3bcd. The clause is now gated on `seriesOptions.some(m => m.key === "heart_rate")`, so it only appears when the heart-rate control is actually offered (not for speed/power-only activities).

fiddur (commented):
✅ Approved

Clean, well-tested refactor that collapses the separate heart-rate chart-image toggle into the heart-rate series control. The extraction of `buildShareBody` and `initialSeriesSelection` into `feed-metrics.ts` is testable and the new unit tests cover the important cases (fold, dedup, backfill, body construction, availability gating).

Correctness looks solid. I traced the key invariant `include_chart === true ⟹ heart_rate ∈ series_metrics` through every path:
- `include_chart` is only true when `canChart && series.has('heart_rate')`.
- `canChart` requires `present.has('heart_rate')` (create) or `keepChart`/`keepSeries` (edit), and in each of those cases `heart_rate` is also present in `seriesOptions`, so the `series_metrics` filter keeps it.

That means the public `chart.png` route (gated on `include_chart`) and the `/series` route (gated on `series_metrics`) can never diverge — no orphaned chart image. The backend confirms the "same data" premise: `chart.png` renders from `getSeries(username, 'heart_rate', …)`. Types line up (`buildShareBody` returns a complete `ShareActivityBody`, assignable to the all-optional `UpdateFeedPostBody` on the PATCH path), and no stale `includeChart`/`setIncludeChart` references or component tests referencing the removed "Heart-rate chart" checkbox remain.

Two non-blocking things worth being aware of (both appear intended and are documented):

1. **Legacy backfill widens exposure on re-save.** Editing a pre-unification post that had `include_chart: true` but an empty `series_metrics` now pre-checks heart rate, and saving writes `series_metrics: ['heart_rate']`. That newly authorizes the public `/series` endpoint to serve the *raw high-resolution* HR trace, where previously only the rendered PNG was exposed. The doc comment (`initialSeriesSelection`) and the "more revealing than a summary" fieldset copy disclose this, and the raw series is arguably equivalent to plotting the PNG — but it is a genuine expansion of what leaves the instance on save, so flagging it for awareness.

2. **Note copy reads unconditionally.** `" Heart rate is also shared as a chart image."` (ShareActivityDialog.tsx:205) renders whenever heart rate is *offered*, regardless of whether the HR box is checked — so a user who unchecks HR still sees the sentence. It describes the coupling rather than the current selection; minor, and the last commit already narrowed it to only show when HR is offered. Not worth blocking.

I could not run the test suite / typecheck locally (this fresh worktree has no `node_modules` installed), but the logic and types were verified by inspection and CI covers the rest.

PR #932, opened 2026-07-13
feat(feed): show a user's public feed on their profile + link local handles (#884)

fiddur:
## Summary

Fourth slice of the #884 tryout follow-ups, addressing Sara's report:

> She couldn't see anything from my public feed or info about me — there needs to be some way to see the info and public feed from a person, and I guess in the timeline it could also be present.

The `/u/:username` profile page already showed identity (avatar + handle) plus shared dashboards/challenges, but **not the person's public posts**, and handles weren't linkable. This slice closes that.

## What changed

- **Backend** — new `GET /public/:username/posts` returning a user's most recent `public`/`unlisted` posts (newest-first, latest page), serialized exactly like the authenticated `/feed` so the web renders them with the same `FeedPostCard`. Reuses the existing `listPublicFeedPostsPage`; short `max-age=60` like the sibling public read surfaces (a deleted/hidden post drops within a minute). Mounted before the generic `/public/:username/:slug` resolver so `posts` isn't taken for a share slug.
- **Profile page** — `/u/:username` now leads with a **Posts** section (the public feed) above dashboards and challenges.
- **Navigation** — a **local** author's name in the home timeline, Following, and Followers lists links to their `/u/:username` page; a **remote** (Mastodon &c.) author isn't linked (no page here). Local-vs-remote is a pure, unit-tested `localProfilePath` helper: the actor URI host matching the browsing host.
- **CSS** — extracted the `.feed-post*` card styles into `FeedPostCard.css` (owned by the card), so the standalone, chrome-less profile page is styled without importing the Feed page's entire stylesheet. The Feed page is unchanged (it renders the card, which now brings its own CSS).

## Discovery path this enables

Sara follows the user → the user's post appears in her home timeline → she clicks the (local) author name → lands on `/u/<user>` → sees their public feed + shared dashboards/challenges.

## Scope notes

- The response reuses the shared `FeedPostsResponse` (which requires `posts`), so the 404 branches carry an empty `posts: []` rather than a bare error body.
- Bounded to the latest 50 posts; cursor pagination for the profile feed is a follow-up if a prolific user needs it (the outbox already paginates for federation).
- Follower/following **counts** and a bio aren't added here — the profile shows identity + public content; richer profile info can be a later slice.

## Tests

- `feed-public-router.integration.test.ts` — serves public + unlisted, excludes followers-only, empty list, malformed-username 404.
- `profile-link.test.ts` — local → `/u/:username`, remote → null, non-actor path → null, malformed → null.
- Full `pnpm check` clean.

### DIFF
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -73,10 +73,12 @@ export const createFeedPublicRouter = (): TypedRouter => {
   // A user's public feed for their profile page: the most recent `public`/
   // `unlisted` posts, newest-first (same set as the ActivityPub outbox — never
   // `followers`-only). Serialized exactly like the authenticated `/feed`, so the
-  // web renders them with the same post card. Bounded to the latest page; a
-  // short cache matches the other public read surfaces (a deleted / hidden post
-  // drops within a minute). Mounted before the generic `/public/:username/:slug`
-  // resolver so `posts` is never mistaken for a share slug.
+  // web renders them with the same post card. Bounded to the latest page.
+  // `no-store` like the sibling `/series` and `/feed/:postId` endpoints: sharing
+  // is revocable, so flipping a post to `followers` or deleting it must drop it
+  // from the profile immediately, never linger in a shared cache. Mounted before
+  // the generic `/public/:username/:slug` resolver so `posts` is never mistaken
+  // for a share slug.
   router.get<{ username: string }, FeedPostsResponse>('/public/:username/posts', async (req, res) => {
     const { username } = req.params
     // The response reuses the authed `/feed` shape (which requires `posts`), so a
@@ -87,7 +89,7 @@ export const createFeedPublicRouter = (): TypedRouter => {
     try {
       const records = await listPublicFeedPostsPage(username, PROFILE_FEED_LIMIT, 0)
       const posts = await Promise.all(records.map((record) => serializeFeedPost(username, record)))
-      res.setHeader('Cache-Control', 'public, max-age=60')
+      res.setHeader('Cache-Control', 'no-store')
       res.json({ posts, success: true })
     } catch (error) {
       if (isMissingDatabase(error)) {
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -14,6 +14,7 @@
  * never mistaken for a share slug.
  */
 import {
+  type FeedPostsResponse,
   type FeedPostStructuredResponse,
   type PublicSeriesQuery,
   publicSeriesQuerySchema,
@@ -21,14 +22,18 @@ import {
 } from '@aurboda/api-spec'
 
 import { isValidUsername } from '../api/auth-routes.ts'
-import { findCoveringSharedSeriesWindow, isMissingDatabase } from '../db/index.ts'
+import { findCoveringSharedSeriesWindow, isMissingDatabase, listPublicFeedPostsPage } from '../db/index.ts'
 import { resolvePublicSeries } from '../services/feed-series.ts'
 import { resolveStructuredPost } from '../services/feed-structured.ts'
+import { serializeFeedPost } from '../services/feed.ts'
 import { queryMetricsBucketed } from '../services/queries/index.ts'
 import { type TypedRouter, typedRouter } from '../typed-router.ts'
 import { validateQuery } from '../validation.ts'
 
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+
+/** Most recent public/unlisted posts returned for a profile's feed (newest-first). */
+const PROFILE_FEED_LIMIT = 50
 
 export const createFeedPublicRouter = (): TypedRouter => {
   const router = typedRouter()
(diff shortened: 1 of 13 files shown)


apps/web/src/pages/Feed/FollowingPanel.tsx
```diff
@@ -27,7 +29,7 @@ const FollowingRow = ({ actor }: { actor: FollowingActor }) => {
         <span class="following-avatar following-avatar--blank" aria-hidden="true" />
       )}
       <span class="following-ident">
-        <span class="following-name">{name}</span>
+        <span class="following-name">{profilePath ? <a href={profilePath}>{name}</a> : name}</span>
```
fiddur, line 32: Minor polish: this new `<a>` inside `.following-name` (and the same one in `FollowersPanel`) has no dedicated CSS, so it renders with the default user-agent link styling (blue + underline) rather than inheriting the name's color like the home-timeline link does. `FeedPostCard.css` adds `.feed-post-name a { color: inherit; text-decoration: none }` for exactly this — worth an equivalent `.following-name a` rule in `style.css` so the linked handle matches the design. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped addition. The new `GET /public/:username/posts` correctly reuses `listPublicFeedPostsPage` (DB-level `visibility IN ('public','unlisted')` filter, so `followers`-only can never leak) and the shared `serializeFeedPost`, mirroring the authed `/feed` shape but bounded to a 50-post page. Router ordering is right — it's mounted before the generic `/public/:username/:slug` resolver, and within the router `posts` can't collide with `feed/:postId`. The integration test covers the visibility filter, the empty case, and the malformed-username 404. `localProfilePath` is a nice, well-tested helper that safely degrades to plain text for remote actors and malformed URIs. The `.feed-post*` CSS move is safe: the only other `.feed-post` consumer (HomeTimeline) always renders on the Feed page, which imports `FeedPostCard` (and thus `FeedPostCard.css`).

One intentional tradeoff worth calling out: unlike `/series` (`no-store`), this endpoint sets `Cache-Control: public, max-age=60`, so a post flipped to `followers`-only or deleted can linger in a shared cache for up to a minute. The response only carries already-public metadata/content (the raw series + images remain `no-store` / live-checked), and the comment documents the 60s window, so this is acceptable — just flagging it as a deliberate choice. Non-blocking otherwise; nice work.

apps/web/src/pages/Feed/FollowingPanel.tsx
fiddur: Fixed in 8458d1a7 — added `.following-name a { color: inherit; text-decoration: none }` (+ hover underline) to style.css, mirroring `.feed-post-name a`, so the linked names in both panels match the design.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -65,6 +70,33 @@ export const createFeedPublicRouter = (): TypedRouter => {
...
+    const { username } = req.params
+    // The response reuses the authed `/feed` shape (which requires `posts`), so a
+    // 404 carries an empty list rather than a bare error body.
+    if (!isValidUsername(username)) {
+      return res.status(404).json({ error: 'Not found', posts: [], success: false })
+    }
+    try {
+      const records = await listPublicFeedPostsPage(username, PROFILE_FEED_LIMIT, 0)
+      const posts = await Promise.all(records.map((record) => serializeFeedPost(username, record)))
+      res.setHeader('Cache-Control', 'public, max-age=60')
```
fiddur: Non-blocking: worth a sanity check that `public, max-age=60` (vs the sibling `/series` and `/feed/:postId` endpoints, which use `no-store` precisely because sharing is revocable) is the intended trade-off. Because it's a `public` response, a shared cache/CDN can keep serving a post for up to ~60s after the owner flips it from public/unlisted to `followers` or deletes it. That's acknowledged in the comment above and is arguably fine for a post-card listing (the content already federated publicly), but the `no-store` neighbors made me pause — confirming it's deliberate.

fiddur (commented):
✅ Approved

Clean, well-tested feature. Verified the substantive aspects:

- **Visibility filter is correct.** `/public/:username/posts` uses `listPublicFeedPostsPage`, which is scoped to `visibility IN ('public','unlisted')` — the same set as the ActivityPub outbox. The integration test explicitly asserts a `followers`-only post is never returned. Matches Mastodon semantics (unlisted posts appear on the profile page / outbox, just not in discovery timelines), so no new data is exposed beyond what already federates.
- **Route ordering holds.** The new route is registered inside `createFeedPublicRouter` (mounted at `rest-routes.ts:167`) before the generic `/public/:username/:slug` resolver in `createPublicSharesRouter` (mounted at `:185`), and before the sibling `/public/:username/feed/:postId` route — so `posts` is never mistaken for a share slug.
- **CSS move preserves styling.** The `.feed-post*` rules moved from `Feed/style.css` into `FeedPostCard.css`. `HomeTimeline`'s `TimelineCard` reuses those classes without importing the CSS itself, but it renders on `Feed/index.tsx`, which imports the `FeedPostCard` module (whose CSS import is bundled with the page), and `PublicProfile` renders `FeedPostCard` directly — so both surfaces get the styles. The new `.feed-post-name a` rule is correctly added for the now-linked author name.
- **Local/remote handle linking is sound.** `localProfilePath` only links actors whose host matches and whose path is `/users/<username>` (the AP actor id), returning null for remote actors and non-actor local URLs; usernames are `[a-z][a-z0-9_]{2,30}` so the encode/decode is defensive but harmless. Good unit coverage.
- Serialization, response shape (`FeedPostsResponse`), `state/api` re-export, and 404 branches all line up with existing patterns; docs updated.

One minor, non-blocking note left inline on the cache header — otherwise this is good to merge.

apps/backend/src/routes/feed-public-router.ts
fiddur: Good call — switched to `no-store` in d435d381 to match the sibling `/series` and `/feed/:postId` endpoints. Sharing is revocable, so a post flipped to followers-only or deleted now drops from the profile immediately rather than lingering ~60s in a shared cache. The listing is bounded (≤50), so re-rendering per request is cheap.

fiddur (commented):
✅ Approved

Clean, well-scoped feature with good test coverage. No blocking issues.

**What I verified**
- **Visibility boundary is enforced server-side.** `GET /public/:username/posts` delegates to the existing `listPublicFeedPostsPage`, whose SQL filters `visibility IN ('public','unlisted')`, so `followers`-only posts can never leak. The new integration test asserts exactly this (public + unlisted returned, followers excluded), plus the empty-list and malformed-username 404 branches.
- **Username validation.** The route calls `isValidUsername` before `username` is used to select a per-user database, matching the sibling `/series` and `/feed/:postId` endpoints — no injection surface. `unlisted` posts appearing on the profile page is consistent with the outbox (and with Mastodon's profile semantics).
- **Route ordering.** `createFeedPublicRouter()` is mounted (rest-routes.ts:167) before `createPublicSharesRouter` (`/public/:username/:slug`), so `posts` is matched as this route, not a share slug.
- **`localProfilePath`** is safe (same-origin path only, encoded username, try/catch on malformed URIs) and well unit-tested; the local/remote link logic in HomeTimeline / Following / Followers is correct.
- **CSS move is safe.** The `.feed-post*` rules moved to `FeedPostCard.css`; the only two consumers (`FeedPostCard` and `HomeTimeline`) both live on the Feed page, whose `index.tsx` imports `FeedPostCard`, so the stylesheet still loads. `serializeFeedPost` handles a deleted activity gracefully, so `Promise.all` won't fail the whole request.

**Non-blocking observations (no change required)**
- The endpoint is unauthenticated, `no-store`, and serializes up to 50 posts each doing its own activity/scalar lookups (an N+1 inherited from `serializeFeedPost`), so each anonymous hit can trigger ~100 DB queries with no cache to absorb it. This mirrors the existing `/series` and `/feed/:postId` public endpoints and is bounded to 50, so it's consistent with the current threat model — just worth keeping in mind if a rate-limit/CDN layer isn't already fronting these.
- `HomeTimeline` renders `.feed-post` markup directly (not via the `FeedPostCard` component) and now relies on `FeedPostCard.css` being loaded by a sibling import on the same page. It works today, but if `HomeTimeline` is ever mounted on a route that doesn't also import `FeedPostCard`, its card styling would silently disappear.

PR #933, opened 2026-07-14
feat(feed): backfill the home timeline with a followee's recent posts on follow (#884)

fiddur:
## Summary

Final slice of the #884 tryout follow-ups. From Sara's report:

> ...and I guess in the timeline it could also be present, even if older things only shared with followers of course won't be present.

Following someone left the home timeline **empty until they next posted**. Now, when a follow you sent is accepted, that person's recent **public** posts are fetched from their ActivityPub outbox and ingested, so your timeline is populated immediately — while their older followers-only/private posts are correctly excluded (exactly Sara's caveat).

## What changed

- **Shared ingest core** — factored `ingestNoteForRecipient` out of `ingestFeedActivity`: the one path that maps a `Note` → timeline entry (sanitise HTML, capture image attachments, best-effort structured-chart enrichment, upsert keyed on `object_uri`). Live delivery and backfill now build entries identically, so a backfilled post that later arrives live just updates in place — no duplicates.
- **New `timeline-backfill.ts`:**
  - `backfillFolloweeTimeline(deps, user, actorUri)` — the orchestration, **Fedify- and DB-free via injected deps**, so it unit-tests without network or a database. Skips non-accepted follows; best-effort (a failed outbox fetch or a single bad note just yields fewer posts).
  - A thin Fedify outbox fetcher — `ctx.lookupObject(actor).outboxId` → `traverseCollection`, bounded to **~20 Notes / 80 scanned items**, accepting `Create{Note}` and bare `Note` entries.
  - `createTimelineBackfiller` — production wiring: real enricher + DB, **time-boxed (15s) and fire-and-forget**.
- **Trigger** — an injected `onFollowAccepted` callback fired from the inbox `Accept` handler (after the accept is durably recorded). This covers **remote** followees (their server's Accept) *and* **local** ones (a local follow auto-accepts back through the same handler), with a single hook. It never blocks or 500s inbox processing.

## Behaviour notes

- **Only the outbox is read**, so `followers`-only and older private posts are never backfilled — a follower sees exactly what a non-follower already could, plus live followers-only posts going forward.
- Backfilled posts are **historical**, so they don't fire the live "N new posts" pill — they appear in the timeline (in published order) on the next load.
- Bounded + time-boxed + fire-and-forget, so a large/slow/hostile outbox can't affect the follow or the inbox.

## Tests

- `timeline-backfill.test.ts` — orchestration: ingests each note & returns the count, passes the followee through, skips non-accepted/missing follows, returns 0 on outbox-fetch or followee-lookup failure, and continues past a single bad note.
- Existing `timeline-ingest` + `federation.integration` suites still green after the `ingestNoteForRecipient` refactor.
- Full `pnpm check` clean.

## Wraps up #884 tryout follow-ups

This is slice 5 of 5 (after #929, #930, #931, #932). Sara can now follow you, immediately see your recent public posts backfilled into her timeline, get live native charts/images going forward, and click through to your `/u/` profile.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -135,14 +135,6 @@ const recordInboundFollow = async (
 }
 
 /**
- * Ingest a `Create`/`Update` of a `Note` into the recipient's home timeline —
- * but ONLY if the sender is an *accepted* followee (so the timeline can't be
- * injected into by a stranger who delivers to our inbox). The author's
- * presentation comes from the cached `feed_following` row; the Note's HTML is
- * sanitised in `noteToTimelineInput`. Best-effort: unresolvable objects and
- * non-Notes are ignored, and a missing DB never 500s (which would invite retries).
- */
-/**
  * Ingest one `Note` from an accepted followee into `user`'s home timeline: map it
  * to a timeline entry (author from the cached `feed_following` row, content
  * sanitised), capture its image attachments, best-effort fetch the native Aurboda
@@ -174,6 +166,14 @@ export const ingestNoteForRecipient = async (
   if (inserted) onNewEntry?.(user)
 }
 
+/**
+ * Ingest a `Create`/`Update` of a `Note` into the recipient's home timeline —
+ * but ONLY if the sender is an *accepted* followee (so the timeline can't be
+ * injected into by a stranger who delivers to our inbox). Resolves the recipient
+ * and the cached followee row, then hands off to `ingestNoteForRecipient`.
+ * Best-effort: unresolvable objects and non-Notes are ignored, and a missing DB
+ * never 500s (which would invite retries).
+ */
 const ingestFeedActivity = async (
   ctx: InboxContext<void>,
   activity: Create | Update,
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -43,6 +43,7 @@ import {
   countFeedFollowers,
   countPublicFeedPosts,
   deleteTimelineEntryByUri,
+  type FeedFollowingRecord,
   getFeedFollowerByActor,
   getFeedFollowingByActor,
   getFeedPostById,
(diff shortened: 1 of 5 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -141,6 +142,38 @@ const recordInboundFollow = async (
  * sanitised in `noteToTimelineInput`. Best-effort: unresolvable objects and
  * non-Notes are ignored, and a missing DB never 500s (which would invite retries).
  */
+/**
```
fiddur, line 137: Minor: the pre-existing doc comment just above ("Ingest a `Create`/`Update` of a `Note`…") described `ingestFeedActivity`, but this refactor inserted `ingestNoteForRecipient` (with its own doc block) between them — so that old comment is now orphaned above this function and `ingestFeedActivity` below is left undocumented. Consider moving the old block down to sit directly above `ingestFeedActivity`.

fiddur (commented):
✅ Approved

Solid, well-tested addition. The refactor cleanly extracts `ingestNoteForRecipient` so backfill and live delivery build timeline entries through the exact same path (sanitisation, image capture, structured enrichment, `object_uri` upsert), and the new orchestration is DB-/Fedify-free via injected deps with good unit coverage (skip-when-unaccepted, best-effort outbox failure, per-note failure isolation).

The design holds up on the risk areas I checked:
- **No SSRF/injection amplification.** Backfill only fetches an outbox *after* `getFollowee` confirms an accepted `feed_following` row, so the fetch target is always an actor the local user chose to follow, not attacker-controlled; `noteToTimelineInput`'s host + `attributedTo` guards still reject cross-actor injection from a hostile outbox, and Fedify's document loader keeps the private-URL protections.
- **Never blocks the inbox.** `withTimeout` + swallowed errors + fire-and-forget mean a slow/hostile/private outbox just yields no backfill.
- **Forward reference is fine.** `backfill` is used inside a closure that only runs on a later inbound Accept, by which point `const backfill` is initialised; the comment documents this.
- **`onNewEntry` correctly omitted** so historical posts don't trigger the "N new posts" pill.
- The `ctx.lookupObject`/`createContext(new URL(origin))` usage matches the established pattern in `following.ts`/`deliver.ts`.

One minor note (non-blocking): the doc's claim that "the outbox only carries public/unlisted posts" is a per-server assumption (a signed fetch as an accepted follower could in principle surface followers-only posts on some implementations) — but since we are a legitimate follower at that point, ingesting such posts is not a privacy leak, so this is fine as-is. Nothing here blocks merge.

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed in d765f9a8 — removed the orphaned block and gave `ingestFeedActivity` its own doc (it now describes resolving the recipient/followee and handing off to `ingestNoteForRecipient`), so each function is documented directly above itself.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The `ingestNoteForRecipient` extraction is behavior-preserving (identical to the old inline block in `ingestFeedActivity`), and sharing it means backfill and live delivery build timeline entries identically — including the `object_uri` upsert that lets a later live delivery replace a backfilled copy in place. The orchestration (`backfillFolloweeTimeline`) is cleanly dependency-injected and covered by seven focused unit tests (accepted-guard, missing-follow, outbox-failure, per-note failure isolation, count). Security is sound: backfill re-checks `accepted`, and `noteToTimelineInput` still enforces `note.id.host === followee host` + attribution, so a hostile outbox can't inject foreign-authored notes; only the outbox (public/unlisted, or `listPublicFeedPostsPage` locally) is read, so followers-only posts aren't backfilled. Best-effort discipline is consistent throughout — fire-and-forget after a durable `markFeedFollowingAccepted`, time-boxed (15s), and bounded by `BACKFILL_LIMIT`/`MAX_SCANNED`. Omitting `onNewEntry` on the backfill path correctly keeps historical posts out of the "N new posts" pill.

Verified: `OrderedCollection extends Collection` in Fedify, so `outbox instanceof Collection` matches a Mastodon outbox; `ctx.traverseCollection`/`lookupObject`/`createContext` match existing usage in `followers.ts`/`deliver.ts`. The forward reference to `backfill` in `api.ts` is safe — the arrow only reads it at call time (on a later inbound Accept), after assignment.

Non-blocking notes (no action required): (1) `withTimeout` is a soft timeout — it stops awaiting but doesn't cancel the in-flight outbox traversal; `MAX_SCANNED` + fire-and-forget bound the work, so this is fine. (2) A duplicate/re-sent `Accept` re-runs a full outbox fetch; it's idempotent via upsert but re-fetches — acceptable given how rarely Accepts repeat. (3) A brand-new post that lands during the backfill window could be ingested by the backfill first and thus miss the live pill; the post still appears, so this is a minor cosmetic edge case. I could not run `pnpm check`/vitest here (no installed node_modules in the review worktree), so CI remains the source of truth for the build.

Issue #934, opened 2026-07-14
Epic: Shareable analysis Articles — long-form feed posts with inline charts


fiddur:
## Vision

Let a user compose and share a **quantified-self analysis** — narrative prose plus multiple charts and correlations over specific time windows — among QS enthusiasts (on Aurboda, the fediverse, and exportable to e.g. r/QuantifiedSelf). Most often **AI-drafted** (Claude, via MCP) but always **user-editable**; posting to Reddit uses the manual analysis text since that community frowns on fully-AI posts.

Builds on the federated feed (#831, #884) and public sharing surfaces (#932).

## Settled design decisions

- **Article _is_ a feed post kind — not a bespoke "analysis" entity.** A post gains an `article` kind: **title + markdown prose + an ordered set of inline chart/correlation blocks**, each anchored to a **locked `[start, end]` window**. The current activity-anchored post is the special case; an article is the general one (N charts + prose, no required activity anchor). This reuses everything the feed already has — visibility (`public`/`unlisted`/`followers`), ActivityPub federation, the home timeline, the public-profile feed (#932), and permalinks — so no separate slug/page surface is needed initially.
- **Lock the chart time window; NO data snapshot.** The fixed window is what makes a chart stable/citable; we re-resolve **live** against it (same as shared dashboards and the `/series` endpoint), so data corrections propagate. Consumers that can't re-resolve (Mastodon, Reddit) get an **on-demand rendered PNG**, not a frozen data blob. Revocation stays trivial (`no-store`).
- **Inline charts on Aurboda; degrade to attached images on Mastodon.** Federate as AS2 **`Article`** carrying prose HTML + **attached chart PNGs** (Mastodon shows prose + images); Aurboda peers render **inline interactive charts** via the native-enrichment channel (#930) + SVG (#901).
- **AI drafting is free from MCP↔API parity.** `create_article`/`update_article` land as MCP tools alongside their REST endpoints in Slice A, so Claude can draft an article conversationally by calling the same tools — there is no separate "AI" slice. (An optional server-side `draft_article` convenience helper could come later.)
- **Prose is rendered through a shared sanitizer** (#910) — AI/user markdown → sanitized HTML.

## Slices

- [ ] **Slice A** (#935) — Article feed posts: entity + block model, MCP+REST (parity), inline rendering, editor.
- [ ] **Slice B** (#936) — Article correlation blocks.
[...]

PR #938, opened 2026-07-14
fix(feed): sharing-page polish — card header purple leak + clipped Share button (#916, #906)

fiddur:
Warm-up polish on the sharing surfaces we shipped in #884/#932, ahead of the analysis-Articles epic (#934). Two `tryout-finding` bugs:

- **#916** — the feed post card's semantic `<header class="feed-post-head">` inherited the global `header { background: #673ab8; display: none }` rule; `display: flex` re-shows it but never clears the background, so a purple banner painted behind the handle (unreadable). Fixed by clearing `background` on `.feed-post-head`.
- **#906** — on narrow/mobile viewports the public-profile header (`display: flex` row: avatar + non-wrapping `<h1>@handle</h1>` + `margin-left:auto` Share button) pushed the Share button off-screen, since a long `@handle` has no spaces to break on. Fixed by `flex-wrap: wrap` on the header plus `min-width: 0; overflow-wrap: anywhere` on the `h1`, so the button wraps to a reachable line and long handles wrap in place.

CSS-only. Closes #916, closes #906.

### DIFF
--- a/apps/web/src/pages/Feed/FeedPostCard.css
+++ b/apps/web/src/pages/Feed/FeedPostCard.css
@@ -12,6 +12,10 @@
   display: flex;
   align-items: center;
   gap: 0.6rem;
+  /* This is a semantic <header>, which the global `header` rule paints purple
+     (a hidden legacy top-bar, re-shown here by `display: flex`). Clear it so the
+     card doesn't get a purple banner behind the handle (#916). */
+  background: none;
 }
 
 .feed-post-avatar {
--- a/apps/web/src/pages/PublicProfile/style.css
+++ b/apps/web/src/pages/PublicProfile/style.css
@@ -13,10 +13,17 @@
   align-items: center;
   gap: 1rem;
   margin-bottom: 1rem;
+  /* Let the Share button wrap to a new line instead of being pushed off-screen
+     by a long, non-wrapping `@handle` on narrow viewports (#906). */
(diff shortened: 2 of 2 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped CSS-only fixes; both are correct and clearly documented.

**`FeedPostCard.css`** — `.feed-post-head` is a semantic `<header>`, and the global `header` rule (`apps/web/src/style.css:50`) paints it `background-color: #673ab8` while hiding it with `display: none`. Because the class re-shows the element via `display: flex` (class specificity beats the element selector), the purple background leaked behind the card handle. `background: none` on the class correctly clears it (higher specificity than the `header` element rule) in both light and dark modes.

**`PublicProfile/style.css`** — the header row holds a fixed-width avatar, `<h1>@handle`, and a Share button pinned right with `margin-left: auto`. A long, unbreakable `@handle` on a narrow viewport forced the row wider than the viewport and pushed the button off-screen. `flex-wrap: wrap` lets the button drop to its own line, and `min-width: 0` (overriding flex's default `min-width: auto`) together with `overflow-wrap: anywhere` lets the handle shrink and break mid-word. All three declarations are needed and work together.

No correctness concerns, no test impact (pure CSS), and the comments cite the tracked issues (#916, #906) per repo convention. Approving.

PR #939, opened 2026-07-14
fix(web): centralize markdown rendering through a DOMPurify sanitizer (#910)

fiddur:
Foundation for the analysis-Articles epic (#934), and closes a latent XSS gap.

## Problem

`marked` (^17) passes raw HTML in the source straight through and has no built-in sanitizer, so the existing `marked.parse(...)` → `dangerouslySetInnerHTML` sinks were unsanitized:

- `EntityDetail/index.tsx` (activity description)
- `EntityDetail/NotesSection.tsx` (notes)
- `MarkdownEditor/index.tsx` (preview)

Low-risk **today** (a user's own notes rendered back to themselves — self-XSS only), but a **stored-XSS** vector the moment a sink renders markdown authored by a *different* user — which analysis **Articles** (#934) will do (shared, federated, AI/other-authored prose).

## Fix

- **One render path** — `renderMarkdown(md)` in `apps/web/src/utils/markdown.ts`: parse (GFM + hard breaks) then **DOMPurify-sanitize**. Centralized so the XSS boundary can't be forgotten at a new sink.
- Routed all three sinks through it; removed their ad-hoc `marked` imports/config (the editor's "no HTML passthrough" comment was misleading — it never sanitized).
- Added `dompurify@^3` (ships its own types).

## Tests

`markdown.test.ts` (jsdom env): renders GFM, honours line breaks, and **strips** `<script>`, event-handler attributes (`onerror`), and `javascript:` URLs while keeping safe links/images. Full `pnpm check` clean.

Closes #910.


fiddur (commented):
✅ Approved

Solid, well-scoped security fix. Routing every markdown sink through a single `renderMarkdown` that runs `marked.parse` output through `DOMPurify.sanitize` closes the stored-XSS gap for user/AI-authored notes and descriptions, and the module doc-comment clearly states why DOMPurify (not marked) is the XSS boundary. The new tests cover the important vectors — `<script>`, `onerror` event-handler attributes, `javascript:` URLs — plus the happy-path GFM/`breaks`/image behaviour, and they correctly pin the jsdom environment so DOMPurify has a `window`.

Coverage looks complete: all three previous `marked.parse` sinks (MarkdownEditor preview, NotesSection, EntityDetail description) now go through `renderMarkdown`, and grep confirms no remaining direct `marked.parse` calls in the web app. The two other `dangerouslySetInnerHTML` sinks (`Feed/FeedPostCard.tsx:75`, `Feed/HomeTimeline.tsx:77`) intentionally render server-generated/server-sanitised AS2 HTML rather than markdown and are documented as such inline, so they are correctly out of scope here. Centralising the `marked.setOptions({ breaks, gfm })` config is also a real improvement over the prior arrangement, where NotesSection/EntityDetail implicitly depended on MarkdownEditor's module being imported to set those global options.

Minor, non-blocking: with `renderMarkdown` now returning `string`, the `rendered as string` cast in `MarkdownEditor/index.tsx:56` is redundant and could be dropped (CLAUDE.md discourages casts) — that line is outside the diff, hence noted here rather than inline.

I could not run `pnpm check`/tests here because the sandbox blocked `pnpm install` (read-only pnpm store), but the logic and types (the `{ async: false }` overload yields `string`, so `DOMPurify.sanitize` receives and returns a string) check out by inspection.

PR #941, opened 2026-07-14
refactor(charts): shared server-side chart SVG renderer (#901, #888)

fiddur:
## What

Extracts the metric line-chart SVG out of `renderChartPng`'s inlined string into a pure, `sharp`-free **`buildChartSvg(series, opts) → string`** in `apps/backend/src/services/charts/chart-svg.ts` — the single implementation the two upcoming foundations share.

## Why

Both foundations need the **same chart from the same data** by two different paths:

- **`image/svg+xml` served directly** for crisp Aurboda-native rendering — `aurboda:charts` (#901).
- **`sharp`-rasterized PNG** for the OG snapshot / Mastodon attachment / Reddit export (#888).

Keeping the chart as one pure string builder avoids two implementations (the exact concern #888 calls out) and — because it's now a plain synchronous string function — makes the chart **directly unit-testable** rather than only assertable through an opaque PNG.

## Changes

- **New** `services/charts/chart-svg.ts`: `buildChartSvg(series, opts)` with configurable colour / label / width / height, plus shared `escapeXml` / `scale` helpers.
- **New** `chart-svg.test.ts`: 10 unit tests on the SVG string — polyline geometry (exact scaled coords), min/max labels, single-point / empty / non-finite handling, custom colour & size, XML-escaped label (SVG-injection guard).
- `renderChartPng` is now a thin `svgToPng(buildChartSvg(...))` wrapper — **byte-identical default output**, so the feed attachment PNG is unchanged (existing feed-images tests still pass).

## Scope / not in this PR

No behaviour change. This is the shared renderer only; the two consumers land next:
- #901 — serve the SVG as `image/svg+xml` + the `aurboda:charts` AS2 property (degrade to PNG for Mastodon).
- #888 — embed a chart snapshot into the OG card (`chartDataUri` slot).

Foundation for the shareable-analysis **Articles** epic (#934).

## Test

`pnpm check` green (0 errors); `chart-svg.test.ts` (10) + `feed-images.test.ts` (8) pass.

### DIFF
--- a/apps/backend/src/services/activitypub/feed-images.ts
+++ b/apps/backend/src/services/activitypub/feed-images.ts
@@ -2,7 +2,9 @@
  * Render feed-post attachment images (a metric line chart and a GPS route map)
  * as PNGs, from raw series / coordinate data.
  *
- * The chart is a pure, self-contained SVG rasterized with `sharp`. The route map
+ * The chart SVG is built by the shared `buildChartSvg` renderer (so the feed PNG
+ * and the crisp `image/svg+xml` path stay one implementation) and rasterized here
+ * with `sharp`. The route map
  * is drawn over an **OpenStreetMap street basemap**: the track is projected into
  * Web Mercator (the tiles' projection), the covering tiles are fetched and
  * composited behind it, and the required "© OpenStreetMap contributors"
@@ -14,10 +16,9 @@
  */
 import sharp from 'sharp'
 
+import { buildChartSvg } from '../charts/chart-svg.ts'
 import { chooseZoom, latToWorldY, lonToWorldX, TILE_SIZE, type TileFetcher } from './osm-tiles.ts'
 
-const CHART_W = 1000
-const CHART_H = 420
 const ROUTE_W = 700
 const ROUTE_H = 700
 const PAD = 40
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Clean, behavior-preserving extraction. The chart SVG-building logic is lifted out of `feed-images.ts` into a new sharp-free `charts/chart-svg.ts` module, and `renderChartPng` becomes a thin `svgToPng(buildChartSvg(...))` wrapper. I verified parity with the original: default dimensions (1000×420), colour (`#ef4444`), label (`Heart rate`), `escapeXml`, `scale`, the `≥2 points` polyline guard, and the SVG structure are all unchanged.

The new unit tests are a genuine improvement — they assert on the SVG string directly rather than an opaque PNG, and cover the meaningful cases (multi-point polyline, single point, non-finite dropping, empty series, custom colour/dimensions, and XML-escaping of the label). I reproduced the polyline coordinate assertions (`40.0,282.9`, `346.7,40.0`, `960.0,380.0`) against the actual scale math and they match exactly.

A couple of things I checked that are fine: dropping `async` from `renderChartPng` doesn't change caller semantics (it still returns a `Promise<Buffer>`, and `buildChartSvg` has no throwing path), and the `stroke="${color}"` interpolation remains unescaped but is not user-controlled (the feed-image router calls `renderChart(series)` with no options, so only the default colour is used) — unchanged from before. Newly exporting `escapeXml`/`scale` is harmless and reasonable for testability.

Note that the `image/svg+xml` serving path (#901) referenced in the module docstring isn't part of this diff; this PR is the extraction that enables it, which is consistent with the stated intent. Tests couldn't be executed here (deps not installed in the review env), but I validated the assertion math manually.

Issue #937, opened 2026-07-14
Article posts (Slice C): federation as AS2 Article + Reddit/markdown export


fiddur:
Part of #934 (Shareable analysis Articles). Depends on Slice A (#935) (+ #888 for PNG rendering).

## Goal

Federate an article and make it exportable, with the agreed degradation: **inline interactive charts for Aurboda peers, attached PNGs for Mastodon**, and a paste-ready **markdown/PNG bundle for Reddit** (manual analysis text, per QS-community norms).

## Scope

- **Federation:** deliver an article as an AS2 **`Article`** (title + prose HTML + **attached chart PNGs**) so Mastodon renders prose + images. Aurboda peers fetch the **native block payload** (prose + chart/correlation blocks with their locked windows) via the structured-enrichment channel (#930 pattern) and render inline interactive charts.
- **Object serving / outbox / timeline:** an article appears in the outbox + followers' home timelines like any post; the timeline card renders inline charts (Aurboda) or the attached images.
- **Export:** a Reddit/markdown export — the article's prose + each chart as a rendered PNG (bundle or embedded links), so the user pastes their own text + the charts. Reuses the chart renderer (#888).
- **Privacy:** reuse the feed visibility + capability-token model; `no-store` on the resolved data endpoints.

🤖 Planned with [Claude Code](https://claude.com/claude-code)

fiddur:
Closing as complete: Slice C merged via #967 and #971. Note the design change from the original title: articles federate as a `Note` (not AS2 `Article`) because Mastodon discards `content` on converted types — documented in docs/features/feed.md. Native structured enrichment for Aurboda peers + Reddit/markdown export (`GET /feed/articles/:postId/export`, `export_article_markdown`) both shipped. Deferred follow-ups: #972, #973, #974, #975.

Issue #935, opened 2026-07-14
Article posts (Slice A): entity, block model, MCP+REST, inline rendering


fiddur:
Part of #934 (Shareable analysis Articles).

## Goal

Introduce the **`article`** feed-post kind: a titled post with markdown prose and inline **chart** blocks over locked time windows, creatable/editable via **both MCP and REST** (parity), rendered inline on Aurboda and reachable at the post permalink + on the public-profile feed (#932).

## Scope

- **Schema (api-spec, single source):** extend the feed-post model with an `article` kind carrying `title`, and an ordered `blocks` array:
  - `{ type: 'prose', markdown }`
  - `{ type: 'chart', metric/selector, window {start,end}, bucket?, caption? }`
  - (correlation blocks land in Slice B / #TBD.)
  - Blocks inherit an article-level default window unless they override it.
- **Backend:** persist blocks (JSONB); `create_article` / `update_article` MCP tools + REST endpoints with parity; shared service resolves each chart block's data live over its locked window (reuse the `/series` resolution). No data snapshot.
- **Web:** a compose/edit view (prose + add/reorder chart blocks, pick metric + window + caption); render an article inline (prose via the shared sanitizer, charts via native SVG/PNG).
- **Reuse:** prose sink = #910 (DOMPurify sanitizer); chart rendering = #901 (SVG) / #888 (PNG).

## Depends on

- #910 (markdown sanitizer) — prerequisite for the prose sink.
- #901 (SVG native chart) — inline chart rendering.

## Out of scope (later slices)

- Correlation blocks → Slice B.
- Federation as AS2 `Article` + Reddit export → Slice C.

🤖 Planned with [Claude Code](https://claude.com/claude-code)

fiddur:
Closing as complete: Slice A merged (feed-post `kind`, article entity + ordered block model, `POST/PATCH /feed/articles` + `create_article`/`update_article` MCP tools, inline web rendering + composer).

PR #942, opened 2026-07-14
feat(feed): serve the crisp chart as image/svg+xml (#901)

fiddur:
## What

Adds `GET /public/:username/feed/:postId/chart.svg` alongside the existing `chart.png`, serving the **same series over the same locked window** as a scalable `image/svg+xml` built by the shared `buildChartSvg` renderer (#941) — no rasterisation, so it's crisp at any size for **Aurboda-native** rendering (#901).

## Details

- Same eligibility gate as the PNG (`resolveImageWindow` — public/unlisted, or followers-only with the capability `?token=`), same `no-store` revocability.
- Its own render-cache key (`chartsvg:…`); the built SVG is stored as UTF-8 bytes so it reuses the existing buffer LRU — the DB series fetch is the cost worth caching, not the string build.
- New injected dep `renderChartSvg: buildChartSvg` (mirrors `renderChart`), wired in the composition root.

## Why now

This is the inline-chart building block **Slice A (#935) depends on** ("SVG native chart — inline chart rendering"): the web app / native consumers embed the crisp SVG, which degrades to the PNG attachment for Mastodon.

## Scope

Delivers the **endpoint** half of #901. The `aurboda:charts` AS2 discovery link (so remote Aurboda peers auto-find this SVG) is deferred to the **federation slice (#937)**, where the article's federation object and the PNG-vs-SVG degrade are wired together — keeping this PR out of the two federation object builders.

## Test

`pnpm check` green (0 errors). `feed-image-router.test.ts` + `chart-svg.test.ts` pass (25); the new route reuses the already-tested `resolveImageWindow`, `createRenderCache`, and `buildChartSvg`.

Foundation for the shareable-analysis **Articles** epic (#934).

### DIFF
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -4,9 +4,13 @@ import { type Response, Router } from 'express'
  * Public feed-post image endpoints (UNAUTHENTICATED).
  *
  * Handles: GET /public/:username/feed/:postId/chart.png
+ *          GET /public/:username/feed/:postId/chart.svg
  *          GET /public/:username/feed/:postId/route.png
  *
- * Rendered on demand from the shared activity's data. An image is served for a
+ * The chart is offered both ways from the same series over the same window: a
+ * rasterised PNG that every consumer understands (Mastodon attaches it) and a
+ * crisp, scalable `image/svg+xml` for Aurboda-native rendering (#901). Rendered
+ * on demand from the shared activity's data. An image is served for a
  * `public`/`unlisted` post that opted into that attachment (`include_chart` /
  * `include_map`); a `followers`-only post is served only when the request carries
  * the post's unguessable capability `?token=` (embedded solely in the Note
@@ -36,6 +40,8 @@ export interface FeedImageDeps {
   getSeries: (user: string, metric: string, start: Date, end: Date) => Promise<[Date, number][]>
   getRoute: (user: string, start: Date, end: Date) => Promise<[number, number][]>
   renderChart: (series: [Date, number][]) => Promise<Buffer>
+  /** Build the crisp `image/svg+xml` chart (same data as `renderChart`, no raster). */
+  renderChartSvg: (series: [Date, number][]) => string
   renderRoute: (coords: [number, number][]) => Promise<Buffer>
 }
 
@@ -113,6 +119,11 @@ const sendPng = (res: Response, png: Buffer) => {
   res.type('png').send(png)
 }
 
+const sendSvg = (res: Response, svg: Buffer) => {
+  res.setHeader('Cache-Control', 'no-store') // revocable, same as the PNG
+  res.type('image/svg+xml').send(svg)
+}
+
 export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
   const router = Router()
   const cached = createRenderCache()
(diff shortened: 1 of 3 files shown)


apps/backend/src/routes/feed-image-router.ts
```diff
@@ -4,9 +4,13 @@ import { type Response, Router } from 'express'
  * Public feed-post image endpoints (UNAUTHENTICATED).
  *
  * Handles: GET /public/:username/feed/:postId/chart.png
+ *          GET /public/:username/feed/:postId/chart.svg
```
fiddur, line 7: The new endpoint is documented here in the router JSDoc, but `docs/features/feed.md` still only lists `chart.png`/`route.png` — both in the endpoint code block (~lines 267–270) and the endpoint table (lines 386–387). Per CLAUDE.md ("keep documentation in `docs/` up to date when changing related code" / "when changing APIs … update the corresponding docs"), please add `chart.svg` there too.

fiddur (commented):
🛑 Changes required

Small, clean change that adds a `chart.svg` sibling to the existing `chart.png` feed-image endpoint, serving the same heart-rate chart as raw `image/svg+xml` for Aurboda-native rendering.

The implementation is solid:
- It reuses `resolveImageWindow`, so the SVG path inherits the exact same eligibility/visibility/capability-token gate as the PNG path (the security-critical logic that's already unit-tested).
- It's fully wired at the single call site in `rest-routes.ts` (`renderChartSvg: buildChartSvg`), and `buildChartSvg` is the very function the PNG is rasterized from (`renderChartPng = svgToPng(buildChartSvg(...))`), so the two formats stay in perfect parity.
- Cache key (`chartsvg:`) is distinct from the PNG/route keys, and `no-store` preserves revocability.
- No XSS concern: the served SVG contains only server-generated numeric data and a default label (and `escapeXml` covers the label text anyway), so no user-controlled string reaches the markup on this path.

One blocking item: **the docs weren't updated.** `docs/features/feed.md` explicitly enumerates the public feed-image endpoints — both the code block (around lines 267–270) and the endpoint table (lines 386–387) list `chart.png` and `route.png` but not the new `chart.svg`. CLAUDE.md requires keeping `docs/` in sync when changing APIs, so please add `chart.svg` in both places (and it may be worth a sentence noting the SVG variant is for Aurboda-native crisp rendering while the PNG is what Mastodon/other peers attach).

Minor (non-blocking): the new route handler itself has no route-level test, but that matches the existing `chart.png`/`route.png` handlers (only the extracted pure helpers are tested), so it's not a regression.

apps/backend/src/routes/feed-image-router.ts
fiddur: Done — added `chart.svg` to docs/features/feed.md in both the endpoint code block and the public-endpoint table, plus a sentence noting the SVG is for Aurboda-native crisp rendering while the PNG is what Mastodon/peers attach (0401cc4c).

fiddur (commented):
✅ Approved

Clean, well-scoped addition. The new `GET /public/:username/feed/:postId/chart.svg` handler faithfully mirrors the existing `chart.png` handler: same eligibility gate via `resolveImageWindow`, same capability-token model for followers-only posts, same empty-series → 404 behaviour, and the same `no-store` revocability. It reuses the already-tested shared `buildChartSvg`, caches under a distinct key (`chartsvg:`) in the shared LRU, and stores the built string as UTF-8 bytes so it fits the existing buffer cache. The new required `renderChartSvg` dep is wired in the only `FeedImageDeps` consumer (`rest-routes.ts`), and `buildChartSvg`'s optional second param keeps the assignment type-safe. Serving `image/svg+xml` inline from the API origin is safe here because the SVG is entirely server-generated from numeric series data with the single label passed through `escapeXml` — no user-controlled markup reaches the output on this path. Docs in `feed.md` are updated consistently.

Minor, non-blocking: there's no route-level test that exercises the new handler end-to-end, but that matches the existing pattern (neither `chart.png` nor `route.png` has one either), and all the non-trivial logic it depends on — `resolveImageWindow`, `createRenderCache`, and `buildChartSvg` — is already well covered by unit tests.

Issue #936, opened 2026-07-14
Article posts (Slice B): correlation blocks


fiddur:
Part of #934 (Shareable analysis Articles). Depends on Slice A (#935).

## Goal

Add a **correlation** block type to articles: a scatter + coefficient (r) + n over a locked window, reusing the existing correlation engine and selectors (`list_correlation_selectors`, `get_metric_correlation`, `get_generic_correlation`, …).

## Scope

- Schema: `{ type: 'correlation', selector, window {start,end}, caption? }` in the article `blocks` union.
- Backend: resolve a correlation block live over its window via the correlation services; MCP+REST parity for adding/editing correlation blocks (falls out of the generic `update_article`).
- Web: render a correlation block (scatter + r + n + caption); compose UI to pick a correlation selector + window.
- Rendering: correlation scatter as native SVG on Aurboda, PNG for Mastodon/export (extends #901/#888).

🤖 Planned with [Claude Code](https://claude.com/claude-code)

fiddur:
Closing as complete: Slice B merged (correlation blocks — trigger/outcome selectors, lag, live Pearson/Spearman + scatter render, block images).

PR #943, opened 2026-07-14
feat(feed): article post kind — schema + JSONB persistence (#935)

fiddur:
First PR of **Slice A** (#935) for the shareable-analysis **Articles** epic (#934): the `article` feed-post kind at the **schema + storage layer**.

An article is a long-form post — a title, markdown prose, and inline **chart blocks** over locked `[start, end]` windows — modelled as a *kind* on the existing `feed_posts` table so it reuses the whole feed (visibility, federation, timeline, permalinks) instead of being a separate entity.

## api-spec (single source)

- `FeedPostKind` (`activity` | `article`).
- `ArticleProseBlock` / `ArticleChartBlock` → discriminated `ArticleBlock`; `ArticleContent` (`title` + optional `default_start`/`default_end` window + ordered `blocks`).
- Chart block = a `metric` + optional `start`/`end` override (inherits the article default window) + optional `bucket`/`caption`. Charts re-resolve **live** against the locked window — no snapshot. Correlation blocks are Slice B (#936).
- `CreateArticleBody` / `UpdateArticleBody`; `FeedPost` gains required `kind` and optional `article`.

## Backend

- `feed_posts` gains `kind VARCHAR(12) DEFAULT 'activity'` + `article JSONB` — idempotent additive migration for existing DBs, registered in the table-creation order.
- `db/feed.ts`: `FeedPostRecord`/`Row` carry `kind` + parsed `article`; new `createArticlePost`; `updateFeedPost` patches the whole `article` JSONB; `FeedPostPatch` gains `article`.
- `serializeFeedPost` surfaces `kind` + `article` (article posts carry no activity, so no resolved `content`).

## Scope

**Storage + schema foundation only.** The create/edit **REST endpoints + MCP tools** (parity) and **web rendering + editor** land in the following Slice A PRs. `docs/features/feed.md` documents the new kind and column.

## Test

`pnpm check` green (0 errors). Feed DB **integration tests** (testcontainers) cover: article create + round-trip (JSONB parse), whole-payload update (replace-not-merge, touches `updated_at`), and public-outbox inclusion/exclusion by visibility — plus a regression assert that an activity share defaults to `kind: 'activity'` with `article: null`.

### DIFF
--- a/packages/api-spec/src/schemas/feed.ts
+++ b/packages/api-spec/src/schemas/feed.ts
@@ -100,6 +100,87 @@ export const updateFeedPostBodySchema = z
 
 export type UpdateFeedPostBody = z.infer<typeof updateFeedPostBodySchema>
 
+// =============================================================================
+// Article posts (long-form: title + markdown prose + inline chart blocks)
+// =============================================================================
+
+/**
+ * The kind of a feed post. `activity` (the default) shares one of the user's
+ * activities; `article` is a long-form post carrying markdown prose and inline
+ * chart blocks over locked time windows. Modelled as a post *kind* rather than a
+ * separate entity so articles reuse the whole feed (visibility, federation, home
+ * timeline, the public-profile feed, permalinks).
+ */
+export const feedPostKindSchema = z
+  .enum(['activity', 'article'])
+  .meta({ description: 'Feed post kind: an activity share or a long-form article', id: 'FeedPostKind' })
+
+export type FeedPostKind = z.infer<typeof feedPostKindSchema>
+
+/** A prose block: a run of markdown, rendered through the shared sanitiser (#910). */
+export const articleProseBlockSchema = z
+  .object({
+    markdown: z
+      .string()
+      .max(20_000)
+      .meta({ description: 'Markdown prose (rendered to sanitised HTML at the sink)' }),
+    type: z.literal('prose'),
+  })
+  .meta({ description: 'A markdown prose block', id: 'ArticleProseBlock' })
+
+/**
+ * A chart block: one shared metric drawn over a locked `[start, end]` window.
+ * The window is what makes the chart stable and citable — data is re-resolved
+ * live against it, never snapshotted. When `start`/`end` are omitted the block
+ * inherits the article's default window.
+ */
+export const articleChartBlockSchema = z
+  .object({
+    bucket: z
+      .string()
+      .regex(/^\d+[smhd]$/, 'Must be {number}{unit} where unit is s, m, h, or d')
+      .optional()
+      .meta({
+        description: 'Bucket granularity (e.g. `5s`, `1h`, `1d`); the server picks a default if omitted',
+      }),
+    caption: z.string().max(280).optional().meta({ description: 'Optional caption shown under the chart' }),
+    end: iso8601DateTimeSchema
+      .optional()
+      .meta({ description: "Window end (ISO 8601); inherits the article's default window if omitted" }),
+    metric: metricTypeSchema.meta({ description: 'The metric to chart' }),
+    start: iso8601DateTimeSchema
+      .optional()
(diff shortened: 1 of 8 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped storage + schema foundation for the `article` post kind: a `kind` discriminator + nullable `article` JSONB column, the api-spec schemas, `createArticlePost`/`updateFeedPost` persistence, serializer wiring, docs, and solid integration tests covering create/round-trip/update/visibility.

- The migration `feed_posts_article_columns` follows the established idempotent `feed_follower_columns` pattern exactly (multi-statement `ADD COLUMN IF NOT EXISTS` via the simple-query protocol with no params). Its placement in `tableCreationOrder` (feed_posts → article_columns → indexes) is correct, and because the pseudo-key is never a real table name it correctly re-runs on every migration for existing DBs.
- `serializeFeedPost` is the single `FeedPost` constructor across REST/MCP/outbox, so the newly-required `kind` field is covered everywhere; the only inline `FeedPostRecord` fixture (`makePost`) was updated with `kind`/`article`. No OpenAPI `id` collisions.

Two forward-looking notes (non-blocking — no production path can create an article yet, since `createArticlePost` and the new body schemas have no route/MCP wiring in this slice):
- **Outbox count/items divergence.** Public/unlisted articles now appear in `listPublicFeedPosts`/`countPublicFeedPosts`, but the outbox dispatcher skips `activity_id == null` posts (`federation.ts:417`). Unlike the soft-deleted-activity case that self-heals, a persistent public article would leave `totalItems` permanently over-counting the emitted items. The federation follow-up should either filter articles from the outbox until they render, or emit an article object.
- **Chart-block authorization.** Article chart blocks embed an arbitrary metric series over a locked window with no equivalent of the `series_metrics` opt-in that this file's header calls "the authorization boundary." That's fine as deliberate author-driven sharing, but the follow-up rendering endpoint must serve only the owner's data and honor the post's visibility (a `followers`-only article's charts must not be publicly fetchable).

packages/api-spec/src/schemas/feed.ts
```diff
@@ -100,6 +100,87 @@ export const updateFeedPostBodySchema = z
...
+      .regex(/^\d+[smhd]$/, 'Must be {number}{unit} where unit is s, m, h, or d')
+      .optional()
+      .meta({
+        description: 'Bucket granularity (e.g. `5s`, `1h`, `1d`); the server picks a default if omitted',
+      }),
+    caption: z.string().max(280).optional().meta({ description: 'Optional caption shown under the chart' }),
+    end: iso8601DateTimeSchema
+      .optional()
+      .meta({ description: "Window end (ISO 8601); inherits the article's default window if omitted" }),
+    metric: metricTypeSchema.meta({ description: 'The metric to chart' }),
```
fiddur, line 150: Forward-looking (non-blocking for this schema-only slice): a chart block lets the author embed any metric series over a locked window, with no equivalent of the `series_metrics` opt-in that this file's header (lines 15-18) calls "the authorization boundary." That's fine as deliberate author-chosen sharing, but make sure the follow-up chart-rendering endpoint serves only the article owner's data and honors the post's visibility — a `followers`-only or `unlisted` article's charts shouldn't be publicly fetchable by URL.

fiddur: Good call — noted as a hard requirement for the chart-rendering slice. The article chart endpoint will mirror the existing `chart.png`/`chart.svg` gate exactly: served only for the article owner's own data, and gated by `resolveImageWindow`-style visibility — `public`/`unlisted` fetchable, a `followers`-only article's charts reachable only via the post's unguessable capability `?token=`, with `no-store` revocability. Resolving as this PR is schema/storage-only; the constraint is captured for A2/A3.

PR #944, opened 2026-07-14
feat(feed): article authoring API — MCP + REST parity (#935)

fiddur:
Second PR of **Slice A** (#935): create + edit **article** posts over both MCP and REST from one shared service (parity), building on the storage model from #943.

## Service (`services/article.ts`, pure + unit-tested)

- **`buildArticleContent`** — resolves each chart block's **effective window** (its own `start`/`end` override, else the article `default_start`/`default_end`) and rejects a chart block with **no bounded window** or **start ≥ end**, at the write boundary (so a malformed article never reaches render time).
- **`mergeArticleContent`** — applies an update patch onto the stored article: provided fields replace, omitted fields keep.

## Parity

- **MCP:** `create_article` / `update_article` (via api-spec `.shape`), mirroring `share_activity` / `update_feed_post`.
- **REST:** `POST /feed/articles`, `PATCH /feed/articles/:postId` (typed router, api-spec-validated), registered before the generic `/:postId` routes. Delete reuses `DELETE /:postId`. A malformed article → **400**; editing a non-article post via the article route → **404**.

## Not in this PR (deliberately)

**No federation.** Articles are authored and shown on the owner's own feed but do not fan out to followers (the `deliver` hooks are unwired for articles). Every activity-assuming outbox / object / structured-post path already guards `activity_id == null`, so a public article is simply skipped there for now — **verified**, no regression. Federation as an AS2 `Article` + Reddit export is Slice C (#937); web rendering + editor is the next Slice A PR.

## Test

- Article **service unit tests** (11): window inheritance, prose-only, windowless-chart rejection, start≥end, 1-based block index, merge semantics.
- **feed-router integration test** (6, testcontainers + supertest): create + round-trip, the 400 windowless-chart rejection, patch-replaces-fields (default window preserved), 404s (unknown id / non-article post), and delete.
- `pnpm check` green (0 errors). Docs updated (API table + MCP tool list + the article note).

### DIFF
--- a/apps/backend/src/db/index.ts
+++ b/apps/backend/src/db/index.ts
@@ -318,7 +318,9 @@ export { emitTimelineNotify, openTimelineChannel } from './timeline-notify.ts'
 
 // Feed posts
 export {
+  type ArticlePostInput,
   countPublicFeedPosts,
+  createArticlePost,
   createFeedPost,
   deleteFeedPost,
   type FeedPostInput,
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -4,10 +4,12 @@
  * and `/feed/following` capabilities.
  */
 import {
+  createArticleBodySchema,
   followActorBodySchema,
   followersQuerySchema,
   shareActivityBodySchema,
   timelineQuerySchema,
+  updateArticleBodySchema,
   updateFeedPostBodySchema,
 } from '@aurboda/api-spec'
 import { z } from 'zod'
@@ -17,6 +19,7 @@ import type { FollowerActions } from '../services/followers.ts'
 import type { FollowActions } from '../services/following.ts'
 
 import {
+  createArticlePost,
   createFeedPost,
   deleteFeedPost,
   getActivityById,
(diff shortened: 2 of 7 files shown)


fiddur (commented):
✅ Approved

Solid slice. The new `article` service cleanly isolates the pure window-resolution/validation logic (`chartBlockWindow` / `buildArticleContent` / `mergeArticleContent`) from the REST and MCP call sites, which keeps the create/update parity trivially symmetric — both build content the same way, validate through `buildArticleContent`, and pass `visibility` separately. Route ordering is correct (`/articles/:postId` PATCH registered before the generic `/:postId`), the merge semantics (omitted field = unchanged) are honoured by `updateFeedPost`'s per-column `!== undefined` guards, and the 404 gating on `kind !== 'article'` is consistent across REST and MCP. Test coverage is good: service unit tests cover per-edge window inheritance, the inverted-window and missing-window rejections, and the 1-based block index in errors; the integration tests exercise create, patch-replace, both 404 paths, and delete against a real DB. Docs are updated in step.

Two minor, non-blocking observations for the follow-up federation slice:

- **Delete of an article fans out a spurious `Delete`.** The article routes intentionally skip `deliver` (create/update never federate yet), but delete reuses the generic `DELETE /:postId`, which calls `deliver?.deleted` → `deliverFeedDelete` unconditionally. Deleting an article therefore sends a `Delete{Tombstone}` to followers for an object whose `Create` was never delivered. It's harmless in practice (AP peers ignore Deletes for unknown objects), but worth reconciling when article federation lands.

- **Article-level default window ordering isn't validated when unused.** `buildArticleContent` only checks `start < end` for windows a chart block actually resolves to. A prose-only article (or one whose chart blocks all override the window) can persist an inverted `default_start`/`default_end`; it would only be rejected later once a block inherits it. This matches the stated contract ("every chart block ends up with a sane bounded window") so it's arguably fine, just noting the asymmetry.

PR #945, opened 2026-07-14
feat(web): article rendering + composer (#935)

fiddur:
Third and final **Slice A** PR: view and author **article** posts in the web app, building on the article API from #944. Completes #935.

## Render (feed / home-timeline card)

- `FeedPostCard` gains a `kind === 'article'` branch → **`ArticleContent`**: the title followed by the ordered blocks. **Prose** renders through the shared DOMPurify sanitiser (#910 — the XSS boundary for AI- or other-authored prose).
- **`ArticleChartBlock`** resolves each chart block's effective window (own override else the article default) and draws the metric **live** via the owner-authenticated bucketed-metrics endpoint + the shared `TrendLineChart` — no snapshot, matching "lock the window, not the data". Because it's owner-scoped, a post's chart data stays behind the same auth until the public/federated surface (Slice C) adds the visibility-gated **server** rendering flagged in #943's review.

## Compose / edit

- **`ArticleEditorDialog`** — title, optional default window, and an ordered **add / remove / reorder** list of **prose** (`MarkdownEditor`) and **chart** (`MetricPicker` + per-block window + caption) blocks, plus the reused `VisibilitySelector`. Create → `POST /feed/articles`, edit → `PATCH /feed/articles/:postId`.
- Feed page: a **New article** button and an **Edit** affordance on article cards.
- API client: `createArticle` / `updateArticle`.

## Testability

The pure state→body logic — `buildArticleBody` (+ validation), `datetime-local` ⇄ ISO conversion, edit-mode seeding, submit-error derivation — lives in **`article-editor-body.ts`** so it is unit-tested without a render harness (**11 tests**), matching the repo's testable-without-mocking preference and the web's pure-logic test convention.

## Scope

Owner-facing rendering + authoring. **Federation** as an AS2 `Article` (prose HTML + attached chart PNGs / native inline) and **Reddit/markdown export** are Slice C (#937); a public article's charts don't yet render for unauthenticated viewers (that's the Slice C visibility-gated server endpoint).

## Test

`pnpm check` green (0 errors). Article editor logic: 11 unit tests. Docs updated (web usage + the article note).

### DIFF
--- a/apps/web/src/components/ArticleEditorDialog.tsx
+++ b/apps/web/src/components/ArticleEditorDialog.tsx
@@ -1,7 +1,7 @@
 import type { FeedPost, FeedVisibility } from '@aurboda/api-spec'
 
 import { useMutation, useQueryClient } from '@tanstack/react-query'
-import { useState } from 'preact/hooks'
+import { useRef, useState } from 'preact/hooks'
 
 /**
  * Compose or edit a long-form **article** feed post: a title, an optional
@@ -22,6 +22,14 @@ import './ShareActivityDialog.css'
 import './ArticleEditorDialog.css'
 
 type BlockPatch = Partial<ChartDraft> & Partial<ProseDraft>
+
+/**
+ * A draft block plus a stable identity, used only as the React `key`. Blocks are
+ * reorderable and their children (MarkdownEditor) hold internal tab/focus state,
+ * so a positional key would strand that state on the wrong block after a move —
+ * hence a per-block id that travels with the block instead of the array index.
+ */
+type KeyedBlock = { key: string; block: BlockDraft }
 
 /** One editable content block (prose or chart) with its reorder / remove controls. */
 const ArticleBlockEditor = ({
@@ -121,12 +129,16 @@ export const ArticleEditorDialog = ({ post, onClose }: { post?: FeedPost; onClos
   const [title, setTitle] = useState(initial.title)
   const [defaultStart, setDefaultStart] = useState(initial.defaultStart)
   const [defaultEnd, setDefaultEnd] = useState(initial.defaultEnd)
-  const [blocks, setBlocks] = useState<BlockDraft[]>(initial.blocks)
+  const nextKey = useRef(0)
+  const keyed = (block: BlockDraft): KeyedBlock => ({ block, key: String(nextKey.current++) })
+  const [blocks, setBlocks] = useState<KeyedBlock[]>(() => initial.blocks.map(keyed))
   const [visibility, setVisibility] = useState<FeedVisibility>(initial.visibility)
   const [validationError, setValidationError] = useState<string | null>(null)
 
   const patchBlock = (i: number, patch: BlockPatch) =>
-    setBlocks((bs) => bs.map((b, j) => (j === i ? ({ ...b, ...patch } as BlockDraft) : b)))
+    setBlocks((bs) =>
+      bs.map((kb, j) => (j === i ? { ...kb, block: { ...kb.block, ...patch } as BlockDraft } : kb)),
+    )
   const removeBlock = (i: number) => setBlocks((bs) => bs.filter((_, j) => j !== i))
   const moveBlock = (i: number, dir: -1 | 1) =>
     setBlocks((bs) => {
@@ -136,13 +148,22 @@ export const ArticleEditorDialog = ({ post, onClose }: { post?: FeedPost; onClos
       ;[next[i], next[j]] = [next[j], next[i]]
       return next
     })
-  const addProse = () => setBlocks((bs) => [...bs, { markdown: '', type: 'prose' }])
+  const addProse = () => setBlocks((bs) => [...bs, keyed({ markdown: '', type: 'prose' })])
   const addChart = () =>
-    setBlocks((bs) => [...bs, { bucket: '', caption: '', end: '', metric: '', start: '', type: 'chart' }])
+    setBlocks((bs) => [
+      ...bs,
+      keyed({ bucket: '', caption: '', end: '', metric: '', start: '', type: 'chart' }),
+    ])
 
   const mutation = useMutation({
     mutationFn: () => {
-      const result = buildArticleBody({ blocks, defaultEnd, defaultStart, title, visibility })
+      const result = buildArticleBody({
+        blocks: blocks.map((kb) => kb.block),
+        defaultEnd,
+        defaultStart,
+        title,
+        visibility,
+      })
       if (!result.ok) {
         setValidationError(result.error)
         return Promise.reject(new Error('validation'))
@@ -207,10 +228,10 @@ export const ArticleEditorDialog = ({ post, onClose }: { post?: FeedPost; onClos
         </fieldset>
 
         <div class="article-blocks">
-          {blocks.map((block, i) => (
+          {blocks.map((kb, i) => (
             <ArticleBlockEditor
-              key={i}
-              block={block}
+              key={kb.key}
+              block={kb.block}
               index={i}
               total={blocks.length}
               onPatch={(patch) => patchBlock(i, patch)}
--- a/apps/web/src/components/ArticleEditorDialog.tsx
+++ b/apps/web/src/components/ArticleEditorDialog.tsx
@@ -1,0 +1,257 @@
+import type { FeedPost, FeedVisibility } from '@aurboda/api-spec'
+
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+import { useState } from 'preact/hooks'
+
+/**
+ * Compose or edit a long-form **article** feed post: a title, an optional
+ * article-level default time window, and an ordered list of prose / chart blocks.
+ * Passing an existing `post` (kind `article`) switches to edit mode (PATCH);
+ * otherwise it creates one (POST). Prose is written in the shared MarkdownEditor;
+ * a chart block picks a metric and an optional window (inheriting the article
+ * default when left blank) — the same shape Claude drafts over MCP.
+ */
+import type { BlockDraft, ChartDraft, ProseDraft } from './article-editor-body'
+
+import { createArticle, updateArticle } from '../state/api'
+import { buildArticleBody, deriveSubmitError, initialArticleEditorState } from './article-editor-body'
+import { MarkdownEditor } from './MarkdownEditor'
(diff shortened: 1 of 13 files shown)


apps/web/src/components/ArticleEditorDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <MarkdownEditor
+        value={block.markdown}
+        onChange={(v) => onPatch({ markdown: v })}
+        placeholder="Write your analysis (markdown supported)…"
+      />
+    ) : (
+      <div class="article-chart-fields">
+        <label class="article-field">
+          <span class="article-field-label">Metric</span>
+          <MetricPicker value={block.metric} onChange={(m) => onPatch({ metric: m })} />
```
fiddur: `MetricPicker` offers the user's custom metrics, but article chart blocks only accept built-in metrics: `articleChartBlockSchema.metric` is the strict `metricTypeSchema` enum, and `buildArticleBody` rejects anything failing `isValidMetric` (built-ins only). So selecting a custom metric here fails with the misleading "Pick a metric for chart block N" (article-editor-body.ts:102), and the server would 400 it as well. Consider excluding custom metrics from the picker in this context, or surfacing a clearer "custom metrics aren't supported in article charts" message.

fiddur (commented):
✅ Approved

Solid, well-structured slice. The pure state→body logic is cleanly split out of the JSX (`article-editor-body.ts`) and unit-tested, the `datetime-local ⇄ ISO` round-trip is timezone-independent, and prose blocks correctly go through the shared `renderMarkdown` sanitiser at every `dangerouslySetInnerHTML` sink (`ArticleContent`, `MarkdownEditor` preview) — the XSS boundary is respected. Edit vs. compose mode, chart-window inheritance (block override → article default → "no time window" fallback), and the live re-resolution of chart data all look correct. A few non-blocking notes:

- **Custom metrics are silently unsupported in chart blocks** (inline). The shared `MetricPicker` lists the user's custom metrics, but `articleChartBlockSchema.metric` is the strict built-in `metricTypeSchema` and `buildArticleBody` gates on `isValidMetric` (built-ins only) — so picking a custom metric yields the misleading error "Pick a metric for chart block N" even though one was picked, and the server would 400 it too. Consider filtering custom metrics out of the picker here or giving a clearer message.
- **A stored default window can't be cleared via the editor.** In edit mode, clearing the From/To default-window fields makes `buildArticleBody` omit `default_start`/`default_end`; `mergeArticleContent` treats an omitted field as "keep", so the previously-stored window persists. This is inherent to the PATCH keep-on-omit design (shared with the MCP/REST path), but the composer presents the fields as clearable, so it can surprise. Worth a follow-up if clearing needs to be supported.
- **Minor:** block reorder/remove uses `key={i}` (array index). For two same-type blocks, transient child UI state (MarkdownEditor's write/preview tab, MetricPicker's open/query state) stays with the position rather than the content after a move. Low impact since all persisted state is lifted up.

apps/web/src/components/ArticleEditorDialog.tsx
fiddur: Fixed in b7af74f: added a `builtinOnly` prop to `MetricPicker` — the article editor now passes it so only built-in `MetricType`s are offered, matching `articleChartBlockSchema`. Also sharpened `buildArticleBody` to distinguish an empty pick ("Pick a metric…") from an unsupported custom metric (clear "custom metrics aren't supported yet" message) for the case where an article edited elsewhere carries one.

apps/web/src/state/api/feed.ts
```diff
@@ -48,6 +50,24 @@ export const updateFeedPost = async (postId: string, body: UpdateFeedPostBody):
   return response.data.post
 }
 
+/** Publish a long-form article (title + prose + inline chart blocks) to the feed. */
+export const createArticle = async (body: CreateArticleBody): Promise<FeedPost> => {
```
fiddur, line 54: `createArticle` (and `updateArticle` below) let the raw `AxiosError` bubble up, so on a 400 the caller sees `error.message === "Request failed with status code 400"` — not the server's `{ error }` body. The article routes return specific window-validation messages (`buildArticleContent`: *"Chart block N (metric) has start on or after end"*, *"…has no time window…"*), and since `buildArticleBody` deliberately skips window validation, those are the expected errors for common author mistakes. `deriveSubmitError` then surfaces the generic HTTP string instead of the actionable message. This file already has `apiErrorMessage(error, fallback)` (used by `followActor`) for exactly this — wrap the axios call in try/catch and rethrow `new Error(apiErrorMessage(error, 'Couldn’t save the article. Please try again.'))` in both functions.

apps/web/src/components/ArticleEditorDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+                onInput={(e) => setDefaultEnd((e.target as HTMLInputElement).value)}
+              />
+            </label>
+          </div>
+        </fieldset>
+
+        <div class="article-blocks">
+          {blocks.map((block, i) => (
+            <ArticleBlockEditor
+              key={i}
```
fiddur: Non-blocking: using the array index as `key` on a reorderable/removable list whose children (`MarkdownEditor`) carry internal Write/Preview tab state means reordering or removing same-typed blocks can leave the wrong tab state (and textarea focus) on a block — the classic index-as-key footgun. Content is unaffected (it's prop-driven), so this is minor, but a stable per-block id in the draft state would avoid it.

fiddur (commented):
🛑 Changes required

Solid, well-structured slice: the state→body logic is extracted into a pure, unit-tested module (`article-editor-body.ts`), prose is routed through the shared markdown sanitiser, and chart blocks re-resolve live over their locked window — all consistent with the documented model.

One functional issue in the error path: `createArticle` / `updateArticle` in `state/api/feed.ts` let the raw `AxiosError` propagate on a non-2xx response. For a 400 the resulting `error.message` is the generic `"Request failed with status code 400"`, not the server's `{ error }` body. Because `buildArticleBody` intentionally delegates chart-window validation to the server (`buildArticleContent`), the most common author mistakes — a chart block with no window, or start ≥ end — come back with specific, actionable messages (e.g. *"Chart block 1 (heart_rate) has start on or after end."*) that the composer then throws away, showing the generic HTTP string via `deriveSubmitError` instead. The same file already has `apiErrorMessage()` (used by `followActor`) for exactly this; wiring it into the two article calls would surface the server's reason. See the inline comment.

Minor (non-blocking): the block list uses the array index as `key`, and `ArticleBlockEditor` renders a `MarkdownEditor` that holds internal Write/Preview tab state — so reordering/removing same-typed blocks can leave that tab state (and textarea focus) attached to the wrong block. Content stays correct since it's prop-driven; only the local tab/focus is affected.

apps/web/src/state/api/feed.ts
fiddur: Fixed in 4b66e68: both `createArticle` and `updateArticle` now wrap the axios call and rethrow `new Error(apiErrorMessage(error, …))` (same pattern as `followActor`), so the server’s window-validation reason (e.g. "Chart block 1 (heart_rate) has start on or after end.") reaches the composer via `deriveSubmitError` instead of the generic HTTP string.

apps/web/src/components/ArticleEditorDialog.tsx
fiddur: Fixed in 4b66e68: blocks are now keyed by a stable per-block id (`KeyedBlock`) that travels with the block through reorder/remove, instead of the array index — so the MarkdownEditor’s Write/Preview tab state and focus stay attached to the right block. The id lives only in the dialog’s local state, so the pure `article-editor-body` module and its tests are untouched.

apps/web/src/components/article-editor-body.ts
```diff
@@ -0,0 +1,131 @@
...
+      ...(b.caption.trim() ? { caption: b.caption.trim() } : {}),
+    })
+  }
+
+  return {
+    body: {
+      blocks,
+      title: state.title.trim(),
+      visibility: state.visibility,
+      ...(toIso(state.defaultStart) ? { default_start: toIso(state.defaultStart) } : {}),
```
fiddur, line 126: Blank default-window fields are omitted here, and the server's `mergeArticleContent` uses `patch.default_start ?? existing.default_start`, so in edit mode a user can't *clear* a default window they previously set — blanking these inputs silently keeps the old value. Non-blocking, but if clearing should be possible you'll need to distinguish "omit" from "set to null" (e.g. send `null`) on the update path.

fiddur (commented):
✅ Approved

Solid, well-structured slice. The pure state→body logic is nicely extracted into `article-editor-body.ts` and unit-tested (mirroring the `buildShareBody` pattern), the JSX stays thin, and the composer/renderer split is clean. Prose goes through the shared markdown sanitiser and charts resolve live over their locked window, matching the article model's design. Type usage aligns with api-spec (`CreateArticleBody`/`UpdateArticleBody`, `isValidMetric`), and the `MetricPicker` `builtinOnly` refactor preserves behaviour while dropping duplicated display-name logic. Reorder keys correctly travel with blocks so `MarkdownEditor` focus/tab state doesn't strand on a move.

One non-blocking behaviour gap: because the editor omits blank `default_start`/`default_end` on save and the server merges with `patch.default_start ?? existing.default_start` (`services/article.ts`), a user editing an article can't actually *clear* a previously-set default window — blanking the fields silently keeps the old value. Worth a follow-up (send an explicit `null`, or document the limitation). Minor: in `builtinOnly` mode `MetricPicker` still fires the `customMetrics` query even though the results are discarded.

Nothing here blocks merge.

apps/web/src/components/article-editor-body.ts
fiddur: Good catch — confirmed: in edit mode, blanking a previously-set default window is a silent no-op. Deferring as a follow-up rather than expanding this approved web-render PR: a proper fix is cross-cutting (`updateArticleBodySchema` must permit `null`, `mergeArticleContent` must treat `null` as "clear" vs `undefined` as "keep", and the web build must send `null` only on the update path). It also interacts with block windows — clearing the article default is only valid when every chart block sets its own window, otherwise the server rightly rejects the now-windowless blocks in `buildArticleContent`. Tracking it for a dedicated follow-up so the null-clear semantics get their own tests. Resolving here.

PR #946, opened 2026-07-14
feat(feed): article correlation blocks — schema + validation (#936)

fiddur:
Part of #934 (Shareable analysis Articles), **Slice B (#936)** — first PR (schema + backend + lossless web integration).

Adds a **correlation** article block: a scatter of a `trigger` predictor against an `outcome` over a locked `[start, end]` window, reusing the exploratory correlation engine's selectors (`CorrelationSelector`) plus an optional `lag_days`. Like a chart block, it inherits the article's default window and re-resolves **live** (no snapshot).

## What's here

**Schema + backend**
- **api-spec** — `articleCorrelationBlockSchema` (`{ type: 'correlation', trigger, outcome, lag_days?, start?, end?, caption? }`) added to the `articleBlockSchema` discriminated union; imports `selectorSchema` from `correlations.ts` (no import cycle — correlations imports only `common`). This gives **MCP↔REST parity automatically**: `create_article`/`update_article` and `POST`/`PATCH /feed/articles` both consume the blocks union, so they accept correlation blocks with no new tool or route.
- **backend** — generalized `chartBlockWindow` → `blockWindow` and extended `buildArticleContent` to validate correlation blocks' effective windows (bounded, `start < end`) via a shared per-block error label.

**Web (lossless, no new UI yet)**
Adding a variant to the shared union forces web to handle it. To keep web compiling **without losing data**:
- `article-editor-body` — a `CorrelationDraft` passthrough carries the stored block verbatim, so `draftsFromPost` → editor → `buildArticleBody` round-trips it. Editing an MCP-authored article no longer risks dropping a correlation block.
- `ArticleEditorDialog` — a correlation block shows as a read-only row (reorder/remove work; a "🔗 Correlation" chip + note).
- `ArticleContent` — correlation blocks are skipped in render for now.

**Tests** — 15 backend `buildArticleContent` cases (incl. 4 correlation-window ones) + 2 web round-trip cases for the passthrough. **Docs** — `docs/features/feed.md` article section + API table updated.

Feed article-block schemas aren't part of the generated OpenAPI/Kotlin (feed endpoints aren't in `openapi.ts`), so `pnpm generate` produced no diff — same as Slice A's #943.

## Follow-ups in this slice

- **Render** — extract the `ScatterPlot` from the Correlations explorer into a reusable component and add `ArticleCorrelationBlock` (client-side authed `fetchContinuousCorrelation` over the block's effective window → scatter + r/ρ/n + caption); wire it into `ArticleContent`.
- **Composer** — replace the read-only editor row with real trigger/outcome selector pickers (+ window + lag), populated from `GET /correlations/selectors`.

### DIFF
--- a/apps/backend/src/services/article.test.ts
+++ b/apps/backend/src/services/article.test.ts
@@ -2,10 +2,18 @@ import type { ArticleContent } from '@aurboda/api-spec'
 
 import { describe, expect, test } from 'vitest'
 
-import { buildArticleContent, chartBlockWindow, mergeArticleContent } from './article.ts'
+import { blockWindow, buildArticleContent, mergeArticleContent } from './article.ts'
 
 const WEEK_START = '2026-07-01T00:00:00.000Z'
 const WEEK_END = '2026-07-07T00:00:00.000Z'
+
+/** A valid correlation block: exercise (trigger) against sleep score (outcome), with an optional window. */
+const corrBlock = (window: { start?: string; end?: string } = {}): ArticleContent['blocks'][number] => ({
+  outcome: { kind: 'metric', metric: 'sleep_score' },
+  trigger: { kind: 'activity', pattern: 'exercise' },
+  type: 'correlation',
+  ...window,
+})
 
 const content = (overrides: Partial<ArticleContent> = {}): ArticleContent => ({
   blocks: [],
@@ -15,9 +23,9 @@ const content = (overrides: Partial<ArticleContent> = {}): ArticleContent => ({
   ...overrides,
 })
 
-describe('chartBlockWindow', () => {
+describe('blockWindow', () => {
   test('uses the block override when present', () => {
-    const win = chartBlockWindow(
+    const win = blockWindow(
       { end: '2026-07-03T00:00:00.000Z', start: '2026-07-02T00:00:00.000Z' },
       { default_end: WEEK_END, default_start: WEEK_START },
     )
@@ -26,7 +34,7 @@ describe('chartBlockWindow', () => {
 
   test('falls back to the article default per-edge', () => {
     // Only `start` is overridden; `end` inherits the article default.
-    const win = chartBlockWindow(
+    const win = blockWindow(
       { start: '2026-07-02T00:00:00.000Z' },
       { default_end: WEEK_END, default_start: WEEK_START },
     )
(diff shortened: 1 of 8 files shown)


fiddur (commented):
✅ Approved

Clean, additive schema-and-validation slice for article **correlation blocks**. The changes hang together well and stay within the documented scope ("render + composer UI land in a follow-up").

What I verified locally:
- `@aurboda/api-spec` builds; `aurboda-backend` and `aurboda-web` both typecheck (`tsgo --noEmit`) with exit 0.
- Unit tests pass: backend `article.test.ts` (15) and web `article-editor-body.test.ts` (14).
- No circular import from `feed.ts` → `correlations.ts` (correlations doesn't import feed); `selectorSchema` is already registered, so the new `ArticleCorrelationBlock` OpenAPI model references an existing `CorrelationSelector`.
- MCP/REST parity is preserved — both paths go through the shared `articleBlockSchema` (validation) and `buildArticleContent`, which now covers `correlation` blocks identically to `chart` blocks.
- Backward compatible: existing prose/chart articles remain valid; the discriminated union just gains a third arm.

Design notes (non-blocking):
- `article.ts` nicely generalises `chartBlockWindow` → `blockWindow` + `blockLabel`, and the validation loop now guards both windowed kinds. The `blockLabel` fallback (`else` ⇒ "Correlation block") is only reachable for chart/correlation because the loop `continue`s on other types, so it's correct.
- The web composer keeps correlation blocks as a verbatim passthrough (`CorrelationDraft`) — a sensible way to round-trip (reorder/remove) a block the UI can't yet author without dropping it. The lossless round-trip is covered by a test.
- `ArticleContent.tsx` skips correlation blocks (`return null`). The editor shows a read-only note for them, so on the public/feed render a correlation block is silently invisible until the follow-up scatter lands. This is explicitly documented as deferred; a small placeholder (like the existing "This chart has no time window" note) could be friendlier, but that's a follow-up UX nicety, not a defect.
- `lag_days` is an unbounded `int`, consistent with the existing `continuousCorrelationBodySchema.lag_days`, so no new concern.
- The `docs/features/feed.md` description reads as the target design with an italic caveat listing what's still pending "within this slice (#936)"; since #936 spans multiple PRs, that roadmap framing is internally consistent.

No blocking issues found.

PR #947, opened 2026-07-14
feat(feed): render article correlation blocks as a live scatter (#936)

fiddur:
Part of #934 (Shareable analysis Articles), **Slice B (#936)** — second PR (web render). Builds on the schema + backend from #946.

Correlation article blocks now **render** on the owner's feed as a live scatter.

## What's here

- **`ArticleCorrelationBlock`** (`pages/Feed/ArticleCorrelationBlock.tsx`) — resolves the block's effective window (own override else the article default), fetches the continuous daily correlation **live** over that window via the owner-authenticated `POST /correlations/continuous`, and draws a compact in-card scatter: points, an OLS regression line, and an `r · ρ · n · p` annotation with selector-labelled axes. Loading / error / "not enough overlapping data" states mirror `ArticleChartBlock`.
- **`ArticleContent`** — renders correlation blocks (replacing the interim skip from #946), with the same "no time window" fallback as chart blocks.
- **CSS** — `.article-scatter*` styles (light + dark) alongside the existing `.article-chart*`.

## Notes

- **DRY** — reuses the shared, unit-tested correlation maths (`linearRegression`, `describeSelectorAxis` from the explorer's `exploreCharts.ts`). The SVG is a purpose-built compact render rather than the explorer's full scatter/box-plot component; unifying the two SVGs would entangle the box-plot mode and the explorer's CSS, so it's left as a possible future cleanup.
- **Live, owner-scoped** — data is re-resolved live (no snapshot). The fetch is owner-authenticated, so this renders only for the article owner; **public/federated correlation rendering is deferred to Slice C (#937)** alongside the public chart-image work — consistent with how chart blocks render client-side today.
- **Tests** — following the codebase convention, fetch/render components (`ArticleChartBlock`) carry no unit test; the reused maths is covered by `exploreCharts.test.ts`.

## Follow-up in this slice

- **Composer** — trigger/outcome selector pickers (+ window + lag) in the article editor, replacing the read-only passthrough row from #946, populated from `GET /correlations/selectors`.

### DIFF
--- a/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
+++ b/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
@@ -38,12 +38,14 @@ const CorrelationScatterSvg = ({ data }: { data: ContinuousCorrelationData }) =>
   const sy = (y: number) => H - PAD.bottom - ((y - yMin) / (yMax - yMin || 1)) * (H - PAD.top - PAD.bottom)
   const reg = linearRegression(xs, ys)
 
+  const label = `Scatter of ${describeSelectorAxis(data.trigger)} versus ${describeSelectorAxis(data.outcome)}; r=${fmt(data.pearson)}, n=${data.n}`
+
   return (
-    <svg viewBox={`0 0 ${W} ${H}`} class="article-scatter" role="img">
+    <svg viewBox={`0 0 ${W} ${H}`} class="article-scatter" role="img" aria-label={label}>
       <line x1={PAD.left} y1={H - PAD.bottom} x2={W - PAD.right} y2={H - PAD.bottom} stroke="#ccc" />
       <line x1={PAD.left} y1={PAD.top} x2={PAD.left} y2={H - PAD.bottom} stroke="#ccc" />
       {data.series.map((p) => (
-        <circle cx={sx(p.trigger)} cy={sy(p.outcome)} r={3} fill={POINT_COLOR} opacity={0.45} />
+        <circle key={p.date} cx={sx(p.trigger)} cy={sy(p.outcome)} r={3} fill={POINT_COLOR} opacity={0.45} />
       ))}
       {reg && (
         <line
@@ -74,8 +76,13 @@ const CorrelationScatterSvg = ({ data }: { data: ContinuousCorrelationData }) =>
   )
 }
 
-/** ISO instant → inclusive `YYYY-MM-DD` regime bound for the correlation query. */
-const toDay = (iso: string): string => new Date(iso).toISOString().slice(0, 10)
+/**
+ * ISO instant → inclusive `YYYY-MM-DD` regime bound for the correlation query.
+ * Take the date straight off the ISO string (every ISO 8601 datetime starts with
+ * `YYYY-MM-DD`) so the bound is the author's wall-clock day — not a UTC day that
+ * could roll back for a window authored with a non-UTC offset near midnight.
+ */
+const toDay = (iso: string): string => iso.slice(0, 10)
 
 export const ArticleCorrelationBlock = ({
   trigger,
--- a/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
+++ b/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
@@ -38,7 +38,14 @@ const CorrelationScatterSvg = ({ data }: { data: ContinuousCorrelationData }) =>
   const sy = (y: number) => H - PAD.bottom - ((y - yMin) / (yMax - yMin || 1)) * (H - PAD.top - PAD.bottom)
   const reg = linearRegression(xs, ys)
 
-  const label = `Scatter of ${describeSelectorAxis(data.trigger)} versus ${describeSelectorAxis(data.outcome)}; r=${fmt(data.pearson)}, n=${data.n}`
+  // For a binary/presence trigger (0/1 — tags, activities, apps) a Pearson r is
+  // misleading, so lead with the present-vs-absent group comparison instead, the
+  // same call the Correlations explorer makes.
+  const gc = data.group_comparison
+  const headline = gc?.trigger_is_binary
+    ? `Δ(present−absent)=${fmt(gc.difference)} · d=${fmt(gc.cohens_d)} · n=${data.n} · p=${fmtP(gc.welch?.p_value)}`
+    : `r=${fmt(data.pearson)} · ρ=${fmt(data.spearman)} · n=${data.n} · p=${fmtP(data.pearson_p)}`
+  const label = `Scatter of ${describeSelectorAxis(data.trigger)} versus ${describeSelectorAxis(data.outcome)}; ${headline}`
 
   return (
     <svg viewBox={`0 0 ${W} ${H}`} class="article-scatter" role="img" aria-label={label}>
@@ -58,7 +65,7 @@ const CorrelationScatterSvg = ({ data }: { data: ContinuousCorrelationData }) =>
         />
       )}
       <text x={PAD.left + 6} y={PAD.top + 12} class="article-scatter-annot">
-        r={fmt(data.pearson)} · ρ={fmt(data.spearman)} · n={data.n} · p={fmtP(data.pearson_p)}
+        {headline}
       </text>
       <text x={(PAD.left + (W - PAD.right)) / 2} y={H - 8} text-anchor="middle" class="article-scatter-axis">
         {describeSelectorAxis(data.trigger)}
--- a/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
+++ b/apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
@@ -1,0 +1,121 @@
+/**
+ * One inline correlation block of an article: a scatter of a `trigger` predictor
+ * against an `outcome`, both resolved **live** over the block's locked
+ * `[start, end]` window (no snapshot) via the owner-authenticated continuous
+ * correlation endpoint. The caller (`ArticleContent`) resolves the effective
+ * window (block override else the article default) and passes concrete ISO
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
```diff
@@ -0,0 +1,121 @@
...
+  start: string
+  end: string
+  caption?: string
+}) => {
+  const { data, isLoading, isError } = useQuery({
+    queryFn: () =>
+      fetchContinuousCorrelation({
+        outcome,
+        period_end: toDay(end),
+        period_start: toDay(start),
```
fiddur, line 114: Non-blocking: `toDay` derives the inclusive `YYYY-MM-DD` bound via `new Date(iso).toISOString().slice(0, 10)`, i.e. the **UTC** calendar day. If a window is authored with a non-UTC offset near midnight (e.g. `2024-06-30T00:00:00+02:00`), the UTC day rolls back to `2024-06-29`, shifting the inclusive bound by a day versus what the author likely intended. Since the blocks are currently authored over MCP/API with full ISO instants, worth confirming the intended day-boundary convention (UTC vs local) so chart-block windows and correlation-block windows stay consistent.

fiddur (commented):
✅ Approved

Clean, well-scoped follow-up that wires up correlation-block rendering. `ArticleContent` now resolves the effective window before branching (so both correlation and chart blocks share the null-window guard), and `ArticleCorrelationBlock` renders a compact live scatter that reuses the shared `linearRegression`/`describeSelectorAxis` maths and the owner-authenticated continuous-correlation endpoint. It closely follows the existing `ArticleChartBlock` and Explore `ScatterPlot` patterns (loading/error/insufficient-data states, degenerate-range guards via `|| 1`, regression drawn only when defined). Types line up with `ContinuousCorrelationBody`/`ContinuousCorrelationData`, the api-spec import paths resolve, and the docs are updated to reflect that correlation blocks now render. No blocking issues.

- `fmt`/`fmtP` are duplicated from `Explore.tsx` (where they're module-private). Small helpers, deliberate compact render — fine to leave, but could be lifted into `exploreCharts.ts` if you want a single source.
- The `<svg role="img">` has no accessible name (`aria-label`/`<title>`); adding one would let screen readers announce the scatter, though the existing Explore scatter omits `role` entirely so this is already an improvement.
- The scatter's `circle` list is rendered without `key`s, matching the existing Explore `ScatterPlot`.

apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
fiddur: Fixed in ba2e06a: `toDay` now takes the date straight off the ISO string (`iso.slice(0, 10)`) instead of converting through a UTC `Date`, so the inclusive regime bound is the author's wall-clock day and a non-UTC-offset window near midnight is no longer rolled back a day. Simpler too — no `Date` construction. (Also folded in the two other observations from your summary: an `aria-label` on the scatter `<svg>` and `key`s on the points.)

apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
```diff
@@ -0,0 +1,128 @@
...
+          x1={sx(xMin)}
+          y1={sy(reg.slope * xMin + reg.intercept)}
+          x2={sx(xMax)}
+          y2={sy(reg.slope * xMax + reg.intercept)}
+          stroke={REG_COLOR}
+          stroke-width={2}
+        />
+      )}
+      <text x={PAD.left + 6} y={PAD.top + 12} class="article-scatter-annot">
+        r={fmt(data.pearson)} · ρ={fmt(data.spearman)} · n={data.n} · p={fmtP(data.pearson_p)}
```
fiddur: The scatter always shows Pearson `r` as the headline stat, but for a binary/presence trigger (tags, activities, productivity apps — all commonly 0/1) that coefficient is misleading. The Explore page handles this by branching on `data.group_comparison?.trigger_is_binary` and hiding `r` in favour of the group comparison, with the note "a correlation coefficient is misleading." Since correlation blocks reuse the same selectors, an author can easily pick a binary trigger here. Consider suppressing or de-emphasising `r`/`p` (or falling back to a present-vs-absent view) when `trigger_is_binary` is true. Non-blocking — this is intentionally a compact render, but worth a decision.

fiddur (commented):
✅ Approved

Clean, well-scoped slice that renders correlation blocks as a live scatter, mirroring the existing `ArticleChartBlock` pattern and reusing the shared correlation maths. Types line up with `ContinuousCorrelationBody`/`CorrelationSelector`, the `[start, end]` window is resolved in the parent and passed as concrete ISO strings, and `toDay()` correctly derives the inclusive `YYYY-MM-DD` bound from the ISO prefix. I confirmed the backend guarantees `series.length === min(n, 1000)`, so the `data.n < 3` guard fully protects the `Math.min(...xs)` spread from an empty array — no degenerate-render bug there.

One substantive design note (non-blocking): the block always renders a scatter with Pearson `r` as the headline annotation, even when the trigger is a binary/presence dimension (tags, activities, productivity apps are commonly 0/1). The Explore page deliberately switches to a group comparison for those (`group_comparison.trigger_is_binary`) precisely because "a correlation coefficient is misleading" on a binary trigger. Since correlation blocks reuse the exact same selectors, an author picking a tag as the trigger will see a misleading `r`. Consider suppressing/annotating the coefficient when `data.group_comparison?.trigger_is_binary` is true, matching Explore's treatment. Flagging inline for the author to decide since the block is intentionally a compact render.

Minor, not worth changing: the query key includes the full-precision `start`/`end` while the request sends the day-truncated dates, so two blocks on the same wall-clock day but different times would produce duplicate cache entries/requests. No test accompanies the component, but that matches the sibling `ArticleChartBlock` (the pure maths is covered in `exploreCharts.test.ts`).

apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
fiddur: Fixed in e3ee119: when `group_comparison.trigger_is_binary` is true, the scatter now leads with the present-vs-absent group comparison — `Δ(present−absent)`, Cohen's d, n and the Welch p-value — instead of the misleading Pearson r/ρ/p, mirroring the explorer's branch. The OLS line stays (through 0/1 x it connects the two group means), and the aria-label uses the same headline. A dedicated present-vs-absent box-plot (like the explorer's) could be a later enhancement, but the headline stat is now correct for binary triggers.

fiddur (commented):
✅ Approved

Clean, well-scoped render slice. `ArticleCorrelationBlock` faithfully mirrors the sibling `ArticleChartBlock` and the Explore `ScatterPlot`, reusing the shared correlation math (`linearRegression`, `describeSelectorAxis`) as its module comment promises. Correctness checks hold up:

- Window nullability is guarded in `ArticleContent` before the block renders, so `start`/`end` reaching the block are always concrete — no need for an `enabled` gate on the query.
- Small samples are handled (`n < 3` note), and the scale functions guard the degenerate `xMax === xMin` / `yMax === yMin` case with `|| 1`, so no divide-by-zero.
- With `n >= 3` the backend always populates `series` (capped at 1000 but never below `n` for small windows), so `Math.min(...xs)` can't hit an empty array.
- Prop types match `articleCorrelationBlockSchema` exactly; the caption is auto-escaped by JSX (no XSS); the endpoint params (`period_start`/`period_end`, optional `lag_days`) match how Explore calls the same endpoint; and the barrel export path resolves.
- `toDay`'s wall-clock-day rationale is consistent with the prior commits in this branch.

One non-blocking observation on the binary-trigger case (inline), and a forward-looking note: `fetchContinuousCorrelation` uses the *viewer's* auth token, so this is only correct while articles render on the owner's own feed (per the doc). When federation (#937) lets followers view articles, the correlation would be computed over the viewer's data rather than the author's — worth carrying into that slice. No changes required for this PR.

apps/web/src/pages/Feed/ArticleCorrelationBlock.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const yMax = Math.max(...ys)
+  const sx = (x: number) => PAD.left + ((x - xMin) / (xMax - xMin || 1)) * (W - PAD.left - PAD.right)
+  const sy = (y: number) => H - PAD.bottom - ((y - yMin) / (yMax - yMin || 1)) * (H - PAD.top - PAD.bottom)
+  const reg = linearRegression(xs, ys)
+
+  // For a binary/presence trigger (0/1 — tags, activities, apps) a Pearson r is
+  // misleading, so lead with the present-vs-absent group comparison instead, the
+  // same call the Correlations explorer makes.
+  const gc = data.group_comparison
+  const headline = gc?.trigger_is_binary
```
fiddur, line 45: Non-blocking / optional: for a binary trigger you correct the *headline* to the group comparison, but the visual is still a scatter of 0/1 x-values (two vertical bands). The Explore explorer deliberately swaps to a `BoxPlot` in exactly this case (`showGroupAsHeadline ? <BoxPlot/> : <ScatterPlot/>`) because the scatter is uninformative there. The compact-render tradeoff is documented, so this is fine to keep — just flagging the divergence in case a box plot (or hiding the regression line) reads better for present/absent triggers.

fiddur: Agreed a present-vs-absent box-plot would read cleaner for binary triggers. Deferring it as a follow-up rather than growing this render PR: doing it well means reusing the explorer's `BoxPlot`, which would pull in the explorer/ScatterPlot extraction (its `CHART_*` constants, `AxisLabels`, CSS) that this PR intentionally left alone. The current state is correct — the headline is the group comparison, the two 0/1 bands still show the split, and the OLS line through 0/1 x connects the two group means. Tracking the box-plot swap for a follow-up. Resolving.

PR #948, opened 2026-07-14
feat(feed): compose correlation blocks in the article editor (#936)

fiddur:
Part of #934 (Shareable analysis Articles), **Slice B (#936)** — final PR (web composer). Builds on the schema/backend (#946) and render (#947).

Correlation article blocks are now **fully authorable in the web composer** — completing Slice B.

## What's here

- **Shared `SelectorPicker`** — lifted the correlation selector picker (kind dropdown + per-kind autocompleted inputs) out of the Correlations explorer into `components/SelectorPicker.tsx` (with its `KIND_LABELS`/`NUTRIENTS`, co-located CSS, and an exported `ALL_SELECTOR_KINDS`). The explorer imports it; behaviour unchanged. Both surfaces now build selectors the same way.
- **Composer** — the article editor replaces the read-only correlation passthrough row (from #946) with a real editor: **trigger** + **outcome** pickers, an optional per-block window, **lag-days**, and a caption; a **+ Correlation** button adds one. The dialog fetches the correlation selectors to autocomplete the pickers.
- **`article-editor-body`** — `CorrelationDraft` is now editable (selectors + `datetime-local` window + `lagDays` + caption) instead of the verbatim passthrough; `draftsFromPost` seeds it and `buildArticleBody` validates both selectors are filled in, then emits the block (empty optional fields omitted). Extracted `buildChartBlock`/`buildCorrelationBlock` to keep `buildArticleBody` under the complexity budget; `emptyCorrelationDraft` for the add button.
- **Tests** — seed→build lossless round-trip, unfilled-trigger rejection, empty-field omission (16 pass). **Docs** — composer bullet + article overview updated.

## Slice B is complete after this

Schema/backend (#946) → live scatter render (#947) → composer (this). What remains for the epic is **Slice C (#937)**: federation (AS2 `Article` + attached PNGs / native inline) and Reddit/markdown export, plus the deferred public/followers-gated server rendering of chart and correlation images.

### DIFF
--- a/apps/web/src/components/article-editor-body.ts
+++ b/apps/web/src/components/article-editor-body.ts
@@ -156,6 +156,9 @@ const buildCorrelationBlock = (b: CorrelationDraft, i: number): BlockResult => {
   const missing = selectorIncomplete(b.trigger) ? 'trigger' : selectorIncomplete(b.outcome) ? 'outcome' : null
   if (missing) return { error: `Correlation block ${i + 1}: choose a ${missing} (metric or pattern).` }
   const lag = Number(b.lagDays)
+  if (b.lagDays.trim() && !Number.isInteger(lag)) {
+    return { error: `Correlation block ${i + 1}: lag days must be a whole number.` }
+  }
   return {
     block: {
       outcome: b.outcome,
@@ -163,7 +166,7 @@ const buildCorrelationBlock = (b: CorrelationDraft, i: number): BlockResult => {
       type: 'correlation',
       ...(toIso(b.start) ? { start: toIso(b.start) } : {}),
       ...(toIso(b.end) ? { end: toIso(b.end) } : {}),
-      ...(b.lagDays.trim() && Number.isFinite(lag) && lag !== 0 ? { lag_days: lag } : {}),
+      ...(b.lagDays.trim() && lag !== 0 ? { lag_days: lag } : {}),
       ...(b.caption.trim() ? { caption: b.caption.trim() } : {}),
     },
   }
--- a/apps/web/src/components/article-editor-body.ts
+++ b/apps/web/src/components/article-editor-body.ts
@@ -4,7 +4,13 @@
  * `buildShareBody`). Handles the `datetime-local` ⇄ ISO conversion, seeding the
  * editor from an existing post, and building/validating the create/update body.
  */
-import type { ArticleBlock, CreateArticleBody, FeedPost, FeedVisibility } from '@aurboda/api-spec'
+import type {
+  ArticleBlock,
+  CorrelationSelector,
+  CreateArticleBody,
+  FeedPost,
+  FeedVisibility,
+} from '@aurboda/api-spec'
 
 import { isValidMetric } from '@aurboda/api-spec'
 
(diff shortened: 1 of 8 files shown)


apps/web/src/components/article-editor-body.ts
```diff
@@ -93,11 +126,55 @@ export const deriveSubmitError = (validationError: string | null, mutationError:
...
+  if (missing) return { error: `Correlation block ${i + 1}: choose a ${missing} (metric or pattern).` }
+  const lag = Number(b.lagDays)
+  return {
+    block: {
+      outcome: b.outcome,
+      trigger: b.trigger,
+      type: 'correlation',
+      ...(toIso(b.start) ? { start: toIso(b.start) } : {}),
+      ...(toIso(b.end) ? { end: toIso(b.end) } : {}),
+      ...(b.lagDays.trim() && Number.isFinite(lag) && lag !== 0 ? { lag_days: lag } : {}),
```
fiddur: `lag_days` is `z.number().int()` in the schema, but the number input has no `step` and this only checks `Number.isFinite(lag)` — a fractional entry like `1.5` passes the form and is rejected server-side with a generic validation error instead of being caught early like the other fields. Consider guarding to integers here (e.g. `Number.isInteger(lag)`, or `Math.round`) and/or adding `step="1"` to the input. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped PR. It (1) extracts the `SelectorPicker` from `Explore.tsx` into a shared component + CSS verbatim (with more specific datalist ids), and (2) turns article correlation blocks from a read-only passthrough into a fully editable draft (trigger/outcome selectors, window, lag, caption). The refactor is behaviour-preserving: `Explore` keeps its own kind lists and just consumes the shared component, the moved CSS retains the same rules, and no dangling imports remain (`NutrientKey`/`CorrelationSelectorsData`/`NUTRIENTS` were removed where no longer used).

The draft-conversion is nicely covered: `draftsFromPost` → `buildArticleBody` round-trips a stored block losslessly (selectors kept verbatim, so optional `agg`/`threshold`/`measure` survive), optional window/lag/caption are correctly omitted when empty, and an unfilled trigger/outcome is rejected before submit. `buildArticleBody` was cleanly decomposed into `buildChartBlock`/`buildCorrelationBlock` with no logic change to the chart path. Docs updated to match.

One minor, non-blocking gap noted inline (fractional `lag_days`). A pre-existing, functionally-harmless nit also carries over: rendering two metric (or two same-kind) selectors produces duplicate `<datalist>` ids in the DOM — invalid HTML but the content is identical so autocomplete still resolves correctly; worth tidying if you touch this again but not required here.

apps/web/src/components/article-editor-body.ts
fiddur: Fixed in acfe238: `buildCorrelationBlock` now rejects a non-integer lag early ("Correlation block N: lag days must be a whole number.") instead of leaving it to the server, and the lag input carries `step="1"`. Added tests for the rejection and the integer round-trip.

fiddur (commented):
✅ Approved

Solid, well-scoped PR that (1) extracts the `SelectorPicker` from `Explore.tsx` into a shared component and (2) turns the previously read-only correlation block into a fully editable composer block, with a follow-up fix that catches non-integer lags client-side.

**Correctness — verified:**
- The `SelectorPicker` extraction is faithful. The datalist ids were renamed (`metric-options` → `selector-metric-options`, `pattern-options-*` → `selector-pattern-options-*`), which correctly avoids colliding with the `metric-options` datalist that still lives in `Explore.tsx` for the event-outcome input.
- The `draftsFromPost` → `buildCorrelationBlock` round-trip is lossless: I traced the fully-populated `correlationBlock` (caption, window, `lag_days: 1`, metric outcome, activity trigger) through both directions and it reconstructs the original object. The minute-boundary `toInputValue`/`toIso` round-trip that this depends on is itself covered by a test.
- Validation is sound: `selectorIncomplete` treats an empty metric/pattern as unfilled while correctly always accepting `nutrition` (its `nutrient` field always has a value), `lag=0` and blank optionals are omitted from the body, and non-integer lags are rejected before hitting the server. `lag_days` has no server-side min, so allowing negatives is intentional.
- Removed imports (`NutrientKey`, `CorrelationSelectorsData`, `NUTRIENTS`, `KIND_LABELS`) have no remaining references in `Explore.tsx`; the moved `.selector-picker` CSS leaves a valid remaining rule.

**Test coverage** is good — the new cases exercise empty-trigger rejection, non-integer lag rejection, integer-lag passthrough, empty-optional omission, and the seed/round-trip.

I could not run `pnpm check`/tests here (the pnpm store is read-only in this sandbox, so `pnpm install` fails), so the above is from manual tracing rather than an executed run — worth confirming CI is green.

**Minor, non-blocking observations (no action required):**
- `ALL_SELECTOR_KINDS` now exists twice with identical contents — exported from `SelectorPicker.tsx` and kept locally in `Explore.tsx` (derived from `EVENT_TRIGGER_KINDS`). Fine as-is; just noting the duplication.
- With multiple correlation blocks, several `<datalist>`s share the same id (e.g. `selector-metric-options`). This is the same pattern the old two-picker Explore page already had, and since each datalist's contents are global (all metrics / all activities), the duplicate ids are harmless — browsers bind to the first match and the options are identical.
- `productivity_app` autocompletes from `activity_types` (the else branch) — pre-existing behavior carried over verbatim from `Explore.tsx`, not introduced here.

Docs (`feed.md`) are updated to reflect that correlation blocks are now composable in the web editor. Nice work.

PR #949, opened 2026-07-15
feat(android): embed web Feed as a native tab via WebView

fiddur:
## What & why

First slice of a **hybrid Android architecture**: keep device-bound screens native, embed the web app for read/config-heavy screens so we don't maintain two implementations. This wires up the **Feed** tab as an embedded web view and lays the reusable foundation (WebView host + web-side embed contract) that future embedded pages (timeline, settings, …) build on.

## Android

- New **Feed** bottom-nav tab hosting `/feed?embed=1` via a reusable `EmbeddedWebScreen` — JavaScript + DOM storage enabled, native loading spinner and retry-able error state, and the system back gesture walks the WebView history before leaving the screen.
- **Shared auth, no second login**: the WebView injects `window.AurbodaNative.getAuth()` returning the app's bearer token; the web app seeds its auth from it. Works because web (localStorage) and Android (EncryptedSharedPreferences) use the same stateless bearer token.
- **External links open in the browser**: links outside the server origin launch `ACTION_VIEW`; same-origin links stay in the WebView. Decision is a pure, unit-tested helper (`WebLink.kt` / `isExternalLink`).

## Web

- **Embed mode** (`?embed=1`, persisted to `sessionStorage`): hides header/sidebar/footer since the native app provides navigation (`apps/web/src/embed.ts`, wired into `AppShell`).
- Seeds auth from the native bridge when embedded, falling back to browser-stored credentials.

## Tests

- Web: `embed.test.ts` covers `computeEmbedded` + `parseNativeAuth` (541 web tests pass).
- Android: `WebLinkTest` covers same-origin/cross-origin/subdomain/scheme/relative/case/port; `AppStateTest` updated for the new tab. Full `testDebugUnitTest` green.

## Docs

- `docs/android-app.md` documents the hybrid architecture and the web↔native embed contract.

## Notes

- A few unrelated web files (`Challenges/*`, `widgets/*`) are reformatted by `oxfmt` (they were merged to develop unformatted); kept per repo convention of committing formatter output.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -1,0 +1,157 @@
+package net.aurboda.ui.screens
+
+import android.annotation.SuppressLint
+import android.content.Intent
+import android.graphics.Bitmap
+import android.webkit.JavascriptInterface
+import android.webkit.WebResourceError
+import android.webkit.WebResourceRequest
+import android.webkit.WebView
+import android.webkit.WebViewClient
+import androidx.activity.compose.BackHandler
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Button
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.viewinterop.AndroidView
+import org.json.JSONObject
+
+/**
+ * Bridge exposed to the embedded web app as `window.AurbodaNative`. The web app
+ * reads [getAuth] at startup to share the native app's bearer token instead of
+ * prompting for a second login (see apps/web/src/embed.ts).
+ */
+private class AuthBridge(private val authJson: String) {
+    @JavascriptInterface
+    fun getAuth(): String = authJson
+}
+
+/**
+ * Hosts a page of the web app inside a WebView so native and web share one
+ * implementation. The native app supplies navigation, the web app renders in
+ * embed mode (chrome hidden). Links to other domains open in the external
+ * browser; same-origin links stay in the WebView.
+ *
+ * @param url the web page to load (already carrying `?embed=1`)
+ * @param baseUrl the server origin, used to decide which links are external
+ */
+@Suppress("ASSIGNED_VALUE_IS_NEVER_READ") // Compose state vars trigger false "assigned but never read" warnings
+@SuppressLint("SetJavaScriptEnabled")
+@Composable
+fun EmbeddedWebScreen(
+    url: String,
+    baseUrl: String,
+    username: String,
+    authToken: String,
+    modifier: Modifier = Modifier,
+) {
+    val context = LocalContext.current
+    var loading by remember { mutableStateOf(true) }
+    var errorMessage by remember { mutableStateOf<String?>(null) }
+    var canGoBack by remember { mutableStateOf(false) }
+    var webView by remember { mutableStateOf<WebView?>(null) }
+
+    val authJson = remember(username, authToken) {
+        JSONObject().put("user", username).put("token", authToken).toString()
+    }
+
+    // Let the WebView consume the system back gesture to walk its own history
+    // before the back press falls through to leaving the screen.
+    BackHandler(enabled = canGoBack) {
+        webView?.goBack()
+    }
+
+    Box(modifier = modifier.fillMaxSize()) {
+        AndroidView(
+            modifier = Modifier.fillMaxSize(),
+            factory = { ctx ->
+                WebView(ctx).apply {
+                    settings.javaScriptEnabled = true
+                    settings.domStorageEnabled = true
+                    addJavascriptInterface(AuthBridge(authJson), "AurbodaNative")
+                    webViewClient = object : WebViewClient() {
+                        override fun shouldOverrideUrlLoading(
+                            view: WebView,
+                            request: WebResourceRequest,
+                        ): Boolean {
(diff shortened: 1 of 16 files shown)


apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    Box(modifier = modifier.fillMaxSize()) {
+        AndroidView(
+            modifier = Modifier.fillMaxSize(),
+            factory = { ctx ->
+                WebView(ctx).apply {
+                    settings.javaScriptEnabled = true
+                    settings.domStorageEnabled = true
+                    addJavascriptInterface(AuthBridge(authJson), "AurbodaNative")
```
fiddur, line 84: The bridge exposes the bearer token to any JavaScript running same-origin in this WebView. It's a reasonable tradeoff here (external navigations are sent to the browser via `shouldOverrideUrlLoading`, the content is first-party, and the same token is already in the page's localStorage after seeding), but note that `shouldOverrideUrlLoading` only gates navigations — external `<script>`/`fetch`/iframe subresources or an XSS on `/feed` would still be able to call `AurbodaNative.getAuth()` and steal the token. Fine to leave as-is given the current threat model; just keep it in mind before the feed renders any third-party/embedded content.

fiddur (commented):
✅ Approved

Solid, well-scoped feature. The hybrid native/web approach is cleanly implemented and the tricky logic (external-link detection, embed-mode detection, native auth parsing) is extracted into pure, well-tested helpers (`WebLink.kt`, `embed.ts`) rather than buried in the Composable/WebView plumbing — exactly the testability the project standards ask for. I verified the mechanics: `MainScreen` has a single call site that's updated, the `MainTab` ordinal shift only affects internal test assertions (tabs are referenced by name, not persisted ordinal), and `credentials` exposes `serverUrl`/`username`/`authToken`. minSdk 34 means the modern WebView defaults apply — mixed content is blocked and file access is off — and the non-deprecated `WebResourceRequest`/`onReceivedError` overloads are the right ones. The embed contract (hide chrome, share bearer token, redirect external nav to the browser) is coherent and matches the docs.

One security note (non-blocking, see inline): `addJavascriptInterface` exposes the bearer token to all JavaScript running in the WebView. This is acceptable here because external navigations are redirected out to the browser and the content is first-party (the same token already lives in the page's localStorage after seeding), but it does mean any script that manages to run same-origin in the WebView (e.g. an external `<script>`/embed on the feed page, or an XSS) can exfiltrate the token — worth keeping in mind as the feed grows to render richer/third-party content.

Minor: the diff bundles a handful of prettier-only reflows in unrelated web files (`MetricCardWidget.tsx`, `PublicWidgetRenderer.tsx`, `Join.tsx`, `PublicChallenge.tsx`) — harmless, presumably from `pnpm fix`, just slight scope creep.

apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
fiddur, line 84: Non-blocking hardening note: a `@JavascriptInterface` object added via `addJavascriptInterface` is injected into **every** frame the WebView loads, including cross-origin iframes — an iframe from another origin could call `AurbodaNative.getAuth()` and read the bearer token (something it could not do via `localStorage`, which is same-origin protected). This is safe *today* because `renderMarkdown`/ingest sanitisation strips `<iframe>`/`<script>` and external links are punted to the system browser, so no untrusted origin ever renders inside this WebView. Worth keeping that invariant explicit: never allow-list `<iframe>` into the feed sanitiser, and consider gating the bridge on the loaded origin. No change required for this PR.

apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
```diff
@@ -0,0 +1,157 @@
...
+                        }
+
+                        override fun onReceivedError(
+                            view: WebView,
+                            request: WebResourceRequest,
+                            error: WebResourceError,
+                        ) {
+                            // Only the main frame failing is a page load error;
+                            // ignore failed sub-resources (images, tiles, etc.).
+                            if (request.isForMainFrame) {
```
fiddur, line 119: Minor: `onReceivedError` only fires for network-level failures, not HTTP error statuses. If the server returns a 4xx/5xx for `/feed` (e.g. an expired/invalid token), this native error state won't show — the WebView will render the server's error body instead. If you want the retry UI to cover that case too, also override `onReceivedHttpError` for the main frame. Not blocking.

fiddur (commented):
✅ Approved

Solid, well-tested feature. The native/web embed contract is clean: auth is shared via a JS bridge rather than a URL param (so no token in logs/history), embed mode is detected once and persisted to `sessionStorage`, and the `isExternalLink` decision is a pure, thoroughly unit-tested helper that correctly handles ports, case, protocol-relative URLs, and `userinfo@host` tricks. I verified the `/feed` route exists, that feed content is sanitised through DOMPurify (`renderMarkdown` / server-side ingest), and that no Android test constructs `MainScreen(` so the new required `feedContent` param doesn't break existing tests. Docs (`docs/android-app.md`) are a nice touch.

No blocking issues. A few non-blocking notes:

- **Token exposure surface of `addJavascriptInterface`** (see inline): the bearer token is reachable from every frame the WebView loads, including cross-origin iframes. It's safe today because the sanitiser strips `<script>`/`<iframe>` and external links open in the system browser, but it's worth staying deliberate about (never allow-list iframes into the feed sanitiser, and consider restricting WebView navigation to the trusted origin).
- **Error handling** only covers main-frame network failures, not HTTP error statuses (see inline) — minor UX gap.
- The reformatting churn in `MetricCardWidget.tsx`, `PublicWidgetRenderer.tsx`, `Join.tsx`, and `PublicChallenge.tsx` is unrelated prettier reflow (no logic change) — harmless, just scope creep in an otherwise Android-focused PR.

apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
fiddur: Agreed, non-blocking. Tracking in a follow-up 'WebView hardening' PR: add the in-code security note documenting this invariant, keep `<iframe>` out of the feed sanitiser, and gate the bridge on the loaded origin. Resolving here.

apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
fiddur: Good catch — folding `onReceivedHttpError` (main-frame 4xx/5xx → native retry, e.g. expired token) into the same follow-up PR. Resolving here.

PR #950, opened 2026-07-15
feat(android): harden the embedded Feed WebView (origin-scoped auth bridge + HTTP errors)

fiddur:
Follow-up to #949, addressing the two non-blocking review notes plus the in-code security doc.

## Changes

1. **Origin-scope the auth bridge.** `window.AurbodaNative` is now injected as a **document-start script scoped to the server origin** via `WebViewCompat.addDocumentStartJavaScript(..., allowedOriginRules)`. The bearer token therefore reaches only the trusted origin's frames — a cross-origin iframe can no longer call `getAuth()`. It also runs *before* the page's own scripts (race-free). Falls back to `addJavascriptInterface` on WebViews without `DOCUMENT_START_SCRIPT` support (safe: the feed sanitiser strips `<iframe>`/`<script>` and external nav opens in the browser). New pure, unit-tested helper `originOf()` derives the origin rule.
2. **Cover HTTP errors in the retry UI.** Added `onReceivedHttpError` for the main frame, so a `401` from an expired token (or any 4xx/5xx on `/feed`) shows the native retry state instead of the server's error body.
3. **In-code security note** documenting the exposure model and the "never allow-list `<iframe>` into the feed sanitiser" invariant (the doc deferred from #949).

## Dependency

Adds `androidx.webkit:webkit:1.12.1` (for `WebViewCompat`/`WebViewFeature`).

## Tests

- `WebLinkTest` extended with `originOf` cases (scheme/host, explicit port, no-host → null).
- Full `:app:testDebugUnitTest` + `:app:assembleDebug` green locally.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -6,6 +6,7 @@ import android.graphics.Bitmap
 import android.webkit.JavascriptInterface
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
+import android.webkit.WebResourceResponse
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.activity.compose.BackHandler
@@ -27,21 +28,22 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
+import androidx.webkit.WebViewCompat
+import androidx.webkit.WebViewFeature
 import org.json.JSONObject
 
 /**
- * Bridge exposed to the embedded web app as `window.AurbodaNative`. The web app
- * reads [getAuth] at startup to share the native app's bearer token instead of
- * prompting for a second login (see apps/web/src/embed.ts).
+ * Fallback bridge exposed to the embedded web app as `window.AurbodaNative` on
+ * WebViews that lack document-start script support (see [installAuthBridge]).
+ * The web app reads [getAuth] at startup to share the native app's bearer token
+ * instead of prompting for a second login (see apps/web/src/embed.ts).
  *
- * Security: [getAuth] hands the bearer token to any JavaScript running
- * same-origin in this WebView. That is acceptable under the current threat
- * model — we only load our own first-party pages, external navigations are sent
- * to the browser ([isExternalLink] + shouldOverrideUrlLoading), and the web app
- * seeds the same token into localStorage anyway, so same-origin scripts gain no
- * new access. Reconsider (e.g. a short-lived/scoped token, or dropping the
- * bridge) before the embedded pages render any third-party or user-embedded
- * content that could run scripts on our origin.
+ * Security: added via `addJavascriptInterface`, [getAuth] is reachable from
+ * every frame the WebView loads, including cross-origin iframes. That is
+ * acceptable because the feed sanitiser strips `<iframe>`/`<script>` and
+ * external navigation is punted to the browser, so no untrusted origin renders
+ * here. The preferred path ([installAuthBridge] via `WebViewCompat`) avoids the
+ * cross-origin exposure by scoping injection to the trusted origin.
  */
 private class AuthBridge(private val authJson: String) {
     @JavascriptInterface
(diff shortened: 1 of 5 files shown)


apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
```diff
@@ -121,6 +157,20 @@ fun EmbeddedWebScreen(
                                 errorMessage = "Could not load page"
                             }
                         }
+
+                        override fun onReceivedHttpError(
+                            view: WebView,
+                            request: WebResourceRequest,
+                            errorResponse: WebResourceResponse,
+                        ) {
+                            // A main-frame HTTP error (e.g. 401 on an expired
```
fiddur: Non-blocking: the `401 on an expired token` example is misleading for this app. The web app is a static Preact/Vite SPA — the main-frame `/feed` document loads unauthenticated (200) and runs `readNativeAuth` client-side; the expired-token 401 happens on a subsequent `fetch` to `/api`, which is not a main-frame navigation, so `isForMainFrame` is false and this branch won't fire for it. The handler is still valuable for genuine main-frame HTTP errors (5xx / 404 on the shell). Consider tweaking the example (here and in docs/android-app.md) to a main-frame 5xx to avoid implying it covers token expiry.

fiddur (commented):
✅ Approved

Solid, well-documented hardening of the embedded Feed WebView. The `installAuthBridge` logic is correct: it prefers the origin-scoped document-start script, falls back to `addJavascriptInterface` only when the feature is unsupported or the origin is unknown, guards the `WebViewCompat` call with `runCatching` so a malformed origin degrades to the fallback rather than crashing, runs exactly one injection path (no double-inject), and installs before `loadUrl`. Both paths satisfy the web contract (`window.AurbodaNative.getAuth()` returning a JSON string, consumed by `parseNativeAuth`).

I verified the security rationale: DOMPurify (the web render sink) and `sanitizeRemoteHtml` (server ingest, strips scripts/iframes) back the "sanitiser strips `<iframe>`/`<script>`" invariant, and `EmbeddedWebScreen` is currently only wired to `/feed`, so the fallback's broader exposure is safe today. `originOf` and its unit tests are sound.

One non-blocking note (inline): the app is a client-rendered Preact/Vite SPA served as a static shell, so the main-frame `/feed` document returns 200 and an expired-token 401 surfaces on client-side `fetch`, not on the main-frame navigation — so the `onReceivedHttpError(isForMainFrame)` handler won't actually catch that case. The handler is still worthwhile (it catches main-frame 5xx/404 so the raw server error body isn't rendered); only the illustrative "401 on an expired token" example in the code comment and docs is inaccurate for this SPA.

fiddur (commented):
✅ Approved

Clean, well-scoped hardening of the embedded Feed WebView. Two independent improvements, both correct:

**1. Origin-scoped auth bridge.** `installAuthBridge` prefers `WebViewCompat.addDocumentStartJavaScript` scoped to the trusted origin so the bearer token no longer reaches cross-origin frames, and falls back to `addJavascriptInterface` only when `DOCUMENT_START_SCRIPT` is unsupported or the origin is unknown. The gating (`origin != null && isFeatureSupported(...)`), the `runCatching { ... }.isSuccess` guard, and the unconditional fallback are all correct — no path leaves `window.AurbodaNative` undefined when a bridge is expected. The injected script returns a JSON *string* via `JSONObject.quote(authJson)`, matching both the `AuthBridge.getAuth(): String` contract and what `parseNativeAuth`/`JSON.parse` in `embed.ts` consumes, so the two paths behave identically. Installing the bridge before `loadUrl` is also required and satisfied.

**2. Main-frame HTTP-error handling.** `onReceivedHttpError` correctly guards on `request.isForMainFrame`, mirroring the existing `onReceivedError`, so a 5xx/404 on the page shell shows the native retry state instead of the server error body. The callback ordering leaves the error state intact, and the comment/docs correctly note that token-expiry 401s hit client-side `/api` fetches, not the main-frame load.

`originOf` is well unit-tested (port / no-port / null-host / unparseable cases), and the docs accurately describe the new behavior and the sanitiser invariant the fallback depends on (DOMPurify strips `<script>` and does not allow `<iframe>` by default).

Minor, non-blocking note: `originOf` scopes the bridge to a port-specific origin while `isExternalLink` treats same-host/different-port as internal. This means on a same-host/different-port navigation the doc-start bridge wouldn't inject (the web app would fall back to browser-stored creds). This is an extreme edge case that doesn't occur against the production single-origin (`aurboda.net`), and including the port is the safer, correct choice for a security scope — so I'd leave it as-is; just flagging the intentional asymmetry.

No blocking issues.

PR #951, opened 2026-07-15
fix(feed): mobile horizontal overflow + WebView system dark mode

fiddur:
Two fixes spotted comparing the Feed on web vs the app WebView on mobile.

## 1. Horizontal overflow (web, affects browser + app)
`.feed-page` had `padding: 1rem` with no `box-sizing` reset (there is no global reset — components set their own), so it extended ~32px past the viewport → slight sideways scroll on both the mobile browser and the embedded WebView. Fixed with `box-sizing: border-box` (+ explicit `width: 100%`).

## 2. WebView ignored system dark mode (Android)
The browser rendered the Feed dark (system preference), but the app's WebView stayed light. Root cause: `Theme.Aurboda` is hardcoded `android:Theme.Material.Light`, and a WebView derives `prefers-color-scheme` from its context's `isLightTheme`.

Fix: build the WebView against a **day/night-themed context** (`Theme.Aurboda.WebView` → `Theme.DeviceDefault.DayNight`) and enable `WebSettingsCompat.setAlgorithmicDarkeningAllowed`. The `Configuration` already carries the system night flag, so `isLightTheme` now follows the system and the page's own dark CSS applies (`index.html` already declares `color-scheme: light dark`). **Native UI stays on the light theme** — only the embedded WebView follows the system.

## Verification
- `:app:assembleDebug` + `:app:testDebugUnitTest` green; web `check` green.
- Dark-mode behavior needs a visual check on-device (toggle system dark mode on the Feed tab) — it can't be asserted in unit tests.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -3,6 +3,7 @@ package net.aurboda.ui.screens
 import android.annotation.SuppressLint
 import android.content.Intent
 import android.graphics.Bitmap
+import android.view.ContextThemeWrapper
 import android.webkit.JavascriptInterface
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
@@ -28,8 +29,10 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
+import androidx.webkit.WebSettingsCompat
 import androidx.webkit.WebViewCompat
 import androidx.webkit.WebViewFeature
+import net.aurboda.R
 import org.json.JSONObject
 
 /**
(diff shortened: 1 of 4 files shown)


fiddur (commented):
✅ Approved

Clean, well-documented, low-risk fix for two separate cosmetic issues. No blocking issues found.

**Android WebView dark mode** (`EmbeddedWebScreen.kt` + `themes.xml`): Building the WebView against a `Theme.DeviceDefault.DayNight` `ContextThemeWrapper` so its `isLightTheme` follows the system setting, combined with a feature-guarded `setAlgorithmicDarkeningAllowed(true)`, is the correct modern approach (replacing the deprecated `setForceDark`). I verified the supporting facts:
- `androidx.webkit:webkit:1.12.1` provides `ALGORITHMIC_DARKENING` / `setAlgorithmicDarkeningAllowed` (added in 1.5.0).
- `targetSdk = 36` (≥ 33), so the API actually takes effect; `minSdk = 34` supports the `Theme.DeviceDefault.DayNight` parent.
- The `R` import namespace (`net.aurboda`) and the `R.style.Theme_Aurboda_WebView` reference resolve correctly, and the native app UI keeps the light `Theme.Aurboda` from the manifest — so only the WebView follows dark mode, as intended.
- The comment/doc claim that "the page declares `color-scheme: light dark`" is accurate: `apps/web/index.html:10` sets `<meta name="color-scheme" content="light dark">`, so WebView respects the page's own `prefers-color-scheme: dark` CSS instead of algorithmically inverting.

**Feed mobile overflow** (`Feed/style.css`): Adding `box-sizing: border-box; width: 100%` is a standard, safe fix — `border-box` keeps the `padding: 1rem` inside the width so it can't overflow the viewport, while `max-width: 720px` + `margin: 0 auto` preserve desktop centering.

Note on local verification: `./gradlew :app:compileDebugKotlin` fails in this worktree, but only in unrelated files (`PasskeyApi.kt`, `SyncWorker.kt`, `AddDataScreen.kt`) that reference generated api-spec Kotlin models not regenerated here — an environment issue, not this PR. `EmbeddedWebScreen.kt` itself compiled without errors.

PR #952, opened 2026-07-15
fix(app): embedded profile overflow + back nav across SPA navigations

fiddur:
Two issues found navigating from the Feed into a follower's public profile inside the app WebView.

## 1. Profile page too wide (web)
`.public-profile` had `padding: 2rem 1.5rem` with no `box-sizing` reset, so it ran ~48px past the viewport on mobile — the Share button got clipped off the right edge. Fixed with `box-sizing: border-box` + `width: 100%`, matching the earlier `.feed-page` fix (#951).

## 2. System back gesture closed the app instead of returning to the feed (Android)
`canGoBack` was only refreshed in `onPageFinished`. The embedded app is a client-side Preact/preact-iso SPA, so in-app navigation (Feed → `/u/:user`) happens via `pushState`, which does **not** fire `onPageFinished` — so `canGoBack` stayed `false`, the `BackHandler` stayed disabled, and the back gesture fell through and left the screen.

Fix: also refresh `canGoBack` in `doUpdateVisitedHistory`, which fires on `pushState`, so the back gesture walks the WebView history back to the feed.

## Verification
- `:app:compileDebugKotlin` + `:app:testDebugUnitTest` green; web `check` green.
- Back-gesture behavior needs an on-device check (Feed → tap a local follower → swipe back → should return to the feed, not exit).

## Note
This is the second page (after Feed) to hit the mobile padding overflow — there's no global `box-sizing` reset in the web app, so each page container with padding overflows. A global reset would fix it once for every current and future embedded page; flagged for discussion rather than bundled here (broad blast radius).

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -159,6 +159,14 @@ fun EmbeddedWebScreen(
                             canGoBack = view.canGoBack()
                         }
 
+                        override fun doUpdateVisitedHistory(view: WebView, url: String?, isReload: Boolean) {
+                            // The embedded app is a client-side SPA: in-app
+                            // navigation happens via pushState, which does not
+                            // fire onPageFinished. Refresh here too so the back
+                            // gesture walks WebView history instead of leaving.
+                            canGoBack = view.canGoBack()
+                        }
+
                         override fun onReceivedError(
                             view: WebView,
                             request: WebResourceRequest,
--- a/apps/web/src/pages/PublicProfile/style.css
+++ b/apps/web/src/pages/PublicProfile/style.css
@@ -1,4 +1,6 @@
 .public-profile {
+  box-sizing: border-box;
+  width: 100%;
   max-width: 720px;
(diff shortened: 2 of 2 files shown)


fiddur (commented):
✅ Approved

Two small, well-scoped fixes, both correct.

**Android (`EmbeddedWebScreen.kt`):** Overriding `doUpdateVisitedHistory` is the right hook for keeping `canGoBack` in sync during client-side SPA navigations (`pushState`), which don't trigger `onPageFinished`. The override signature matches the framework method and the nullability convention used by the neighboring `onPageStarted`/`onPageFinished` overrides, so it compiles cleanly. The state also self-corrects: `webView.goBack()` re-fires `doUpdateVisitedHistory`, so once history is exhausted `canGoBack` returns to `false` and the back gesture again falls through to leaving the screen. The comment clearly explains the rationale.

**Web (`PublicProfile/style.css`):** `box-sizing: border-box` + `width: 100%` is the standard, low-risk pattern to fold the horizontal padding into the element width and prevent overflow on narrow viewports, while `max-width: 720px` still caps it on wide ones. This complements the existing header wrapping rules.

No behavioral tests exist for the Compose WebView callback or the CSS, so there's nothing to add there. No blocking issues found.

PR #953, opened 2026-07-15
feat(feed): link all followers, follow-back, drop profile Share button

fiddur:
Three small social-UX changes from using the embedded feed in the app.

## 1. Link every actor, including remote ones
The Followers (and Following) lists only linked *local* actors; remote fediverse actors rendered as plain text. Now every actor's name links to their profile:
- **Local** → `/u/:username` (stays in the app WebView).
- **Remote** → the actor's own URL, `target="_blank"` — a new tab in the browser, and the **system browser from the app** (the WebView's `isExternalLink` handling sends off-origin links out).

New shared `actorProfileLink()` helper + `ActorName` component, reused by both panels (removes the duplicated inline link logic).

## 2. Follow back
Accepted followers you don't already follow get a **Follow back** button (uses the follower's handle, or actor URI — the `/feed/following` endpoint accepts either). It disappears once the follow lands (the `['following']` query refetches). The panel shares that query key with the Following panel, so it's a single fetch.

## 3. Remove the profile Share button
Dropped the `ShareLinkButton` from `/u/:username` — it wasn't useful there. (The `.public-profile` overflow was fixed separately in #952.)

## Tests
- `actorProfileLink` unit tests (local / remote-external / non-http+malformed → null); 548 web tests pass, `check` green.
- The follow-back and external-link-out behaviors are UI wiring — worth an on-device check (Feed → Followers → Follow back; tap a remote follower → opens system browser).

### DIFF
--- a/apps/web/src/pages/Feed/FollowersPanel.tsx
+++ b/apps/web/src/pages/Feed/FollowersPanel.tsx
@@ -11,8 +11,8 @@ import type { FollowerActor } from '@aurboda/api-spec'
 
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 
-import { approveFollower, fetchFollowers, rejectFollower } from '../../state/api'
-import { localProfilePath } from './profile-link'
+import { approveFollower, fetchFollowers, fetchFollowing, followActor, rejectFollower } from '../../state/api'
+import { ActorName } from './ActorName'
 
 const Avatar = ({ actor }: { actor: FollowerActor }) =>
   actor.avatar_url ? (
@@ -23,10 +23,11 @@ const Avatar = ({ actor }: { actor: FollowerActor }) =>
 
 const Ident = ({ actor }: { actor: FollowerActor }) => {
   const name = actor.display_name ?? actor.handle ?? actor.actor_uri
-  const profilePath = localProfilePath(actor.actor_uri, window.location.host)
   return (
     <span class="following-ident">
-      <span class="following-name">{profilePath ? <a href={profilePath}>{name}</a> : name}</span>
+      <span class="following-name">
+        <ActorName name={name} actorUri={actor.actor_uri} />
+      </span>
       {actor.handle && <span class="following-handle">{actor.handle}</span>}
     </span>
   )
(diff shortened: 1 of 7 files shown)


apps/web/src/pages/Feed/FollowersPanel.tsx
```diff
@@ -55,25 +56,45 @@ const PendingRow = ({ actor }: { actor: FollowerActor }) => {
...
+const AcceptedRow = ({ actor, isFollowing }: { actor: FollowerActor; isFollowing: boolean }) => {
   const queryClient = useQueryClient()
   const remove = useMutation({
     mutationFn: () => rejectFollower(actor.id),
     onSuccess: () => queryClient.invalidateQueries({ queryKey: ['followers'] }),
   })
+  // Follow back with the follower's handle (or actor URI — the endpoint accepts
+  // either). Once it lands, the ['following'] list refetches and `isFollowing`
+  // flips true, hiding this button.
+  const followBack = useMutation({
```
fiddur, line 68: Minor/non-blocking: unlike `FollowingPanel` (which renders `follow.isError`), this `followBack` mutation surfaces no feedback if the follow fails — the button just re-enables silently. Consider showing the error message (`followBack.error`) on failure for consistency.

fiddur (commented):
✅ Approved

Clean, well-scoped change with good test coverage. The extraction of `actorProfileLink`/`ActorName` correctly links local actors to `/u/:username` and remote actors to their own URL (opened externally), falling back to plain text for non-http(s) URIs. The "Follow back" logic is sound: mutual detection keys on `actor_uri`, which is the same URI for a follower and the followee record created by following them back, and the shared `['following']` query key means a single fetch drives both panels and refetches after follow-back. Removing `ShareLinkButton` from `PublicProfile` leaves no dangling reference — it's still used by `PublicDashboard`.

Non-blocking notes:

- The `followBack` mutation surfaces no error to the user (button just re-enables on failure), whereas `FollowingPanel` renders `follow.isError`. A failed follow-back (network, rate limit, etc.) is silent. Minor UX inconsistency — see inline comment.
- Until the `['following']` query resolves, `followingUris` is empty, so "Follow back" briefly shows for accounts you already follow. Harmless flicker.

Neither is blocking; approving.

apps/web/src/pages/Feed/FollowersPanel.tsx
fiddur: Done — follow-back now shows the error message on its own line under the row (mobile-friendly, matching FollowingPanel). Resolving.

fiddur (commented):
✅ Approved

Solid, well-factored change. The new `actorProfileLink` helper cleanly generalizes `localProfilePath` (local `/u/:username` vs. external remote URL), the `ActorName` component de-duplicates the name-linking logic across both panels, and the follow-back flow reuses the existing mutation/error patterns.

Correctness and security check out:
- `actorProfileLink` validates the protocol is `http:`/`https:` before returning an external `href`, so `javascript:`/`mailto:` URIs fall through to plain text — no injectable anchor href. External links correctly use `target="_blank" rel="noopener noreferrer"`.
- Follow-back matches followers against following by `actor_uri` (shared `['following']` query key, one fetch), and invalidates `['following']` on success so the button hides once the follow lands. Error handling mirrors the existing FollowingPanel form (`role="alert"`, server-message fallback).
- New CSS (`#991b1b`, `flex-wrap`) is consistent with the existing `.feed-error`/`.following-error` styling. `ShareLinkButton` remains used by PublicDashboard, and `localProfilePath` remains used by HomeTimeline, so neither removal leaves dead code. Tests cover the new helper's local/remote/malformed cases.

Non-blocking observations:
- `HomeTimeline.tsx` still uses `localProfilePath` directly and links only local authors — remote authors there stay plain text, unlike the panels now. Consider migrating it to `ActorName` in a follow-up for consistency.
- `actorProfileLink` treats a same-host URI whose path isn't `/users/:username` as `external` (opens in a new tab). Fine for aurboda's `<host>/users/<username>` actor URIs; just noting the heuristic keys off path-matchability, not host equality.

PR #954, opened 2026-07-15
feat(feed): link remote timeline authors via ActorName

fiddur:
Follow-up to #953 (flagged by that review).

Home timeline post authors only linked **local** actors; remote (Mastodon &c.) authors rendered as plain text. Migrated `TimelineCard` to the shared `ActorName` component, so remote authors now link to their own profile (opened externally, via the WebView's off-origin handling), consistent with the Followers/Following panels.

- No behavior change for local authors (still `/u/:username`).
- Uses the existing `.feed-post-name a` styling (`color: inherit`, hover underline).
- `localProfilePath` is still used internally by `actorProfileLink`, so nothing is orphaned.

`check` green (0 errors).

### DIFF
--- a/apps/web/src/pages/Feed/HomeTimeline.tsx
+++ b/apps/web/src/pages/Feed/HomeTimeline.tsx
@@ -14,7 +14,7 @@ import { formatDistanceToNow } from 'date-fns'
 import { useState } from 'preact/hooks'
 
 import { fetchTimeline } from '../../state/api'
-import { localProfilePath } from './profile-link'
+import { ActorName } from './ActorName'
 import { TimelineStructured } from './TimelineStructured'
 import { useTimelineLive } from './useTimelineLive'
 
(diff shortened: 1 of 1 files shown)


fiddur (commented):
✅ Approved

Clean, minimal refactor. `TimelineCard` now renders author names through the existing `ActorName` component instead of the inline `localProfilePath` link logic, so remote (fediverse) authors get an external profile link too — matching the PR intent. The removed `localProfilePath` import isn't used anywhere else in the file, and the underlying `actorProfileLink` behavior (local vs. external vs. null) is covered by `profile-link.test.ts`, so no new tests are needed for this wiring change.

Non-blocking observation (in unchanged `ActorName.tsx`, so noting here rather than inline): the external author link uses `rel="noopener noreferrer"`, whereas the sibling handle/time link in this card uses `rel="noopener noreferrer nofollow"`. If `nofollow` on outbound user-authored links is intentional elsewhere, you may want the same on the author link for consistency — purely optional.

PR #955, opened 2026-07-15
feat(android): add Home (dashboard) tab, default to it on launch

fiddur:
Adds the main **dashboard** to the app and opens to it by default.

- New **Home** bottom-nav tab (leftmost) rendering the web app's `/?embed=1` in a WebView. For a logged-in user, `/` renders `<Dashboard />`, so the tab shows the main dashboard — reusing the existing `EmbeddedWebScreen` (auth bridge, dark mode, external-link + back handling all apply).
- **Default tab on launch is now Home** (was Sync): `MainTab.Home` is the fallback in `rememberAppState`/`AppState`, and `logout()` resets to it. The widget/shortcut deep-link into `Add` (`EXTRA_OPEN_TAB`) still works.
- Tab order: **Home**, Sync, Add, Feed, Live, Account (existing tabs keep their relative order).

`AppStateTest` updated for the new tab (count, ordinals, names). `:app:assembleDebug` + `:app:testDebugUnitTest` green.

## Note
This makes six bottom-nav tabs. That's at the upper end for a Material bottom bar — workable, but if it feels cramped on a phone we could consolidate later (e.g. fold Sync/Live into a menu). Flagging for your call after seeing it on-device.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/AppState.kt
+++ b/apps/android/app/src/main/java/net/aurboda/AppState.kt
@@ -14,6 +14,7 @@ enum class AppScreen {
 }
 
 enum class MainTab {
+    Home,
     Sync,
     Add,
     Feed,
@@ -24,7 +25,7 @@ enum class MainTab {
 class AppState(
     private val context: Context,
     initialScreen: AppScreen,
-    initialTab: MainTab = MainTab.Sync
+    initialTab: MainTab = MainTab.Home
 ) {
     var currentScreen by mutableStateOf(initialScreen)
         private set
@@ -45,7 +46,7 @@ class AppState(
 
     fun logout() {
         CredentialsManager.clearCredentials(context)
-        currentTab = MainTab.Sync
+        currentTab = MainTab.Home
         currentScreen = AppScreen.Login
     }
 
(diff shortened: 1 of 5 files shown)


fiddur (commented):
✅ Approved

Clean, focused change that adds a `Home` tab (ordinal 0) rendering the web dashboard `/?embed=1` in a WebView and makes it the default launch tab.

Verified:
- `homeContent` mirrors the existing `feedContent` pattern exactly (same `EmbeddedWebScreen` signature and auth params, differing only in the URL path), and the web app already routes `/` to the dashboard and honors `embed=1` to hide chrome — so the embed target is valid.
- The default is applied consistently in `AppState`'s constructor default, `logout()`, and the `rememberAppState` fallback.
- The only `MainScreen` caller (`MainActivity`) is updated with the new required `homeContent` parameter, and the sole `@Preview` targets `HealthConnectScreen`, so nothing else fails to compile.
- The deep-link intent path (`TAB_ADD → MainTab.Add`, else `null → Home`) still works, and no code persists tab ordinals, so inserting `Home` at index 0 is safe.
- Tests (`AppStateTest`) and docs (`android-app.md`) are updated to match the new ordering and count.

One non-blocking note (product decision, not a defect): the default launch tab is now an embedded WebView, so first paint depends on the network, whereas the previous default (native Sync) rendered offline. This is intentional per the PR title/docs, just worth being aware of.

PR #960, opened 2026-07-15
fix(android): give embedded WebView a proper layout viewport (timeline squeeze)

fiddur:
## Symptom
Full-height embedded pages — the **timeline** — render squeezed into a small band in the app, while identical in the browser.

## Diagnosis
The embed CSS is **not** at fault. I rendered `/timeline?embed=1` in **headless Chromium** (same engine as Android WebView) at phone size and it fills full height correctly:

> chart spans 00:00→20:00 over the whole viewport — container fills as intended.

So it's an **Android WebView runtime** difference: with `useWideViewPort` off (the default), the WebView doesn't build a layout viewport from the page's `<meta name="viewport" content="width=device-width, initial-scale=1">`, so `100vh` / `%` / flex-fill heights get no stable reference and collapse. Content-height pages (Home, Feed) never exposed this because they don't depend on viewport height.

## Fix
Enable on the embedded WebView:
```kotlin
settings.useWideViewPort = true
settings.loadWithOverviewMode = true
```
The viewport meta already pins `width=device-width, initial-scale=1`, so scale stays 1:1; Home/Feed are unaffected.

## Verification
- `:app:compileDebugKotlin` + `:app:testDebugUnitTest` green.
- **Needs an on-device check** (I can't run the Android WebView here): open More → Timeline and confirm the chart now fills the screen. If it's still squeezed, the fallback is a post-load relayout nudge — I'll iterate.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -125,6 +125,13 @@ fun EmbeddedWebScreen(
                 WebView(webViewContext).apply {
                     settings.javaScriptEnabled = true
                     settings.domStorageEnabled = true
+                    // Establish a proper layout viewport from the page's
+                    // `<meta viewport>` (width=device-width, initial-scale=1).
+                    // Without this the WebView doesn't give viewport-height /
+                    // flex-fill layouts a stable height, collapsing full-height
+                    // pages like the timeline (scale stays 1:1 either way).
+                    settings.useWideViewPort = true
+                    settings.loadWithOverviewMode = true
                     // Let the embedded web app's own dark styles apply when the
                     // system is in dark mode (the page declares color-scheme
                     // support, so WebView uses its CSS rather than auto-inverting).


fiddur (commented):
✅ Approved

Clean, well-scoped Android fix. Enabling `useWideViewPort` + `loadWithOverviewMode` establishes a proper layout viewport from the page's `<meta viewport>`, which resolves the collapsing full-height (`100vh`/flex-fill) layouts like the timeline.

Verified the two conditions that make this safe:

- The embedded WebView only renders the aurboda web app's own pages — external links are routed out to a browser via `shouldOverrideUrlLoading`/`isExternalLink`, so `useWideViewPort` never falls back to the 980px default viewport.
- That web app's SPA entry point (`apps/web/index.html:9`) declares `width=device-width, initial-scale=1.0`, so the layout viewport tracks device width and `loadWithOverviewMode`'s overview scale is effectively a no-op (matching the comment's "scale stays 1:1 either way").

No tests are warranted (pure WebView configuration) and no web counterpart needs syncing since this is a native rendering fix rather than a functional change. The explanatory comment is accurate and helpful. LGTM.

PR #961, opened 2026-07-15
fix(android): pin embedded WebView shell height to window.innerHeight (timeline squeeze)

fiddur:
The **real, verified** fix for the squeezed timeline (supersedes the ineffective #960 approach).

## Root cause (reproduced + instrumented in an emulator)
I booted the API 36 emulator, pointed a WebView at `/timeline?embed=1`, and injected a measurement overlay. Result:

```
iH=915   (window.innerHeight — correct!)
html h=0px  body h=0px
#app oh=0  h=0px  min-height=0px   ← collapsed
.app-content=293  main=293  chart=150
```

So `window.innerHeight` is correct, but the CSS **`100vh` unit and the `%` height chain resolve to 0** in this WebView — a known Android WebView race where Blink's layout viewport height for `vh`/ICB is 0 even though the widget is sized. `#app{height:100vh}` → 0, everything collapses to content height. **#960's `useWideViewPort` did not fix it** (measured `#app=0` with it on). Deferring `loadUrl` until laid-out didn't help either. Setting `<html>` height from `innerHeight` px **did** (`chart 150→772`).

## Fix
A document-start script (injected **only in the WebView**, so browsers are untouched) pins `<html>` height to `window.innerHeight`, lets `body`/`#app` fill via `%`, and re-applies on `resize` (+ re-asserted in `onPageFinished` for WebViews without document-start support).

## Verified
On the emulator, across 3 launches: `chart=772  #app=915  iH=915` — the timeline fills the screen every time (before: `chart=150 #app=0`).

## Follow-up
#960 (`useWideViewPort`) is now redundant for this bug but harmless; leaving it. Could revert separately if desired.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
+++ b/apps/android/app/src/main/java/net/aurboda/ui/screens/EmbeddedWebScreen.kt
@@ -54,6 +54,27 @@ private class AuthBridge(private val authJson: String) {
 }
 
 /**
+ * Pins the app-shell height to `window.innerHeight`. Android WebView can resolve
+ * CSS viewport units (`100vh`) and the `%` height chain to 0 even when the
+ * viewport is sized correctly, which collapses full-height pages (the timeline).
+ * `window.innerHeight` stays reliable, so we set `<html>` to it (in px) and let
+ * `body`/`#app` fill via `%`, re-applying on resize. Injected only in the
+ * WebView, so browsers are unaffected.
+ */
+private const val VIEWPORT_FIX_JS =
+    "(function(){function f(){document.documentElement.style.height=window.innerHeight+'px';" +
+        "if(document.body){document.body.style.height='100%';var a=document.getElementById('app');if(a)a.style.height='100%';}}" +
+        "window.addEventListener('resize',f);document.addEventListener('DOMContentLoaded',f);f();})();"
+
+private fun installViewportFix(webView: WebView, origin: String?) {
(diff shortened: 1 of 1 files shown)


fiddur (commented):
✅ Approved

Small, well-scoped Android-only fix that pins the embedded WebView's `<html>` height to `window.innerHeight` (in px) and lets `body`/`#app` fill via `%`, working around the WebView collapsing `100vh`/`%` height chains to 0 (the timeline squeeze).

The approach holds up against the actual web layout: `#app` is the app shell (`apps/web/index.html`, `style.css` `#app { height: 100vh }`), and on the phone/mobile layout (`@media max-width: 60em`) `.app-content` is `height:auto; flex:1` inside a column `#app`, so pinning `#app` to `innerHeight` flows height down correctly. Injection is defensive and correctly ordered: a document-start script (when supported and origin is known), an immediate `f()`, plus `DOMContentLoaded`/`resize` listeners and an `onPageFinished` re-run — all guarded (`if(document.body)`, `if(a)`), so it also covers WebViews without document-start support and post-load layout. Setting `documentElement.style.height` doesn't change `window.innerHeight`, so there's no resize feedback loop. Imports and patterns (`WebViewCompat`/`WebViewFeature`/`runCatching`/`DOCUMENT_START_SCRIPT`) already exist and mirror `installAuthBridge`.

One minor, non-blocking note: because the IIFE runs both at document-start and again in `onPageFinished`, each run registers a fresh `resize`/`DOMContentLoaded` listener. This is bounded per document (a new document clears prior listeners) and harmless, so no change needed. No tests are added, which is reasonable given WebView JS injection isn't unit-testable in this file.

Looks good to merge.

PR #962, opened 2026-07-15
feat(feed): per-actor notify_on_post preference on follows (#957 slice 1)

fiddur:
**Slice 1 of #957** (post notifications). Adds a per-followed-account preference for whether that actor's new posts should notify — the backend + web foundation the Android background poller (Slice 2) will read.

## Layers (API/MCP parity)
- **DB**: `notify_on_post BOOLEAN NOT NULL DEFAULT true` on `feed_following`, with an idempotent additive migration. Kept out of the upsert `SET`, so **muting survives a re-follow**.
- **api-spec**: `notify_on_post` on `FollowingActor`; new `UpdateFollowingBody` schema.
- **backend**: `updateFeedFollowingNotify` DB fn → `PATCH /feed/following/:id` (REST) and `set_following_notify` (MCP); `serializeFollowing` exposes the field.
- **web**: a 🔔/🔕 toggle per row in the **Following** panel + `updateFollowingNotify` client.
- **docs**: `docs/features/feed.md`.

## Tests
- Integration: default `true`, toggle, persistence, **preserved across re-follow**, null for unknown id.
- Affected suites green: backend unit + `feed-following` integration + `feed-router`, web (548).

## Next (Slice 2)
Android WorkManager poller reading `notify_on_post`.

## Note
`oxfmt` reflowed a few pre-unformatted files (braces only) — kept per repo convention.

### DIFF
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -27,6 +27,7 @@ import {
   listFeedFollowers,
   listFeedFollowing,
   listFeedPosts,
+  updateFeedFollowingNotify,
   updateFeedPost,
 } from '../db/index.ts'
 import { buildArticleContent, mergeArticleContent } from '../services/article.ts'
@@ -186,6 +187,20 @@ export const registerFeedTools = (
       const removed = await followActions.unfollow(user, id)
       if (!removed) return errorResponse('Not following that actor')
       return jsonResponse({ id, unfollowed: true })
+    },
+  )
+
+  server.tool(
+    'set_following_notify',
+    'Enable or disable notifications for new posts from a followed actor (by the local follow id from `list_following`).',
+    {
+      id: z.string().uuid().describe('Local follow id (the `id` from list_following)'),
+      notify_on_post: z.boolean().describe('Whether new posts from this actor should notify'),
+    },
+    async ({ id, notify_on_post }) => {
+      const record = await updateFeedFollowingNotify(user, id, notify_on_post)
+      if (!record) return errorResponse('Not following that actor')
+      return jsonResponse(serializeFollowing(record))
     },
   )
 
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -11,6 +11,7 @@ import {
   timelineQuerySchema,
   updateArticleBodySchema,
   updateFeedPostBodySchema,
+  updateFollowingBodySchema,
 } from '@aurboda/api-spec'
 import { z } from 'zod'
 
@@ -195,7 +196,7 @@ export const registerFeedTools = (
     'Enable or disable notifications for new posts from a followed actor (by the local follow id from `list_following`).',
     {
       id: z.string().uuid().describe('Local follow id (the `id` from list_following)'),
-      notify_on_post: z.boolean().describe('Whether new posts from this actor should notify'),
+      ...updateFollowingBodySchema.shape,
     },
     async ({ id, notify_on_post }) => {
       const record = await updateFeedFollowingNotify(user, id, notify_on_post)
--- a/apps/backend/src/db/feed-following.integration.test.ts
+++ b/apps/backend/src/db/feed-following.integration.test.ts
@@ -13,6 +13,7 @@ import {
   markFeedFollowingAccepted,
   removeFeedFollowing,
   removeFeedFollowingByActor,
+  updateFeedFollowingNotify,
   upsertFeedFollowing,
 } from './feed-following.ts'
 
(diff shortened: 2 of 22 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped slice. The `notify_on_post` preference is threaded end-to-end (DB column + `updateFeedFollowingNotify`, additive migration, MCP `set_following_notify`, `PATCH /feed/following/:id`, api-spec schemas, web toggle) with matching docs.

Verified the potentially risky parts:
- **Migration** mirrors the existing `feed_follower_columns` idiom exactly. The `feed_following_notify` pseudo-table key runs its idempotent `ADD COLUMN IF NOT EXISTS` both in `initializeSchema` (fresh DBs — harmless no-op) and in `migrateSchema`'s create loop (existing DBs, since `existingTableNames.has('feed_following_notify')` is always false for a non-table key). A `42703` on the widened `FEED_FOLLOWING_COLUMNS` select also triggers the migrate-and-retry path. Correct.
- **Authz**: each user has their own database, so `updateFeedFollowingNotify`'s `WHERE id = $2` (no user filter) is safe and consistent with `removeFeedFollowing`/`getFeedFollowing`. An unknown id returns null → 404 / `errorResponse`.
- **Re-follow semantics**: `upsertFeedFollowing`'s `ON CONFLICT DO UPDATE` deliberately omits `notify_on_post`, so a muted follow stays muted — covered by the integration test.
- Generated OpenAPI/Kotlin artifacts aren't committed (built in CI), so no regenerated files were expected. The new `updateFollowingBodySchema`/`UpdateFollowingBody` are re-exported via `export * from './feed.ts'`.

Two non-blocking observations: (1) the MCP tool inlines `notify_on_post: z.boolean()` instead of spreading `...updateFollowingBodySchema.shape` per the api-spec convention (see inline note); (2) the PR bundles a handful of unrelated formatting-only diffs (challenge-federation.ts, chart-data.test.ts, mutations.test.ts, safe-fetch.ts, shared-dashboard-data.ts, admin.ts) — all behavior-preserving, likely `pnpm fix` drift, just review noise.

apps/backend/src/mcp/feed-tools.ts
```diff
@@ -189,6 +190,20 @@ export const registerFeedTools = (
     },
   )
 
+  server.tool(
+    'set_following_notify',
+    'Enable or disable notifications for new posts from a followed actor (by the local follow id from `list_following`).',
+    {
+      id: z.string().uuid().describe('Local follow id (the `id` from list_following)'),
+      notify_on_post: z.boolean().describe('Whether new posts from this actor should notify'),
```
fiddur: Optional/non-blocking: per the api-spec convention in CLAUDE.md, an MCP tool with extra fields should reuse the body schema — `{ id: z.string().uuid().describe(...), ...updateFollowingBodySchema.shape }` — rather than re-declaring `notify_on_post`. Right now the field and its description are duplicated from `updateFollowingBodySchema` (the descriptions even differ slightly), which defeats the single-source-of-truth intent.

fiddur (commented):
✅ Approved

Clean, well-scoped slice-1 that adds a per-actor `notify_on_post` preference on follows. The change is consistent across all layers and follows the project's documented conventions.

**Verified good:**
- **Migration** — the `feed_following_notify` entry (a fake table key whose "create statement" is an idempotent `ALTER TABLE ... ADD COLUMN IF NOT EXISTS`) exactly mirrors the existing `feed_follower_columns` precedent. It runs on fresh DBs via `initializeSchema` (a harmless no-op, since `feed_following` already defines the column and is ordered first) and on existing DBs via the `migrateSchema` loop (which runs it because no real table is named `feed_following_notify`). Existing rows correctly default to `true`.
- **Re-follow preservation** — `upsertFeedFollowing`'s `ON CONFLICT DO UPDATE` deliberately doesn't touch `notify_on_post`, so a mute survives a re-follow; the integration test asserts exactly this.
- **API/MCP parity** — `set_following_notify` (MCP) and `PATCH /feed/following/:id` (REST) both added, sharing `updateFollowingBodySchema.shape` per the api-spec convention. Response shapes are valid (`baseResponseSchema` carries `error`/`success`), and the 404/`Not following that actor` paths are consistent between the two surfaces.
- **Scoping/security** — `updateFeedFollowingNotify` omits a user predicate in the WHERE clause, but that matches every sibling function: each user has an isolated per-user database, so row ownership is enforced by the connection, not the query.
- **Tests & docs** — DB function has integration coverage (default, toggle, persistence, unknown-id → null), the serializer is unit-tested, and `docs/features/feed.md` is updated. Generated api-spec artifacts (OpenAPI YAML / Kotlin) are gitignored and regenerated by CI, so their absence from the diff is expected.

**Minor, non-blocking:**
- Several unrelated prettier-only reformats are bundled in (`admin.ts`, `safe-fetch.ts`, `shared-dashboard-data.ts`, `challenge-federation.ts`, `chart-data.test.ts`, `mutations.test.ts`). Behavior-neutral, just diff noise.
- The bell toggle's accessible name comes only from the emoji + `title`; an explicit `aria-label` would read more clearly to screen readers, but it's already a proper `aria-pressed` toggle button, so this is optional.

No blocking issues — good to go.

PR #963, opened 2026-07-15
feat(android): notify on new posts from followed accounts (#957 slice 2)

fiddur:
**Slice 2 of #957** (completes it). The Android app now notifies the user when accounts they follow post to their home timeline — honouring the per-account `notify_on_post` flag from slice 1 (#962).

## How it works
- **`NotificationWorker`** — a periodic WorkManager job (15 min, `CONNECTED` constraint, mirroring `SyncWorker`). Each run fetches `/feed/following` (which actors still have notifications on) + `/feed/timeline` (newest posts), decides what's new, and posts notifications.
- **`PostNotifications.kt`** — hand-written kotlinx.serialization DTOs (only the fields needed; `ignoreUnknownKeys`) + the **pure, unit-tested `decideNotifications`**: notify posts newer than a stored **high-water mark** from actors in the notify-on set; the **first run only records the mark** (no backlog dump). Plus `htmlToSnippet` for the body.
- **Opt-in toggle** — a *"Notify me about new posts"* switch on the Account screen requests `POST_NOTIFICATIONS` (Android 13+) and schedules/cancels the worker; `AurbodaApplication` resumes it on app start. Default **off**.
- **Tap → Feed tab** (`MainActivity.TAB_FEED`).
- **Per-account muting** happens on the web Feed page (the 🔔 toggle from #962).

## Tests
- `PostNotificationsTest`: first-run-records-only, newer-than-high-water + notify-on filtering, oldest-first ordering, high-water never regresses, unparseable timestamps ignored, `htmlToSnippet`. `:app:testDebugUnitTest` + `:app:assembleDebug` green.

## Needs on-device verification
End-to-end delivery needs a real account with a followed account that posts — please verify once built: toggle on (permission prompt), then a new post from a followed (un-muted) account raises a notification that opens the Feed; muting that account on the web stops it.

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/AurbodaApplication.kt
+++ b/apps/android/app/src/main/java/net/aurboda/AurbodaApplication.kt
@@ -26,6 +26,11 @@ class AurbodaApplication : Application() {
     } else {
       Log.d(TAG, "Background sync is not enabled")
     }
+
+    // Resume the home-timeline post notifier across restarts if the user left it on.
+    if (isPostNotificationsEnabled(this)) {
+      NotificationWorker.schedule(this)
+    }
   }
 
   private fun isBackgroundSyncEnabled(context: Context): Boolean {
--- a/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
+++ b/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
@@ -143,6 +143,7 @@ class MainActivity : ComponentActivity() {
   companion object {
     const val EXTRA_OPEN_TAB = "open_tab"
     const val TAB_ADD = "add"
+    const val TAB_FEED = "feed"
   }
 
   override fun onCreate(savedInstanceState: Bundle?) {
@@ -150,6 +151,7 @@ class MainActivity : ComponentActivity() {
     val initialTab =
       when (intent?.getStringExtra(EXTRA_OPEN_TAB)) {
         TAB_ADD -> MainTab.Add
+        TAB_FEED -> MainTab.Feed
         else -> null
       }
     setContent {
--- a/apps/android/app/src/main/java/net/aurboda/NotificationApi.kt
+++ b/apps/android/app/src/main/java/net/aurboda/NotificationApi.kt
@@ -1,0 +1,44 @@
+package net.aurboda
+
+import android.util.Log
+import io.ktor.client.HttpClient
+import io.ktor.client.request.get
+import io.ktor.client.request.headers
+import io.ktor.client.statement.bodyAsText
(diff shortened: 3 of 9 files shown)


fiddur (commented):
✅ Approved

Clean, well-factored slice. The which-posts-to-notify logic is a pure, DI-friendly function (`decideNotifications`) with good unit coverage (first-run backlog suppression, high-water monotonicity, muting, ordering, unparseable timestamps), and the worker is thin I/O + notification plumbing around it. Verified the integration points hold up:

- The Kotlin lite-models line up with the api-spec source of truth: `TimelineEntry` (`object_uri`/`actor_uri`/`published_at`/`content`/nullable `display_name`+`handle`), `TimelinePage.entries`, `FollowingActorLite` (`actor_uri`/`notify_on_post`), and `FollowingList.following` all match `timelineEntrySchema`/`timelineResponseSchema`/`followingActorSchema`/`followingResponseSchema`. `appJson` has `ignoreUnknownKeys = true`, so the deliberately-partial models decode fine.
- Endpoint paths are correct: routers mount `/feed/following` and `/feed/timeline`, and `credentials.apiUrl` already appends `/api`, matching the existing `"$apiUrl/…"` convention.
- `published_at` is emitted via `record.published_at.toISOString()` (Z-suffixed), which `OffsetDateTime.parse` handles — no risk of every post being silently dropped as unparseable.
- `POST_NOTIFICATIONS` is declared in `AndroidManifest.xml`, so the runtime request launcher works; the permission/toggle state machine keeps the switch off until the grant lands, and the worker re-checks `areNotificationsEnabled()` before posting.
- Docs and the app-restart resume path (`AurbodaApplication`) are updated.

Two non-blocking design notes to consider for a later slice (not required for this PR):

1. **Page-cap gap.** The poll fetches only the newest timeline page (default `limit=20`, no cursor) and advances the high-water to the newest post seen. If a user follows very active accounts and more than a page of posts arrive within a 15-minute cycle, the oldest of those new posts fall off the page yet the high-water jumps past them, so they are never notified. Fine as a best-effort v1; worth a comment or a follow-up if it becomes a real gap.
2. **`takeLast(MAX_PER_RUN)` silent drop.** When more than 8 eligible posts exist in a run, the oldest eligible ones are dropped (and, per the high-water advance, won't re-notify). This is a reasonable anti-spam cap; just flagging the intentional data loss.

Minor: `htmlToSnippet` only decodes a handful of named entities and `&#39;` (other numeric entities pass through), and decodes `&amp;` before `&lt;`/`&gt;` so a double-escaped sequence like `&amp;lt;` collapses to `<`. Cosmetic only — the source is already server-sanitised and this is just notification-preview text.

PR #964, opened 2026-07-16
feat(android): open Goals from the goal widget, keep screen if app already open

fiddur:
Tapping the goal widget now opens the app on the **Goals** page (which is embedded in the app now), but doesn't yank you off your current screen if the app's already open.

## Changes
- **Widget → Goals**: `HrZoneWidgetProvider` now deep-links to the embedded Goals page (`More → /goals`) via `EXTRA_OPEN_TAB=more` + `EXTRA_MORE_PATH=/goals` (was opening the Add tab).
- **Keep current screen if already open**: `MainActivity` is now `launchMode=singleTop`, and the widget launches with `FLAG_ACTIVITY_SINGLE_TOP`. When the app is already running it's reused via `onNewIntent` (not overridden), so it stays on whatever screen it's on. The Goals deep link is applied **only on a cold start** (`onCreate`).
- **One-shot deep link**: `MoreScreen` takes an `initialPath` consumed once (`LaunchedEffect` + `onInitialPathConsumed`), so a later *manual* visit to More still shows the hub, not Goals.

## Verification
`:app:assembleDebug` + `:app:testDebugUnitTest` green. Deep-link/launch behavior needs an on-device check (place the goal widget, tap it cold → lands on Goals; tap it while the app is open on another screen → stays put).

### DIFF
--- a/apps/android/app/src/main/AndroidManifest.xml
+++ b/apps/android/app/src/main/AndroidManifest.xml
@@ -103,6 +103,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:label="@string/app_name"
+            android:launchMode="singleTop"
             android:theme="@style/Theme.Aurboda">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
+++ b/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
@@ -144,23 +144,33 @@ class MainActivity : ComponentActivity() {
     const val EXTRA_OPEN_TAB = "open_tab"
     const val TAB_ADD = "add"
     const val TAB_FEED = "feed"
+    const val TAB_MORE = "more"
+
+    /** With [EXTRA_OPEN_TAB] = [TAB_MORE], the More web page to open (e.g. "/goals"). */
+    const val EXTRA_MORE_PATH = "more_path"
   }
 
+  // Deep links (widget / notification) only steer the initial screen on a cold
+  // start. The activity is `singleTop`, so when it's already open the launch is
+  // delivered to onNewIntent (not overridden) and the current screen is kept.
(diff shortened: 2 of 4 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped change that makes the goals widget open the More → Goals web page on cold start while preserving the current screen when the app is already running.

**Correctness looks good:**
- The consume-once flow is sound. `pendingMorePath` is seeded from the intent, and `MoreScreen.destination` is `remember`ed, so `onInitialPathConsumed()` nulling `pendingMorePath` triggers a recomposition without resetting the already-open web page. `MainScreen` renders tab content via `when(currentTab)`, so leaving and re-entering More disposes/recreates `MoreScreen`, giving the intended "later manual visit shows the hub" behavior.
- The global `android:launchMode="singleTop"` is consistent with the existing entry points: `NotificationWorker` (TAB_FEED) and `ble/SensorService` already launch `MainActivity` with `FLAG_ACTIVITY_SINGLE_TOP` and rely on cold-start `onCreate` for tab steering. Since `onNewIntent` is intentionally not overridden, an already-open app keeps its screen for the widget, notification, and sensor intents alike — no deep-link regression.
- The widget deep-link path `/goals` matches the existing `MoreItem("Goals", MoreDestination.Web("/goals"))` in `moreGroups`, and `FLAG_ACTIVITY_NEW_TASK` is correctly retained for the non-activity widget context.

**Minor, non-blocking notes:**
- If the widget is tapped while logged out, the deep link is preserved and Goals opens after login (More tab + pending path survive to the first Main composition) — reasonable, just worth being aware of.
- As with the pre-existing tab state, `pendingMorePath`/`destination` use `remember` (not `rememberSaveable`), so a config change/rotation while on the deep-linked Goals page re-opens Goals from the intent. This matches the existing TAB_ADD/TAB_FEED pattern, so it's not a new issue.
- No automated tests are added, which is understandable for Compose/Intent navigation without instrumentation; the logic is simple enough to accept as-is.

PR #965, opened 2026-07-16
fix(analyze): reflect the selected context in the tab label + intro

fiddur:
On the Analyze page, switching **Context** to Heart rate / Stress still showed the tab as "HRV context" and the intro still said "heart rate variability (HRV) and heart rate" — both were hardcoded.

Now both are driven by the selected `contextMetric`:
- Tab label → "HRV context" / "Heart rate context" / "Stress context" (reuses the existing `CONTEXT_METRIC_LABELS`).
- Intro → "…correlate with your heart rate variability (HRV) / heart rate / stress level."

`check` green.

> Note: the separate "loads for minutes with no useful feedback" issue is a bigger caching/async change — proposal coming separately.

### DIFF
--- a/apps/web/src/pages/Correlations/index.tsx
+++ b/apps/web/src/pages/Correlations/index.tsx
@@ -32,6 +32,13 @@ const CONTEXT_METRIC_LABELS: Record<HrvContextMetric, string> = {
   stress_level: 'Stress',
 }
 
+// The metric phrased for the intro sentence ("…correlate with your <phrase>.").
+const CONTEXT_METRIC_PHRASES: Record<HrvContextMetric, string> = {
+  heart_rate: 'heart rate',
+  hrv_rmssd: 'heart rate variability (HRV)',
+  stress_level: 'stress level',
+}
+
 // Active tab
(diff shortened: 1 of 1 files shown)


fiddur (commented):
✅ Approved

Small, well-scoped UI fix that makes the Analyze tab label and intro sentence reflect the currently selected context metric instead of hard-coding "HRV".

The change is correct:
- The new `CONTEXT_METRIC_PHRASES` is typed as `Record<HrvContextMetric, string>`, so TypeScript enforces that all three enum values (`hrv_rmssd`, `heart_rate`, `stress_level`) are covered — no risk of an `undefined` lookup at runtime.
- Splitting the Title-case tab label (`CONTEXT_METRIC_LABELS`, e.g. "Heart rate context") from the natural-language intro phrase (`CONTEXT_METRIC_PHRASES`, e.g. "heart rate variability (HRV)") reads well and matches the intent of the PR.
- It mirrors the existing `CONTEXT_METRIC_LABELS` pattern, and no tests assert on the old strings, so nothing breaks.

No blocking issues found.

PR #966, opened 2026-07-16
perf(analyze): don't block correlations on inline provider syncs

fiddur:
Addresses the "Analyze spins for minutes" issue (the "optimize first" direction).

## Root cause
`getHrvActivitiesCorrelation` (the Analyze **context** tab) and `getActivityImpact` (the row-click before/during/after detail) **awaited external freshness syncs** (Oura tags/sessions, RescueTime, calendars) *inline* before computing. A first/stale sync hits external APIs and can take minutes — long enough to blow the HTTP timeout — so the page just shows "Analyzing correlations…" indefinitely.

The fire-and-forget **`triggerCorrelationSyncs`** helper already exists for exactly this ("Correlation analysis is retrospective … it must never block the response on live external syncs") and is used by `explore.ts` / `generic.ts` — these two services simply never got migrated.

## Fix
Swap the blocking `await Promise.all([...syncs])` for `triggerCorrelationSyncs(sync, user)` in both. The (retrospective) analysis now returns from existing DB data immediately; the syncs refresh in the background so data is fresh next time. Both services already fire the exact sync set the helper does, so no behavior change beyond not blocking.

## Tests
All 88 correlation tests pass; typecheck clean.

## Follow-up (if still slow)
With syncs no longer blocking, the remaining cost is the raw HRV/HR fetch over the period (continuous HR can be large). If that's still slow, the next step is bucketing that fetch (`getTimeSeriesBucketed`) and/or caching results — flagged, not needed yet. (`event-probability.ts` has the same blocking pattern with a different sync set; left out of scope.)

### DIFF
--- a/apps/backend/src/services/correlations/activity-impact.ts
+++ b/apps/backend/src/services/correlations/activity-impact.ts
@@ -7,6 +7,7 @@ import type { ActivityImpactResult, TimeWindowStats } from './types.ts'
 
 import { getAllActivitiesInRange, getProductivity, getTimeSeries } from '../../db/index.ts'
 import { getPlaceVisits } from '../locations.ts'
+import { triggerCorrelationSyncs } from './background-sync.ts'
 import { getDataInRange, mean, stddev } from './utils.ts'
 
 /**
@@ -27,15 +28,9 @@ export async function getActivityImpact(
   start.setDate(start.getDate() - periodDays)
   start.setHours(0, 0, 0, 0)
 
-  // Auto-sync if provider available
-  if (sync) {
-    await Promise.all([
-      sync.syncOuraIfNeeded(user, 'tags'),
-      sync.syncOuraIfNeeded(user, 'sessions'),
-      sync.syncRescueTimeIfNeeded(user),
-      sync.syncCalendarsIfNeeded(user),
-    ])
-  }
+  // Freshness syncs run in the background — never block this retrospective
+  // analysis on a slow external sync (see background-sync.ts).
+  triggerCorrelationSyncs(sync, user)
 
   // Fetch HRV/HR/stress data
   const [hrvData, hrData, stressData] = await Promise.all([
--- a/apps/backend/src/services/correlations/hrv-activities.ts
+++ b/apps/backend/src/services/correlations/hrv-activities.ts
@@ -14,6 +14,7 @@ import type {
 
 import { getAllActivitiesInRange, getProductivity, getTimeSeries } from '../../db/index.ts'
 import { getPlaceVisits } from '../locations.ts'
+import { triggerCorrelationSyncs } from './background-sync.ts'
 import { addBaselineDelta, calculateHrvStats, getDataInRange, pearsonCorrelation } from './utils.ts'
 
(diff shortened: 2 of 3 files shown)


fiddur (commented):
✅ Approved

Solid, low-risk change. `getActivityImpact` and `getHrvActivitiesCorrelation` now fire the freshness syncs via the existing `triggerCorrelationSyncs` fire-and-forget helper instead of `await Promise.all([...])`, matching the pattern already used in `generic.ts` and `explore.ts`. This correctly avoids blocking a retrospective analysis on a slow first/stale external sync, and the deliberate tradeoff (current request uses existing data; fresh data lands on the next request) is documented both here and in `background-sync.ts`.

Verification notes:
- The `hrv-activities.test.ts` "calls sync provider when provided" assertions still hold: `triggerCorrelationSyncs` invokes each `sync*IfNeeded` synchronously (inside the `Promise.allSettled` array), so the `toHaveBeenCalledWith` expectations pass even though the promises are no longer awaited.
- The non-MCP caller `shared-dashboard-data.ts:202` passes no `sync` provider, so `triggerCorrelationSyncs` is a no-op there — no behavioral change.
- Rejections are absorbed by `Promise.allSettled`, so no unhandled-rejection risk.
- I could not execute the test suite (dependencies are not installed in this worktree and the npm cache is read-only), so I verified the behavior by inspection rather than a run.

One non-blocking follow-up: `docs/features/correlations.md:137` still reads "Auto-sync is triggered before correlation computation, which may add a brief delay on the first load." After this PR that is no longer accurate — the sync no longer blocks computation; instead the first load may show slightly stale data while the sync runs in the background. Per the repo's "update the corresponding docs when changing sync behavior" standard, consider updating that line (it was arguably already stale from the earlier `generic`/`explore` migration, so this PR is a good place to correct it). That file isn't part of this diff, hence noting it here rather than inline.

fiddur (commented):
✅ Approved

Clean, well-scoped perf change. `getActivityImpact` and `getHrvActivitiesCorrelation` are migrated from the blocking inline `await Promise.all([...syncs])` to the fire-and-forget `triggerCorrelationSyncs` helper, which already exists on develop (PR #792) and is well covered by `background-sync.test.ts`. This makes them consistent with the other correlation queries and with the query-layer pattern in `services/queries/tags.ts` / `daily-summary.ts`.

Correctness looks good:
- The `sync` parameter is still used (passed into the helper), so no unused-import/param regressions.
- The old `if (sync)` guard is preserved inside the helper (`if (!sync) return`), so behavior when no provider is supplied is unchanged.
- The helper uses `Promise.allSettled` + `void`, so a rejected background sync can't surface as an unhandled rejection.
- Existing unit tests call these functions without a `sync` provider, so they remain valid.
- The docs update in `correlations.md` accurately describes the new fire-and-forget semantics and the trade-off (first load after a stale period may reflect slightly out-of-date data; freshness lands on the next request).

The trade-off (current request returns existing data instead of waiting for fresh sync) is intentional, documented, and consistent with the precedent set in PR #792. No blocking issues found.

PR #967, opened 2026-07-27
feat(feed): server-side article-block rendering + AS2 Article federation (#937 Slice C, part 1)

fiddur:
> [!NOTE]
> **Correction (2026-08-19, see #970):** contrary to the C2 description below, the shipped code federates an article as a **`Create{Note}`** at the shared post object path (`…/users/<user>/feed/<id>`) — not as an AS2 `Article` at a separate `…/article` id. Mastodon discards `content` on converted types like `Article`, so a `Note` is what actually shows the prose + attached images. `docs/features/feed.md` (updated in this PR) describes the real behaviour; the C2 prose below is stale from a pre-review revision.

Part of #934 (Shareable analysis Articles). This is the first of two PRs for **Slice C (#937)** — consolidated into larger PRs to minimise Docker rebuilds on `develop`. This one delivers **server-side article-block image rendering (C1)** and **AS2 `Article` federation (C2)**; a follow-up covers native peer enrichment (C3) + Reddit/markdown export (C4).

## C1 — gated server-side article-block images

New endpoints, rendered on demand:

```
GET /public/:username/feed/:postId/blocks/:index/image.{png,svg}
```

- **Security gate (lands the #943-deferred work).** Unlike an activity chart, an article block has **no `include_chart` opt-in flag** — a chart/correlation block can embed *any* metric over *any* window, so the post's **visibility** is the whole authorization boundary: `public`/`unlisted` open; `followers`-only reachable only via the post's unguessable `?token=` (constant-time compare); `no-store`. Mirrors the existing `chart.png` gate (`resolveImageWindow` → new `resolveArticleBlock`).
- **Chart blocks** reuse `buildChartSvg` over a bucketed metric series matching the web's live render (shared `defaultArticleChartBucket`).
- **Correlation blocks** get a new pure server scatter renderer (`scatter-svg.ts`) — the server twin of the web `ArticleCorrelationBlock`: same OLS line + binary-trigger group-comparison headline, on the dark card background.
- The render cache keys on the post's `updated_at`, so editing an article serves a fresh image (articles, unlike shared activities, are mutable).
- **DRY:** moved `metricLabels`/`getMetricDisplayName`, `describeSelectorAxis`, and `linearRegression` into `@aurboda/api-spec` as the single source; web re-exports so existing imports keep working.

## C2 — AS2 Article federation

- An article federates as an AS2 **`Article`** (title `name` + sanitised prose HTML `content` + one PNG attachment per chart/correlation block), so Mastodon shows prose + images. Its object id is its own dispatcher path (`…/feed/<id>/article`), distinct from the `Note` path.
- Prose markdown → HTML via `marked` + the shared `sanitizeRemoteHtml` allowlist (the #910 XSS boundary, server side).
- Outbox now lists articles as `Create{Article}` (previously dropped on the `activity_id` guard); the `Article` object dispatcher serves them; article delete tombstones the article object id.
- `FeedDeliver` gains `createdArticle`/`updatedArticle`; the REST `/feed/articles` routes and the `create_article`/`update_article` MCP tools fan out (parity).

## Testing

- `pnpm check` green (tsgo + oxlint) across api-spec / backend / web.
- Unit: `resolveArticleBlock` (visibility gate, window inheritance, prose/out-of-range/non-article rejection), `buildScatterSvg`, article content/attachments, the Article builders.
- Integration (testcontainers): an article appears in the outbox as `Create{Article}` and resolves as an `Article`; existing federation + feed-router integration suites still pass.

Adds `marked` to the backend. Docs updated (`docs/features/feed.md`).

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -57,7 +57,15 @@ import { stravaClient } from './integrations/strava/client.ts'
 import { createMcpRouter } from './mcp.ts'
 import { createFeedTombstoneRouter } from './routes/feed-tombstone-router.ts'
 import { createOAuthRouter } from './routes/oauth-router.ts'
-import { deliverFeedDelete, deliverFeedPost, deliverFeedUpdate } from './services/activitypub/deliver.ts'
+import {
+  deliverFeedArticleDelete,
+  deliverFeedArticlePost,
+  deliverFeedArticleUpdate,
+  deliverFeedDelete,
+  deliverFeedPost,
+  deliverFeedUpdate,
+  toDeliverableArticle,
+} from './services/activitypub/deliver.ts'
 import { createFeedFederation } from './services/activitypub/federation.ts'
 import { createTimelineBackfiller } from './services/activitypub/timeline-backfill.ts'
 import { auditError } from './services/audit-log.ts'
@@ -352,8 +360,14 @@ const main = async () => {
         await deliverFeedPost(feedDeps, user, post, resolved)
       })().catch(onDeliverError('create', user, post.id))
     },
+    // An article has no linked activity, so its Delete is tombstoned from the
+    // article object id; every other post from the Note id.
     deleted: (user, post) => {
-      void deliverFeedDelete(feedDeps, user, post).catch(onDeliverError('delete', user, post.id))
+      const article = toDeliverableArticle(post)
+      const run = article
+        ? deliverFeedArticleDelete(feedDeps, user, article)
+        : deliverFeedDelete(feedDeps, user, post)
+      void run.catch(onDeliverError('delete', user, post.id))
     },
     // Resolve the (merged-span) activity inside the fire-and-forget boundary so a
     // lookup failure never bubbles into the (already-committed) edit's response.
@@ -363,6 +377,20 @@ const main = async () => {
         const activity = await resolveFeedActivity(user, post.activity_id)
         if (activity) await deliverFeedUpdate(feedDeps, user, post, activity)
       })().catch(onDeliverError('update', user, post.id))
+    },
+    // Articles fan out as an AS2 Article (built purely from stored content — no
+    // activity to resolve). Best-effort, same as the Note path.
+    createdArticle: (user, post) => {
+      const article = toDeliverableArticle(post)
+      if (article) {
+        void deliverFeedArticlePost(feedDeps, user, article).catch(onDeliverError('create', user, post.id))
+      }
+    },
+    updatedArticle: (user, post) => {
+      const article = toDeliverableArticle(post)
+      if (article) {
+        void deliverFeedArticleUpdate(feedDeps, user, article).catch(onDeliverError('update', user, post.id))
+      }
     },
   }
   // The network-requiring follow operations, bound to the same federation +
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -58,7 +58,6 @@ import { createMcpRouter } from './mcp.ts'
 import { createFeedTombstoneRouter } from './routes/feed-tombstone-router.ts'
 import { createOAuthRouter } from './routes/oauth-router.ts'
 import {
-  deliverFeedArticleDelete,
   deliverFeedArticlePost,
   deliverFeedArticleUpdate,
   deliverFeedDelete,
@@ -360,14 +359,10 @@ const main = async () => {
         await deliverFeedPost(feedDeps, user, post, resolved)
       })().catch(onDeliverError('create', user, post.id))
     },
-    // An article has no linked activity, so its Delete is tombstoned from the
-    // article object id; every other post from the Note id.
+    // Articles and activities share the Note object id, so a deleted post of
+    // either kind tombstones there — one path.
     deleted: (user, post) => {
-      const article = toDeliverableArticle(post)
-      const run = article
-        ? deliverFeedArticleDelete(feedDeps, user, article)
-        : deliverFeedDelete(feedDeps, user, post)
-      void run.catch(onDeliverError('delete', user, post.id))
+      void deliverFeedDelete(feedDeps, user, post).catch(onDeliverError('delete', user, post.id))
     },
     // Resolve the (merged-span) activity inside the fire-and-forget boundary so a
     // lookup failure never bubbles into the (already-committed) edit's response.
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -373,8 +373,8 @@ const main = async () => {
         if (activity) await deliverFeedUpdate(feedDeps, user, post, activity)
       })().catch(onDeliverError('update', user, post.id))
     },
-    // Articles fan out as an AS2 Article (built purely from stored content — no
-    // activity to resolve). Best-effort, same as the Note path.
+    // Articles fan out as a Create{Note}/Update{Note} built purely from stored
+    // content (no activity to resolve). Best-effort, same as the activity path.
     createdArticle: (user, post) => {
       const article = toDeliverableArticle(post)
       if (article) {
--- a/apps/backend/src/api/rest-routes.ts
+++ b/apps/backend/src/api/rest-routes.ts
@@ -16,7 +16,6 @@ import type { FollowerActions } from '../services/followers.ts'
 import type { FollowActions } from '../services/following.ts'
 import type { InvitationAuth } from '../services/invitation.ts'
 import type { OuraWebhookManager } from '../services/oura-webhook-manager.ts'
-import type { SyncProvider } from '../services/queries/index.ts'
 import type { TimelineHub } from '../services/timeline-hub.ts'
 import type { WebAuthnService } from '../services/webauthn.ts'
 import type { AnyMiddleware } from '../typed-router.ts'
@@ -63,12 +62,15 @@ import { createTrainingLoadRouter } from '../routes/training-load-router.ts'
 import { createTrendsRouter } from '../routes/trends-router.ts'
 import { createWebAuthnRouter } from '../routes/webauthn-router.ts'
 import { createWellKnownRouter, type WellKnownConfig } from '../routes/well-known-router.ts'
-import { renderChartPng, renderRoutePng } from '../services/activitypub/feed-images.ts'
+import { renderChartPng, renderRoutePng, renderScatterPng } from '../services/activitypub/feed-images.ts'
 import { fetchOsmTile } from '../services/activitypub/osm-tiles.ts'
 import { loadAvatarDataUri } from '../services/avatar-resolve.ts'
 import { buildChartSvg } from '../services/charts/chart-svg.ts'
+import { buildScatterSvg } from '../services/charts/scatter-svg.ts'
+import { getContinuousCorrelation } from '../services/correlations/index.ts'
 import { resolveFeedActivity } from '../services/feed.ts'
 import { createOgImageRenderer } from '../services/og-image.ts'
+import { type SyncProvider, queryMetricsBucketed } from '../services/queries/index.ts'
 import { createTemplateLoader } from '../services/web-template.ts'
 
 interface RestRoutesDeps {
@@ -173,6 +175,39 @@ export const mountRestRouters = ({
       // Merged-span window so the rendered chart/route cover what the user
       // shared, matching the Note's duration/metrics (#881).
       getActivity: resolveFeedActivity,
+      // An article chart block's bucketed metric series over its locked window
+      // (mirrors the web's live bucketed render).
+      getArticleChartSeries: async (user, metric, start, end, bucket) => {
+        const result = await queryMetricsBucketed(user, [metric], start, end, bucket)
+        const series: [Date, number][] = []
+        for (const b of result.buckets) {
+          const avg = b.metrics[metric]?.avg
+          if (avg != null) series.push([new Date(b.start), avg])
+        }
+        return series
+      },
+      // An article correlation block's continuous scatter over its locked window;
+      // null when too sparse to be meaningful (n < 3), which 404s the image.
+      getCorrelationScatter: async (user, { end, lagDays, outcome, start, trigger }) => {
+        const c = await getContinuousCorrelation(user, {
+          lagDays,
+          outcome,
+          periodEnd: end.toISOString().slice(0, 10),
+          periodStart: start.toISOString().slice(0, 10),
+          trigger,
+        })
+        if (c.n < 3) return null
+        return {
+          group_comparison: c.group_comparison,
+          n: c.n,
+          outcome,
+          pearson: c.pearson,
+          pearson_p: c.pearson_p,
+          series: c.series,
+          spearman: c.spearman,
+          trigger,
+        }
+      },
       getPost: getFeedPostById,
       getRoute: async (user, start, end) =>
         (await getLocations(user, start, end)).locations.map((l) => l.coordinates),
@@ -182,6 +217,8 @@ export const mountRestRouters = ({
       // Draw the route over the OSM basemap; falls back to a bare shape if tiles
       // can't be fetched (e.g. offline).
       renderRoute: (coords) => renderRoutePng(coords, { fetchTile: fetchOsmTile }),
+      renderScatter: renderScatterPng,
+      renderScatterSvg: buildScatterSvg,
     }),
   )
   httpd.use(createPublicSharesRouter(webHost))
--- a/apps/backend/src/api/rest-routes.ts
+++ b/apps/backend/src/api/rest-routes.ts
@@ -200,12 +200,12 @@ export const mountRestRouters = ({
       },
       // An article correlation block's continuous scatter over its locked window;
       // null when too sparse to be meaningful (n < 3), which 404s the image.
-      getCorrelationScatter: async (user, { end, lagDays, outcome, start, trigger }) => {
+      getCorrelationScatter: async (user, { lagDays, outcome, periodEnd, periodStart, trigger }) => {
         const c = await getContinuousCorrelation(user, {
           lagDays,
           outcome,
-          periodEnd: end.toISOString().slice(0, 10),
-          periodStart: start.toISOString().slice(0, 10),
+          periodEnd,
+          periodStart,
           trigger,
         })
         if (c.n < 3) return null
--- a/apps/backend/src/api/rest-routes.ts
+++ b/apps/backend/src/api/rest-routes.ts
@@ -20,7 +20,13 @@ import type { TimelineHub } from '../services/timeline-hub.ts'
 import type { WebAuthnService } from '../services/webauthn.ts'
 import type { AnyMiddleware } from '../typed-router.ts'
 
-import { getFeedPostById, getLocations, getTimeSeries, markActivityDetailSynced } from '../db/index.ts'
+import {
+  getFeedPostById,
+  getLocations,
+  getTimeSeries,
+  getUserSettings,
+  markActivityDetailSynced,
+} from '../db/index.ts'
 import { processActivityDetail } from '../integrations/garmin/process.ts'
 import { createActivitiesRouter } from '../routes/activities-router.ts'
 import { createActivityTypesRouter } from '../routes/activity-types-router.ts'
@@ -176,9 +182,15 @@ export const mountRestRouters = ({
       // shared, matching the Note's duration/metrics (#881).
       getActivity: resolveFeedActivity,
       // An article chart block's bucketed metric series over its locked window
-      // (mirrors the web's live bucketed render).
+      // (mirrors the web's live bucketed render). Bucket in the author's own
+      // timezone (`device_timezone`, IANA — auto-detected from their device) so a
+      // `1d` bucket splits on the author's calendar days, matching the web render
+      // (which sends the browser tz); falls back to UTC when it's unset.
       getArticleChartSeries: async (user, metric, start, end, bucket) => {
-        const result = await queryMetricsBucketed(user, [metric], start, end, bucket)
+        const settings = await getUserSettings(user)
+        const result = await queryMetricsBucketed(user, [metric], start, end, bucket, {
+          tz: settings?.device_timezone ?? undefined,
+        })
         const series: [Date, number][] = []
         for (const b of result.buckets) {
           const avg = b.metrics[metric]?.avg
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -1,5 +1,3 @@
-import { type Response, Router } from 'express'
-
 /**
  * Public feed-post image endpoints (UNAUTHENTICATED).
  *
@@ -20,10 +18,17 @@ import { type Response, Router } from 'express'
  * effect immediately — the untoken'd public URL then 404s). Mounted before the
  * generic `/public/:username/:slug` resolver.
  */
+import type { ArticleContent, CorrelationSelector, MetricType } from '@aurboda/api-spec'
+
+import { defaultArticleChartBucket, getMetricDisplayName } from '@aurboda/api-spec'
+import { type Response, Router } from 'express'
+
 import type { FeedPostRecord } from '../db/index.ts'
+import type { ScatterSvgData } from '../services/charts/scatter-svg.ts'
 
 import { isValidUsername } from '../api/auth-routes.ts'
 import { isMissingDatabase } from '../db/index.ts'
+import { blockWindow } from '../services/article.ts'
 import { isCapabilityAuthorized } from '../services/feed-capability.ts'
 
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
@@ -34,16 +39,60 @@ export interface ImageActivity {
   end_time?: Date
 }
 
+/** Optional chart styling — an article block labels its chart with the metric. */
+export interface ChartRenderOpts {
+  label?: string
+  color?: string
+}
+
+/** The window-resolved inputs for one article correlation block's scatter. */
+export interface CorrelationBlockParams {
+  trigger: CorrelationSelector
+  outcome: CorrelationSelector
+  lagDays?: number
+  start: Date
+  end: Date
+}
+
 export interface FeedImageDeps {
   getPost: (user: string, postId: string) => Promise<FeedPostRecord | null>
   getActivity: (user: string, activityId: string) => Promise<ImageActivity | null>
   getSeries: (user: string, metric: string, start: Date, end: Date) => Promise<[Date, number][]>
   getRoute: (user: string, start: Date, end: Date) => Promise<[number, number][]>
-  renderChart: (series: [Date, number][]) => Promise<Buffer>
+  renderChart: (series: [Date, number][], opts?: ChartRenderOpts) => Promise<Buffer>
   /** Build the crisp `image/svg+xml` chart (same data as `renderChart`, no raster). */
-  renderChartSvg: (series: [Date, number][]) => string
+  renderChartSvg: (series: [Date, number][], opts?: ChartRenderOpts) => string
   renderRoute: (coords: [number, number][]) => Promise<Buffer>
+  /** A bucketed metric series for an article chart block over its locked window. */
+  getArticleChartSeries: (
+    user: string,
+    metric: MetricType,
+    start: Date,
+    end: Date,
+    bucket: string,
+  ) => Promise<[Date, number][]>
+  /** The continuous correlation for an article correlation block, or null when too sparse (n < 3). */
+  getCorrelationScatter: (user: string, params: CorrelationBlockParams) => Promise<ScatterSvgData | null>
+  renderScatter: (data: ScatterSvgData) => Promise<Buffer>
+  renderScatterSvg: (data: ScatterSvgData) => string
 }
+
+/**
+ * One article chart/correlation block resolved to its render inputs: the block's
+ * kind and its effective `[start, end]` window (its own override, else the
+ * article default). Prose blocks and out-of-range indices don't resolve.
+ */
+export type ResolvedArticleBlock =
+  | { type: 'chart'; metric: MetricType; bucket?: string; start: Date; end: Date; updatedAt: Date }
+  | {
+      type: 'correlation'
+      trigger: CorrelationSelector
+      outcome: CorrelationSelector
+      lagDays?: number
+      start: Date
+      end: Date
+      updatedAt: Date
+    }
 
 /**
  * A tiny memoising render cache with in-flight de-duplication (mirrors
@@ -110,6 +159,117 @@ export const resolveImageWindow = async (
   return activity
 }
 
+/**
+ * Resolve one article chart/correlation block to its render inputs, or `null`
+ * when the request isn't eligible: invalid username / non-UUID id / bad index,
+ * missing DB, missing post, a non-article post, a `followers`-only post without a
+ * matching capability `token`, an out-of-range or prose block, or a block whose
+ * effective window is unbounded / non-increasing. Pure of Express — unit-testable.
+ *
+ * Unlike a shared activity's chart, an article block has NO `include_chart`
+ * opt-in flag: a chart/correlation block can embed any metric over any window, so
+ * the post's visibility (public/unlisted open; followers-only via the unguessable
+ * `token`) is the whole authorization boundary (#943).
+ */
+/**
+ * Load the article behind an image request and authorize it, or `null` when
+ * ineligible (invalid username / non-UUID id, missing DB, missing post, a
+ * non-article post, or a `followers`-only post without a matching capability
+ * `token`). Split out of `resolveArticleBlock` to keep each piece simple.
+ */
+const loadArticleForImage = async (
+  deps: Pick<FeedImageDeps, 'getPost'>,
+  username: string,
+  postId: string,
+  token?: string,
+): Promise<{ article: ArticleContent; updatedAt: Date } | null> => {
+  if (!isValidUsername(username) || !UUID_RE.test(postId)) return null
+  let post: FeedPostRecord | null
+  try {
+    post = await deps.getPost(username, postId)
+  } catch (error) {
+    if (isMissingDatabase(error)) return null
+    throw error
+  }
+  if (post == null || post.kind !== 'article' || post.article == null) return null
+  if (!isCapabilityAuthorized(post, token)) return null
+  return { article: post.article, updatedAt: post.updated_at }
+}
+
+export const resolveArticleBlock = async (
+  deps: Pick<FeedImageDeps, 'getPost'>,
+  username: string,
+  postId: string,
+  index: number,
+  token?: string,
+): Promise<ResolvedArticleBlock | null> => {
+  if (!Number.isInteger(index) || index < 0) return null
+  const loaded = await loadArticleForImage(deps, username, postId, token)
+  if (loaded == null) return null
+  const block = loaded.article.blocks[index]
+  if (block == null || (block.type !== 'chart' && block.type !== 'correlation')) return null
+  const { end, start } = blockWindow(block, loaded.article)
+  if (start == null || end == null) return null
+  const startDate = new Date(start)
+  const endDate = new Date(end)
+  if (startDate.getTime() >= endDate.getTime()) return null
+  const updatedAt = loaded.updatedAt
+  if (block.type === 'chart') {
+    return {
+      bucket: block.bucket,
+      end: endDate,
+      metric: block.metric,
+      start: startDate,
+      type: 'chart',
+      updatedAt,
+    }
+  }
+  return {
+    end: endDate,
+    lagDays: block.lag_days,
+    outcome: block.outcome,
+    start: startDate,
+    trigger: block.trigger,
+    type: 'correlation',
+    updatedAt,
+  }
+}
+
+/** Article-chart line colour, matching the web inline render (`ArticleChartBlock`). */
+const ARTICLE_CHART_COLOR = '#673ab8'
+
+/**
+ * Render one resolved article block to its image bytes (PNG or the crisp SVG),
+ * or `null` when the block has too little data to draw (a chart needs ≥ 2 points;
+ * a correlation needs n ≥ 3, surfaced by `getCorrelationScatter` returning null).
+ * Pure of Express so the routes stay thin.
+ */
+const renderArticleBlockImage = async (
+  deps: FeedImageDeps,
+  user: string,
+  block: ResolvedArticleBlock,
+  format: 'png' | 'svg',
+): Promise<Buffer | null> => {
+  if (block.type === 'chart') {
+    const bucket = block.bucket ?? defaultArticleChartBucket(block.start, block.end)
+    const series = await deps.getArticleChartSeries(user, block.metric, block.start, block.end, bucket)
+    if (series.length < 2) return null
+    const opts: ChartRenderOpts = { color: ARTICLE_CHART_COLOR, label: getMetricDisplayName(block.metric) }
+    return format === 'png'
+      ? deps.renderChart(series, opts)
+      : Buffer.from(deps.renderChartSvg(series, opts), 'utf8')
+  }
+  const data = await deps.getCorrelationScatter(user, {
+    end: block.end,
+    lagDays: block.lagDays,
+    outcome: block.outcome,
+    start: block.start,
+    trigger: block.trigger,
+  })
+  if (data == null) return null
+  return format === 'png' ? deps.renderScatter(data) : Buffer.from(deps.renderScatterSvg(data), 'utf8')
+}
+
 const notFound = (res: Response) => res.status(404).json({ error: 'Not found', success: false })
 
 const sendPng = (res: Response, png: Buffer) => {
@@ -171,6 +331,35 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
     })
     if (!png) return notFound(res)
     sendPng(res, png)
+  })
+
+  // Article chart/correlation block images. Gated by post visibility + capability
+  // token only (no `include_chart` flag — a block can embed any metric, so
+  // visibility is the whole boundary, #943). Cache key includes the post's
+  // `updated_at` so editing the article serves a fresh render (articles, unlike
+  // shared activities, are mutable).
+  router.get('/public/:username/feed/:postId/blocks/:index/image.png', async (req, res) => {
+    const { index, postId, username } = req.params
+    const token = typeof req.query.token === 'string' ? req.query.token : undefined
+    const block = await resolveArticleBlock(deps, username, postId, Number(index), token)
+    if (!block) return notFound(res)
+    const png = await cached(`blockpng:${username}:${postId}:${index}:${block.updatedAt.getTime()}`, () =>
+      renderArticleBlockImage(deps, username, block, 'png'),
+    )
+    if (!png) return notFound(res)
+    sendPng(res, png)
+  })
+
+  router.get('/public/:username/feed/:postId/blocks/:index/image.svg', async (req, res) => {
+    const { index, postId, username } = req.params
+    const token = typeof req.query.token === 'string' ? req.query.token : undefined
+    const block = await resolveArticleBlock(deps, username, postId, Number(index), token)
+    if (!block) return notFound(res)
+    const svg = await cached(`blocksvg:${username}:${postId}:${index}:${block.updatedAt.getTime()}`, () =>
+      renderArticleBlockImage(deps, username, block, 'svg'),
+    )
+    if (!svg) return notFound(res)
+    sendSvg(res, svg)
   })
 
   return router
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -53,10 +53,12 @@ export interface CorrelationBlockParams {
   outcome: CorrelationSelector
   lagDays?: number
   /**
-   * Inclusive day bounds (`YYYY-MM-DD`), sliced from the block's ISO window so they
-   * are the author's wall-clock days — NOT a UTC-normalised day. Matches the web
-   * scatter's `iso.slice(0, 10)`, so an offset-bearing ISO (e.g. from `create_article`
-   * over MCP) covers the same day range in-app and in the federated/exported PNG.
+   * Inclusive day bounds (`YYYY-MM-DD`), sliced straight from the block's ISO
+   * window (`iso.slice(0, 10)`) to match the web scatter's `toDay()`. Block windows
+   * are `Z`-only (`iso8601DateTimeSchema` = `z.iso.datetime()`, offset false), so
+   * this is the UTC day — but slicing the raw string rather than round-tripping
+   * through `Date` keeps that parity exact and stays correct if the schema ever
+   * gains offset support.
    */
   periodStart: string
   periodEnd: string
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -52,8 +52,14 @@ export interface CorrelationBlockParams {
   trigger: CorrelationSelector
   outcome: CorrelationSelector
   lagDays?: number
-  start: Date
-  end: Date
+  /**
+   * Inclusive day bounds (`YYYY-MM-DD`), sliced from the block's ISO window so they
+   * are the author's wall-clock days — NOT a UTC-normalised day. Matches the web
+   * scatter's `iso.slice(0, 10)`, so an offset-bearing ISO (e.g. from `create_article`
+   * over MCP) covers the same day range in-app and in the federated/exported PNG.
+   */
+  periodStart: string
+  periodEnd: string
 }
 
 export interface FeedImageDeps {
@@ -91,8 +97,8 @@ export type ResolvedArticleBlock =
       trigger: CorrelationSelector
       outcome: CorrelationSelector
       lagDays?: number
-      start: Date
-      end: Date
+      periodStart: string
+      periodEnd: string
       updatedAt: Date
     }
 
@@ -128,6 +134,29 @@ export const createRenderCache = (maxEntries = 200) => {
     } finally {
       inFlight.delete(key)
     }
+  }
+}
+
+/**
+ * A bounded set of cache keys whose render produced NO image (a sparse block →
+ * 404). `createRenderCache` deliberately never caches a `null`, so a
+ * publicly-listed sparse correlation block would otherwise re-run the full
+ * `getContinuousCorrelation` (two selector resolutions over the whole window) on
+ * every unauthenticated hit, uncached and unthrottled. Remembering the negative
+ * under the same hourly-bucketed key bounds that to one render per hour — the key
+ * rolls over hourly, so a block that later gains data re-renders.
+ */
+export const createNegativeCache = (maxEntries = 500) => {
+  const seen = new Map<string, true>()
+  return {
+    add: (key: string) => {
+      if (seen.size >= maxEntries) {
+        const oldest = seen.keys().next().value
+        if (oldest !== undefined) seen.delete(oldest)
+      }
+      seen.set(key, true)
+    },
+    has: (key: string) => seen.has(key),
   }
 }
 
@@ -227,10 +256,12 @@ export const resolveArticleBlock = async (
     }
   }
   return {
-    end: endDate,
+    // Slice the day off the RAW ISO (author's wall-clock day), like the web
+    // scatter — not the UTC day a Date round-trip would give.
     lagDays: block.lag_days,
     outcome: block.outcome,
-    start: startDate,
+    periodEnd: end.slice(0, 10),
+    periodStart: start.slice(0, 10),
     trigger: block.trigger,
     type: 'correlation',
     updatedAt,
@@ -267,10 +298,10 @@ export const renderArticleBlockImage = async (
       : Buffer.from(deps.renderChartSvg(series, opts), 'utf8')
   }
   const data = await deps.getCorrelationScatter(user, {
-    end: block.end,
     lagDays: block.lagDays,
     outcome: block.outcome,
-    start: block.start,
+    periodEnd: block.periodEnd,
+    periodStart: block.periodStart,
     trigger: block.trigger,
   })
   if (data == null) return null
@@ -354,6 +385,9 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
   // comparatively expensive render.
   const blockCacheKey = (kind: string, username: string, postId: string, index: number, updatedAt: Date) =>
     `${kind}:${username}:${postId}:${index}:${updatedAt.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
+  // Remembers the "no image" outcome (a sparse block 404s) under the same hourly
+  // key, so a public sparse block doesn't re-run the render engine on every hit.
+  const negativeBlocks = createNegativeCache()
 
   router.get('/public/:username/feed/:postId/blocks/:index/image.png', async (req, res) => {
     const { index, postId, username } = req.params
@@ -361,10 +395,13 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
     const idx = Number(index)
     const block = await resolveArticleBlock(deps, username, postId, idx, token)
     if (!block) return notFound(res)
-    const png = await cached(blockCacheKey('blockpng', username, postId, idx, block.updatedAt), () =>
-      renderArticleBlockImage(deps, username, block, 'png'),
-    )
-    if (!png) return notFound(res)
+    const key = blockCacheKey('blockpng', username, postId, idx, block.updatedAt)
+    if (negativeBlocks.has(key)) return notFound(res)
+    const png = await cached(key, () => renderArticleBlockImage(deps, username, block, 'png'))
+    if (!png) {
+      negativeBlocks.add(key)
+      return notFound(res)
+    }
     sendPng(res, png)
   })
 
@@ -374,10 +411,13 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
     const idx = Number(index)
     const block = await resolveArticleBlock(deps, username, postId, idx, token)
     if (!block) return notFound(res)
-    const svg = await cached(blockCacheKey('blocksvg', username, postId, idx, block.updatedAt), () =>
-      renderArticleBlockImage(deps, username, block, 'svg'),
-    )
-    if (!svg) return notFound(res)
+    const key = blockCacheKey('blocksvg', username, postId, idx, block.updatedAt)
+    if (negativeBlocks.has(key)) return notFound(res)
+    const svg = await cached(key, () => renderArticleBlockImage(deps, username, block, 'svg'))
+    if (!svg) {
+      negativeBlocks.add(key)
+      return notFound(res)
+    }
     sendSvg(res, svg)
   })
 
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -4,6 +4,8 @@
  * Handles: GET /public/:username/feed/:postId/chart.png
  *          GET /public/:username/feed/:postId/chart.svg
  *          GET /public/:username/feed/:postId/route.png
+ *          GET /public/:username/feed/:postId/blocks/:index/image.png
+ *          GET /public/:username/feed/:postId/blocks/:index/image.svg
  *
  * The chart is offered both ways from the same series over the same window: a
  * rasterised PNG that every consumer understands (Mastodon attaches it) and a
@@ -160,18 +162,6 @@ export const resolveImageWindow = async (
 }
 
 /**
- * Resolve one article chart/correlation block to its render inputs, or `null`
- * when the request isn't eligible: invalid username / non-UUID id / bad index,
- * missing DB, missing post, a non-article post, a `followers`-only post without a
- * matching capability `token`, an out-of-range or prose block, or a block whose
- * effective window is unbounded / non-increasing. Pure of Express — unit-testable.
- *
- * Unlike a shared activity's chart, an article block has NO `include_chart`
- * opt-in flag: a chart/correlation block can embed any metric over any window, so
- * the post's visibility (public/unlisted open; followers-only via the unguessable
- * `token`) is the whole authorization boundary (#943).
- */
-/**
  * Load the article behind an image request and authorize it, or `null` when
  * ineligible (invalid username / non-UUID id, missing DB, missing post, a
  * non-article post, or a `followers`-only post without a matching capability
@@ -196,6 +186,18 @@ const loadArticleForImage = async (
   return { article: post.article, updatedAt: post.updated_at }
 }
 
+/**
+ * Resolve one article chart/correlation block to its render inputs, or `null`
+ * when the request isn't eligible: invalid username / non-UUID id / bad index,
+ * missing DB, missing post, a non-article post, a `followers`-only post without a
+ * matching capability `token`, an out-of-range or prose block, or a block whose
+ * effective window is unbounded / non-increasing. Pure of Express — unit-testable.
+ *
+ * Unlike a shared activity's chart, an article block has NO `include_chart`
+ * opt-in flag: a chart/correlation block can embed any metric over any window, so
+ * the post's visibility (public/unlisted open; followers-only via the unguessable
+ * `token`) is the whole authorization boundary (#943).
+ */
 export const resolveArticleBlock = async (
   deps: Pick<FeedImageDeps, 'getPost'>,
   username: string,
--- a/apps/backend/src/routes/feed-image-router.ts
+++ b/apps/backend/src/routes/feed-image-router.ts
@@ -348,15 +348,20 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
   // whole process lifetime — the block re-resolves its data live (#934), matching
   // the web inline render within ≤ 1h. Articles, unlike shared activities, are
   // mutable and their windows can gain data after publish.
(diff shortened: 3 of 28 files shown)


apps/backend/src/services/activitypub/federation.ts
```diff
@@ -390,6 +397,28 @@ export const createFeedFederation = (
     },
   )
 
+  // Individual article object, served at its own path (distinct from the Note
+  // dispatcher above) as an AS2 `Article`. Same public/unlisted-only gate: a
+  // `followers`-only article is delivered with its object inline, so its id never
+  // needs an unauthenticated fetch.
+  federation.setObjectDispatcher(
```
fiddur: A deleted public/unlisted article's id will answer **404, not 410 Gone**.

`createFeedTombstoneRouter` catches the dispatcher's `next()` fall-through only for `GET /users/:username/feed/:postId` — Express won't match that against `/users/:u/feed/:id/article`, so the tombstone row recorded on delete is never consulted for an article id. `buildTombstoneJsonLd` also hardcodes `formerType: 'Note'`.

Net effect: a remote server dereferencing a deleted article gets a transient-looking 404 forever instead of the permanent-deletion signal, which is exactly the distinction the tombstone router exists to make. The `Delete{Tombstone}` fan-out is correct, so this only bites servers that dereference rather than act on the delivered `Delete` — but `docs/features/feed.md:127-131` (changed in this PR) now documents the article id as covered by the 410 behaviour.

Suggest extending the tombstone router with a `/users/:username/feed/:postId/article` route that emits `formerType: 'Article'` at the `…/article` id.

docs/features/feed.md
```diff
@@ -33,10 +33,10 @@ A feed post has a **`kind`**:
   `POST /feed/articles` / `PATCH /feed/articles/:postId` (and the `create_article` /
   `update_article` MCP tools), or in the web app's article composer; prose renders
   through the shared markdown sanitiser and each windowed block resolves its data live
-  over the locked window.
-  _Federation as an AS2 `Article` + Reddit/markdown export is the remaining slice
-  (#937), so an article is authored and shown on the owner's own feed but does not yet
-  fan out to followers._
+  over the locked window. An article **federates as an AS2 `Article`** (title + prose
+  HTML + a rendered PNG per chart/correlation block) so followers on Mastodon see prose
+  + images; Aurboda peers get the richer inline render via structured enrichment.
```
fiddur: "Aurboda peers get the richer inline render via structured enrichment" reads as present-tense, but inbound `Create{Article}` is dropped today: `ingestFeedActivity` (federation.ts) does `if (!(object instanceof Note)) return`, and Fedify's `Article` is a sibling of `Note`, not a subclass. `timeline-backfill.ts` likewise only collects `Note`s from a followee's outbox.

So an Aurboda follower of an Aurboda author sees nothing at all for an article — strictly worse than a Mastodon follower, who gets prose + attachments. `article-object.ts`'s own docstring is accurate here ("a later slice"); this line should match it.

Per CLAUDE.md, either handle `Article` in `ingestFeedActivity`/backfill in this PR, or file the follow-up issue and reword this to "will get the richer inline render (#<issue>)" so the gap isn't silent.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,89 @@
...
+import { SCATTER_HEIGHT, SCATTER_WIDTH } from '../charts/scatter-svg.ts'
+import { isPubliclyVisible } from './object.ts'
+import { sanitizeRemoteHtml } from './timeline-ingest.ts'
+
+// GFM + hard line breaks, matching the web's `renderMarkdown` (#910) so an
+// article reads the same in-app and federated.
+marked.setOptions({ breaks: true, gfm: true })
+
+/** Render one run of authored markdown to sanitised, federation-safe HTML. */
+const renderProse = (markdown: string): string => sanitizeRemoteHtml(marked.parse(markdown, { async: false }))
```
fiddur: `sanitizeRemoteHtml` was tuned for *inbound* Mastodon-style content, and its allowlist has no `img`, `table`/`thead`/`tbody`/`tr`/`td`/`th`, `hr`, `h5`, or `h6`. With `disallowedTagsMode: 'discard'` the tags go but the text stays, so authored GFM that renders fine in-app (the web sink uses DOMPurify via `utils/markdown.ts`, which keeps all of these) federates degraded:

- a markdown table → one run of concatenated cell text with no structure
- `alt` → nothing at all (no text content to keep)
- `---` / `##### h5` → the rule vanishes, the heading flattens to a paragraph

Reusing the inbound sanitiser as the XSS boundary is the right call; the issue is only its tag set. Consider an outbound allowlist that extends the inbound one with the table tags, `hr`, `h5`/`h6` and `img` (`src`/`alt`, http(s) only) — Mastodon renders all of those in `content`. Worth a test on the table case either way.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -110,6 +159,117 @@ export const resolveImageWindow = async (
...
+ * missing DB, missing post, a non-article post, a `followers`-only post without a
+ * matching capability `token`, an out-of-range or prose block, or a block whose
+ * effective window is unbounded / non-increasing. Pure of Express — unit-testable.
+ *
+ * Unlike a shared activity's chart, an article block has NO `include_chart`
+ * opt-in flag: a chart/correlation block can embed any metric over any window, so
+ * the post's visibility (public/unlisted open; followers-only via the unguessable
+ * `token`) is the whole authorization boundary (#943).
+ */
+/**
```
fiddur, line 195: Two doc blocks stacked here: the one at 162-173 describes `resolveArticleBlock` (its eligibility rules and the #943 no-`include_chart` rationale) but now sits directly above `loadArticleForImage`, so it documents the wrong function and `resolveArticleBlock` at 199 has no docstring. Looks like the extract-helper refactor left the original comment behind — move the 162-173 block down to `resolveArticleBlock`.

fiddur (commented):
🛑 Changes required

Solid, well-factored slice: the new `resolveArticleBlock` / `renderArticleBlockImage` split keeps the image routes thin and unit-testable, `buildScatterSvg` is a genuine shared-source twin of the web scatter (nice that `linearRegression` / `describeSelectorAxis` / `metricLabels` / `defaultArticleChartBucket` all moved into `@aurboda/api-spec` rather than being duplicated), and the authorization story on the new public endpoints is right — `isCapabilityAuthorized` runs *before* the render cache is consulted, the cache key includes `updated_at` so an edit can't serve a stale image, and `no-store` keeps everything revocable. Test coverage of the pure pieces is good (window inheritance, prose/out-of-range/negative/NaN index, followers-token matrix, sanitiser boundary, Article/Create/Delete shapes, outbox integration). CI is green.

Three things I'd like resolved before merge, plus one nit:

1. **A deleted article's canonical id never serves `410 Gone`.** `feed-tombstone-router` only matches `/users/:username/feed/:postId`, so `…/feed/<id>/article` falls through to a plain 404 after deletion, and `buildTombstoneJsonLd` hardcodes `formerType: 'Note'`. The docs changed in this PR now put the article id inside the very bullet that promises 410-on-delete (`docs/features/feed.md:127-131`), so the documented behaviour and the code disagree. Inline comment on the dispatcher.

2. **An inbound `Create{Article}` is silently dropped.** `ingestFeedActivity` bails on `if (!(object instanceof Note)) return`, and `timeline-backfill.ts` likewise only collects `Note`s from a followee's outbox. Fedify's `Article` is not a `Note` subclass, so an Aurboda user following another Aurboda user sees *nothing* for a federated article — worse than a Mastodon follower, who at least gets prose + attachments. `docs/features/feed.md:38` states "Aurboda peers get the richer inline render via structured enrichment" in the present tense, which isn't true yet. Per CLAUDE.md ("file a GitHub issue tracking the gap rather than letting them silently diverge"), either ingest `Article` here or file the follow-up and reword the doc to match `article-object.ts`'s own "(a later slice)" caveat.

3. **Prose fidelity loss in the federated `content`.** `renderProse` reuses `sanitizeRemoteHtml`, whose allowlist has no `img`, `table`/`tr`/`td`, `hr`, `h5`, `h6`. The web sink (`utils/markdown.ts`) uses DOMPurify, which keeps all of those. A GFM table in an article's prose federates as a run of concatenated cell text and markdown images vanish entirely — see inline comment.

Minor / non-blocking:

- The `feed-image-router.ts` header comment (lines 4-6, outside the diff) still lists only the `chart.png` / `chart.svg` / `route.png` paths; the two new `blocks/:index/image.{png,svg}` routes should be added there.
- `PATCH /feed/:postId` (the generic one) can still change an article's `visibility`, and that path calls `deliver.updated`, which no-ops for an article (`post.activity_id` is null). So flipping a federated article public↔followers never federates an `Update`, and a public→followers flip leaves already-delivered attachment URLs 404ing for existing followers. Worth routing an article through `updatedArticle` there too.
- `buildScatterSvg` (unlike `buildChartSvg`) doesn't filter non-finite values; a single NaN in `series` would make `Math.min`/`Math.max` NaN and produce an SVG that `sharp` rejects. Not reachable from the current correlation engine, but a `Number.isFinite` filter would be cheap insurance.
- A block whose render yields `null` (chart < 2 points, correlation n < 3) isn't cached, so every unauthenticated hit re-runs the full `getContinuousCorrelation` over the author's window. Same shape as the existing `chart.png` path, so consistent — just noting it as the one unbounded-work surface here.

apps/backend/src/routes/feed-image-router.ts
fiddur: Fixed in c6d60e78: moved the eligibility/#943 docstring down onto `resolveArticleBlock` (it was orphaned above `loadArticleForImage` after the extract), and added the two `blocks/:index/image.{png,svg}` routes to the file header comment.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Fixed in c6d60e78: outbound prose now goes through a dedicated `sanitizeArticleHtml` allowlist that extends the inbound one with table tags, `img` (src/alt, http(s) only), `hr`, and h5/h6. Added a test asserting a GFM table + image survive. Still a hard XSS boundary (no script/style/iframe/handlers).

docs/features/feed.md
fiddur: Filed #968 for inbound Create{Article} ingestion + backfill (the inbound half of C3), and reworded docs/features/feed.md to stop claiming present-tense Aurboda-peer render — it now says the richer inline render is a later slice (#968) and that an Aurboda follower doesn't yet see a federated article.

apps/backend/src/services/activitypub/federation.ts
fiddur: Fixed in c6d60e78: feed-tombstone-router now has a `/users/:username/feed/:postId/article` route emitting `formerType: 'Article'` at the `…/article` id (buildTombstoneJsonLd is parameterised). Unit test + an integration test that a deleted public article's `…/article` id returns 410.

fiddur (commented):
✅ Approved

Solid slice. I traced the authorization, addressing and id paths end-to-end and found no security or correctness blockers.

**What I verified**

- **Capability gating** — `resolveArticleBlock` runs `isCapabilityAuthorized` (constant-time token compare) *before* the render cache is consulted, so an unshared / visibility-flipped article 404s immediately and never serves a cached buffer. The `Article` object dispatcher and the outbox both gate on `isPubliclyVisible` / `listPublicFeedPostsPage`, so a `followers`-only article's `?token=` never leaks into an unauthenticated fetch or the outbox. Good.
- **Web/server parity** — `getArticleChartSeries` mirrors `ArticleChartBlock` exactly (`metrics[metric]?.avg`, `< 2` points → nothing), and `getCorrelationScatter`'s `n < 3` cutoff plus the binary-trigger group-comparison headline match `ArticleCorrelationBlock`. The `toISOString().slice(0,10)` day truncation is equivalent to the web's `iso.slice(0,10)` because `iso8601DateTimeSchema` (`z.iso.datetime()`) only admits `Z` offsets — worth keeping in mind if that schema ever gains `offset: true`.
- **Ids and tombstones** — the `Article` object id is on its own `…/article` path so it can't collide with the `Note` dispatcher; `#create` / `#update-<epoch>` / `#delete` fragments follow the existing Note convention; the tombstone handler refactor keeps the `next()` fall-through semantics intact. Inbound `Create{Article}` is silently dropped by `ingestFeedActivity`'s `instanceof Note` guard, which matches what the docs now say about #968.
- **Article `Update` routing** — nicely caught that the *generic* `PATCH /:postId` and the MCP `update_feed_post` also need the article path (a visibility flip would otherwise no-op, since `updated` bails on `!post.activity_id`).

**Non-blocking notes**

1. **Test coverage.** `resolveArticleBlock` is thoroughly unit-tested, but `renderArticleBlockImage` is private and untested, and neither new route has a supertest case. That leaves the chart-vs-correlation renderer dispatch, the `series.length < 2` / `getCorrelationScatter → null` 404 paths, and the `updated_at`-in-the-cache-key behaviour unexercised. Given "backend should be well covered with tests", exporting `renderArticleBlockImage` (like `resolveArticleBlock`) and adding a couple of cases would close the gap cheaply.
2. **Unauthenticated cost.** An article block's window and `bucket` are entirely author-chosen and unbounded, and both new endpoints are unauthenticated — a correlation block over a multi-year window runs the full selector-resolution engine per cold cache key. Related: `bucket: "0s"` passes `articleChartBlockSchema`'s `^\d+[smhd]$` regex and reaches Postgres as `date_bin('0 seconds', …)`, which errors → 500 rather than 404 (Express 5 forwards it, so no crash risk; the same hole already exists on the authenticated metrics route). Also worth noting the outbox re-runs `marked` + `sanitize-html` over every article's prose on each page fetch, with no memoisation.
3. **Docs** are updated well — the endpoint tables, the "no `include_chart` flag" rationale, and the #968 hand-off all read accurately.

The `queryFrequentMeals` fixture de-rot (relative dates instead of hardcoded April 2026) is a good fix with a comment explaining why — that class of test rot is easy to reintroduce.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -173,5 +335,34 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
...
+  // token only (no `include_chart` flag — a block can embed any metric, so
+  // visibility is the whole boundary, #943). Cache key includes the post's
+  // `updated_at` so editing the article serves a fresh render (articles, unlike
+  // shared activities, are mutable).
+  router.get('/public/:username/feed/:postId/blocks/:index/image.png', async (req, res) => {
+    const { index, postId, username } = req.params
+    const token = typeof req.query.token === 'string' ? req.query.token : undefined
+    const block = await resolveArticleBlock(deps, username, postId, Number(index), token)
+    if (!block) return notFound(res)
+    const png = await cached(`blockpng:${username}:${postId}:${index}:${block.updatedAt.getTime()}`, () =>
```
fiddur: The cache key busts on `updated_at`, which covers article *edits* — but not new *data* landing in the window, and an article block is explicitly "lock the window, re-resolve the data live" (#934). `createRenderCache` has no TTL, so once a key is populated the buffer is frozen for the process lifetime (or until it falls out of the 200-entry LRU).

Concretely: publish an article with a chart block over the last 7 days, let something fetch `blocks/0/image.png`, then a delayed Garmin/Oura sync backfills yesterday inside that window. The web inline render (`ArticleChartBlock`, live `useQuery`) shows the new points; this endpoint keeps serving the old render indefinitely, so the federated/exported image and the in-app one diverge. The `createRenderCache` docstring's premise ("the source data (a past activity's series/GPS) is effectively immutable") holds for the activity chart/route but no longer for article blocks.

A coarse time bucket folded into the key (e.g. `:${Math.floor(Date.now() / 3_600_000)}`) or a TTL on the entry would bound the staleness without giving up the in-flight de-duplication.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,152 @@
...
+import { marked } from 'marked'
+import sanitizeHtml from 'sanitize-html'
+
+import { CHART_HEIGHT, CHART_WIDTH } from '../charts/chart-svg.ts'
+import { SCATTER_HEIGHT, SCATTER_WIDTH } from '../charts/scatter-svg.ts'
+import { isPubliclyVisible } from './object.ts'
+
+// GFM + hard line breaks, matching the web's `renderMarkdown` (#910) so an
+// article reads the same in-app and federated.
+marked.setOptions({ breaks: true, gfm: true })
```
fiddur: This mutates `marked`'s process-wide singleton at import time, which the repo standards steer away from ("avoid global variables, avoid module level variables, avoid singletons") — and it silently applies `breaks: true` to any other module that imports `marked` later.

`marked.parse` merges per-call options over the defaults, so you can drop the module-level side effect and make `renderProse` self-contained:

```ts
const renderProse = (markdown: string): string =>
  sanitizeArticleHtml(marked.parse(markdown, { async: false, breaks: true, gfm: true }))
```

apps/backend/src/routes/feed-image-router.ts
fiddur: Good catch — the immutable-source premise doesn't hold for article blocks. Fixed in cc42f26d: the block-image cache key now folds in a coarse hourly bucket (`Math.floor(Date.now()/3_600_000)`) alongside `updated_at`, so backfilled data inside a locked window can't be served stale for the process lifetime — it re-resolves live within ≤1h, matching the web inline render. In-flight de-dup is preserved.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Fixed in cc42f26d: dropped the module-level `marked.setOptions` and pass `{ async:false, breaks:true, gfm:true }` per `marked.parse` call in `renderProse` — no process-wide singleton mutation, no leak of `breaks:true` to other importers.

apps/backend/src/api/rest-routes.ts
```diff
@@ -173,6 +175,39 @@ export const mountRestRouters = ({
       // Merged-span window so the rendered chart/route cover what the user
       // shared, matching the Note's duration/metrics (#881).
       getActivity: resolveFeedActivity,
+      // An article chart block's bucketed metric series over its locked window
+      // (mirrors the web's live bucketed render).
+      getArticleChartSeries: async (user, metric, start, end, bucket) => {
+        const result = await queryMetricsBucketed(user, [metric], start, end, bucket)
```
fiddur: `queryMetricsBucketed` is called without `options.tz`, so it falls back to `'UTC'` (`services/queries/metrics.ts`: `const tz = options.tz ?? 'UTC'`). The web path sends the viewer's `tz` (`state/api/metrics.ts` → `tz: browserTz`), and `getTimeSeriesBucketed` genuinely bins on `time AT TIME ZONE $5`, so for a `1d` bucket the boundaries are local midnight in-app and UTC midnight here.

`defaultArticleChartBucket` returns `'1d'` for any window over two days, so this is the common case: for an author in `Europe/Stockholm` every point in the federated PNG averages 00:00–24:00 UTC instead of 00:00–24:00 local — different values, and metrics logged near midnight land on the neighbouring day. That contradicts the comment two lines up ("mirrors the web's live bucketed render").

The viewer's timezone isn't knowable server-side, but the *author's* is: `user_settings.tz` (IANA, already stored). Suggest resolving it once and passing `{ tz }`, which also makes the render deterministic across viewers rather than dependent on whoever happened to open the page.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -173,5 +340,40 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
...
+  // Article chart/correlation block images. Gated by post visibility + capability
+  // token only (no `include_chart` flag — a block can embed any metric, so
+  // visibility is the whole boundary, #943). Cache key includes the post's
+  // `updated_at` (busts on an edit) AND a coarse hourly bucket, so a locked
+  // window that later gains backfilled data can't serve a stale render for the
+  // whole process lifetime — the block re-resolves its data live (#934), matching
+  // the web inline render within ≤ 1h. Articles, unlike shared activities, are
+  // mutable and their windows can gain data after publish.
+  const blockCacheKey = (kind: string, username: string, postId: string, index: string, updatedAt: Date) =>
+    `${kind}:${username}:${postId}:${index}:${updatedAt.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
```
fiddur, line 389: The cache key interpolates the raw `:index` path segment, but authorization/resolution uses `Number(index)`. Those disagree on spelling: `/blocks/0/image.png`, `/blocks/00/…`, `/blocks/000/…`, `/blocks/0x0/…`, `/blocks/1e0/…` and `/blocks/%200/…` all resolve to the same block (`Number.isInteger` passes for every one) but produce distinct cache keys.

So the hourly bucket you added on the line below doesn't actually bound work for a public article: an unauthenticated caller can walk `0`, `00`, `000`, … and force a fresh `getCorrelationScatter` (a full continuous-correlation query) plus a `sharp` rasterisation on every request, while also evicting the shared 200-entry LRU that the activity chart/route images depend on.

Keying on the parsed value fixes both the churn and the duplicate entries, e.g. hoist `const blockIndex = Number(index)` in each handler and pass `blockIndex` instead of `index` (typing the param `number`).

fiddur (commented):
🛑 Changes required

Solid slice: the AS2 `Article` object, its own dispatcher path, the `Create`/`Update`/`Delete{Tombstone}` triad, the `formerType: Article` tombstone, and the gated block-image endpoints all hang together, and the shared-source moves (`metricLabels`, `describeSelectorAxis`, `linearRegression`, `defaultArticleChartBucket` into `@aurboda/api-spec`) genuinely remove duplication rather than adding an indirection. Unit + integration coverage is good, docs are updated, and the security-relevant bits check out: the outbound sanitiser is a real allowlist (no `style`/event handlers/`javascript:`), `followers`-only attachment URLs only carry `?token=` in the delivered object (the outbox and the object dispatcher both refuse non-public posts), inbound `Create` still hard-filters to `Note` so a peer's `Article` can't be half-ingested, and the backfiller does the same.

Two things I'd like changed before merge, both in the new server-side render path:

1. **Daily buckets are computed in UTC, not the author's timezone**, so a federated/exported article chart doesn't actually mirror the in-app render it claims to (`rest-routes.ts:181`). `1d` is the default bucket for every window over two days.
2. **The block-image cache key uses the raw `:index` path string**, so `Number`-equivalent spellings (`0`, `00`, `0x0`, `1e0`, …) each get their own cache entry — which defeats the hourly-bucket rate limit on an unauthenticated, comparatively expensive render (`feed-image-router.ts:352`).

Two smaller observations, no change needed unless you disagree:

- `renderArticleContentHtml` runs a block's `caption` through `renderProse` (markdown → HTML), while the web renders it as plain text in a `<figcaption>`. A caption containing markdown or raw HTML therefore reads differently in-app vs federated. Not a security issue (the sanitiser catches it), just a fidelity gap in the opposite direction from the prose fix.
- `articleImageAttachments` emits one `Image` per chart/correlation block unconditionally, but `renderArticleBlockImage` 404s when the block has < 2 points (or n < 3). A sparse block therefore ships a dead attachment URL to Mastodon. Probably unavoidable without a synchronous pre-render, but worth a note in `docs/features/feed.md` so it isn't debugged twice.

apps/backend/src/routes/feed-image-router.ts
fiddur: Fixed: the block-image cache key now uses Number(index) (normalised) rather than the raw path string, so 0/00/1e0/0x0 collapse to one entry and can't each dodge the hourly-bucket rate limit.

apps/backend/src/api/rest-routes.ts
fiddur: Fixed: getArticleChartSeries now passes the author's device_timezone (IANA) to queryMetricsBucketed, so a 1d bucket splits on the author's calendar days — matching the web render (which sends the browser tz). Falls back to UTC when unset.

fiddur (commented):
🛑 Changes required

Solid, well-documented slice. The AS2 `Article` object, its own object-dispatcher path, the `formerType: Article` tombstone, the outbox `Create{Article}` branch, and the visibility-only gate on block images all hang together, and the shared-code extraction (`metricLabels`/`getMetricDisplayName` → `common.ts`, `linearRegression`/`describeSelectorAxis` → `correlations.ts`, `defaultArticleChartBucket` → `feed.ts`) genuinely removes duplication rather than adding a second source of truth. Test coverage is good: new unit tests for `resolveArticleBlock`, `renderArticleBlockImage`, `buildScatterSvg`, `articleImageAttachments`, plus integration coverage for the outbox/object/tombstone paths.

I verified locally on this head: `pnpm install`, `pnpm --filter @aurboda/api-spec build`, `tsgo --noEmit` (backend) all clean; backend unit tests 2232 passed / 158 files, web 548 passed / 50 files. Integration tests were not run (no Docker in this environment).

Three things I'd like addressed, ranked:

1. **Negative renders are never cached** (`feed-image-router.ts`) — `createRenderCache` deliberately doesn't cache a `null` result, and `articleImageAttachments` attaches an `Image` for *every* chart/correlation block regardless of whether it has data (documented in `docs/features/feed.md` as "a sparse block ships an attachment URL that resolves to 404"). So every publicly-listed sparse correlation block is an unauthenticated URL that re-runs a full `getContinuousCorrelation` — two selector resolutions over the whole window — on each hit, with no cache and no rate limit. This is the one I'd call blocking; the fix is cheap (cache the "no image" outcome under the same hourly-bucketed key, e.g. a sentinel empty buffer or a parallel `Set` of negative keys).

2. **Attachment URLs aren't versioned by `updated_at`** (`article-object.ts`) — the render cache busts on edit, but the URL a follower's server already stored is byte-identical after an edit, so a remote media cache (Mastodon re-hosts attachments at receipt) keeps the pre-edit PNG. The `Update{Article}` path this PR adds therefore doesn't visibly update charts on the fediverse.

3. **Correlation-block day bounds are derived in UTC** (`rest-routes.ts`) — the web block slices the stored ISO string directly, with an explicit comment about not using a UTC day; the server re-derives via `new Date(...).toISOString().slice(0, 10)`. The web composer normalises to `Z`, so this can't bite for web-authored articles, but `create_article` over MCP accepts an offset-bearing ISO datetime, and then the in-app scatter and the federated PNG cover different day ranges.

Two smaller notes, no action needed if you disagree:

- `docs/features/feed.md` says block charts bucket in "the author's timezone (`settings.tz`)", and the code comment says it matches the web render. `device_timezone` is only ever written by the Android device sync (`sync-router.ts`), so for a web-only author it's unset and the server buckets in UTC while the in-app render uses the browser tz. The code comment is honest about the fallback; the doc reads as a stronger parity claim than holds.
- `GET /users/:username/feed/:postId/article` returns `410` with `formerType: Article` for *any* deleted post id, including one that only ever existed as a `Note` (and, symmetrically, the bare `…/feed/:postId` id now 410s as a `Note` for a deleted article that never resolved there). The inline comment acknowledges this; harmless in practice since only an article's `…/article` id is ever advertised.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -173,5 +340,46 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
...
+  const blockCacheKey = (kind: string, username: string, postId: string, index: number, updatedAt: Date) =>
+    `${kind}:${username}:${postId}:${index}:${updatedAt.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
+
+  router.get('/public/:username/feed/:postId/blocks/:index/image.png', async (req, res) => {
+    const { index, postId, username } = req.params
+    const token = typeof req.query.token === 'string' ? req.query.token : undefined
+    const idx = Number(index)
+    const block = await resolveArticleBlock(deps, username, postId, idx, token)
+    if (!block) return notFound(res)
+    const png = await cached(blockCacheKey('blockpng', username, postId, idx, block.updatedAt), () =>
```
fiddur: `createRenderCache` intentionally does not cache a `null` result ("`produce` returning `null` (no data) is NOT cached"), and `renderArticleBlockImage` returns `null` for a sparse block — a correlation with `n < 3` in particular, which is only known *after* `getCorrelationScatter` has resolved both selectors over the full window.

Combined with `articleImageAttachments`, which attaches an `Image` for every chart/correlation block unconditionally (and `docs/features/feed.md` explicitly says a sparse block ships a URL that 404s), this publishes unauthenticated URLs whose every hit re-runs a full `getContinuousCorrelation` with no cache and no rate limit. `inFlight` only collapses *concurrent* hits; sequential polling re-runs the query each time.

Suggest caching the negative outcome under the same hourly-bucketed key — e.g. keep a `Set<string>` of keys that produced `null` in the same closure and short-circuit on it, so a sparse block costs one query per hour like a successful render does.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,157 @@
...
+  article.blocks.forEach((block, index) => {
+    if (block.type !== 'chart' && block.type !== 'correlation') return
+    const [width, height] =
+      block.type === 'chart' ? [CHART_WIDTH, CHART_HEIGHT] : [SCATTER_WIDTH, SCATTER_HEIGHT]
+    images.push(
+      new Image({
+        height,
+        mediaType: 'image/png',
+        name: attachmentName(block),
+        url: new URL(`${base}/blocks/${index}/image.png${query}`),
```
fiddur, line 175: The attachment URL is stable across edits, so it doesn't carry the freshness that `blockCacheKey`'s `updated_at` component buys. A follower's server (Mastodon downloads and re-hosts attachments at receipt) compares attachments by `url`; an `Update{Article}` with an identical URL won't trigger a re-fetch, so an edited chart stays visually stale on the remote side even though this instance would now serve a new PNG.

Appending a version to the URL would close it, e.g. `?v=<updated_at epoch>` (merged with the existing `token` query when followers-only). The endpoint ignores unknown query params, so it costs nothing server-side.

apps/backend/src/api/rest-routes.ts
```diff
@@ -173,6 +181,45 @@ export const mountRestRouters = ({
...
+        }
+        return series
+      },
+      // An article correlation block's continuous scatter over its locked window;
+      // null when too sparse to be meaningful (n < 3), which 404s the image.
+      getCorrelationScatter: async (user, { end, lagDays, outcome, start, trigger }) => {
+        const c = await getContinuousCorrelation(user, {
+          lagDays,
+          outcome,
+          periodEnd: end.toISOString().slice(0, 10),
```
fiddur: This derives the day bound in UTC, which is the exact thing the web counterpart deliberately avoids — `ArticleCorrelationBlock.toDay` slices the stored ISO string directly, with the comment "so the bound is the author's wall-clock day — not a UTC day that could roll back for a window authored with a non-UTC offset near midnight."

The web composer normalises windows to `Z` (`toIso` → `new Date(local).toISOString()`), so this can't diverge for articles authored in the app. But `create_article` / `update_article` over MCP accept any ISO 8601 datetime, including an offset form: a block ending `2026-07-08T01:00:00+02:00` gives the web `2026-07-08` and this path `2026-07-07`, so the in-app scatter and the federated PNG are computed over different windows (different `n`, different coefficients).

Easiest fix is to keep the block's raw ISO strings alongside the `Date`s in `ResolvedArticleBlock` and slice those, so both sides use the identical expression.

fiddur: Fixed: correlation blocks now derive period_start/period_end from the RAW ISO window (iso.slice(0,10) — the author's wall-clock day), matching the web scatter, instead of new Date(...).toISOString().slice(0,10) (UTC). Added an offset-window test. Chart-block bucketing already passes the author device_timezone.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Fixed: articleImageAttachments now appends ?v=<updated_at epoch> to each block image URL. The render cache already busts on edit; this makes the URL itself change so a remote media cache (Mastodon re-hosts at receipt) re-fetches after an Update{Article}. The endpoint ignores v (reads only token).

apps/backend/src/routes/feed-image-router.ts
fiddur: Fixed: added a bounded createNegativeCache keyed by the same hourly-bucketed block key. A sparse block's 404 is remembered, so a public correlation block re-runs getContinuousCorrelation at most once per hour instead of on every unauthenticated hit.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,164 @@
+/**
+ * The AS2 representation of an article's prose and inline blocks for federation.
+ *
+ * An article federates as an AS2 `Article` (built in `deliver.ts`): the title is
+ * its `name`, the prose is HTML `content`, and each chart/correlation block is an
+ * attached PNG so a plain fediverse client (Mastodon) shows prose + images. Aurboda
```
fiddur: This premise doesn't hold for Mastodon specifically. Mastodon lists `Article` in `ActivityPub::Activity::Create::CONVERTED_TYPES` (alongside `Image`/`Audio`/`Video`/`Page`/`Event`), and for a converted type it ignores `content` and builds the status text from the object's `name` + `url` instead. So the carefully-sanitised prose HTML this module produces will be dropped on the receiving side; the follower sees the title, a link, and the attached PNGs.

That's the long-standing reason WordPress-ActivityPub, Plume and WriteFreely posts show up on Mastodon as title+link (and why the WP plugin added a "send as Note" option).

If prose fidelity on Mastodon is the goal, federating as a `Note` (prose in `content`, block PNGs attached) would actually achieve it, with `Article` kept for the object dispatcher / future Aurboda-peer ingestion. If `Article` is the deliberate choice, this doc and `docs/features/feed.md` should say what a Mastodon follower really gets.

apps/backend/src/services/activitypub/deliver.ts
```diff
@@ -251,3 +254,168 @@ export const deliverFeedDelete = async (
...
+      post.article,
+    ),
+    attribution: ctx.getActorUri(user),
+    ccs: cc,
+    content: renderArticleContentHtml(post.article),
+    id: objectId,
+    name: post.article.title,
+    published: dateToTemporalInstant(post.created_at),
+    tos: to,
+    url: objectId,
```
fiddur: `url` points at the AS2 object id. For a `Note` that's mostly harmless (Mastodon renders `content` and only uses `url` for "open original"), but for an `Article` Mastodon *substitutes* `name` + `url` for the body — so this URL becomes the main clickable thing a follower sees.

It isn't browsable: a `Accept: text/html` GET on `/users/:u/feed/:id/article` misses Fedify's content negotiation, `@fedify/express` calls `next()` from `onNotAcceptable`, the tombstone router `next()`s too, and Express returns 404. So the link is dead in a browser.

There's no per-post web route today (`apps/web/src/index.tsx` has `/u/:username` and `/u/:username/:slug` only), so the pragmatic options are pointing `url` at `buildProfileUrl(webHost, user)` or adding a per-post public page.

docs/features/feed.md
```diff
@@ -33,10 +33,12 @@ A feed post has a **`kind`**:
   `POST /feed/articles` / `PATCH /feed/articles/:postId` (and the `create_article` /
   `update_article` MCP tools), or in the web app's article composer; prose renders
   through the shared markdown sanitiser and each windowed block resolves its data live
-  over the locked window.
-  _Federation as an AS2 `Article` + Reddit/markdown export is the remaining slice
-  (#937), so an article is authored and shown on the owner's own feed but does not yet
-  fan out to followers._
+  over the locked window. An article **federates as an AS2 `Article`** (title, prose
+  HTML, and a rendered PNG per chart/correlation block) so followers on Mastodon see the
```
fiddur: "so followers on Mastodon see the prose and images" — Mastodon drops `content` for `Article` (a converted type) and renders `name` + `url` plus the attachments, so followers see the title, a link and the images but not the prose. Please reword to match whatever object type the code ends up sending; as written this sets the wrong expectation for anyone reading the doc later.

fiddur (commented):
🛑 Changes required

Solid, well-structured slice: the render path is cleanly split into pure, unit-testable pieces (`resolveArticleBlock` / `renderArticleBlockImage` / `buildScatterSvg`), the shared maths and labels genuinely move to `@aurboda/api-spec` instead of being duplicated, the capability-token gate is reused verbatim from the activity image endpoints, and the negative cache + hourly-bucketed key is a sensible answer to "unauthenticated, expensive, re-resolved live". Docs are updated in the same change. I verified locally: `pnpm check` is clean (0 errors) and the backend unit tests for the touched files (69) plus the whole web suite (548) pass. The federation integration test couldn't run here (no Docker in this sandbox) — CI will cover it.

**The one substantive problem is the federation surface's headline claim.** Mastodon treats AS2 `Article` as a *converted type* (`CONVERTED_TYPES = %w(Image Audio Video Article Page Event)` in `ActivityPub::Activity::Create`): for those types it discards the object's `content` entirely and synthesises the status body from `name` + `url`. So a Mastodon follower will see the article **title and a link**, plus the PNG attachments — not the prose. This is the same well-known behaviour that makes WordPress/Plume/WriteFreely articles render as title+link on Mastodon. Compounding it, `url` is set to the AS2 object id, and a browser request (`Accept: text/html`) to `/users/:u/feed/:id/article` falls through Fedify's `onNotAcceptable` → `next()` → the tombstone router's `next()` → a plain Express 404. So the only thing Mastodon *does* render for the body is a dead link.

Worth deciding explicitly before merge (all three are defensible, but the docs and comments currently assert an outcome that won't happen):

1. Keep `Article` and accept that Mastodon shows title + link + images — then fix `docs/features/feed.md` and the `article-object.ts` module doc, and give `url` a browsable target.
2. Federate as a `Note` (prose HTML in `content`, images attached) so the body actually renders, keeping `Article` for the object dispatcher / Aurboda peers.
3. Add a per-post public web page (there's no `/u/:username/feed/:postId` route today — only `/u/:username` and `/u/:username/:slug`) and point `url` at it.

Smaller notes, none blocking:

- The route handlers themselves (negative-cache wiring, `Number(index)` normalisation, 404 paths) have no supertest-level coverage — only the extracted pure functions do. A couple of end-to-end route tests would lock in the cache/404 behaviour the comments describe.
- Attachments are emitted unconditionally, so a sparse block ships an `Image` whose URL 404s. That's documented as intentional; on Mastodon it leaves a `MediaAttachment` whose download fails, which is a bit worse than "one fewer image" — fine as a tradeoff, just noting the actual remote-side effect.
- `escapeHtml` in `article-object.ts` duplicates `escapeXml` from `charts/chart-svg.ts` (same four replacements). Minor.

Everything else I probed checked out: the `?v=<updated_at>` cache-busting matches the render key, followers-only tokens never reach the outbox (`listPublicFeedPostsPage` is public+unlisted only), inbound `Create{Article}` is safely ignored by `ingestFeedActivity`'s `instanceof Note` guard, `/users/` stays inside `isFederationRequest` so the new dispatcher is reachable, the correlation `periodStart`/`periodEnd` day slicing matches `ArticleCorrelationBlock`'s `toDay()` exactly, the web really does send `browserTz` (so the `device_timezone` parity note in the docs is accurate), and `buildScatterSvg` degrades safely on an all-non-finite series.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Resolved by federating the article as a Note instead of an AS2 Article (your call in the round-5 question). Mastodon renders a Note's content, so the prose + attached charts now show inline — the headline claim is accurate. buildArticleNote replaces the Article builder; the article shares the standard Note object id (one dispatcher, one tombstone path); docs updated throughout. Commit 5688847a. Also deduped escapeHtml → reuse escapeXml from chart-svg.

apps/backend/src/services/activitypub/deliver.ts
fiddur: Resolved by federating the article as a Note instead of an AS2 Article (your call in the round-5 question). Mastodon renders a Note's content, so the prose + attached charts now show inline — the headline claim is accurate. buildArticleNote replaces the Article builder; the article shares the standard Note object id (one dispatcher, one tombstone path); docs updated throughout. Commit 5688847a.

docs/features/feed.md
fiddur: Resolved by federating the article as a Note instead of an AS2 Article (your call in the round-5 question). Mastodon renders a Note's content, so the prose + attached charts now show inline — the headline claim is accurate. buildArticleNote replaces the Article builder; the article shares the standard Note object id (one dispatcher, one tombstone path); docs updated throughout. Commit 5688847a.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,162 @@
...
+  sanitizeArticleHtml(marked.parse(markdown, { async: false, breaks: true, gfm: true }))
+
+/**
+ * The AS2 `content` HTML for an article: each prose block rendered from markdown,
+ * and each chart/correlation block's caption (when set) as a paragraph. A caption
+ * is emitted as plain escaped text — NOT markdown — to match the web, which shows
+ * it as text in a `<figcaption>`; the chart images themselves ride as attachments
+ * (`articleImageAttachments`), which a fediverse client lays out after the content.
+ */
+export const renderArticleContentHtml = (article: ArticleContent): string =>
```
fiddur: The article title never renders on Mastodon. `buildArticleNote` puts it in `name`, but Mastodon's `ActivityPub::Parser::StatusParser#processed_text` only uses `title`/`name` for *converted* object types (`Article`, `Page`, `Image`, `Audio`, `Video`, `Event`); for a `Note` it returns `text || ''` — i.e. `content` only — and `name` is dropped. So the very reason for switching away from AS2 `Article` (Mastodon rendering `name` + link instead of `content`) now costs the title in the other direction.

Worse for an article whose blocks are all charts/correlations without captions: this function returns `''`, so the delivered status has neither title nor body — only attachment URLs, some of which 404 when a block is sparse.

Suggest prepending the title here (e.g. `<p><strong>${escapeXml(article.title)}</strong></p>` or an `<h2>`, mirroring how Mastodon itself renders converted types) so `content` is self-contained; `name` can stay for Aurboda peers.

apps/backend/src/api.ts
```diff
@@ -364,6 +373,20 @@ const main = async () => {
         if (activity) await deliverFeedUpdate(feedDeps, user, post, activity)
       })().catch(onDeliverError('update', user, post.id))
     },
+    // Articles fan out as an AS2 Article (built purely from stored content — no
```
fiddur: Stale after the final `Article` → `Note` refactor — these fan out as a `Create{Note}` / `Update{Note}`, not an AS2 `Article`.

apps/backend/src/routes/feed-router.ts
```diff
@@ -160,11 +164,10 @@ export const createFeedRouter = (
...
   )
 
-  // Article posts (long-form: title + prose + inline chart blocks). Create/edit
-  // have their own body shape; delete reuses `DELETE /:postId`. Registered before
-  // the generic `/:postId` routes. Federation of articles is a later slice, so
-  // these do not (yet) call `deliver` — an article is authored and shown on the
-  // owner's feed but does not fan out to followers until then.
+  // Article posts (long-form: title + prose + inline chart/correlation blocks).
+  // Create/edit have their own body shape; delete reuses `DELETE /:postId`.
+  // Registered before the generic `/:postId` routes. Federated as an AS2 `Article`
```
fiddur: Same stale wording: articles federate as a `Note` (deliberately not an AS2 `Article`, per the module docs in `article-object.ts`), so this comment now contradicts the implementation.

fiddur (commented):
🛑 Changes required

Solid slice overall: the block-image endpoints are gated exactly like the existing activity images (visibility + capability token, no-store, index normalised into the cache key), the render/negative caches are bounded and busted by `updated_at` + hourly bucket, the shared `linearRegression` / `describeSelectorAxis` / `metricLabels` / `defaultArticleChartBucket` moves into `@aurboda/api-spec` genuinely de-duplicate web ↔ backend, `tz` is parameterised into the SQL (no injection via `device_timezone`), and the server chart/scatter paths match the web renders point-for-point (`avg` per bucket, `iso.slice(0,10)` wall-clock day bounds, same `n < 3` / `< 2 points` thresholds). Docs and tests are thorough.

I verified locally: `pnpm --filter @aurboda/api-spec build`, backend `tsgo --noEmit`, `pnpm --filter aurboda-web check`, `pnpm --filter @aurboda/api-spec check` all clean, and the four touched/new unit test files pass (60 tests).

One functional issue blocks: **the article title never reaches a Mastodon follower.** The switch to `Note` fixes `content` being discarded, but Mastodon's `StatusParser#processed_text` only falls back to `title` (`name`) for *converted* object types (`Article`/`Page`/`Video`/`Event`); for a `Note` it returns `text || ''`, i.e. `content` alone — `name` is ignored outright (see `app/lib/activitypub/parser/status_parser.rb`). So `buildArticleNote` sets `name: post.article.title` and it is silently dropped. For an article whose blocks are all charts with no captions this is worse: `renderArticleContentHtml` returns `''`, so the federated status has no title and no text at all — just attachment URLs (some of which may 404 for sparse blocks). Prepending the title into `content` (e.g. `<p><strong>…</strong></p>` or an `<h2>`, matching how Mastodon renders converted types) would fix both, and keeps `name` for Aurboda peers.

Two stale comments left over from the final `Article` → `Note` refactor still say articles federate as an AS2 `Article` (inline below). Also, `apps/backend/src/routes/feed-router.integration.test.ts:13` still carries "Federation is a later slice, so no `deliver` is wired here" — that file isn't in the diff, but the statement is no longer true and the article create/update routes now have an unexercised `deliver` call; a `deliver` spy there (or in `feed-router.test.ts`) would cover the `kind === 'article'` branch in `PATCH /:postId`, which is currently untested.

Non-blocking notes: PNG and SVG use distinct cache keys (`blockpng`/`blocksvg`), so the underlying `getContinuousCorrelation` runs twice per hour per correlation block; and the `bucket` regex `^\d+[smhd]$` still admits absurd values (`99999999999d`) that would blow up `date_bin` — only reachable via the author's own block, and equally true of the web render, so just noting it.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Fixed: renderArticleContentHtml now leads the content with the title as a bold heading (<p><strong>title</strong></p>), since Mastodon's StatusParser ignores a Note's name. An all-charts/no-caption article now federates the title rather than an empty status. name is still set on the Note for peers. Unit + integration tests assert the title-in-content.

apps/backend/src/routes/feed-router.ts
fiddur: Fixed the stale comment — articles federate as a Note. Also added a deliver-spy integration test covering the kind==='article' branch in the generic PATCH /:postId (routes through updatedArticle).

apps/backend/src/api.ts
fiddur: Fixed the stale comment — these fan out as Create{Note}/Update{Note}.

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Sanitise authored article prose for outbound federation. Extends the inbound
+ * fediverse allowlist (`sanitizeRemoteHtml`) with the block elements a QS write-up
+ * uses — GFM tables, images, `hr`, h5/h6 — all of which Mastodon renders in
+ * `content` and the web DOMPurify sink already keeps, so authored formatting isn't
+ * silently flattened on the way out. Still a hard XSS boundary: no script/style/
+ * iframe/event handlers survive, and link/image URLs are http(s)/mailto only.
+ */
+const sanitizeArticleHtml = (html: string): string =>
```
fiddur, line 45: The doc comment says this "extends the inbound fediverse allowlist (`sanitizeRemoteHtml`)", but it's actually a full copy of that config with additions — the 20-tag list, the `a`/`ol`/`span` attribute map, `allowedSchemes`, `disallowedTagsMode` and the `transformTags` link hardening are all duplicated from `timeline-ingest.ts`.

That's two independent XSS allowlists that are meant to stay related. A future hardening applied to one (say dropping `target="_blank"`, or adding a scheme restriction) silently won't reach the other, and nothing fails to catch the drift.

Suggest exporting the shared base from one place and literally extending it, e.g. a `BASE_ALLOWLIST` in `timeline-ingest.ts` (or a small shared module) that this one spreads and adds `img`/`hr`/`h5`/`h6`/table tags plus the `img`/`td`/`th` attributes to. Behaviour identical, but the "extends" in the comment becomes true in the code.

docs/features/feed.md
```diff
@@ -303,6 +309,36 @@ crisp, scalable `image/svg+xml` for **Aurboda-native** rendering (it stays sharp
...
+article (unlike a shared activity) is editable **and** its locked window can later gain
+backfilled data, the render cache keys on the post's `updated_at` **and** a coarse hourly
+bucket, so an edit or new data serves a fresh image within ≤ 1h; a `null` (no-data) render
+is remembered under the same key so a sparse public block can't re-run the render engine
+on every hit. The AS2 `Image` attachment URL also carries `?v=<updated_at>` so a remote
+media cache (which re-hosts the PNG at receipt) re-fetches after an edit.
+
+A block image endpoint **404s** when the block is too sparse to draw (a chart with < 2
+points, a correlation with n < 3) or has a zero-duration bucket. The article's `Note`
+attaches one `Image` per chart/correlation block **unconditionally** (the attachment list
```
fiddur, line 342: Worth also noting the receiving-side cap here: Mastodon stops at `Status::MEDIA_ATTACHMENTS_LIMIT` (4) when processing a remote `Create`'s `attachment` array, so an article with more than four chart/correlation blocks only shows the first four images on Mastodon — the rest are silently dropped, not just the sparse ones. That's a slightly different failure mode from the 404 case this paragraph describes, and it also softens the claim at the top of the file that "followers on Mastodon see the prose and images inline".

No code change needed (nothing sensible to do about it at this layer), but a reader picking block counts for an article would want to know.

fiddur (commented):
✅ Approved

Solid slice. I read the whole diff and verified it locally: `pnpm install`, `pnpm --filter @aurboda/api-spec build`, `pnpm --filter aurboda-backend check` (tsgo + oxlint, 0 errors), `pnpm --filter aurboda-web check` (0 errors), and the full backend unit suite (158 files / 2232 tests, all green, including the 61 new ones).

**Security surface** — the new `/public/:username/feed/:postId/blocks/:index/image.{png,svg}` endpoints are unauthenticated, so I traced the gate end to end and it holds:

- `resolveArticleBlock` validates username/UUID/index *before* any DB hit, rejects non-article posts, prose blocks, out-of-range/negative/non-integer indices, and unbounded or non-increasing windows.
- Authorization is `isCapabilityAuthorized` (constant-time `timingSafeEqual`), same boundary the existing chart/route images use. Dropping the `include_chart` flag for articles is correct and documented (#943) — a block can name any metric, so visibility *is* the boundary.
- The `image_token` only ever reaches a Note delivered to followers: the object dispatcher and outbox both filter to `public`/`unlisted` before `buildArticleNote`, and `articleImageAttachments` only sets `token` when `!isPubliclyVisible`. Verified there's no path that emits a token'd URL publicly.
- `sanitizeArticleHtml` replaces (not merges with) sanitize-html's defaults and adds no `'*'` attribute key, so no `style`/`id`/`on*` survives; `allowedSchemesAppliedToAttributes` defaults cover `href`/`src`, and `script`/`style` are in `nonTextTags` so their text content is dropped too.

**Correctness spot-checks that passed**: attachment `width`/`height` (1000×420 chart, 900×600 scatter) match what `svgToPng` actually rasterizes 1:1; `queryMetricsBucketed` sorts buckets, so the polyline can't zigzag; the web `ArticleChartBlock` also uses `metrics[m].avg` and a `< 2` points threshold, so in-app and federated renders agree; `buildScatterSvg` degrades safely on an all-non-finite series (no `NaN` in the output — `sx`/`sy` aren't reached); the outbox `setCounter` (`countPublicFeedPosts`) already counted articles, so including them in `items` actually *fixes* a `totalItems`/items divergence; `createArticlePost` gets `image_token` from the column default, so `toDeliverableArticle` can't produce an empty token.

**Non-blocking observations** (nothing I'd hold the PR for):

- `renderArticleContentHtml` runs marked + sanitize-html on every unauthenticated outbox page fetch and object dereference, with no caching. Worst case is 20 posts/page × 100 blocks × 20 000 chars of markdown. Almost certainly fine in practice for a self-hosted single-user instance, but if outbox crawls ever show up hot, memoising the rendered HTML on `(postId, updated_at)` is the cheap fix.
- The `meals.integration.test.ts` fixture de-rot (relative dates instead of hardcoded April 2026) is a good catch and the comment explains why — worth keeping in mind for other date-anchored fixtures.
- I could not run the integration tests (`federation.integration.test.ts`, `feed-router.integration.test.ts`) here, so the new article outbox/object/410 cases and the fan-out spy test are unverified locally — CI covers them.

docs/features/feed.md
fiddur: Added in 3b6b9737: docs now note Mastodon's MEDIA_ATTACHMENTS_LIMIT (4) — an article with >4 chart/correlation blocks shows only its first four images on Mastodon; Aurboda peers (native payload) aren't capped.

apps/backend/src/services/activitypub/article-object.ts
fiddur: Fixed in 3b6b9737: reworded the doc comment — sanitizeArticleHtml is a superset of the inbound allowlist maintained separately, not a literal shared/extended config (inbound is intentionally stricter).

apps/backend/src/services/activitypub/article-object.ts
```diff
@@ -0,0 +1,177 @@
...
+
+import { CHART_HEIGHT, CHART_WIDTH, escapeXml } from '../charts/chart-svg.ts'
+import { SCATTER_HEIGHT, SCATTER_WIDTH } from '../charts/scatter-svg.ts'
+import { isPubliclyVisible } from './object.ts'
+
+/**
+ * Sanitise authored article prose for outbound federation. A superset of the
+ * inbound fediverse allowlist (`sanitizeRemoteHtml`) — the same safe base, plus
+ * the block elements a QS write-up uses (GFM tables, images, `hr`, h5/h6), all of
+ * which Mastodon renders in `content` and the web DOMPurify sink already keeps, so
```
fiddur: The justification here (and in the module docstring at line 15) isn't accurate for Mastodon. Its inbound status sanitiser (`Sanitize::Config::MASTODON_STRICT`) allows roughly `p br span a del s pre blockquote code b strong u i em ul ol li ruby rt rp h1..h6` — `img`, `hr`, and the whole `table`/`thead`/`tbody`/`tr`/`td`/`th` family are **not** on that list. Since `Sanitize` strips a disallowed element but keeps its text, a GFM table federates to Mastodon as run-together cell text and an inline `img` vanishes entirely.

The widened outbound allowlist is still the right call (it matches the web DOMPurify sink, and non-Mastodon consumers plus the future Aurboda-peer path do render these) — it's the stated reason that's wrong. Worth rewording to something like "…which the web DOMPurify sink keeps, so authored formatting isn't flattened on the way out; Mastodon's own sanitiser will still drop tables/images/`hr`." Same caveat is probably worth a line in `docs/features/feed.md`, since a QS write-up with a results table is exactly the case that silently degrades there.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -34,17 +41,67 @@ export interface ImageActivity {
...
+
+/** The window-resolved inputs for one article correlation block's scatter. */
+export interface CorrelationBlockParams {
+  trigger: CorrelationSelector
+  outcome: CorrelationSelector
+  lagDays?: number
+  /**
+   * Inclusive day bounds (`YYYY-MM-DD`), sliced from the block's ISO window so they
+   * are the author's wall-clock days — NOT a UTC-normalised day. Matches the web
+   * scatter's `iso.slice(0, 10)`, so an offset-bearing ISO (e.g. from `create_article`
```
fiddur: The "offset-bearing ISO (e.g. from `create_article` over MCP)" case can't actually occur: block `start`/`end` are validated by `iso8601DateTimeSchema` = `z.iso.datetime()`, which in Zod 4 defaults to `offset: false` and so only accepts a `Z` suffix — I confirmed `'2026-07-01T23:30:00-05:00'` fails to parse. Both the REST bodies (`createArticleBodySchema` / `updateArticleBodySchema`) and the MCP tools go through `articleBlockSchema`, so every stored window is UTC.

The code is fine either way — `slice(0, 10)` on a `Z` string is the UTC day, which is exactly what the web scatter's `toDay()` produces, so parity holds. But the comment (and the `resolveArticleBlock` test built on the `-05:00` fixture, which bypasses validation by constructing the record directly) reads as if offsets are a supported input. Either drop the rationale to just "matches the web scatter's `iso.slice(0, 10)`", or loosen the schema to `{ offset: true }` if offset-bearing windows are actually wanted.

fiddur (commented):
✅ Approved

Solid, well-scoped slice. The article → `Note` federation, the gated per-block image endpoints, and the shared-single-source extractions (`metricLabels`, `describeSelectorAxis`, `linearRegression`, `defaultArticleChartBucket` into `@aurboda/api-spec`) all line up with the repo's conventions, and the reasoning left in the comments is genuinely useful.

**Verified locally** (after `pnpm install` + `pnpm --filter @aurboda/api-spec build`):
- `pnpm --filter aurboda-backend check`, `--filter aurboda-web check`, `--filter @aurboda/api-spec check` — 0 errors (warning counts unchanged from `develop`'s pre-existing baseline).
- Backend unit suite: 158 files / 2232 tests pass.
- `federation.integration.test.ts` + `feed-router.integration.test.ts`: 26 tests pass (article outbox `Create{Note}`, object dispatcher, 410 `Tombstone` after delete, and the `createdArticle` / `updatedArticle` fan-out branches all green).
- `oxfmt --check`: only `packages/api-spec/dist/**` (generated, untracked) is unformatted; every source file is clean.

**Things I checked and found correct:**
- Authorization: block images reuse `isCapabilityAuthorized`, so `followers`-only articles need the `image_token`, and the `include_chart` flag is deliberately not consulted — matching the stated #943 boundary. The token never leaks into the object dispatcher or outbox (both gate on `isPubliclyVisible` first).
- Web/server render parity: `avg` per bucket, `< 2` points → no image, `iso.slice(0, 10)` day bounds, `n < 3` sparse guard, `#673ab8` line colour, and the shared `defaultArticleChartBucket` all match `ArticleChartBlock` / `ArticleCorrelationBlock` exactly.
- XSS boundary: `escapeXml` on title/captions, `sanitize-html` with an explicit allowlist and `http|https|mailto` schemes, `marked` options passed per call rather than via the process-wide `setOptions` — good catch on that one.
- Cache keying: normalised numeric index + `updated_at` + hourly bucket, with the negative cache covering the sparse-block path. The `?v=<updated_at>` on the attachment URL correctly handles Mastodon re-hosting media at receipt.

**Non-blocking notes** (two are inline):
1. The sanitiser rationale overstates what Mastodon keeps — see the inline comment on `article-object.ts`.
2. The offset-bearing-ISO rationale in `feed-image-router.ts` describes an input the write boundary rejects — see the inline comment.
3. `isZeroDurationBucket` guards `0s`/`00m`, but the block schema also accepts an absurd bucket like `99999999999999999999d`, which `parseBucketSize` turns into an interval Postgres `date_bin` rejects — that would surface as a 500 on the new unauthenticated endpoint rather than a 404. Only the article's own author can create that, so it's self-inflicted and low priority, but an upper bound alongside the zero-duration guard would close the last hole in "a bad bucket means no image".
4. Nice unrelated fix bundling `queryFrequentMeals`' date-relative fixtures — that had genuinely rotted.

apps/backend/src/services/activitypub/article-object.ts
fiddur: You're right — reworded both the module docstring and the sanitizeArticleHtml comment: the wide allowlist matches the web DOMPurify sink and non-Mastodon / Aurboda-peer renderers, and Mastodon's OWN inbound sanitiser strips tables/images/hr (charts still arrive as attachments). Added the caveat to docs/features/feed.md too.

apps/backend/src/routes/feed-image-router.ts
fiddur: Confirmed — iso8601DateTimeSchema is Z-only, so the offset case can't occur. Reworded the comment (the day is iso.slice(0,10) = the UTC day; slicing the raw string keeps exact parity with the web toDay and stays correct if offsets are ever allowed), and retargeted the test off the -05:00 fixture (which bypassed validation) to a Z window.

apps/backend/src/routes/feed-image-router.ts
```diff
@@ -173,5 +373,55 @@ export const createFeedImageRouter = (deps: FeedImageDeps): Router => {
...
+  // window that later gains backfilled data can't serve a stale render for the
+  // whole process lifetime — the block re-resolves its data live (#934), matching
+  // the web inline render within ≤ 1h. Articles, unlike shared activities, are
+  // mutable and their windows can gain data after publish.
+  // Cache key uses the NORMALISED numeric index (not the raw path string), so
+  // `Number`-equivalent spellings (`0`, `00`, `1e0`, `0x0`) collapse to one entry
+  // and can't each dodge the hourly-bucket rate limit on this unauthenticated,
+  // comparatively expensive render.
+  const blockCacheKey = (kind: string, username: string, postId: string, index: number, updatedAt: Date) =>
+    `${kind}:${username}:${postId}:${index}:${updatedAt.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
```
fiddur, line 389: Worth a thought: folding the hourly bucket into the key means every article block mints a **new** entry in the shared `createRenderCache` every hour, while `chart:`/`route:` keys for activity posts never rotate. Since eviction there is insertion-order FIFO over 200 entries (not LRU), the rotating block keys will steadily push the long-lived activity entries out.

The sharper edge is the miss path: a correlation block's render runs `getContinuousCorrelation` (two full `resolveSelector` passes over the block window) on an unauthenticated, unthrottled endpoint. `negativeBlocks` bounds the *empty* case nicely, but a populated block is only bounded while its entry survives the FIFO — and `blocks` is capped at 100 per article, so two public articles of wide-window correlation blocks are enough to exceed 200 hot keys and turn every request into a full recompute.

Not blocking (an attacker has to author the articles first, and nginx fronts this), but a dedicated cache for block renders — or a plain TTL map keyed without the hour, so entries expire instead of accumulating — would decouple this from the activity-image cache and cap the worst case.

fiddur (commented):
✅ Approved

Solid, well-tested slice. I traced the security-relevant paths end to end and found no blocking issues.

**Verified correct**

- **Capability-token containment.** `articleImageAttachments` only appends `?token=` when `!isPubliclyVisible(visibility)`, and every path that can expose a `Note` to an unauthenticated fetcher refuses non-public posts first: the object dispatcher (`federation.ts` — `isPubliclyVisible` check precedes `toDeliverableArticle`) and the outbox (`listPublicFeedPostsPage` filters `visibility IN ('public','unlisted')` in SQL). So a followers-only article's token never reaches the public surface.
- **Block-image authorization.** `resolveArticleBlock` re-uses `isCapabilityAuthorized` (constant-time compare), validates username/UUID before any DB hit, rejects non-integer/negative indices, and rejects unbounded or non-increasing windows. The cache key omits the token, but that's safe because an unauthorized request 404s in `resolveArticleBlock` before the cache is consulted.
- **Outbound sanitiser.** `sanitizeArticleHtml` is a genuine allowlist: `script`/`style` are outside `allowedTags` and are covered by sanitize-html's default `nonTextTags`, so their text content is dropped too (asserted by the `alert(1)` test); no event handlers survive `allowedAttributes`; `href`/`src` are restricted to http/https/mailto, so `data:`/`javascript:` images can't ride out. Passing marked's options per call instead of `marked.setOptions` correctly avoids mutating process-wide state for other importers.
- **Web/server render parity.** I diffed the server render against `ArticleChartBlock`/`ArticleCorrelationBlock` line by line: same colour `#673ab8`, same `avg` aggregate, same `< 2 points` / `n < 3` thresholds, same `getMetricDisplayName` label, same `iso.slice(0, 10)` day bounds, same headline formatting and binary-trigger branch. Hoisting `linearRegression`/`describeSelectorAxis`/`metricLabels`/`defaultArticleChartBucket` into api-spec is the right way to keep that from drifting, and the re-export shims keep existing web imports working with no behaviour change.
- **Degenerate inputs.** `buildScatterSvg` with an all-filtered-out series is safe rather than NaN-poisoned: `circles` and `regLine` are both empty, and only constant axis coordinates remain, so `sharp` still gets valid markup.

**Non-blocking notes**

1. *Doc juxtaposition* — in `docs/features/feed.md`, the pre-existing sentence "An image is served when the matching flag was opted into." now sits directly after the new article-blocks section, which states articles have **no** opt-in flag. A reader will take the former as a global rule contradicting the latter. Scoping it to "An **activity** post's image is served when…" would remove the ambiguity.
2. *Unvalidated `device_timezone`* — `getArticleChartSeries` feeds `settings.device_timezone` into `getTimeSeriesBucketed`, where it is a bound parameter (`AT TIME ZONE $5`), so there's no injection. But it's stored verbatim from the Android sync payload with no IANA validation, so a bogus value would now surface as a 500 on an *unauthenticated* endpoint rather than only on authenticated ones. Pre-existing, and low risk given the client is first-party.

The integration tests covering the outbox `Create{Note}`, the served article object, and the 410 tombstone are exactly the right level for this, and the `queryFrequentMeals` fixture de-rot (relative dates + a comment explaining why) is a good fix for a real CI time bomb.

apps/backend/src/routes/feed-image-router.ts
fiddur: Good analysis. Non-blocking as you note (author-gated, single-user self-hosted instance, nginx-fronted, process-lifetime cache), and the empty case is already bounded by negativeBlocks. Filed #969 to track the proper fix — a dedicated TTL map keyed without the hour (entries expire rather than accumulate), decoupling block renders from the activity-image cache and capping the worst-case recompute at once-per-TTL-per-block. Keeping it out of this PR so it gets a focused change rather than bolting a second cache primitive onto this one.

PR #971, opened 2026-07-28
feat(feed): article structured enrichment for peers + Reddit/markdown export (#937 Slice C, part 2)

fiddur:
Completes **Slice C (#937)** of epic #934 (Shareable Articles) — the second and final PR, combining **C3** (native structured enrichment for Aurboda peers) and **C4** (Reddit/markdown export). Follows PR 1 (#967, merged), where articles federate as a **Note** (so Mastodon already renders prose + attached chart PNGs, and peers already ingest them as timeline Notes) — C3 layers the *richer* native render on top.

## C3 — native structured enrichment for Aurboda peers
- The public structured endpoint (`GET /public/:username/feed/:postId` → `resolveStructuredPost`) now handles **articles**. New `FeedStructuredArticle` variant in api-spec; `structured` (both the endpoint response and the timeline entry) becomes a **`kind`-discriminated union** (`FeedStructuredActivity | FeedStructuredArticle`), so activity payloads now carry `kind:'activity'`.
- Each block resolves **live over its locked window**, mirroring the C1 image path exactly: chart blocks bucket via `queryMetricsBucketed` (author `device_timezone`, shared `defaultArticleChartBucket` / `isZeroDurationBucket`, `samplesFromBucketedResult`); correlation blocks run `getContinuousCorrelation` (wall-clock day bounds). Same **visibility + capability-token gate** as the block images. Prose ships **raw markdown** so the receiving peer sanitises it itself (#910) — never trusting remote HTML.
- Web: `TimelineArticle` renders a peer's article inline (prose + native chart + scatter); `CorrelationScatterSvg` is exported with a shared structural type so it draws either a live fetch or a structured payload with no adapter. **Legacy timeline entries** (stored `structured` without a `kind`) still render as activity.

## C4 — Reddit/markdown export
- `buildArticleMarkdown` (pure) → title + prose + one image link per chart/correlation block, via the shared `articleBlockImageUrl` (the same URL the AS2 attachment points at). **REST** `GET /feed/articles/:postId/export` + **MCP** `export_article_markdown` (parity), plus an `ArticleExportDialog` with copy-to-clipboard.

## DRY
`isZeroDurationBucket` + `articleBlockImageUrl` moved to `services/article.ts`; `samplesFromBucketedResult` extracted from `resolvePublicSeries`.

## Testing
`pnpm check` clean (0 errors, api-spec/backend/web); unit tests (article, article-export, timeline-enrich, timeline-structured) + integration tests (feed-structured, feed-router) all green locally.

_Provenance: the bulk was produced by a scheduled agent that couldn't push from its sandbox; applied, reviewed, and fixed here (the legacy-entry `structuredChartSeries` guard) before pushing._

### DIFF
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -16,6 +16,7 @@
 import {
   type FeedPostsResponse,
   type FeedPostStructuredResponse,
+  type FeedStructuredPost,
   type PublicSeriesQuery,
   publicSeriesQuerySchema,
   type PublicSeriesResponse,
@@ -29,6 +30,7 @@ import { serializeFeedPost } from '../services/feed.ts'
 import { queryMetricsBucketed } from '../services/queries/index.ts'
 import { type TypedRouter, typedRouter } from '../typed-router.ts'
 import { validateQuery } from '../validation.ts'
+import { createRenderCache } from './feed-image-router.ts'
 
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
 
@@ -37,6 +39,14 @@ const PROFILE_FEED_LIMIT = 50
 
 export const createFeedPublicRouter = (): TypedRouter => {
   const router = typedRouter()
+  // Caches the serialised structured payload (JSON bytes) for the unauthenticated
+  // `/feed/:postId` endpoint, which resolves up to 100 blocks per request — the
+  // same bounded LRU + in-flight de-dup the block images use. Keyed on the
+  // capability token (so a public request never sees a followers-only payload, and
+  // a wrong token 404s and isn't cached) and a coarse hourly bucket (staleness on
+  // an edit is bounded to ≤1h — acceptable for a peer-fetched enrichment payload;
+  // an `updated_at`-keyed version + a per-block sample cap is #972).
+  const structuredCache = createRenderCache<FeedStructuredPost>()
 
   router.get<{ username: string }, PublicSeriesResponse, unknown, PublicSeriesQuery>(
     '/public/:username/series',
@@ -115,7 +125,8 @@ export const createFeedPublicRouter = (): TypedRouter => {
       }
       const token = typeof req.query.token === 'string' ? req.query.token : undefined
       try {
-        const structured = await resolveStructuredPost(username, postId, token)
+        const key = `structured:${username}:${postId}:${token ?? ''}:${Math.floor(Date.now() / 3_600_000)}`
+        const structured = await structuredCache(key, () => resolveStructuredPost(username, postId, token))
         if (!structured) {
           return res.status(404).json({ error: 'Not found', success: false })
         }
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -25,7 +25,7 @@ import {
 import { isValidUsername } from '../api/auth-routes.ts'
 import { findCoveringSharedSeriesWindow, isMissingDatabase, listPublicFeedPostsPage } from '../db/index.ts'
 import { resolvePublicSeries } from '../services/feed-series.ts'
-import { resolveStructuredPost } from '../services/feed-structured.ts'
+import { loadAuthorizedStructuredPost, resolveStructuredContent } from '../services/feed-structured.ts'
 import { serializeFeedPost } from '../services/feed.ts'
 import { queryMetricsBucketed } from '../services/queries/index.ts'
 import { type TypedRouter, typedRouter } from '../typed-router.ts'
@@ -39,14 +39,16 @@ const PROFILE_FEED_LIMIT = 50
 
 export const createFeedPublicRouter = (): TypedRouter => {
   const router = typedRouter()
-  // Caches the serialised structured payload (JSON bytes) for the unauthenticated
+  // Caches the resolved structured payload OBJECT for the unauthenticated
   // `/feed/:postId` endpoint, which resolves up to 100 blocks per request — the
-  // same bounded LRU + in-flight de-dup the block images use. Keyed on the
-  // capability token (so a public request never sees a followers-only payload, and
-  // a wrong token 404s and isn't cached) and a coarse hourly bucket (staleness on
-  // an edit is bounded to ≤1h — acceptable for a peer-fetched enrichment payload;
-  // an `updated_at`-keyed version + a per-block sample cap is #972).
-  const structuredCache = createRenderCache<FeedStructuredPost>()
+  // same bounded LRU + in-flight de-dup the block images use. The route authorizes
+  // the post BEFORE consulting this cache (like the sibling image routes), so the
+  // capability token is NOT part of the key: a `followers`-only payload never
+  // reaches the cache via a public request, and an anonymous `?token=…` walk can't
+  // inflate the key space. Keyed on `updated_at` (busts on an edit, no ≤1h edit
+  // staleness) — a smaller cap than the image LRU because a payload can be large
+  // (a per-block sample cap is #972). Producing `null` (no content) isn't cached.
+  const structuredCache = createRenderCache<FeedStructuredPost>(50)
 
   router.get<{ username: string }, PublicSeriesResponse, unknown, PublicSeriesQuery>(
     '/public/:username/series',
@@ -125,8 +127,15 @@ export const createFeedPublicRouter = (): TypedRouter => {
       }
       const token = typeof req.query.token === 'string' ? req.query.token : undefined
       try {
-        const key = `structured:${username}:${postId}:${token ?? ''}:${Math.floor(Date.now() / 3_600_000)}`
-        const structured = await structuredCache(key, () => resolveStructuredPost(username, postId, token))
+        // Authorize BEFORE the cache: an unauthorized post 404s and never reaches
+        // it, so the token stays out of the key and a revoked/flipped post stops
+        // resolving immediately. Cache keyed on `updated_at` so an edit busts it.
+        const post = await loadAuthorizedStructuredPost(username, postId, token)
+        if (!post) {
+          return res.status(404).json({ error: 'Not found', success: false })
+        }
+        const key = `structured:${username}:${postId}:${post.updated_at.getTime()}`
+        const structured = await structuredCache(key, () => resolveStructuredContent(username, post))
         if (!structured) {
           return res.status(404).json({ error: 'Not found', success: false })
         }
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -45,9 +45,14 @@ export const createFeedPublicRouter = (): TypedRouter => {
   // the post BEFORE consulting this cache (like the sibling image routes), so the
   // capability token is NOT part of the key: a `followers`-only payload never
   // reaches the cache via a public request, and an anonymous `?token=…` walk can't
-  // inflate the key space. Keyed on `updated_at` (busts on an edit, no ≤1h edit
-  // staleness) — a smaller cap than the image LRU because a payload can be large
-  // (a per-block sample cap is #972). Producing `null` (no content) isn't cached.
+  // inflate the key space. Keyed on `updated_at` AND a coarse hourly bucket — the
+  // SAME two-part key the block-image cache uses for the same live-resolved article
+  // data: `updated_at` busts on an edit, and the hourly term bounds staleness from
+  // *backfilled measurements inside an already-locked window* (which don't touch
+  // `updated_at`) to ≤1h, so this endpoint can't freeze into a lifetime snapshot
+  // while its own PNG stays live (#934 locks the window, not the data). A smaller
+  // cap than the image LRU because a payload can be large (per-block sample cap is
+  // #972). Producing `null` (no content) isn't cached.
   const structuredCache = createRenderCache<FeedStructuredPost>(50)
 
   router.get<{ username: string }, PublicSeriesResponse, unknown, PublicSeriesQuery>(
@@ -134,7 +139,7 @@ export const createFeedPublicRouter = (): TypedRouter => {
         if (!post) {
           return res.status(404).json({ error: 'Not found', success: false })
         }
-        const key = `structured:${username}:${postId}:${post.updated_at.getTime()}`
+        const key = `structured:${username}:${postId}:${post.updated_at.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
         const structured = await structuredCache(key, () => resolveStructuredContent(username, post))
         if (!structured) {
           return res.status(404).json({ error: 'Not found', success: false })
--- a/apps/backend/src/services/article-export.ts
+++ b/apps/backend/src/services/article-export.ts
@@ -15,8 +15,17 @@ import type { ArticleContent, FeedVisibility } from '@aurboda/api-spec'
 
 import { articleBlockImageUrl, articleBlockLabel } from './article.ts'
 
-/** Collapse newlines (markdown alt/caption text is a single line) and escape `]` so it can't close the image syntax early. */
-const inlineText = (text: string): string => text.replaceAll(/\s+/g, ' ').trim().replaceAll(']', '\\]')
+/**
+ * Collapse newlines (markdown alt/caption text is a single line) and escape both
+ * `[` and `]` so an unbalanced bracket can't break the `![alt](url)` syntax — a
+ * lone `[` (e.g. a tag selector's regex `pattern` surfaced by `describeSelectorAxis`,
+ * or a caption) would otherwise let CommonMark match the `]` against the inner `[`.
+ */
+const inlineText = (text: string): string =>
+  text
+    .replaceAll(/\s+/g, ' ')
+    .trim()
+    .replaceAll(/([[\]])/g, '\\$1')
 
 /**
  * Render an article as paste-ready markdown: an H1 title, each prose block's
--- a/apps/backend/src/services/article-export.ts
+++ b/apps/backend/src/services/article-export.ts
@@ -43,7 +43,7 @@ export const buildArticleMarkdown = (
   updatedAt: Date,
   article: ArticleContent,
 ): string => {
-  const sections = [`# ${article.title}`]
+  const sections = [`# ${inlineText(article.title)}`]
   article.blocks.forEach((block, index) => {
     if (block.type === 'prose') {
       sections.push(block.markdown)
--- a/apps/backend/src/services/article-export.ts
+++ b/apps/backend/src/services/article-export.ts
@@ -11,21 +11,12 @@
  * on-demand image endpoint rather than embedding rendered data itself, so no
  * chart/correlation resolution happens here.
  */
-import type { ArticleBlock, ArticleContent, FeedVisibility } from '@aurboda/api-spec'
+import type { ArticleContent, FeedVisibility } from '@aurboda/api-spec'
 
-import { describeSelectorAxis, getMetricDisplayName } from '@aurboda/api-spec'
-
-import { articleBlockImageUrl } from './article.ts'
+import { articleBlockImageUrl, articleBlockLabel } from './article.ts'
 
 /** Collapse newlines (markdown alt/caption text is a single line) and escape `]` so it can't close the image syntax early. */
 const inlineText = (text: string): string => text.replaceAll(/\s+/g, ' ').trim().replaceAll(']', '\\]')
-
-/** A human label for a chart/correlation block's image: its caption, else a label from its data. */
-const blockLabel = (block: Extract<ArticleBlock, { type: 'chart' | 'correlation' }>): string =>
-  block.caption ??
-  (block.type === 'chart'
-    ? getMetricDisplayName(block.metric)
-    : `${describeSelectorAxis(block.trigger)} vs ${describeSelectorAxis(block.outcome)}`)
 
 /**
  * Render an article as paste-ready markdown: an H1 title, each prose block's
@@ -49,7 +40,7 @@ export const buildArticleMarkdown = (
       sections.push(block.markdown)
       return
     }
-    const label = inlineText(blockLabel(block))
+    const label = inlineText(articleBlockLabel(block))
     const url = articleBlockImageUrl(apiBaseUrl, user, postId, visibility, imageToken, updatedAt, index)
     const lines = [`![${label}](${url})`]
     if (block.caption) lines.push(`*${inlineText(block.caption)}*`)
--- a/apps/backend/src/services/article-export.ts
+++ b/apps/backend/src/services/article-export.ts
@@ -1,0 +1,59 @@
+/**
+ * Reddit/markdown export (C4): a paste-ready markdown rendering of an article —
+ * title, prose, and one image link per chart/correlation block (via the C1
+ * `blocks/:index/image.png` endpoint) — for pasting into a text-only
+ * destination like r/QuantifiedSelf, where the author adds their own write-up
+ * around the linked charts. Reuses the exact same block-image URL the AS2
+ * attachment builder points at (`articleBlockImageUrl`), so the export and the
+ * federated image always agree.
+ *
+ * Pure and synchronous: the markdown only LINKS to the already-gated,
+ * on-demand image endpoint rather than embedding rendered data itself, so no
+ * chart/correlation resolution happens here.
+ */
+import type { ArticleBlock, ArticleContent, FeedVisibility } from '@aurboda/api-spec'
+
+import { describeSelectorAxis, getMetricDisplayName } from '@aurboda/api-spec'
+
+import { articleBlockImageUrl } from './article.ts'
+
+/** Collapse newlines (markdown alt/caption text is a single line) and escape `]` so it can't close the image syntax early. */
+const inlineText = (text: string): string => text.replaceAll(/\s+/g, ' ').trim().replaceAll(']', '\\]')
+
+/** A human label for a chart/correlation block's image: its caption, else a label from its data. */
+const blockLabel = (block: Extract<ArticleBlock, { type: 'chart' | 'correlation' }>): string =>
+  block.caption ??
+  (block.type === 'chart'
+    ? getMetricDisplayName(block.metric)
+    : `${describeSelectorAxis(block.trigger)} vs ${describeSelectorAxis(block.outcome)}`)
+
+/**
+ * Render an article as paste-ready markdown: an H1 title, each prose block's
+ * markdown verbatim, and each chart/correlation block as a markdown image link
+ * to its rendered PNG (the C1 endpoint), with its caption (else a data label)
+ * as both the image's alt text and an italic line underneath it — the markdown
+ * equivalent of the web/federated render's `<figcaption>`.
+ */
+export const buildArticleMarkdown = (
+  apiBaseUrl: string,
+  user: string,
+  postId: string,
+  visibility: FeedVisibility,
+  imageToken: string,
+  updatedAt: Date,
+  article: ArticleContent,
+): string => {
+  const sections = [`# ${article.title}`]
+  article.blocks.forEach((block, index) => {
+    if (block.type === 'prose') {
+      sections.push(block.markdown)
+      return
+    }
+    const label = inlineText(blockLabel(block))
+    const url = articleBlockImageUrl(apiBaseUrl, user, postId, visibility, imageToken, updatedAt, index)
+    const lines = [`![${label}](${url})`]
+    if (block.caption) lines.push(`*${inlineText(block.caption)}*`)
+    sections.push(lines.join('\n'))
+  })
+  return sections.join('\n\n')
+}
--- a/apps/backend/src/services/article.ts
+++ b/apps/backend/src/services/article.ts
@@ -11,6 +11,8 @@
  */
 import type { ArticleBlock, ArticleContent, FeedVisibility, UpdateArticleBody } from '@aurboda/api-spec'
 
+import { isPubliclyVisible } from './activitypub/object.ts'
+
 export type BuildArticleResult = { ok: true; article: ArticleContent } | { ok: false; error: string }
 
 /**
@@ -41,7 +43,8 @@ export const articleBlockImageUrl = (
 ): string => {
   const base = `${apiBaseUrl.replace(/\/+$/, '')}/public/${encodeURIComponent(user)}/feed/${postId}`
   const params = new URLSearchParams({ v: String(updatedAt.getTime()) })
-  if (visibility === 'followers') params.set('token', imageToken)
+  // Allowlist (public/unlisted are open); any other visibility needs the token.
+  if (!isPubliclyVisible(visibility)) params.set('token', imageToken)
   return `${base}/blocks/${index}/image.${format}?${params.toString()}`
 }
 
--- a/apps/backend/src/services/article.ts
+++ b/apps/backend/src/services/article.ts
@@ -9,9 +9,41 @@
  * article is rejected at the write boundary rather than failing later at render
  * time. Pure and synchronous — unit-testable without a DB.
  */
-import type { ArticleBlock, ArticleContent, UpdateArticleBody } from '@aurboda/api-spec'
+import type { ArticleBlock, ArticleContent, FeedVisibility, UpdateArticleBody } from '@aurboda/api-spec'
 
 export type BuildArticleResult = { ok: true; article: ArticleContent } | { ok: false; error: string }
+
+/**
+ * A zero-duration bucket (`0s`, `00m`, …) passes the block schema regex but is
+ * invalid for `date_bin` (`date_bin('0 seconds', …)` errors) — treat it as no
+ * data. Shared by the block-image renderer and the structured-enrichment
+ * resolver, both of which bucket a chart block over its locked window.
+ */
+export const isZeroDurationBucket = (bucket: string): boolean => /^0+[smhd]$/.test(bucket)
+
+/**
+ * The public URL of one article chart/correlation block's rendered image (the
+ * C1 endpoint), including the cache-busting `?v=<updated_at>` and, for a
+ * `followers`-only post, the capability `?token=`. Shared by the outbound AS2
+ * attachment builder (`articleImageAttachments`) and the markdown export
+ * (`article-export.ts`), so the two surfaces can never point at different URLs
+ * for the same block.
+ */
+export const articleBlockImageUrl = (
+  apiBaseUrl: string,
+  user: string,
+  postId: string,
+  visibility: FeedVisibility,
+  imageToken: string,
+  updatedAt: Date,
+  index: number,
+  format: 'png' | 'svg' = 'png',
+): string => {
+  const base = `${apiBaseUrl.replace(/\/+$/, '')}/public/${encodeURIComponent(user)}/feed/${postId}`
+  const params = new URLSearchParams({ v: String(updatedAt.getTime()) })
+  if (visibility === 'followers') params.set('token', imageToken)
+  return `${base}/blocks/${index}/image.${format}?${params.toString()}`
+}
 
 /** A windowed block's effective window: its own override, else the article default. */
 export const blockWindow = (
--- a/apps/backend/src/services/article.ts
+++ b/apps/backend/src/services/article.ts
@@ -11,9 +11,27 @@
  */
 import type { ArticleBlock, ArticleContent, FeedVisibility, UpdateArticleBody } from '@aurboda/api-spec'
 
+import { describeSelectorAxis, getMetricDisplayName } from '@aurboda/api-spec'
+
 import { isPubliclyVisible } from './activitypub/object.ts'
 
 export type BuildArticleResult = { ok: true; article: ArticleContent } | { ok: false; error: string }
+
+/**
+ * A human label for a chart/correlation block: its caption, else a label from its
+ * data (the metric display name, or `trigger vs outcome`). Shared single source
+ * for the AS2 image attachment name (`articleImageAttachments`) and the markdown
+ * export (`article-export.ts`) so the two can't diverge. A blank caption falls
+ * back to the data label.
+ */
+export const articleBlockLabel = (
+  block: Extract<ArticleBlock, { type: 'chart' | 'correlation' }>,
+): string => {
+  if (block.caption) return block.caption
+  return block.type === 'chart'
+    ? getMetricDisplayName(block.metric)
+    : `${describeSelectorAxis(block.trigger)} vs ${describeSelectorAxis(block.outcome)}`
+}
 
 /**
  * A zero-duration bucket (`0s`, `00m`, …) passes the block schema regex but is
--- a/apps/backend/src/services/feed-structured.integration.test.ts
+++ b/apps/backend/src/services/feed-structured.integration.test.ts
@@ -3,11 +3,13 @@ import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest'
 /**
  * Integration test for the structured-post resolver served at
  * `GET /public/:user/feed/:postId` — the payload an Aurboda peer fetches to
- * render a native chart. Exercises the real scalar + series resolution and the
- * public/unlisted gate against a live database.
+ * render a native chart or article. Exercises the real scalar + series
+ * resolution (activity posts) and the real bucketed-chart + continuous-
+ * correlation resolution (article posts), plus the shared public/unlisted gate,
+ * against a live database.
  */
 import { insertActivity } from '../db/activities/index.ts'
-import { createFeedPost } from '../db/feed.ts'
+import { createArticlePost, createFeedPost } from '../db/feed.ts'
 import { insertTimeSeries } from '../db/time-series.ts'
 import { cleanTestDb, getTestUser, startTestDb, stopTestDb } from '../test/db-test-helper.ts'
 import { resolveStructuredPost } from './feed-structured.ts'
@@ -65,18 +67,19 @@ describe('resolveStructuredPost', () => {
 
     const structured = await resolveStructuredPost(user, post.id)
     expect(structured).not.toBeNull()
-    expect(structured?.activity_type).toBe('exercise')
-    expect(structured?.start_time).toBe(START.toISOString())
-    expect(structured?.end_time).toBe(END.toISOString())
-    expect(structured?.duration_seconds).toBe(2400)
+    if (structured?.kind !== 'activity') throw new Error('expected an activity payload')
+    expect(structured.activity_type).toBe('exercise')
+    expect(structured.start_time).toBe(START.toISOString())
+    expect(structured.end_time).toBe(END.toISOString())
+    expect(structured.duration_seconds).toBe(2400)
 
-    const byKey = Object.fromEntries((structured?.metrics ?? []).map((m) => [m.key, m.value]))
+    const byKey = Object.fromEntries(structured.metrics.map((m) => [m.key, m.value]))
     expect(byKey.heart_rate_avg).toBe(150)
     expect(byKey.duration).toBe(2400)
 
     // The opted-in series is inlined with non-empty samples at the floored bucket.
-    expect(structured?.series).toHaveLength(1)
-    const hr = structured?.series[0]
+    expect(structured.series).toHaveLength(1)
+    const hr = structured.series[0]
     expect(hr?.metric).toBe('heart_rate')
     expect(hr?.bucket).toBe('5s')
     expect(hr?.samples.length).toBeGreaterThan(0)
@@ -95,8 +98,9 @@ describe('resolveStructuredPost', () => {
       visibility: 'public',
     })
     const structured = await resolveStructuredPost(user, post.id)
-    expect(structured?.metrics.map((m) => m.key)).toEqual(['heart_rate_avg'])
-    expect(structured?.series).toEqual([])
+    if (structured?.kind !== 'activity') throw new Error('expected an activity payload')
+    expect(structured.metrics.map((m) => m.key)).toEqual(['heart_rate_avg'])
+    expect(structured.series).toEqual([])
   })
 
   test('gates a followers-only post on the capability token', async () => {
@@ -115,12 +119,126 @@ describe('resolveStructuredPost', () => {
     expect(await resolveStructuredPost(user, post.id, 'wrong')).toBeNull()
     // The post's own capability token (delivered to followers) → resolves.
     const structured = await resolveStructuredPost(user, post.id, post.image_token)
-    expect(structured?.series.map((s) => s.metric)).toEqual(['heart_rate'])
-    expect(structured?.metrics.map((m) => m.key)).toEqual(['heart_rate_avg'])
+    if (structured?.kind !== 'activity') throw new Error('expected an activity payload')
+    expect(structured.series.map((s) => s.metric)).toEqual(['heart_rate'])
+    expect(structured.metrics.map((m) => m.key)).toEqual(['heart_rate_avg'])
   })
 
   test('returns null for an unknown post id', async () => {
     const user = getTestUser()
     expect(await resolveStructuredPost(user, '00000000-0000-4000-8000-000000000000')).toBeNull()
+  })
+})
+
+describe('resolveStructuredPost — article posts', () => {
+  beforeAll(async () => {
+    await startTestDb()
+  }, CONTAINER_TIMEOUT)
+
+  afterAll(async () => {
+    await stopTestDb()
+  })
+
+  beforeEach(async () => {
+    await cleanTestDb()
+  })
+
+  test('resolves prose verbatim and a chart block’s bucketed samples over its window', async () => {
+    const user = getTestUser()
+    // Dense heart-rate samples across the window (5-minute cadence, 150 bpm) so
+    // the default `1h` bucket (a ≤2-day window) has real data to aggregate.
+    await insertTimeSeries(
+      user,
+      Array.from({ length: 8 }, (_, i) => ({
+        metric: 'heart_rate' as const,
+        source: 'garmin' as const,
+        time: new Date(START.getTime() + i * 5 * 60_000),
+        value: 150,
+      })),
+    )
+    const post = await createArticlePost(user, {
+      article: {
+        blocks: [
+          { markdown: 'Slept **well**.', type: 'prose' },
+          { end: END.toISOString(), metric: 'heart_rate', start: START.toISOString(), type: 'chart' },
+        ],
+        title: 'My analysis',
+      },
+      visibility: 'public',
+    })
+
+    const structured = await resolveStructuredPost(user, post.id)
+    expect(structured?.kind).toBe('article')
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    expect(structured.title).toBe('My analysis')
+    expect(structured.blocks).toEqual([
+      { markdown: 'Slept **well**.', type: 'prose' },
+      expect.objectContaining({ metric: 'heart_rate', type: 'chart' }),
+    ])
+    const chart = structured.blocks[1]
+    if (chart.type !== 'chart') throw new Error('expected a chart block')
+    expect(chart.bucket).toBe('1h')
+    expect(chart.samples.length).toBeGreaterThan(0)
+    expect(chart.samples[0].avg).toBe(150)
+  })
+
+  test('resolves a correlation block (present unconditionally, even with no overlapping data)', async () => {
+    const user = getTestUser()
+    const post = await createArticlePost(user, {
+      article: {
+        blocks: [
+          {
+            end: END.toISOString(),
+            outcome: { kind: 'metric', metric: 'sleep_score' },
+            start: START.toISOString(),
+            trigger: { kind: 'metric', metric: 'steps' },
+            type: 'correlation',
+          },
+        ],
+        title: 'Steps vs sleep',
+      },
+      visibility: 'public',
+    })
+
+    const structured = await resolveStructuredPost(user, post.id)
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    expect(structured.blocks).toHaveLength(1)
+    const block = structured.blocks[0]
+    if (block.type !== 'correlation') throw new Error('expected a correlation block')
+    // No data seeded for either dimension — n is 0, not an error, so the block
+    // still resolves (unlike the block-image endpoint, which would 404 below n < 3).
+    expect(block.n).toBe(0)
+    expect(block.series).toEqual([])
+    expect(block.trigger).toEqual({ kind: 'metric', metric: 'steps' })
+    expect(block.outcome).toEqual({ kind: 'metric', metric: 'sleep_score' })
+  })
+
+  test('gates a followers-only article on the capability token', async () => {
+    const user = getTestUser()
+    const post = await createArticlePost(user, {
+      article: { blocks: [{ markdown: 'Private thoughts', type: 'prose' }], title: 'Private' },
+      visibility: 'followers',
+    })
+    expect(await resolveStructuredPost(user, post.id)).toBeNull()
+    expect(await resolveStructuredPost(user, post.id, 'wrong')).toBeNull()
+    const structured = await resolveStructuredPost(user, post.id, post.image_token)
+    expect(structured?.kind).toBe('article')
+  })
+
+  test('omits a chart/correlation block whose window is invalid (start on or after end)', async () => {
+    const user = getTestUser()
+    const post = await createArticlePost(user, {
+      article: {
+        blocks: [
+          { markdown: 'Intro', type: 'prose' },
+          { end: START.toISOString(), metric: 'heart_rate', start: END.toISOString(), type: 'chart' },
+        ],
+        title: 'Broken window',
+      },
+      visibility: 'public',
+    })
+    const structured = await resolveStructuredPost(user, post.id)
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    expect(structured.blocks).toEqual([{ markdown: 'Intro', type: 'prose' }])
   })
 })
--- a/apps/backend/src/services/feed-structured.integration.test.ts
+++ b/apps/backend/src/services/feed-structured.integration.test.ts
@@ -182,6 +182,42 @@ describe('resolveStructuredPost — article posts', () => {
     expect(chart.samples[0].avg).toBe(150)
   })
 
+  test('keeps a `1d` chart bucket as authored (not rewritten)', async () => {
+    const user = getTestUser()
+    await insertTimeSeries(
+      user,
+      Array.from({ length: 8 }, (_, i) => ({
+        metric: 'heart_rate' as const,
+        source: 'garmin' as const,
+        time: new Date(START.getTime() + i * 5 * 60_000),
+        value: 150,
+      })),
+    )
+    const post = await createArticlePost(user, {
+      article: {
+        blocks: [
+          {
+            bucket: '1d',
+            end: END.toISOString(),
+            metric: 'heart_rate',
+            start: START.toISOString(),
+            type: 'chart',
+          },
+        ],
+        title: 'Daily',
+      },
+      visibility: 'public',
+    })
+    const structured = await resolveStructuredPost(user, post.id)
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    const chart = structured.blocks[0]
+    if (chart.type !== 'chart') throw new Error('expected a chart block')
+    // `1d` is passed through — NOT rewritten (a broken floor turned it into `5s`).
+    expect(chart.bucket).toBe('1d')
+    expect(chart.samples.length).toBeGreaterThan(0)
+    expect(chart.samples[0].avg).toBe(150)
+  })
+
   test('resolves a correlation block (present unconditionally, even with no overlapping data)', async () => {
     const user = getTestUser()
     const post = await createArticlePost(user, {
--- a/apps/backend/src/services/feed-structured.integration.test.ts
+++ b/apps/backend/src/services/feed-structured.integration.test.ts
@@ -218,6 +218,34 @@ describe('resolveStructuredPost — article posts', () => {
     expect(chart.samples[0].avg).toBe(150)
   })
 
+  test('floors a sub-5s chart bucket on this unauthenticated endpoint', async () => {
+    const user = getTestUser()
+    await insertTimeSeries(user, [
+      { metric: 'heart_rate' as const, source: 'garmin' as const, time: START, value: 150 },
+    ])
+    const post = await createArticlePost(user, {
+      article: {
+        blocks: [
+          {
+            bucket: '1s',
+            end: END.toISOString(),
+            metric: 'heart_rate',
+            start: START.toISOString(),
+            type: 'chart',
+          },
+        ],
+        title: 'Dense',
+      },
+      visibility: 'public',
+    })
+    const structured = await resolveStructuredPost(user, post.id)
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    const chart = structured.blocks[0]
+    if (chart.type !== 'chart') throw new Error('expected a chart block')
+    // `1s` is floored to the 5s public-series minimum (not queried at 1s).
+    expect(chart.bucket).toBe('5s')
+  })
+
   test('resolves a correlation block (present unconditionally, even with no overlapping data)', async () => {
     const user = getTestUser()
     const post = await createArticlePost(user, {
--- a/apps/backend/src/services/feed-structured.ts
+++ b/apps/backend/src/services/feed-structured.ts
@@ -36,7 +36,7 @@ import { resolveActivityScalars } from './activitypub/feed-activity.ts'
 import { blockWindow, isZeroDurationBucket } from './article.ts'
 import { getContinuousCorrelation } from './correlations/explore.ts'
 import { isCapabilityAuthorized } from './feed-capability.ts'
-import { resolvePublicSeries, samplesFromBucketedResult } from './feed-series.ts'
+import { floorSeriesBucket, resolvePublicSeries, samplesFromBucketedResult } from './feed-series.ts'
 import { resolveFeedActivity } from './feed.ts'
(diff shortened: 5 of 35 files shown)


apps/web/src/pages/Feed/TimelineStructured.tsx
```diff
@@ -1,16 +1,21 @@
...
+import type { FeedStructuredPost } from '@aurboda/api-spec'
 
 import { TrendLineChart } from '../../components/charts/TrendLineChart'
 import { structuredChartSeries } from './timeline-structured'
+import { TimelineArticle } from './TimelineArticle'
 import './TimelineStructured.css'
 
-export function TimelineStructured({ structured }: { structured: FeedStructured }) {
+export function TimelineStructured({ structured }: { structured: FeedStructuredPost }) {
+  if (structured.kind === 'article') return <TimelineArticle article={structured} />
```
fiddur, line 17: An article ends up rendered twice on the receiving peer.

`HomeTimeline.tsx:77` renders `entry.content` unconditionally:

```jsx
<div class="feed-post-content" dangerouslySetInnerHTML={{ __html: entry.content }} />
{entry.structured ? <TimelineStructured structured={entry.structured} /> : …}
```

and for an article the delivered Note's content is `renderArticleContentHtml(post.article)` — `<p><strong>title</strong></p>` plus every prose block and caption — which ingest stores verbatim (`timeline-ingest.ts:102` sanitises but keeps `<p>`/`<strong>`/…). So the reader sees the bold title + all prose from the HTML, then `TimelineArticle` renders the `<h3>` title + the same prose (through `renderMarkdown`) + the charts again.

For an `activity` post this doesn't happen (the Note content is a short summary the chart complements), but for an article the structured render is a full replacement. `HomeTimeline.tsx` isn't in this diff, so it needs a change there — e.g. skip the `feed-post-content` div when `entry.structured?.kind === 'article'`.

packages/api-spec/src/schemas/feed-structured.ts
```diff
@@ -88,3 +91,125 @@ export const feedStructuredSchema = z
...
 export type FeedStructured = z.infer<typeof feedStructuredSchema>
+
+/**
+ * `FeedStructured` tagged with its post kind, so a consumer of the discriminated
+ * `FeedStructuredPost` union (below) can narrow on `kind` before reading the
+ * activity-shaped fields. The wire shape is otherwise identical to
+ * `FeedStructured` — this is purely the `kind` tag for the union.
+ */
+export const feedStructuredActivitySchema = feedStructuredSchema
+  .extend({ kind: z.literal('activity').meta({ description: 'Discriminator: an activity-share post' }) })
```
fiddur, line 102: Making `kind` required breaks enrichment from a peer that hasn't upgraded yet.

`enrichFromAurboda` parses the fetched body with `feedPostStructuredResponseSchema.safeParse` (`timeline-enrich.ts`), which now resolves to the `kind`-discriminated union. An Aurboda instance still running the previous release serves the old `FeedStructured` shape with no `kind`, so `safeParse` fails, `enrichFromAurboda` returns `null`, and the receiver silently falls back to the HTML — losing the native chart that used to work.

Worth noting that the render path was deliberately made tolerant of exactly this (`timeline-structured.ts`: "a legacy entry stored before the `kind` tag existed … gate on `article` (not `!== 'activity'`)"); the ingest parse has the same exposure but no such tolerance. A default/fallback to `'activity'` when `kind` is absent would keep cross-version enrichment working.

apps/web/src/components/ArticleExportDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <div class="share-dialog" onClick={(e) => e.stopPropagation()}>
+        <div class="share-dialog-header">
+          <h2>Export markdown</h2>
+          <button type="button" class="share-dialog-close" onClick={onClose} aria-label="Close">
+            ×
+          </button>
+        </div>
+
+        <p class="share-dialog-note">
+          Paste this into r/QuantifiedSelf or a similar text-only destination — add your own write-up around
```
fiddur, line 48: For a `followers`-only article the exported markdown contains the post's capability `?token=` in every image URL (covered intentionally by `article-export.test.ts` → "carries the capability token for a followers-only post"). This copy says to paste the result into r/QuantifiedSelf, so following the instruction publishes that token — which also authorizes `GET /public/:user/feed/:postId?token=` (the full structured payload, i.e. raw bucketed min/max/count/sum per block and the complete correlation data), not just the PNGs.

The token is needed for the images to render for a public reader, so the trade-off is deliberate — but the dialog should say so when the post isn't `public`/`unlisted` (the parent has `post`, so `post.visibility` is available), rather than leaving the user to discover it.

apps/backend/src/services/article.ts
```diff
@@ -9,10 +9,42 @@
...
+  postId: string,
+  visibility: FeedVisibility,
+  imageToken: string,
+  updatedAt: Date,
+  index: number,
+  format: 'png' | 'svg' = 'png',
+): string => {
+  const base = `${apiBaseUrl.replace(/\/+$/, '')}/public/${encodeURIComponent(user)}/feed/${postId}`
+  const params = new URLSearchParams({ v: String(updatedAt.getTime()) })
+  if (visibility === 'followers') params.set('token', imageToken)
```
fiddur: This replaces `!isPubliclyVisible(visibility)` with an equality check, inverting a deliberate allowlist into a denylist. `isPubliclyVisible` documents the intent: "An allowlist, so any future non-public visibility is excluded by default." With `visibility === 'followers'`, a future non-public value (e.g. `private`/`direct`) would silently get a token-less URL and its images/export would 404 rather than fail closed at review time.

Behaviour is identical today (`FeedVisibility` is `public | unlisted | followers`). If the goal was to keep `article.ts` free of the `activitypub/object.ts` (fedify) import, `isPubliclyVisible` is also reachable via `feed-capability.ts`'s dependency graph, or the helper could be moved somewhere neutral.

fiddur (commented):
🛑 Changes required

Solid slice: the structured payload becomes a proper discriminated union, the block-image URL and the zero-bucket guard are hoisted into `article.ts` so the AS2 attachment builder, the block-image renderer and the new export can't drift, `samplesFromBucketedResult` is shared with the public series path, and REST + MCP parity is kept for the export. Test coverage is good (unit tests for `buildArticleMarkdown`/`articleBlockImageUrl`, integration tests for the article structured resolver and the export route), and docs are updated.

One blocking issue: **a federated article renders twice on a receiving Aurboda peer.** `HomeTimeline` renders `entry.content` unconditionally, and the delivered Note's content for an article is already `renderArticleContentHtml(...)` — bold title + every prose block + captions. Adding `TimelineArticle` on top means the receiver shows title + prose from the sanitised HTML and then the whole article again natively. `HomeTimeline.tsx` isn't in this diff, so the fix needs to go there (or the branch in `TimelineStructured` needs to suppress the HTML), e.g. skip `entry.content` when `entry.structured?.kind === 'article'`.

Other things worth a look (details inline):

- Requiring `kind` on `FeedStructuredActivity` makes enrichment from a peer running the previous release fail to parse, silently dropping native charts. The web render path was deliberately made tolerant of pre-`kind` payloads (`timeline-structured.ts`), but the ingest parse isn't.
- Exporting a `followers`-only article emits the post's capability token into text the dialog explicitly tells the user to paste publicly.
- `articleBlockImageUrl` inverts the documented `isPubliclyVisible` allowlist into a `visibility === 'followers'` denylist.

Two smaller notes, no inline comment:

- `buildArticleMarkdown` carefully collapses whitespace and escapes `]` in captions/labels, but interpolates `article.title` raw into `# ${title}`. Titles are `z.string().min(1).max(200)` with no newline restriction, so a title containing a newline breaks the H1 into a paragraph. Passing the title through `inlineText` too would make it uniform.
- `GET /public/:username/feed/:postId` is unauthenticated and `no-store`, and now resolves up to 100 blocks per request (a bucketed query or a full `getContinuousCorrelation` each), sequentially. Per-block cost matches the existing image endpoint, but a single request can now chain 100 of them, and unlike `resolvePublicSeries` the article chart path doesn't floor the bucket, so an author-chosen `1s` bucket over a long window has no payload bound. Probably fine for now given blocks are author-authored, but it's a bigger unauthenticated surface than before.

apps/web/src/pages/Feed/TimelineStructured.tsx
fiddur: Fixed: HomeTimeline now skips `entry.content` when `entry.structured?.kind === 'article'`, so the native TimelineArticle (same title + prose + captions) isn't rendered on top of the sanitised Note HTML. Activity posts and enrichment-failed articles still show `content`.

packages/api-spec/src/schemas/feed-structured.ts
fiddur: Fixed: feedStructuredPostSchema now preprocesses a payload with no `kind` into an activity before discriminating, so enrichFromAurboda's safeParse still accepts the un-tagged shape a peer on the previous release emits (no silently-dropped chart during a rolling upgrade). The web render path was already tolerant; both sides now are.

apps/web/src/components/ArticleExportDialog.tsx
fiddur: Fixed: the export (REST route + MCP tool) now 400s for a followers-only article — its images need the private capability token, which must not land in publicly-pasted text. The dialog only ever exports public/unlisted. Flipped the unit test (public URLs carry no token) and added a route integration test for the 400.

apps/backend/src/services/article.ts
fiddur: Fixed: articleBlockImageUrl now uses `!isPubliclyVisible(visibility)` (the documented allowlist) instead of `=== 'followers'`, so a future non-public visibility is token-gated by default rather than silently treated as public.

apps/backend/src/services/feed-structured.ts
```diff
@@ -59,21 +80,147 @@ const resolveStructuredSeries = async (
...
+ */
+const resolveStructuredArticle = async (
+  user: string,
+  article: ArticleContent,
+): Promise<FeedStructuredArticle> => {
+  const settings = await getUserSettings(user)
+  const tz = settings?.device_timezone ?? undefined
+
+  const blocks: FeedStructuredArticleBlock[] = []
+  for (const block of article.blocks) {
```
fiddur, line 193: This loop is unauthenticated, uncached and unbounded work on a public endpoint.

`articleContentSchema` allows up to 100 blocks, and every chart block runs a `queryMetricsBucketed` while every correlation block runs a full `getContinuousCorrelation` (two `resolveSelector` queries + the stats). Serially, on every `GET /public/:username/feed/:postId`, which sets `Cache-Control: no-store`. Public post ids are enumerable from `/public/:username/posts`, so one cheap request can fan out to ~200 DB round trips.

The sibling block-image endpoint was deliberately hardened against exactly this — `blockCacheKey` mixes in `Math.floor(Date.now() / 3_600_000)` plus a negative cache, with the comment "can't each dodge the hourly-bucket rate limit on this unauthenticated, comparatively expensive render." The structured path bypasses all of it.

There's a functional consequence too: a many-block article resolved serially can easily exceed the peer's `ENRICH_TIMEOUT_MS` (12 s), so enrichment silently returns `null` and the article never renders natively for the very case this slice exists to serve.

Suggest reusing the same hourly-bucketed render cache keyed on `(user, postId, updated_at)`, and/or capping the number of chart/correlation blocks resolved per payload (logging what was dropped).

apps/web/src/components/ArticleExportDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+          the linked charts.
+        </p>
+
+        {isLoading && <p>Loading…</p>}
+        {isError && <p class="share-dialog-error">Couldn't export this article. Please try again.</p>}
```
fiddur: The generic error swallows the one message the user needs to act on. The backend returns a 400 with `error: 'A followers-only article can't be exported — its charts need a private link. Make it public or unlisted first.'`, but `fetchArticleExport` throws a bare axios error and this branch renders "Couldn't export this article. Please try again." — advice that will never work.

The Export markdown button in `Feed/index.tsx` is rendered for every article regardless of visibility, so a followers-only article is a reachable dead end. Either surface `error.response.data.error` here, or hide/disable the button when `post.visibility` isn't public/unlisted.

apps/backend/src/services/article-export.ts
```diff
@@ -0,0 +1,59 @@
...
+import { describeSelectorAxis, getMetricDisplayName } from '@aurboda/api-spec'
+
+import { articleBlockImageUrl } from './article.ts'
+
+/** Collapse newlines (markdown alt/caption text is a single line) and escape `]` so it can't close the image syntax early. */
+const inlineText = (text: string): string => text.replaceAll(/\s+/g, ' ').trim().replaceAll(']', '\\]')
+
+/** A human label for a chart/correlation block's image: its caption, else a label from its data. */
+const blockLabel = (block: Extract<ArticleBlock, { type: 'chart' | 'correlation' }>): string =>
+  block.caption ??
```
fiddur: `blockLabel` is a copy of `attachmentName` in `activitypub/article-object.ts`, and the copy diverges: `attachmentName` uses `if (block.caption)` (truthy) while this uses `??` (nullish). `caption` is `z.string().max(280).optional()`, so an empty-string caption is storable — it yields the metric/selector label in the AS2 attachment but an empty `` alt here.

Given this PR extracted `articleBlockImageUrl` precisely so the two surfaces can't disagree about a block, the label deserves the same treatment — export the helper from `article-object.ts` (or alongside `articleBlockImageUrl`) and use it in both.

packages/api-spec/src/schemas/feed-structured.ts
```diff
@@ -88,3 +91,137 @@ export const feedStructuredSchema = z
...
+ * resolved blocks. `kind` tells a consuming peer (and the web timeline card)
+ * which native render to use.
+ *
+ * A payload with NO `kind` is treated as an `activity` before discriminating, so
+ * enrichment fetched from a peer running the *previous* release (which emitted
+ * the un-tagged `FeedStructured` shape) still parses instead of silently dropping
+ * the native chart during a rolling upgrade.
+ */
+export const feedStructuredPostSchema = z
+  .preprocess(
```
fiddur, line 221: This rolling-upgrade shim is the one piece of the union with no test. `timeline-enrich.test.ts` was updated to send `kind: 'activity'` explicitly, so nothing exercises the path this preprocess exists for — a peer on the previous release emitting an untagged `FeedStructured`. A regression that drops the `.preprocess(...)` wrapper (or changes the guard) would silently stop enriching every pre-upgrade peer and no test would fail.

A one-liner in `timeline-enrich.test.ts` asserting that a payload with the `kind` key deleted still parses to `kind: 'activity'` would lock it in.

fiddur (commented):
✅ Approved

Solid, well-documented slice. I traced the three new surfaces end to end and found no correctness or security defects:

- **Authorization** is unchanged and correct: `resolveStructuredPost` now runs `isCapabilityAuthorized` *before* branching on `kind`, so a `followers`-only article resolves only with its `image_token` — same boundary as the block images and the outbox. Integration tests cover the token/no-token/wrong-token cases.
- **XSS boundary** is right: the article's prose crosses the wire as raw markdown and `TimelineArticle` renders it through `renderMarkdown` (marked → DOMPurify), i.e. the receiver sanitises rather than trusting a remote instance's HTML. Suppressing `entry.content` for `structured.kind === 'article'` correctly avoids the double render.
- **Backward compat** is handled: the `z.preprocess` shim tags a `kind`-less payload as `activity`, and `structuredChartSeries` gates on `=== 'article'` rather than `!== 'activity'`, so pre-existing `timeline_entry.structured` rows (which are never re-parsed through Zod on read) still render their native chart.
- **Export** refuses `followers`-only at both the REST route and the MCP tool before ever calling `buildArticleMarkdown`, so the capability token can't land in publicly-pasted text. Extracting `articleBlockImageUrl` so the AS2 attachment and the export can't diverge is a good call.
- Chart bucketing matches the block-image renderer (same `defaultArticleChartBucket`, same `device_timezone`, same `isZeroDurationBucket` guard, now shared), and the correlation path reuses the same `getContinuousCorrelation` call with the same raw-ISO `slice(0, 10)` day bounds. Docs, API/MCP parity, and tests are all in place.

The one thing I'd actually act on before this sees traffic is the caching/bounding gap on the article branch of the public structured endpoint (inline). The other three notes are minor.

Small extra: `ArticleExportDialog` uses `class="article-input"` but only imports `ShareActivityDialog.css` — that class lives in `ArticleEditorDialog.css` and only happens to be loaded because `Feed/index.tsx` imports both dialogs into the same chunk. Worth importing the CSS it actually uses.

packages/api-spec/src/schemas/feed-structured.ts
fiddur: Fixed: added an enrichFromAurboda test that feeds a kind-less payload (the previous release's FeedStructured shape) and asserts it parses tagged as activity — exercising the z.preprocess shim end to end.

apps/backend/src/services/feed-structured.ts
fiddur: Bounded the concrete case: the chart bucket is now floored to the public-series minimum (5s) in resolveStructuredChartBlock, like resolvePublicSeries, so a sub-5s bucket can't 5×-balloon the sample count. The remaining fuller work (a per-series payload cap — window÷bucket is still unbounded, same as resolvePublicSeries — and a structured-payload cache) is filed as #972 (sibling to #969). Non-blocking on a single-user, nginx-fronted, once-per-ingest surface.

apps/web/src/components/ArticleExportDialog.tsx
fiddur: Fixed: fetchArticleExport now runs failures through apiErrorMessage (same as createArticle/updateArticle), and the dialog renders error.message — so a followers-only export shows the server's actionable 'make it public or unlisted first' text instead of the generic message.

apps/backend/src/services/article-export.ts
fiddur: Fixed: extracted a single articleBlockLabel into services/article.ts (truthy-caption fallback) and pointed both articleImageAttachments and buildArticleMarkdown at it, removing the two divergent copies.

apps/backend/src/services/feed-structured.ts
```diff
@@ -59,21 +80,151 @@ const resolveStructuredSeries = async (
...
+  if (start == null || end == null) return null
+  const startDate = new Date(start)
+  const endDate = new Date(end)
+  if (startDate.getTime() >= endDate.getTime()) return null
+  const rawBucket = block.bucket ?? defaultArticleChartBucket(startDate, endDate)
+  if (isZeroDurationBucket(rawBucket)) return null
+  // Floor to the public-series minimum (5s), like `resolvePublicSeries`, so an
+  // author-chosen sub-5s bucket can't balloon the sample count on this
+  // unauthenticated endpoint (see #972 for a fuller payload cap + cache).
+  const bucket = floorSeriesBucket(rawBucket)
```
fiddur: `floorSeriesBucket` silently rewrites any `d`-unit bucket to `5s`, so this breaks the common article case rather than bounding it.

Its regex is `/^(\d+)([smh])$/` (`feed-series.ts:87`) — no `d`. `'1d'` doesn't match, hits `if (!match) return '5s'`, and is returned as `5s`. But `defaultArticleChartBucket` returns `'1d'` for any window longer than two days (`feed.ts:166`), and `articleChartBlockSchema` documents `1d` as a valid author-set bucket.

Concretely, for the PR's own fixture shape (a 7-day `heart_rate` chart block, no explicit bucket):
- `rawBucket` = `'1d'` → `bucket` = `'5s'`
- `queryMetricsBucketed` returns up to 120 960 buckets (7 × 86400 / 5), all non-empty for 1 Hz Garmin HR
- the payload is roughly 15 MB, so `safeFetchGet`'s 1 MB `maxContentLength` (`safe-fetch.ts:77`) rejects it and `enrichFromAurboda` swallows it to `null` — the peer never renders the native article at all
- and where it *does* fit, the peer draws a 5-second chart while the block's own PNG (`renderArticleBlockImage`, which passes the bucket through unfloored) draws a daily one, contradicting the "same bucketing the block's own PNG uses" claim in the schema doc and the "a `1d` bucket splits on the author's calendar days" comment on `resolveStructuredArticle` below.

The intent here ("an author-chosen sub-5s bucket can't balloon the sample count") is right — it's the reuse of `floorSeriesBucket` that's wrong, since that helper was written for the `seriesBucketSchema` domain (`s`/`m`/`h` only), not the article-chart domain (`s`/`m`/`h`/`d`). Either teach `floorSeriesBucket` the `d` unit (`parseBucketSize` already handles `smhdwM`), or floor with an article-specific helper and leave the public-series one alone. Worth adding a `1d` case to the integration test either way.

apps/backend/src/services/feed-structured.integration.test.ts
```diff
@@ -115,12 +119,126 @@ describe('resolveStructuredPost', () => {
...
+    expect(structured?.kind).toBe('article')
+    if (structured?.kind !== 'article') throw new Error('expected an article payload')
+    expect(structured.title).toBe('My analysis')
+    expect(structured.blocks).toEqual([
+      { markdown: 'Slept **well**.', type: 'prose' },
+      expect.objectContaining({ metric: 'heart_rate', type: 'chart' }),
+    ])
+    const chart = structured.blocks[1]
+    if (chart.type !== 'chart') throw new Error('expected a chart block')
+    expect(chart.bucket).toBe('1h')
```
fiddur, line 180: This asserts the `1h` branch only — `START`..`END` is a 40-minute window, so `defaultArticleChartBucket` returns `'1h'`, which is the one case `floorSeriesBucket` passes through unchanged. The `1d` default (any window > 2 days, i.e. the normal article shape) is untested, which is why the flooring bug above slipped through. A second case with a multi-day window asserting `chart.bucket === '1d'` would pin it down.

fiddur (commented):
🛑 Changes required

Solid slice overall — the shared-helper extraction (`articleBlockLabel` / `articleBlockImageUrl` / `isZeroDurationBucket` / `samplesFromBucketedResult`) genuinely removes the duplication risk between the AS2 attachment builder, the block-image renderer and the new markdown export; the `kind` discriminator with a preprocess shim for pre-upgrade peers is a nice touch, and it's covered by a test. Docs are thorough and in sync. The `followers`-only export refusal (both REST and MCP) is the right call and correctly keeps the capability token out of publicly-pasted text.

**One blocking bug**, in `resolveStructuredChartBlock`: it runs the block's bucket through `floorSeriesBucket`, whose regex is `/^(\d+)([smh])$/` — it has no `d` branch, so any `d`-unit bucket falls through the `if (!match)` guard and is replaced with `5s`. `defaultArticleChartBucket` returns `'1d'` for every window longer than two days, and `articleChartBlockSchema` explicitly allows an author-set `1d`. So the common article case ("a week of sleep and HRV" — the PR's own fixture title) resolves at 5-second granularity instead of daily. Details inline; the integration test only exercises a 40-minute window, so the `1d` path is entirely uncovered.

Two smaller, non-blocking observations for the summary:

- **Unbounded, uncached per-block work on an unauthenticated endpoint.** `resolveStructuredArticle` loops over up to 100 blocks sequentially, issuing a `queryMetricsBucketed` or a full `getContinuousCorrelation` per block, and `GET /public/:username/feed/:postId` sets `Cache-Control: no-store` with no server-side render cache (unlike the block-image route, which has `createRenderCache`). Public post IDs are discoverable via the profile feed listing, so this is a new amplification path. The code comment already points at #972 for "a fuller payload cap + cache" — just noting the exposure is real today, and it gets much worse combined with the bucket bug above.
- **The export emits an image link for every chart/correlation block**, including ones `renderArticleBlockImage` will 404 on (chart with < 2 points, correlation with n < 3, zero-duration bucket). The pasted markdown then carries a dead image. This matches the existing `articleImageAttachments` behaviour, so it's consistent rather than a regression — but it's more visible in a paste-and-publish flow than in an AS2 attachment list.

I did not run the test suite (no `node_modules` in this worktree); the finding above is from reading `floorSeriesBucket` and confirming the regex behaviour directly.

apps/backend/src/services/feed-structured.integration.test.ts
fiddur: Fixed by the revert (the floor is gone). Also added an integration test with an explicit `bucket: '1d'` block asserting `chart.bucket === '1d'` — covering the daily branch the existing 40-minute/`1h` test missed, and guarding against a future bucket-rewrite regression.

apps/backend/src/services/feed-structured.ts
fiddur: Good catch — floorSeriesBucket's regex is s/m/h only, so it turned the common `1d` bucket into `5s`, breaking daily charts. Reverted: the chart buckets as-authored (like the block-image path, which handles `1d`). The floor only ever touched sub-5s buckets, and the meaningful payload bound (a sample cap + cache for both this endpoint and the block images) is #972 / #969 — kept out of this PR so both surfaces get one fix.

fiddur (commented):
✅ Approved

Solid, well-documented slice. The `FeedStructured` → `FeedStructuredPost` discriminated-union migration is done consistently across api-spec, the DB layer, enrichment, and the web timeline, and the rolling-upgrade shim (`z.preprocess` tagging a kind-less payload as `activity`) plus the deliberate `kind === 'article'` (rather than `!== 'activity'`) gate in `structuredChartSeries` correctly protect timeline entries stored before the tag existed. Sharing `articleBlockImageUrl` / `articleBlockLabel` / `isZeroDurationBucket` / `samplesFromBucketedResult` removes the real divergence risk between the AS2 attachment, the block image, and the new export. Route ordering, the `isCapabilityAuthorized` gate for articles, and the followers-only export refusal (so the capability token never lands in publicly-pasted text) all look right, and the export/MCP parity is in place per CLAUDE.md.

**Verification run locally** (deps installed, api-spec built): `pnpm check` — 0 errors; backend unit suite — 2246 tests / 159 files passing; web — 549 tests passing; and the two touched integration files (`feed-structured.integration.test.ts`, `feed-router.integration.test.ts`) — 20 tests passing against a real Postgres container, including the new `1d`-bucket regression test that pins the round-3 revert.

Two non-blocking notes are inline. Separately, the unauthenticated `GET /public/:user/feed/:postId` now resolves up to 100 article blocks per request (a bucketed metrics query or a full continuous-correlation computation each) with `Cache-Control: no-store` and no server-side cache or sample cap — that is a real amplification surface, but it is the same one the block-image endpoint already has on `develop`, and it is tracked in the open #972 / #969, so deferring it here (rather than repeating the round-2 attempt that broke `1d` buckets) is the right call.

apps/web/src/pages/Feed/TimelineArticle.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { TrendLineChart } from '../../components/charts/TrendLineChart'
+import { renderMarkdown } from '../../utils/markdown'
+import { getMetricDisplayName } from '../../utils/metricLabels'
+import { CorrelationScatterSvg } from './ArticleCorrelationBlock'
+
+const CHART_COLOR = '#673ab8'
+
+const TimelineArticleBlock = ({ block }: { block: FeedStructuredArticleBlock }) => {
+  if (block.type === 'prose') {
+    return <div class="article-prose" dangerouslySetInnerHTML={{ __html: renderMarkdown(block.markdown) }} />
```
fiddur: Remote-peer markdown renders under DOMPurify's **default** allowlist, which is broader than the policy this codebase deliberately chose for the very same posts. `sanitizeRemoteHtml` (`timeline-ingest.ts`) drops `` outright and forces `rel="nofollow noopener noreferrer"` / `target` on links — its docstring calls out images explicitly — while DOMPurify's defaults keep both.

Concretely: a followee's instance serves `structured.blocks[0] = { type: 'prose', markdown: '' }`. The same content in `entry.content` would have had the `` stripped server-side; here the reader's browser fetches the tracker on timeline render. Not XSS (scripts/handlers are still removed), and remote `` already loads via `entry.images` for non-article entries — but for an article entry that branch is skipped, so this becomes the only image vector for the card, and it's arbitrary markdown rather than the declared attachment list.

Worth a dedicated `renderRemoteMarkdown` (or a `FORBID_TAGS: ['img']` + link-hardening variant) that mirrors `sanitizeRemoteHtml`'s allowlist, so the two inbound-federation surfaces enforce one policy.

apps/web/src/pages/Feed/TimelineArticle.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * Prose is the author's raw markdown, rendered through the same sanitising
+ * `renderMarkdown` the author's own composer uses (the #910 XSS boundary),
+ * since a remote peer's markdown is untrusted content.
+ */
+import type { FeedStructuredArticle, FeedStructuredArticleBlock } from '@aurboda/api-spec'
+
+import { TrendLineChart } from '../../components/charts/TrendLineChart'
+import { renderMarkdown } from '../../utils/markdown'
+import { getMetricDisplayName } from '../../utils/metricLabels'
+import { CorrelationScatterSvg } from './ArticleCorrelationBlock'
```
fiddur, line 21: The `.article-prose` / `.article-chart*` / `.article-scatter*` classes this component uses live in `FeedPostCard.css`, which this module never imports (`TimelineStructured.css` only defines `.timeline-*`). It renders correctly today only because there's no route-level code splitting, so every stylesheet ends up in one bundle — the same "same-chunk coincidence" the round-2 commit deliberately removed for `ArticleExportDialog`. An explicit `import './FeedPostCard.css'` here would keep it consistent.

apps/web/src/pages/Feed/TimelineArticle.tsx
fiddur: Fixed: added renderRemoteMarkdown (DOMPurify with FORBID_TAGS: ['img']), mirroring the backend sanitizeRemoteHtml policy, and TimelineArticle now uses it — so a remote article's prose can't smuggle a tracking-pixel ; remote images arrive only via the declared attachment list. (marked emits <a> without target=_blank, so there's no tabnabbing vector to harden.) Tests cover the img-forbid + that safe formatting/script-stripping still hold.

apps/web/src/pages/Feed/TimelineArticle.tsx
fiddur: Fixed: TimelineArticle now does an explicit import './FeedPostCard.css' (where .article-prose / .article-chart* / .article-scatter* live), rather than relying on the same-chunk coincidence — same fix as the round-2 ArticleExportDialog CSS import.

apps/web/src/utils/markdown.ts
```diff
@@ -19,3 +19,15 @@ marked.setOptions({ breaks: true, gfm: true })
...
+ * Render markdown authored by a REMOTE peer (e.g. a federated article's prose) to
+ * sanitised HTML. Like `renderMarkdown`, but **forbids `<img>`** — a remote
+ * instance must not smuggle a tracking pixel into the reader's timeline card;
+ * remote images arrive only via the post's declared attachment list. Mirrors the
+ * backend's inbound `sanitizeRemoteHtml` policy (which also drops `<img>`), so
+ * both inbound-federation surfaces enforce one allowlist. (marked emits `<a>`
+ * without `target`, so there's no tabnabbing vector to harden here.)
+ */
+export const renderRemoteMarkdown = (md: string): string =>
+  DOMPurify.sanitize(marked.parse(md, { async: false }), { FORBID_TAGS: ['img'] })
```
fiddur: `FORBID_TAGS: ['img']` doesn't close the hole the comment above describes. It's a one-item denylist on top of DOMPurify's **default** allowlist, which still keeps `<style>`, the `style=` attribute, `<video poster>`, `<audio>/<source srcset>`, `<svg><image href>` and `<input type="image" src>` — and `marked` passes raw HTML through verbatim, so a remote peer fully controls what reaches the sanitiser.

Concretely, a followed peer publishes an article whose prose block is:

```html
<div style="background-image:url(https://tracker.example/px.png?u=alice)"></div>
<video poster="https://tracker.example/px.gif?u=alice"></video>
```

Ingest stores `markdown` verbatim (it's just a string in `feedStructuredArticleProseBlockSchema`), `TimelineArticle` renders it through this function, and the reader's browser fetches both URLs — the tracking pixel this is meant to prevent (IP, UA, exact read time).

The doc comment's "mirrors the backend's inbound `sanitizeRemoteHtml` policy … so both inbound-federation surfaces enforce one allowlist" also isn't accurate: `sanitizeRemoteHtml` is a strict ~20-tag allowlist with attributes only on `a`/`span`/`ol`. Suggest actually mirroring it:

```ts
DOMPurify.sanitize(marked.parse(md, { async: false }), {
  ALLOWED_TAGS: ['p','br','a','span','em','strong','b','i','del','s','u','pre','code','blockquote','ul','ol','li','h1','h2','h3','h4','table','thead','tbody','tr','th','td'],
  ALLOWED_ATTR: ['href','rel','target','class','translate','start'],
})
```

(or at minimum `FORBID_TAGS: ['img','video','audio','source','picture','track','input','style','svg','math']` + `FORBID_ATTR: ['style','poster','srcset','background']`). The parenthetical about tabnabbing is also only true for marked-generated anchors — raw `<a target="_blank">` in the remote markdown survives DOMPurify's defaults (modern browsers imply `noopener`, so this part is largely moot, but the comment overstates it).

apps/backend/src/services/feed-structured.ts
```diff
@@ -59,21 +80,151 @@ const resolveStructuredSeries = async (
...
+  const { end, start } = blockWindow(block, content)
+  if (start == null || end == null) return null
+  const startDate = new Date(start)
+  const endDate = new Date(end)
+  if (startDate.getTime() >= endDate.getTime()) return null
+  // Bucket as authored (same as the block-image path — including `1d`, which
+  // `floorSeriesBucket` doesn't understand). Bounding this unauthenticated
+  // payload (a sample cap + cache) is tracked in #972, alongside the block-image
+  // endpoint's #969 — kept out of this PR so both surfaces get one fix.
+  const bucket = block.bucket ?? defaultArticleChartBucket(startDate, endDate)
```
fiddur: The round-3 revert threw out the bound along with the bug. `floorSeriesBucket`'s regex is `/^(\d+)([smh])$/` (and `BUCKET_UNIT_SECONDS` has no `d`), so `1d` fell through to the `!match` branch and became `5s` — that's the breakage round-3 fixed, but the fix was to drop the floor entirely rather than teach the helper about `d`:

```ts
const BUCKET_UNIT_SECONDS: Record<string, number> = { d: 86400, h: 3600, m: 60, s: 1 }
const match = bucket.match(/^(\d+)([smhd])$/)
```

which keeps `1d` intact *and* restores the sub-5s guard here. As it stands an author-stored `bucket: '1s'` over the article's default window (nothing in `articleChartBlockSchema` forbids it — the regex is just `^\d+[smhd]$`) yields ~600k buckets for a week, on an endpoint that is unauthenticated and `Cache-Control: no-store`.

Same thread, wider: `resolveStructuredArticle` walks up to 100 blocks sequentially per request, each a `queryMetricsBucketed` or a full `getContinuousCorrelation` (two selector resolutions), with no cache at all — while the sibling block-image endpoint got `createRenderCache` + `createNegativeCache` + an hourly cache-bucket precisely because it's "unauthenticated, comparatively expensive" and it only ever does **one** block per request. Deferring the full cap to #972 is reasonable, but at minimum the bucket floor (above) and reusing the same hourly-bucketed cache keyed on `(user, postId, updated_at)` would put this surface on par with the one it borrows its authorization model from.

fiddur (commented):
🛑 Changes required

Solid slice overall. The `FeedStructured` → discriminated `FeedStructuredPost` migration is done thoroughly (DB record types, federation enricher, web renderers, docs all updated), the rolling-upgrade `preprocess` shim for kind-less peer payloads is a nice touch and is tested, `articleBlockImageUrl` / `articleBlockLabel` / `isZeroDurationBucket` / `samplesFromBucketedResult` being lifted into shared helpers genuinely removes the divergence risk between the AS2 attachment builder, the block-image renderer and the new export, and the export route's followers-only refusal (so the capability token never lands in publicly-pasted text) is the right call and is covered by an integration test. Docs in `docs/features/feed.md` are updated in step with the code.

Two things I'd like addressed before merge:

1. **`renderRemoteMarkdown` doesn't achieve what its doc comment claims** (inline). `FORBID_TAGS: ['img']` is a one-item denylist layered over DOMPurify's *default* allowlist, which still permits `<style>`, the `style=` attribute, `<video poster>`, `<svg><image href>`, `<input type=image src>` and `<source srcset>` — all of which cause the reader's browser to fetch a remote-peer-controlled URL. Since `marked` passes raw HTML straight through, a hostile peer only has to put that HTML in an article prose block. This is a privacy leak (reader IP / UA / read time to a remote server), not XSS, but it's exactly the tracking pixel the comment says is blocked, and the claim that it "mirrors the backend's inbound `sanitizeRemoteHtml` policy … so both inbound-federation surfaces enforce one allowlist" isn't accurate — `sanitizeRemoteHtml` is a ~20-tag, ~5-attribute allowlist.

2. **The unauthenticated structured endpoint lost the bounds its sibling has** (inline). `GET /public/:username/feed/:postId` is unauthenticated and `no-store`, and for an article it now resolves *every* block in one request — up to 100 blocks (`articleContentSchema.blocks.max(100)`), each either a `queryMetricsBucketed` over the block's window or a full `getContinuousCorrelation` (two selector resolutions), sequentially, with no result cache. The block-image path deliberately got an LRU + negative cache + hourly cache-bucket for being "unauthenticated, comparatively expensive"; the structured path is strictly more expensive per request and has none of that. The round-2 bucket floor was reverted wholesale in round-3 rather than fixed, so the sub-5s guard is gone too — see the inline note for the one-line root cause.

Minor, non-blocking:

- `feed-router.integration.test.ts` and `feed-structured.integration.test.ts` each add a second top-level `describe` with its own `startTestDb()`/`stopTestDb()`, so each file now boots the postgis testcontainer twice. Hoisting the container hooks to file level (and keeping `cleanTestDb()` per-test) would cut a couple of minutes off CI.
- `ArticleExportDialog` uses react-query defaults, so the deterministic 400 for a followers-only article is retried 3× with backoff before the (good) actionable message appears; `retry: false` — or hiding the Export button for `visibility === 'followers'` — would be snappier.
- `TimelineArticle`'s `.filter((p): p is { date: string; value: number } => p.value != null)` is a no-op: `publicSeriesSampleSchema.avg` is `z.number()`, not nullable.
- Nothing to do for OpenAPI/Kotlin regeneration — `packages/api-spec/generated/` is gitignored and `openapi.ts` doesn't cover the feed schemas.

I could not run `pnpm check` / the test suites here (no `node_modules` in this worktree), so the above is static review only.

apps/backend/src/services/feed-structured.ts
fiddur: Done both: taught floorSeriesBucket about `d` (regex + BUCKET_UNIT_SECONDS) so `1d` stays intact, and restored the floor in resolveStructuredChartBlock (a `1s` bucket → `5s`). Also added the cache you suggested: createRenderCache is now generic and the structured endpoint caches its payload under an hourly-bucketed, token-keyed key (public never sees a followers-only payload; a wrong token 404s uncached) — on par with the block images. Tests: 1d passthrough, 1s→5s floor, floorSeriesBucket('1d'). The fuller per-block sample cap + updated_at keying stays in #972.

apps/web/src/utils/markdown.ts
fiddur: Right — a one-item denylist on DOMPurify's default still leaks trackers via <style background-image>/<video poster>/<source srcset>/<svg image>/<input type=image>. Switched renderRemoteMarkdown to a strict ALLOWED_TAGS/ALLOWED_ATTR allowlist mirroring sanitizeRemoteHtml (no media tags at all), and dropped the overstated tabnabbing note. Test now asserts the style/video/svg vectors are stripped too, and that the safe allowlist (incl. tables) + script-stripping still hold.

fiddur (commented):
🛑 Changes required

Solid slice overall — the `FeedStructuredPost` discriminated union with the kind-less preprocess shim is a nice rolling-upgrade story, `articleBlockLabel`/`articleBlockImageUrl`/`isZeroDurationBucket`/`samplesFromBucketedResult` genuinely de-duplicate logic that had started to drift between the AS2 attachment builder, the block-image renderer and the new export, the `renderRemoteMarkdown` allowlist (rather than a `FORBID_TAGS: ['img']` denylist) is the right shape for an untrusted-peer sink, and `docs/features/feed.md` is kept properly in sync. Test coverage for the new pure pieces is good.

**Blocking:** the structured-payload cache added on `GET /public/:username/feed/:postId` in the r5 commit has two problems, both on an unauthenticated endpoint and both fixed by the same restructuring.

1. The cache key interpolates the raw `?token=` query value, but `isCapabilityAuthorized` *ignores* the token for `public`/`unlisted` posts — so any token value resolves and gets its own cache entry. An attacker can walk `?token=1`, `?token=2`, … against a public article and force a full miss every time, re-running up to 100 block resolutions (bucketed metric queries + `getContinuousCorrelation`) per request while evicting the 200-entry LRU. That's exactly the cost the cache was added to bound.
2. Authorization now happens *inside* the `produce` callback. The sibling image endpoints deliberately authorize first (`resolveImageWindow` / `resolveArticleBlock` run before `cached(...)`), and this handler sets `Cache-Control: no-store` precisely because "sharing is revocable … must take effect immediately". As written, flipping a public post to `followers` (or deleting it) still serves the full structured payload to unauthenticated callers until the hour bucket rolls.

Both go away if the post is loaded + authorized before consulting the cache, with the cache keyed on `username:postId:updated_at` (no token, and `updated_at` also removes the ≤1h edit staleness the comment currently apologises for).

**Non-blocking notes**

- `createRenderCache`'s default `maxEntries = 200` was sized for PNG buffers; the structured payloads it now holds are unbounded (100 blocks × up to 20 000 chars of prose, plus samples). Worth a smaller cap for this instance, or folding it into the #972 sample-cap work.
- Related: the emitting side applies no sample cap, while the receiving side's `safeFetchGet` caps responses at 1 MB — an article with a fine-grained chart bucket over a wide window will silently fail enrichment on the peer. Default buckets (`1h`/`1d`) keep the common case well under, so this is a real but narrow gap; #972 covers it.
- The comment above `structuredCache` says it caches "the serialised structured payload (JSON bytes)"; it caches the object.

**Verification:** `pnpm check` passes (0 errors across backend/web/api-spec; only pre-existing lint warnings). The new/changed unit tests pass — `article-export.test.ts`, `article.test.ts`, `feed-series.test.ts`, `timeline-enrich.test.ts` (52 passed) and web `markdown.test.ts`, `timeline-structured.test.ts` (13 passed). I could not run the testcontainers integration tests here (the sandbox can't pull the postgis image), so `feed-structured.integration.test.ts` and `feed-router.integration.test.ts` are unverified locally — CI covers them. `pnpm generate` is unaffected: `openapi.ts` doesn't reference the feed schemas, so the new `z.preprocess`-based `feedStructuredPostSchema` can't break OpenAPI/Kotlin generation.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -115,7 +125,8 @@ export const createFeedPublicRouter = (): TypedRouter => {
       }
       const token = typeof req.query.token === 'string' ? req.query.token : undefined
       try {
-        const structured = await resolveStructuredPost(username, postId, token)
+        const key = `structured:${username}:${postId}:${token ?? ''}:${Math.floor(Date.now() / 3_600_000)}`
```
fiddur: The raw `token` in the cache key gives an unauthenticated caller an unbounded key space on a **public** post.

`isCapabilityAuthorized` short-circuits on `isPubliclyVisible(post.visibility)` and never looks at the token, so for a `public`/`unlisted` post *every* `?token=<anything>` resolves successfully — and each distinct value is a distinct key. `GET /public/fiddur/feed/<id>?token=1`, `?token=2`, … is a 100%-miss loop that re-runs the whole resolve (up to 100 blocks: `queryMetricsBucketed` per chart + `getContinuousCorrelation` per correlation) on every request, and each result is inserted into the shared 200-entry LRU, evicting the legitimately cached block images too.

Since the token only affects the *outcome* for `followers`-only posts, it shouldn't be in the key at all — see the note on the next line; authorizing before the cache lets the key be `structured:${username}:${postId}:${post.updated_at.getTime()}`.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -115,7 +125,8 @@ export const createFeedPublicRouter = (): TypedRouter => {
...
+        const structured = await structuredCache(key, () => resolveStructuredPost(username, postId, token))
```
fiddur: Authorization runs *inside* `produce`, so a cached payload outlives revocation.

The sibling image routes deliberately do the opposite — `resolveImageWindow` / `resolveArticleBlock` authorize *before* `cached(...)` is consulted, which is what `createRenderCache`'s own doc comment promises ("The caller checks eligibility BEFORE consulting this, so an unshared/visibility-changed post 404s and never reaches the cache"). Here the check is inside the callback, so once key `structured:<user>:<id>::<hour>` is populated by an anonymous hit, flipping the post to `followers` or deleting it doesn't stop the payload being served to anonymous callers until the hour rolls over. That directly contradicts the `no-store` two lines below, which is set because "sharing is revocable … and must take effect immediately".

Suggest splitting `resolveStructuredPost` into a load+authorize step and a resolve step: fetch the post, run `isCapabilityAuthorized(post, token)`, 404 on failure, and only then `structuredCache(key, () => resolveStructuredContent(user, post))` keyed on `updated_at` instead of the hour bucket. That fixes revocation, drops the token from the key, and removes the ≤1h edit staleness in one go.

apps/web/src/utils/markdown.ts
```diff
@@ -19,3 +19,54 @@ marked.setOptions({ breaks: true, gfm: true })
...
+  'h3',
+  'h4',
+  'table',
+  'thead',
+  'tbody',
+  'tr',
+  'th',
+  'td',
+]
+const REMOTE_ALLOWED_ATTR = ['href', 'rel', 'target', 'class', 'translate', 'start']
```
fiddur, line 62: Minor: the comment above claims this is "the same strict tag/attr set the backend's inbound `sanitizeRemoteHtml` uses", but the two differ in a security-relevant way. `sanitizeRemoteHtml` forces `rel="nofollow noopener noreferrer" target="_blank"` on every `<a>` via `transformTags`; here `rel`/`target` are merely *allowed*, and `marked` emits `<a href=…>` with neither — so a remote peer's article prose produces links that leak a `Referer` to the target and carry no `nofollow`, unlike the same peer's HTML content on the same timeline card. Allowing `target` also lets raw `<a target="_blank">` through unaccompanied by `noopener` (modern browsers imply it, so low impact).

The tag lists also aren't identical — this one adds `table`/`thead`/`tbody`/`tr`/`th`/`td`, which is fine and intentional (GFM tables), but worth saying so rather than claiming parity.

Easiest fix: post-process links (or use a `marked` renderer) to stamp `rel="nofollow noopener noreferrer"`, and reword the comment to "a stricter superset of the backend's inbound allowlist".

apps/backend/src/routes/feed-public-router.ts
fiddur: Fixed by authorizing before the cache. Split `resolveStructuredPost` into `loadAuthorizedStructuredPost` (load + `isCapabilityAuthorized`) and `resolveStructuredContent`; the route calls the first, 404s on failure, then keys the cache `structured:${username}:${postId}:${post.updated_at.getTime()}` — no token in the key, so a `?token=1,2,3…` walk against a public post can no longer force a full-miss loop or churn the LRU.

apps/backend/src/routes/feed-public-router.ts
fiddur: Done exactly as suggested. Authorization now runs before `structuredCache(...)` (matching `resolveImageWindow`/`resolveArticleBlock`), keyed on `updated_at` instead of the hour bucket — so flipping a post to `followers` or deleting it stops serving the payload to anonymous callers immediately (honouring the `no-store` two lines down), and the ≤1h edit staleness is gone. `resolveStructuredPost` stays as a thin load-then-resolve wrapper so the existing resolver tests are unchanged.

apps/web/src/utils/markdown.ts
fiddur: Good catch on the parity gap. Added `hardenRemoteLink` — an `afterSanitizeAttributes` hook that stamps `rel="nofollow noopener noreferrer"` + `target="_blank"` on every `<a>`, matching `sanitizeRemoteHtml`'s `transformTags`. The hook is added/removed around the one synchronous `DOMPurify.sanitize` (no `await` between) so it never leaks onto `renderMarkdown` (own content) — a test asserts both the stamping and that isolation. Also reworded the comment to "a stricter superset" (it adds the GFM table tags) rather than claiming exact parity.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -37,6 +39,16 @@ const PROFILE_FEED_LIMIT = 50
...
+  // Caches the resolved structured payload OBJECT for the unauthenticated
+  // `/feed/:postId` endpoint, which resolves up to 100 blocks per request — the
+  // same bounded LRU + in-flight de-dup the block images use. The route authorizes
+  // the post BEFORE consulting this cache (like the sibling image routes), so the
+  // capability token is NOT part of the key: a `followers`-only payload never
+  // reaches the cache via a public request, and an anonymous `?token=…` walk can't
+  // inflate the key space. Keyed on `updated_at` (busts on an edit, no ≤1h edit
+  // staleness) — a smaller cap than the image LRU because a payload can be large
+  // (a per-block sample cap is #972). Producing `null` (no content) isn't cached.
+  const structuredCache = createRenderCache<FeedStructuredPost>(50)
```
fiddur, line 56: The key here is `…:${post.updated_at.getTime()}` only, but the sibling block-image cache — which resolves the *same* live article data — deliberately adds a coarse hourly bucket:

```ts
`${kind}:${username}:${postId}:${index}:${updatedAt.getTime()}:${Math.floor(Date.now() / 3_600_000)}`
```

with the rationale “a locked window that later gains backfilled data can't serve a stale render for the whole process lifetime — the block re-resolves its data live (#934) … Articles, unlike shared activities, are mutable and their windows can gain data after publish.”

The comment above justifies dropping it as “no ≤1h edit staleness”, but the hourly bucket in the image key isn't about *edits* — `updated_at` already covers those. It's about *backfilled measurements inside an already-locked window*, which `updated_at` does **not** cover. Without it, a chart block's samples (and a correlation block's n/scatter) stay frozen for the whole process lifetime, i.e. the structured endpoint quietly becomes a snapshot while its own PNG stays live — the opposite of #934's “lock the window, not the data”.

In the normal federation flow this rarely bites (a re-fetch follows an `Update`, which bumps `updated_at`), but any direct GET of this public endpoint will diverge from the block image for the same article. Appending the same `Math.floor(Date.now() / 3_600_000)` term would make the two endpoints agree and costs nothing.

fiddur (commented):
✅ Approved

Solid, well-documented slice. I traced the whole chain — schema union → resolver → public route → enrichment → web render → markdown export — and found no blocking correctness or security defect.

Things I specifically verified:

- **Authorization ordering.** `loadAuthorizedStructuredPost` runs before `structuredCache`, so the capability token stays out of the cache key, a `followers`-only payload can't be reached by an anonymous request, and a flipped/deleted post 404s immediately. `isCapabilityAuthorized` is the same constant-time gate the image routes use.
- **Bucket handling.** `articleChartBlockSchema` allows `^\d+[smhd]$` and `floorSeriesBucket` now matches the same alphabet, so `1d`/`1h` pass through and only sub-5s floors — the round-3 regression is genuinely fixed and covered by the `keeps a 1d chart bucket as authored` integration test. `seriesBucketSchema` is still `[smh]`, so the public `/series` endpoint is unaffected by the widened regex.
- **Rolling-upgrade shim.** The `z.preprocess` `kind`-tagging is correct for the ingest path (`feedPostStructuredResponseSchema.parse`), and both untagged-payload readers (`structuredChartSeries`, `HomeTimeline`'s `!== 'article'` guard) are written so legacy DB rows without `kind` still render the native chart.
- **Remote markdown boundary.** `renderRemoteMarkdown`'s allowlist is a real superset of the backend's inbound `sanitizeRemoteHtml` (plus GFM tables) with no media tags at all — `img`/`svg`/`video`/`style` are absent from `ALLOWED_TAGS` rather than denylisted, so the tracking-pixel vectors are closed at the root. DOMPurify's default URI regexp still blocks `javascript:` since `ALLOWED_ATTR` doesn't touch it, and the `addHook`/`removeHook` pair is genuinely scoped to one synchronous sanitise (the test for hook leakage into `renderMarkdown` is the right test).
- **Parity.** The structured chart resolver buckets in `device_timezone` exactly like `getArticleChartSeries` in `rest-routes.ts`, and both the peer render and the author's own `ArticleChartBlock` plot `avg`, so the native render really does match.
- **Export gating.** REST and MCP both refuse a non-public article before ever building a URL, so `image_token` can't reach publicly-pasted text; `articleBlockImageUrl` is now the single source shared with the AS2 attachment builder, which removes the previous divergence risk.
- No circular imports introduced (`services/article.ts` → `activitypub/object.ts` is a leaf; `activitypub/article-object.ts` → `services/article.ts` doesn't close a cycle), no stale `FeedStructured` references left, generated api-spec artifacts are gitignored so nothing is missing from the diff, and the removed imports in `article-object.ts` are genuinely unused.

Four non-blocking notes inline. The most substantive is the structured cache key omitting the hourly bucket that the sibling block-image cache deliberately carries for the same live-resolved article data.

One broader observation for the summary: the structured payload still has no sample cap (tracked in #972), and the receiving side's `safeFetchGet` caps at 1 MB. A single one-year `1h` chart block is already ~900 KB of JSON, so a moderately large article will silently fail enrichment on the peer and fall back to the federated HTML + PNGs. That's graceful (it's the pre-PR behaviour), but it does mean the headline feature quietly doesn't apply to the larger articles it's most useful for — worth folding into #972's scope.

apps/backend/src/services/feed-structured.ts
```diff
@@ -60,20 +83,171 @@ const resolveStructuredSeries = async (
...
+  // — an author-stored `1s` bucket over a wide window would otherwise yield
+  // ~600k buckets on this unauthenticated endpoint. `floorSeriesBucket` now
+  // understands `d`, so the common `1d`/`1h` buckets pass through unchanged. A
+  // fuller sample cap + cache is tracked in #972 (with the block image's #969).
+  const authored = block.bucket ?? defaultArticleChartBucket(startDate, endDate)
+  if (isZeroDurationBucket(authored)) return null
+  const bucket = floorSeriesBucket(authored)
+
+  const result = await queryMetricsBucketed(user, [block.metric], startDate, endDate, bucket, { tz })
+  const samples = samplesFromBucketedResult(result, block.metric)
```
fiddur, line 113: Worth pinning a concrete number on the #972 follow-up: the receiver's `safeFetchGet` has `MAX_RESPONSE_BYTES = 1_000_000`, and one `publicSeriesSampleSchema` entry serialises to roughly 105 bytes. So a single chart block at `1h` over a year (~8 760 samples) is already ~900 KB, and a `5s` block over a one-day window (~17 280 samples) is ~1.8 MB.

Past that point `enrichFromAurboda` swallows the failure and returns `null`, so the peer silently falls back to the federated HTML + attached PNGs — no error anywhere, the article just never renders natively. Since the emit side has no cap and the ingest side does, the two are mismatched by construction. A per-block sample cap (e.g. coarsen the bucket until `samples.length` fits a budget, the way `defaultArticleChartBucket` already coarsens by span) would keep large articles working rather than dropping them; worth folding into #972's scope.

apps/backend/src/services/article-export.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Pure and synchronous: the markdown only LINKS to the already-gated,
+ * on-demand image endpoint rather than embedding rendered data itself, so no
+ * chart/correlation resolution happens here.
+ */
+import type { ArticleContent, FeedVisibility } from '@aurboda/api-spec'
+
+import { articleBlockImageUrl, articleBlockLabel } from './article.ts'
+
+/** Collapse newlines (markdown alt/caption text is a single line) and escape `]` so it can't close the image syntax early. */
+const inlineText = (text: string): string => text.replaceAll(/\s+/g, ' ').trim().replaceAll(']', '\\]')
```
fiddur: `inlineText` escapes `]` but not `[`, so an unbalanced `[` still breaks the image syntax. `articleBlockLabel` falls back to `describeSelectorAxis`, which interpolates the selector's **regex** `pattern` — e.g. a tag selector `run[` yields `run[ (count) vs sleep_score`, where CommonMark matches the `]` against the inner `[` and leaves `![run` as literal text plus a stray link. A caption containing a lone `[` does the same.

One-character fix: `.replaceAll(/([[\]])/g, '\\$1')` instead of `.replaceAll(']', '\\]')` (and the existing `escapes a ] in the caption` test extends naturally to `[`).

apps/web/src/components/ArticleExportDialog.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useState } from 'preact/hooks'
+
+import { fetchArticleExport } from '../state/api'
+import './ArticleEditorDialog.css' // for `.article-input`
+import './ShareActivityDialog.css'
+
+export const ArticleExportDialog = ({ postId, onClose }: { postId: string; onClose: () => void }) => {
+  const { data, isLoading, isError, error } = useQuery({
+    queryFn: () => fetchArticleExport(postId),
+    queryKey: ['article-export', postId],
```
fiddur, line 17: This `useQuery` uses the default `retry: 3`, so the deterministic 400 from a `followers`-only article gets retried with exponential backoff before the message appears — the user stares at “Loading…” for several seconds before seeing the actionable “make it public or unlisted first”. Same for the 503 and the 404. Adding `retry: false` here makes the error surface immediately; none of the failure modes this route can produce are transient.

apps/backend/src/services/feed-structured.ts
fiddur: Agreed this belongs in #972 — I've added your concrete byte math there (≈105 B/sample → ~900 KB for a 1h/1-year block, ~1.8 MB for a 5s/1-day block vs the 1 MB `safeFetchGet` ceiling) plus the proposed emit-side per-block sample budget (coarsen the bucket until it fits, like `defaultArticleChartBucket`). Comment: https://github.com/fiddur/aurboda/issues/972#issuecomment-5102259857

apps/backend/src/routes/feed-public-router.ts
fiddur: Good distinction — you're right that the hourly term covers backfill *inside* a locked window, which `updated_at` doesn't. Restored it: the key is now `structured:${username}:${postId}:${updated_at}:${Math.floor(Date.now()/3_600_000)}`, the exact two-part key the block-image cache uses, so the two endpoints agree and neither freezes into a lifetime snapshot. Reworded the comment to say so.

apps/web/src/components/ArticleExportDialog.tsx
fiddur: Added `retry: false`. None of this route's failures (400 non-public, 404, 503) are transient, so retrying just delayed the actionable message behind exponential backoff — it now surfaces immediately.

apps/backend/src/services/article-export.ts
fiddur: Fixed exactly as suggested: `inlineText` now does `.replaceAll(/([[\]])/g, '\\$1')`, escaping both brackets. Extended the existing caption test to expect `![HR \[bpm\]]` and added a lone-`[` case (`run[ start` → `![run\[ start]`) so the selector-pattern path is covered too.

apps/web/src/pages/Feed/HomeTimeline.tsx
```diff
@@ -73,11 +73,16 @@ function TimelineCard({ entry }: { entry: TimelineEntry }) {
         </div>
       </header>
 
-      {/* Sanitised server-side on ingest (timeline-ingest.ts) — safe to render. */}
-      <div class="feed-post-content" dangerouslySetInnerHTML={{ __html: entry.content }} />
+      {/* Sanitised server-side on ingest (timeline-ingest.ts) — safe to render.
+          Suppressed for a native article render (below), whose `content` is the
+          same title + prose + captions — else the article shows twice. */}
+      {entry.structured?.kind !== 'article' && (
```
fiddur, line 79: Suppressing `content` makes the card fully dependent on `structured` being as fresh as `content` — but the ingest upsert deliberately breaks that coupling:

```sql
structured = COALESCE(EXCLUDED.structured, timeline_entry.structured),
```

(`apps/backend/src/db/timeline.ts`) keeps the last-known payload when re-enrichment fails, while `content` is overwritten unconditionally. So for an edited article whose `Update{Note}` arrives but whose enrichment fetch times out (12s bound, failures swallowed), the reader now sees the **stale** title + prose and never the fresh `content` — previously the stale `structured` only affected the chart and the fresh prose still showed.

Narrow, and only until the next successful redelivery, but it's a behaviour change introduced here. Cheapest fix is probably to let the article branch of the upsert overwrite rather than `COALESCE` (an article's `structured` is the whole body, not an additive enhancement), or to stamp the payload with the object's `published_at`/edit marker and fall back to `content` when it doesn't match. Fine to defer to a follow-up issue if you'd rather not widen this PR.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -37,6 +39,21 @@ const PROFILE_FEED_LIMIT = 50
...
+  // reaches the cache via a public request, and an anonymous `?token=…` walk can't
+  // inflate the key space. Keyed on `updated_at` AND a coarse hourly bucket — the
+  // SAME two-part key the block-image cache uses for the same live-resolved article
+  // data: `updated_at` busts on an edit, and the hourly term bounds staleness from
+  // *backfilled measurements inside an already-locked window* (which don't touch
+  // `updated_at`) to ≤1h, so this endpoint can't freeze into a lifetime snapshot
+  // while its own PNG stays live (#934 locks the window, not the data). A smaller
+  // cap than the image LRU because a payload can be large (per-block sample cap is
+  // #972). Producing `null` (no content) isn't cached.
+  const structuredCache = createRenderCache<FeedStructuredPost>(50)
```
fiddur, line 56: Entry count is bounded but payload size isn't — a single chart block floored to `5s` over a multi-month window is on the order of 10^5 samples, and `articleContentSchema` allows 100 blocks. 50 such entries pinned for an hour is a much larger resident set than the 200-buffer image LRU it's modelled on. Since only the instance owner authors articles this isn't attacker-reachable, but it does mean #972's per-block sample cap is now load-bearing for memory, not just response size. Worth a `log`/metric on payload bytes, or dropping the cap to something like 20, until that lands.

fiddur (commented):
✅ Approved

Solid slice. I traced the three new surfaces end to end — structured article enrichment, the markdown export, and the native `TimelineArticle` render — and found no blocking correctness or security defect.

**What I verified**

- **Authorization ordering.** `loadAuthorizedStructuredPost` runs `isCapabilityAuthorized` *before* the new `structuredCache` is consulted, and the token is not in the cache key. A public request for a `followers`-only post 404s and never reaches the cache; flipping visibility or deleting the post drops it immediately. The old `activity_id == null` short-circuit is preserved (moved into `resolveStructuredContent`), so activity behaviour is unchanged.
- **Shared URL/label builders.** `articleBlockImageUrl` reproduces the exact query string the old inline builder in `article-object.ts` produced (`v` always, `token` only when `!isPubliclyVisible`), and `articleBlockLabel` matches the old `attachmentName`. No cycle is introduced: `article.ts` → `activitypub/object.ts` is a leaf (types only).
- **`floorSeriesBucket` accepting `d`.** `seriesBucketSchema` still restricts the public `/series` endpoint to `s|m|h`, so the public-series path is behaviourally unchanged; only article chart blocks (regex `^\d+[smhd]$`) can reach the new `d` branch, and `1w`/`1M` are unrepresentable there, so there is no silent fall-through to `5s` for a coarse bucket.
- **Correlation parity.** `periodStart/periodEnd = start.slice(0, 10)` matches `resolveArticleBlock`'s raw-ISO day slice, and the chart path buckets in `device_timezone` exactly like `getArticleChartSeries` in `rest-routes.ts`, so the structured payload and the block PNG agree.
- **`renderRemoteMarkdown`.** The hook is added and removed within one synchronous call, so it can't leak onto `renderMarkdown`; `ALLOWED_TAGS` carries no media element at all (the tests cover `style`/`video poster`/`svg image`), and DOMPurify's URI scheme check still applies to the allowlisted `href`. The rolling-upgrade `preprocess` shim (kind-less → `activity`) is correct, and the web gates on `kind === 'article'` rather than `!== 'activity'`, so legacy `timeline_entry.structured` rows keep rendering their chart.

**Non-blocking notes**

1. **Structured payload size vs. the new cache.** A chart block floored to `5s` over a wide window still yields ~120k samples, and an article may hold 100 blocks. That was transient before; the 50-entry LRU now pins up to 50 such payloads in memory with no byte accounting. Since only the instance owner can author articles this is self-inflicted rather than attacker-controlled, but it's worth making sure #972 lands the per-block sample cap before an article with a long window and a fine bucket exists in the wild.
2. **Correlation `series` below n = 3.** The block-image endpoint 404s at `n < 3` and `TimelineArticleBlock` hides the scatter, but the structured JSON still ships the aligned daily points for n = 1–2. That's consistent with the author having published the block over that window, so it isn't a leak, but the author's own view ("not enough overlapping data") doesn't hint that the underlying values are in the public payload. Worth a line in `docs/features/feed.md` if you want the two surfaces to read the same.
3. **Docs table cell** — `| \`GET /feed/articles/:postId/export\`|` in `docs/features/feed.md` is missing the space before the closing pipe (renders fine, just inconsistent with its neighbours).

apps/web/src/pages/Feed/HomeTimeline.tsx
fiddur: Deferred to #973 — narrow edge case (edit + failed re-enrichment within the 12s budget), self-correcting on next successful redelivery, not attacker-reachable. Fixing properly needs kind-aware upsert logic (activities still want the COALESCE-preserve) plus tests, so tracking separately rather than widening this PR.

apps/backend/src/routes/feed-public-router.ts
fiddur: Folded into #972 (added a comment there) — the cache-pinning angle is the same root cause as that issue's per-block sample cap, so the fix belongs together with it rather than as a separate cap here.

PR #976, opened 2026-08-03
fix(garmin): resolve versioned typeKeys and never FK-fail a sync batch

fiddur:
## Problem

Garmin sync reported one error per date:

```json
{ "date": "2026-08-03", "error": "insert or update on table \"activities\" violates foreign key constraint \"fk_activities_type\"" }
```

The raw payload showed the culprit:

```json
{ "activityName": "Gothenburg Rowing", "activityType": { "typeKey": "rowing_v2" } }
```

`rowing` and `rowing_machine` both exist as builtin definitions — Garmin has just moved to a versioned key. `processActivities` passed the raw `typeKey` straight through to `activities.activity_type`, which has a FK to `activity_type_definitions(name)`, so the insert threw.

Blast radius was larger than one activity:

- activities are fetched by offset/limit, not per day (`sync.ts:310-312`), so the same bad activity re-threw on **every** day of the sync range;
- `processActivities` had no per-activity isolation, so every activity behind it in the batch was skipped too.

## Fix

Activity type resolution (`resolveActivityType`):

1. override map on the raw `typeKey` (unchanged behaviour);
2. **strip Garmin's version suffix** — `rowing_v2` → `rowing`, `indoor_rowing_v2` → the existing `indoor_rowing` override → `rowing_machine`;
3. **fall back to `other_workout`** when the resolved name has no definition, keeping the original key in `data.garmin_type_key` so unmapped sports are discoverable rather than silently lost.

Results are cached per batch, so it costs one `activityTypeExists` lookup per distinct typeKey.

Each activity is also processed in its own `try`/`catch` with the failure audit-logged, so one bad activity can no longer take the batch with it. This incidentally fixes `act.activityType?.typeKey ?? 'unknown'`, which was itself a type with no definition.

## Tests

- Unit (`process.test.ts`): version-suffix stripping, override-after-strip, fallback + `garmin_type_key`, resolved type passed to `deleteGarminActivityWithWrongType`, per-typeKey lookup caching, batch survives a failing activity.
- Integration (`process.integration.test.ts`, new): runs `processGarminData` against a real PostgreSQL container **with the FK enforced**, so it reproduces the original failure and proves the three paths land rows.
- Docs: new "Activity Type Mapping" section in `docs/garmin.md`.

`pnpm check` clean; 2253 backend unit tests pass.

## Follow-up

Nothing to clean up in the DB — the failed inserts never landed. A Garmin resync after deploy picks the activity up (the raw records were already stored).

### DIFF
--- a/apps/backend/src/integrations/garmin/process.integration.test.ts
+++ b/apps/backend/src/integrations/garmin/process.integration.test.ts
@@ -1,0 +1,128 @@
+/**
+ * Integration test for Garmin activity type resolution.
+ *
+ * Runs processGarminData against a real PostgreSQL instance so the
+ * activities → activity_type_definitions foreign key is actually enforced:
+ * an unmapped Garmin typeKey must degrade to a defined type instead of
+ * failing the insert and taking the rest of the batch with it.
+ */
+import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest'
+
+import { query } from '../../db/connection.ts'
+import {
+  activityTypeExists,
+  deleteGarminActivityWithWrongType,
+  insertActivity,
+  insertLocations,
+  insertRawRecord,
+  insertTimeSeries,
+} from '../../db/index.ts'
+import { softDeleteLocationRange } from '../../db/locations.ts'
+import { auditError } from '../../services/audit-log.ts'
+import { cleanTestDb, getTestUser, startTestDb, stopTestDb } from '../../test/db-test-helper.ts'
+import { processGarminData } from './process.ts'
+
+const CONTAINER_TIMEOUT = 120_000
+
+const realDeps = {
+  activityTypeExists,
+  auditError,
+  deleteGarminActivityWithWrongType,
+  insertActivity,
+  insertLocations,
+  insertRawRecord,
+  insertTimeSeries,
+  softDeleteLocationRange,
+}
+
+const makeActivity = (overrides: Record<string, unknown> = {}) => ({
+  activityId: 12345,
+  activityName: 'Morning Row',
+  activityType: { typeKey: 'rowing_v2' },
+  averageHR: 145,
+  beginTimestamp: 1736924400000,
+  calories: 350,
+  distance: 5000,
+  duration: 1800,
+  elapsedDuration: 1800,
+  elevationGain: 0,
+  maxHR: 175,
+  startTimeGMT: '2025-01-15T07:00:00.000',
+  steps: 0,
+  vO2MaxValue: 0,
+  ...overrides,
+})
+
+const getStoredActivities = async (user: string) => {
+  const result = await query(
+    user,
+    `SELECT activity_type, title, data FROM activities WHERE source = 'garmin' ORDER BY start_time, title`,
+    [],
+  )
+  return result.rows as { activity_type: string; data: Record<string, unknown>; title: string }[]
+}
+
+describe('Garmin activity type resolution (integration)', () => {
+  beforeAll(async () => {
+    await startTestDb()
+  }, CONTAINER_TIMEOUT)
+
+  afterAll(async () => {
+    await stopTestDb()
+  })
+
+  beforeEach(async () => {
+    await cleanTestDb()
+  })
+
+  test('stores a versioned typeKey under its unversioned activity type', async () => {
+    const user = getTestUser()
+
+    const count = await processGarminData(user, 'activities', [makeActivity()], realDeps)
+
+    expect(count).toBe(1)
+    const stored = await getStoredActivities(user)
+    expect(stored).toHaveLength(1)
+    expect(stored[0].activity_type).toBe('rowing')
+    expect(stored[0].data.garmin_type_key).toBeUndefined()
+  })
+
+  test('stores an unknown typeKey as other_workout instead of violating the foreign key', async () => {
+    const user = getTestUser()
+
+    const count = await processGarminData(
+      user,
+      'activities',
+      [makeActivity({ activityName: 'Kite Day', activityType: { typeKey: 'kitesurfing' } })],
+      realDeps,
+    )
+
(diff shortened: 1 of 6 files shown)


apps/backend/src/integrations/garmin/process.integration.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+
+  test('a failing activity does not stop the rest of the batch', async () => {
+    const user = getTestUser()
+
+    // title is VARCHAR(255) — an over-long one fails at the DB, standing in for
+    // any per-activity error Garmin might hand us.
+    const data = [
+      makeActivity({ activityId: 1, activityName: 'x'.repeat(300) }),
+      makeActivity({ activityId: 2, activityName: 'Evening Row' }),
```
fiddur, line 114: Both fixtures inherit the same `startTimeGMT` and both resolve to `rowing`, so they collide on the `idx_activities_type_time` unique index (`source, activity_type, start_time` — Garmin activities have no `external_id`). That means `toHaveLength(1)` and `title === 'Evening Row'` would pass even if the first activity had *succeeded*: the second would just upsert over it. The claim is still carried by `count === 1` plus the audit row, but giving the second activity a distinct `startTimeGMT` would make the "the rest of the batch is stored" assertion actually distinguishing.

apps/backend/src/integrations/garmin/process.integration.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const count = await processGarminData(user, 'activities', data, realDeps)
+
+    expect(count).toBe(1)
+    const stored = await getStoredActivities(user)
+    expect(stored).toHaveLength(1)
+    expect(stored[0].title).toBe('Evening Row')
+
+    const audit = await query(user, `SELECT message, details FROM audit_log WHERE level = 'error'`, [])
+    expect(audit.rows).toHaveLength(1)
+    expect(audit.rows[0].message).toContain('1')
```
fiddur, line 126: `toContain('1')` passes on any message containing the digit 1 — including `Failed to process Garmin activity 12345`, the id every other fixture uses. Assert the identifying substring instead (`toContain('activity 1')` or the full message), and consider asserting `details.error` too since it's already selected but unused.

Also worth knowing: `cleanTestDb` doesn't truncate `audit_log`, so `toHaveLength(1)` silently depends on no earlier test in this file writing an error entry. Scoping the query (e.g. `WHERE level = 'error' AND message LIKE '%Garmin activity 1'`) would make it order-independent.

apps/backend/src/integrations/garmin/process.ts
```diff
@@ -99,6 +105,47 @@ const garminTypeKeyOverrides: Record<string, string> = {
...
+  cache: Map<string, ResolvedActivityType>,
+): Promise<ResolvedActivityType> => {
+  const cached = cache.get(typeKey)
+  if (cached) return cached
+
+  const unversioned = stripVersionSuffix(typeKey)
+  const mapped = garminTypeKeyOverrides[typeKey] ?? garminTypeKeyOverrides[unversioned] ?? unversioned
+  const resolved: ResolvedActivityType = (await deps.activityTypeExists(user, mapped))
+    ? { activity_type: mapped }
+    : { activity_type: fallbackActivityType, unmapped_key: typeKey }
```
fiddur, line 143: The degrade is silent — no log, and the sync now reports success — so an unmapped sport is only discoverable by grepping `data.garmin_type_key` after the fact, as the doc describes. An `auditWarn` here (once per distinct key, which the cache already gives you for free) would surface new/unmapped Garmin keys without the manual sweep. Non-blocking.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. The three pieces — strip the `_vN` suffix, resolve against `activity_type_definitions` with a `other_workout` fallback, and isolate each activity in a try/catch — all hold up, and the extraction of `processActivity` keeps the loop readable without changing the existing ordering (raw record → resolve → delete-wrong-type → insert → vo2max series).

Things I checked that could have broken but don't:

- **The fallback target always exists.** `other_workout` is in the built-in seed (`apps/backend/src/schema/activities.ts`), so the degraded path can't itself trip the FK.
- **A failed insert can't poison the rest of the loop.** `db/connection.ts#query` runs each statement standalone (no surrounding transaction), so the `INSERT` that violates a constraint doesn't leave the connection in an aborted-transaction state for the next activity. The integration test exercises this against real Postgres, which is the right call.
- **The fallback doesn't create timeline duplicates.** Cross-source merge (`db/activities/merge.ts#eligibleAnchor`/`isCrossMergePair`) pairs on `display_category`, not on the exact type name, so a Garmin `other_workout` still merges with e.g. a Strava `biking` inside the 120 s window. (Note `isGenericExercise` only treats `'exercise'` as generic despite its comment mentioning `other_workout` — pre-existing, and not reachable here because the cross-source pass already collapsed the pair.)
- **`deps.auditError(...)` is not awaited**, but that matches the established convention (every audit call in the backend is fire-and-forget, and `auditLog` swallows its own errors), so no unhandled rejection. Ordering still holds in the new integration test because the pg client serializes queries and the audit insert is enqueued before the second activity's statements.

One non-blocking design thought: the fallback is now *silent*. Before this change an unmapped sport blew up the day and was at least visible in `sync_state.error_message`; now the sync reports success and the only trace is `data.garmin_type_key`, which the doc says to go grep manually. Several very common Garmin keys have no seeded definition and no override — `cycling`, `mountain_biking`/`road_biking`, `indoor_cycling`, `trail_running`, `lap_swimming` — so they'll quietly land in `other_workout` even though good targets exist (`biking`, `biking_stationary`, `running`, `swimming_pool`). Worth either extending `garminTypeKeyOverrides` with those now, or emitting an `auditWarn` the first time a key falls back per batch so unmapped sports surface on their own.

Minor test-quality notes inline. Nothing blocking.

PR #977, opened 2026-08-03
fix(locations): activity GPS takes precedence over phone tracking

fiddur:
## Problem

The activity-detail map interleaved the Garmin watch track with much coarser phone positions, e.g. on `/detail/activity/merged:6f969c7b-…`. Three separate causes, all in play at once:

1. **The read path never filtered soft-deletes.** `getRawLocationPoints` — the query behind `ActivityMap` — was the only read of `locations` without `AND deleted_at IS NULL`. Points the sync had *already* soft-deleted rendered anyway, so the existing precedence logic was invisible on the map.
2. **The delete range was too narrow.** It covered `gpsPoints[0].time`..`gpsPoints[last].time`. The stored track is downsampled to 60-second intervals, so its first and last fixes sit *inside* the activity and phone points survived at both ends.
3. **Only `owntracks` was cleared**, hardcoded by name.

## Fix

When an activity brings its own GPS, `softDeleteOtherSourceLocations(user, keepSource, start, end)` clears every *other* source across the activity's **full span** — the union of the activity span and the track itself, via a new shared `gpsPrecedenceSpan` helper.

The span is threaded in from the callers that already had the activity row, so no extra lookup: `syncActivityDetails` (has `start_time`/`end_time`) and the `resync-detail` route (uses the row that supplied the `garmin_activity_id`, which for a merged activity is the Garmin source, not the merge). Strava gets the same treatment from `start_date` + `elapsed_time`.

**Nothing is hard-deleted**, and `insertLocations` now revives a source's own soft-deleted rows on conflict (`DO UPDATE SET deleted_at = NULL WHERE locations.deleted_at IS NOT NULL`). Without that, a session held by *both* Garmin and Strava could end up with an empty span: each soft-deletes the other's points, and the old `DO NOTHING` meant a re-sync could never bring its own track back. Now whichever synced last wins and re-syncing the other switches it back — covered by an integration test.

## Also, from the #976 review

- **`auditWarn` when a Garmin typeKey degrades to `other_workout`** — once per distinct key, which the resolution cache already gives for free. New Garmin sports now surface in the audit log instead of only in `data.garmin_type_key`.
- **`auditInfo` recording how many foreign points an activity track replaced**, so the deletion isn't silent either.
- The two test nits: the batch-survival fixtures shared a `startTimeGMT` and collided on `idx_activities_type_time` (so the assertion couldn't distinguish a survived batch from an overwritten one), and `toContain('1')` matched any message containing the digit. Both tightened, and the audit query is now scoped since `cleanTestDb` doesn't truncate `audit_log`.

## Tests

- `gps-precedence.test.ts` (new): span widening, track points outside the span, no-span fallback, single point, unordered points.
- `locations-gps-precedence.integration.test.ts` (new, real Postgres): soft-delete scoping and count, range bounds, idempotency, the Garmin↔Strava revive case, and **`getRawLocationPoints` excluding soft-deleted points** — the regression test for cause #1.
- Garmin unit tests: full-span precedence, the replaced-points audit line, warn-once-per-key, no warning for mapped keys.
- Strava unit test now asserts the full 1-hour span rather than just that a delete happened.

`pnpm check` clean; 2265 backend unit tests plus the affected integration suites pass.

## Backfilling existing activities

The `deleted_at` filter immediately hides everything previously soft-deleted. Points that survived at the span edges need a detail re-sync to clear: `sync_garmin({ full_resync: true })`, or the per-activity resync-detail action. Note `getActivitiesNeedingDetail` caps at 10 rows ordered by `start_time DESC` even with `forceAll`, so a full resync only revisits the 10 most recent activities — older ones need the per-activity action. Pre-existing limit, not touched here.


fiddur (commented):
🛑 Changes required

Solid, well-documented change: the read-path fix in `getRawLocationPoints` (the only `FROM locations` query that still leaked soft-deleted rows — I checked the others, they all filter), the extracted `gpsPrecedenceSpan` helper with a real unit test, the new integration test covering delete/revive/read, and the docs updates across `data-sources.md` / `garmin.md` / `owntracks.md` / `strava.md` all hang together. The unmapped-typeKey warning is a nice addition and is warned once per distinct key via the existing batch cache.

One thing I'd want fixed before merge: switching `insertLocations` from `ON CONFLICT ... DO NOTHING` to `DO UPDATE` silently drops Postgres' tolerance for duplicate conflict keys *within one statement*. Details inline — the Garmin path keeps every GPS sample (`GPS_DOWNSAMPLE_MS = 0`), so two samples sharing a `directTimestamp` (or a repeated `timestampGMT` in the `geoPolylineDTO` fallback) now abort the whole location insert with SQLSTATE 21000 instead of being ignored. A dedupe by `(source, time)` before formatting the VALUES list restores the old guarantee for every caller.

Two design points worth a decision rather than a fix (also inline): `source <> $1` makes Garmin and Strava delete *each other's* tracks, and since Strava's track is downsampled to 60 s while Garmin's keeps every sample, "most recent sync wins" can permanently demote a full-resolution track to a coarse one (Garmin won't re-fetch detail after `markActivityDetailSynced`, and Strava's incremental `after` won't revisit the activity). And `gpsPrecedenceSpan` takes an unclamped min/max over track times, so one bogus timestamp in a detail response widens the soft-delete window arbitrarily.

Smaller notes, no action needed if you disagree:
- The two places that now *derive* the span (`sync.ts:242` and `activities-router.ts:618` → `rest-routes.ts:147`) have no test; `gpsPrecedenceSpan` and `processActivityDetail` are covered, but nothing asserts that the resync route passes the Garmin *source* activity's span (which is the subtle part — it picks `garminSource`, not the requested activity).
- `docs/data-storage.md` still shows the `locations` DDL without the `deleted_at` column. Pre-existing, but soft-delete is now central to how locations behave, so it'd be worth adding there too.
- Note the asymmetry (which I think is correct, just undocumented): OwnTracks writes via `insertLocation`, which kept `DO NOTHING`, so a re-posted phone point does *not* revive. The docs sentence "re-inserting a source's own points revives them" only holds for the batch path.

apps/backend/src/db/locations.ts
```diff
@@ -70,25 +70,37 @@ export const insertLocations = async (user: string, locations: Location[]): Prom
               v.accuracy::double precision, v.altitude::double precision,
               v.velocity::double precision, v.regions::text[]
        FROM (VALUES %L) AS v(source, time, lon, lat, accuracy, altitude, velocity, regions)
-       ON CONFLICT (source, time) DO NOTHING`,
+       ON CONFLICT (source, time) DO UPDATE SET deleted_at = NULL
```
fiddur: `DO UPDATE` makes this batch insert intolerant of duplicate `(source, time)` rows *inside the same statement*, which `DO NOTHING` silently absorbed. If two proposed rows share a conflict key and the row is new, Postgres raises `ON CONFLICT DO UPDATE command cannot affect row a second time` (SQLSTATE 21000) and the whole insert fails.

That's reachable from the Garmin path: `GPS_DOWNSAMPLE_MS = 0` in `garmin/process.ts`, so `ts - lastGpsTime >= 0` never filters anything and every metrics row with lat/lon becomes a point — two samples carrying the same `directTimestamp` (multisport transitions, >1 sample/s devices) or a repeated `timestampGMT` in the `extractPolylineGps` fallback is enough. The failure mode is a lost GPS track for the activity (swallowed by the `try/catch` in `syncActivityDetails`) or a 500 on `POST /activities/:id/resync-detail`. Neither the fixture (120 distinct timestamps) nor the new tests exercise a duplicate, so CI won't catch it.

Cheapest fix is to dedupe in this function so every caller benefits, e.g. build `values` from a `Map` keyed by `` `${source}|${time.getTime()}` ``.

apps/backend/src/db/locations.ts
```diff
@@ -70,25 +70,37 @@ export const insertLocations = async (user: string, locations: Location[]): Prom
...
   end: Date,
-): Promise<void> => {
-  await query(
+): Promise<number> => {
+  const result = await query(
     user,
     `UPDATE locations SET deleted_at = NOW()
-     WHERE source = $1 AND time >= $2 AND time <= $3 AND deleted_at IS NULL`,
-    [source, start, end],
+     WHERE source <> $1 AND time >= $2 AND time <= $3 AND deleted_at IS NULL`,
```
fiddur: `source <> $1` means Garmin and Strava now soft-delete each other, not just OwnTracks. For a user whose Strava mirrors their Garmin activities that's a quality regression, not just a swap: Strava's track is downsampled to 60 s (`GPS_DOWNSAMPLE_MS = 60_000`) while Garmin keeps every sample, so if the Strava queue processes the mirrored activity after Garmin's detail sync, a full-resolution track is replaced by a coarse one — and it stays that way, since `markActivityDetailSynced` stops Garmin from re-fetching the detail and Strava's incremental `after` stops it from revisiting the activity. A Garmin `fullResync` has the mirror-image effect on every Strava track.

The docs cover "most recent sync wins", but not that it can cost resolution. If that's acceptable, fine as documented; otherwise consider only clearing passive-tracking sources (i.e. an explicit "phone sources" list) rather than everything that isn't `keepSource`.

apps/backend/src/integrations/gps-precedence.ts
```diff
@@ -0,0 +1,30 @@
...
+ * when the span is unknown.
+ */
+export const gpsPrecedenceSpan = (
+  gpsPoints: { time: Date }[],
+  activitySpan?: ActivitySpan | null,
+): ActivitySpan => {
+  const times = gpsPoints.map((p) => p.time.getTime())
+  if (activitySpan) times.push(activitySpan.start.getTime(), activitySpan.end.getTime())
+
+  return { end: new Date(Math.max(...times)), start: new Date(Math.min(...times)) }
```
fiddur: The min/max is taken over the raw track times with no bound, so a single bogus timestamp in a detail response (`extractGpsPoint` only checks `ts > 0`) widens the soft-delete window to cover everything between that outlier and the activity — potentially wiping months of phone history in one `UPDATE`. Clamping the track contribution to the activity span plus a small tolerance (rather than an unbounded union) would keep the intended "track may overhang the span slightly" behaviour without that tail risk.

Two smaller things on the same line: `Math.max(...times)` spreads one argument per point, and since the Garmin path keeps every sample, an ultra-long activity (≳24 h at 1 Hz) can exceed V8's argument limit and throw. A `reduce` avoids both this and the outlier problem. Also, `gpsPrecedenceSpan([])` with no span yields `Invalid Date` for both ends — unreachable from today's callers (both guard on `gpsPoints.length > 0`), but the exported signature allows it and the `[]` test only covers the with-span case.

apps/backend/src/db/locations.ts
fiddur: Fixed by reverting the `DO UPDATE` rather than deduping around it.

You were right that it was reachable — `GPS_DOWNSAMPLE_MS = 0`, so nothing filters same-timestamp samples in either the metrics path or the `extractPolylineGps` fallback.

The `DO UPDATE` only existed to let a source revive its own track after another *activity* source had deleted it. Since activity sources no longer supersede each other (see the thread below), that revive is unreachable, so keeping it would have meant carrying a footgun for dead behaviour. Both insert paths are `DO NOTHING` again.

Pinned with an integration test that inserts two `garmin` points sharing a timestamp and asserts the batch lands.

apps/backend/src/db/locations.ts
fiddur: Agreed, and fixed rather than documented — losing resolution permanently is worse than the cosmetic problem the change was meant to solve.

`softDeleteSupersededLocations` now takes `keepSources: string[]`, and both callers pass `activityTrackSources` (`['garmin', 'strava']`) rather than just their own source. Passive tracking is superseded; activity tracks never supersede each other. Two tracks of the same route sit within GPS error of each other, so overlaying them is cosmetic — nothing like the phone-vs-watch divergence this PR is about.

Deliberately not a "phone sources" allowlist: enumerating passive sources is open-ended (a new phone integration would silently stop being superseded, which is the original bug), while the activity-track set is closed — it is exactly the integrations that write locations from an activity. Documented on the constant and in `data-sources.md`, and covered by an integration test asserting a mirrored Strava point survives a Garmin sync.

apps/backend/src/integrations/gps-precedence.ts
fiddur: All three fixed.

- **Outlier**: track times are now clamped to the activity span ±5 min (`TRACK_OVERHANG_TOLERANCE_MS`), keeping the intended slight overhang without the tail risk. Tests cover an outlier a year after and one a year before.
- **Spread limit**: replaced with a loop. Added a test with a 24 h 1 Hz track (86 400 points), which would have thrown with `Math.max(...times)`.
- **`Invalid Date`**: the return type is now `ActivitySpan | null`, null when there is nothing to derive a range from, and callers skip the soft-delete on null.

One knock-on your comment led me to: both callers derived the span as `end_time ?? start_time`, so an activity with no `end_time` produced a start-only span — which under the new clamp would have shrunk the range to ±5 min around the start and let phone points survive mid-activity. They now pass `null` in that case, falling back to the track's own range.

NaN from an unparseable Date fails both comparisons and is skipped, so a bad timestamp can no longer poison the range either — covered by a test.

apps/backend/src/integrations/gps-precedence.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Sources that record a GPS track for an activity, as opposed to tracking
+ * position continuously in the background.
+ *
+ * These never supersede each other. Their tracks describe the same route within
+ * GPS error, so overlaying two of them is cosmetic, while letting them delete
+ * each other is not: Strava downsamples to 60 s where Garmin keeps every sample,
+ * and neither integration revisits an activity once synced — so "most recent
+ * sync wins" would permanently demote a full-resolution track to a coarse one.
+ */
+export const activityTrackSources = ['garmin', 'strava']
```
fiddur, line 22: Consider `as const` (or typing as `readonly string[]`) here. The entire "tracks never supersede each other" invariant rests on this array's contents, and as a mutable module-level export any code could `push`/`splice` it — the project style rules also call out avoiding mutable module-level state. Freezing it makes the invariant structural rather than conventional.

apps/backend/src/integrations/strava/process.ts
```diff
@@ -175,11 +181,21 @@ const processGpsStream = async (
...
+      ? await deps.softDeleteSupersededLocations(user, activityTrackSources, span.start, span.end)
+      : 0
     await deps.insertLocations(user, gpsPoints)
+
+    if (span && replaced > 0) {
+      deps.auditInfo(
+        user,
+        'sync',
+        `🛰️ Strava GPS took precedence over ${replaced} location point(s) from other sources`,
+        { end: span.end.toISOString(), start: span.start.toISOString() },
```
fiddur, line 197: The Garmin counterpart includes `garmin_activity_id` in the audit details, but this entry has no way to trace back to which Strava activity triggered the supersede. Plumbing `activity.id` into `processGpsStream` (or logging from the caller) would make the audit entries symmetric and debuggable.

fiddur (commented):
✅ Approved

Solid, carefully reasoned change. The inversion from "delete owntracks in track range" to "delete everything except activity-track sources across the activity span" is the right fix for tracks superseding each other, and I verified the supporting details hold up: every caller of the renamed `softDeleteSupersededLocations` is updated (no stale references), all location read paths (`getLocations`, `getLocationPoints`, `getRawLocationPoints`, the regions query) now filter `deleted_at IS NULL`, and OwnTracks is currently the only passive source writing to `locations`, so the except-list semantics are safe today — with the `activityTrackSources` unit test acting as the tripwire for future sources. `gpsPrecedenceSpan` handles the nasty edges (outlier timestamps clamped to ±5 min, NaN dates skipped, loop instead of spread for long tracks, start-only spans rejected at the callers), and the merged-activity resync correctly uses the Garmin source row's span rather than the wrapper's. Docs are thorough and accurate.

1. **Late-arriving phone points can still interleave.** OwnTracks uploads are often delayed/batched; if the Garmin/Strava sync runs before the phone flushes, points posted *afterwards* with timestamps inside the activity span insert as new rows (ON CONFLICT only blocks previously-superseded exact timestamps) and stay live — and since neither integration revisits a synced activity, nothing re-supersedes them short of a manual resync-detail. Probably acceptable, but worth knowing the zig-zag can still reappear for recent activities.

2. **Soft-delete runs before the replacement insert.** If `insertLocations` throws, the span is left with no live track until the retry succeeds (Garmin retries since the activity isn't marked synced, so this is transient; recoverable via `deleted_at` regardless). Inserting first would close the window, but it's a minor robustness point.

Minor pre-existing note: `deps.auditError`/`auditWarn`/`auditInfo` are fire-and-forget, and `process.integration.test.ts` asserts the audit row immediately after `processGarminData` — there's a theoretical race there, though it predates this PR and in practice the batch's later awaits give the insert time to land.

apps/backend/src/db/locations.ts
fiddur: Garmin should definitely trump strava.

fiddur:
Correction to this description (tracked in #978): the merged code does **not** revive soft-deleted location rows on re-insert. `insertLocations` / `insertLocation` are `ON CONFLICT (source, time) DO NOTHING`, and the shipped integration test asserts the non-revival. The Garmin↔Strava empty-span scenario is prevented by `keepSources = activityTrackSources` (neither track source supersedes the other) rather than by revival. To restore a superseded track, clear `deleted_at` directly, as `docs/data-sources.md` says.

Issue #658, opened 2026-04-20
Timeline: pixel-zoom gating for hierarchy collapse (vs range-based)


fiddur:
## Background

Follow-up to #650. The v1 collapse gate is **range-based**: if the fetched range spans > 3 days, collapse. This trades continuous responsiveness for a much simpler implementation — the memoized activities pipeline only needs to react to the fetch window, not to live zoom gestures.

The original design called out a **pixel-based** gate: when `pixelsPerHour` drops below a threshold, collapse. That would let the collapse react smoothly to zoom gestures, but `pixelsPerHour` lives inside the D3 `draw` closure and isn't accessible in the memoized `useTimelineData` pipeline.

## Current state

```typescript
const shouldCollapseHierarchy = useMemo(() => {
  const spanDays = (fetchEnd.getTime() - fetchStart.getTime()) / (24 * 60 * 60 * 1000)
  return spanDays > 3
}, [fetchStart, fetchEnd])
```

## Desired state

A pixel-aware gate. Zooming in past the threshold (while keeping the same fetched range) should un-collapse; zooming out should re-collapse — without a data re-fetch.

## Design options

1. **Lift zoom state**: expose a quantized `pixelsPerHour` bucket from `useTimelineZoom` as React state. `useTimelineData` depends on it. Memos recompute only when crossing bucket boundaries.
2. **Collapse inside `draw`**: apply the collapse after packing, inside the D3 `draw` closure where `pixelsPerHour` is live. Risk: `draw` runs on every pan frame, so the collapse + re-pack needs to be cheap or memoized.
3. **Hybrid**: keep the range-based gate as the primary, but let `draw` further collapse / un-collapse cosmetically (e.g. by hiding subtypes) when `pixelsPerHour` crosses a finer threshold.

## Acceptance

- Zooming in on a weekly view reveals the subtypes (running / strength / yoga) without a data refetch.
- Zooming back out re-collapses smoothly.
- Pan doesn't trigger redundant collapse recomputations.

Issue #657, opened 2026-04-20
Timeline: tooltip enrichment for collapsed/merged bars


fiddur:
## Background

Follow-up to #650. When adjacent `running` + `strength_training` + `stretching` activities collapse into a single `exercise` bar, the rendered tooltip loses the information that the bar actually represents three distinct subtypes. Currently the merged bar shows up as if it were a single `exercise` activity.

## Current state

`collapseToParentType` produces a single `Activity` with the parent's `activity_type`, the first member's id and title, and the merged `start_time` / `end_time` span. No provenance is carried.

## Desired state

When a tooltip hovers over a collapsed bar, it should indicate what subtypes it merged, e.g.:
> **Exercise** (10:00 – 11:40)
> Merged: Running, Strength Training, Yoga

## Design options

1. **Store subtype breakdown in `source_records`**: requires conforming to the existing `SourceRecord` shape (id, source, start_time, end_time, title), which is doable but verbose.
2. **Add a `collapsed_from?: string[]` field** to the Activity surface specifically for this case: carries the original activity_type names.
3. **Keep a lighter breakdown**: `collapsed_types?: Array<{ type: string; count: number }>`: enough for the tooltip without carrying per-activity detail.

Option 3 is probably the right trade-off — tooltip only needs aggregate info.

## Acceptance

- Hovering a merged `exercise` bar shows the constituent subtypes.
- The breakdown is correct (order and counts match what got merged).
- Non-collapsed bars (single-subtype or non-hierarchical) render tooltips as today.

PR #979, opened 2026-08-03
fix(timeline): let zoom separate adjacent same-type activities

fiddur:
## Problem

Two yoga sessions on 2026-07-31 rendered as a single timeline bar spanning 21:11–22:07 at **every** zoom level, and the bar linked only to the first — leaving the second session (and its Garmin + Health Connect sources) unreachable:

| Session | Time | Duration | Garmin ID |
|---|---|---|---|
| 1 | 21:11:11 → 21:24:01 | 13 min | 23804369937 |
| 2 | 21:32:52 → 22:07:48 | 35 min | 23804689313 |

Gap: **8 min 51 s**.

The backend is fine — `query_activities` returns two separate merged activities, and `mergeOverlappingActivities` only merges on genuine overlap (`currentEnd >= nextStart`). This is entirely the client-side, zoom-aware merge, which is intended behaviour that just couldn't be zoomed out of:

```ts
const mergeGapMs = useMemo(() => {
  const days = differenceInCalendarDays(effectiveViewEnd, effectiveViewStart)
  if (days > 50) return 4 * 60 * 60 * 1000
  if (days > 2) return 60 * 60 * 1000
  return 10 * 60 * 1000        // ← floor
}, [effectiveViewStart, effectiveViewEnd])
```

`differenceInCalendarDays` is 0 for *any* view inside a single day, so zooming from 24 hours down to one minute never changed the threshold, and any pair of sessions closer than 10 minutes stayed welded together forever. Its own comment describes the intent it couldn't deliver: *"we keep the gap small enough that genuinely separate sessions stay separate."*

`collapseDepth` was migrated to pixels-per-hour in #658 precisely so density decisions adapt to zoom and container width; `mergeGapMs` was left behind on the day-based scale.

## Fix

`mergeGapForZoom(days, pixelsPerHour)` keeps the coarse tiers for multi-day views, but inside a single day derives the gap from pixels-per-hour — bridging only gaps under ~7 px, which are too small to draw anyway:

| View span | Gap bridged |
|---|---|
| > 50 days | 4 h |
| > 2 days | 1 h |
| single day or closer | `7px / pixelsPerHour`, capped at 10 min |

7 px is tuned so a 24 h view in a 1000 px container yields ~10 minutes — **exactly** the old value, so the day view is unchanged. The cap means a wide container can never merge *more* than before. Above ~47 pixels-per-hour the yoga pair splits, and back-to-back screentime slivers still read as one bar (their gaps are seconds).

## Also: the merge was invisible

`collapseToParentType` dropped `collapsed_types` whenever its single entry matched the bar's own type. That's right for `Merged: Yoga` — redundant — but it threw away the **×2**, which is the only signal that a bar holds two sessions and that clicking reaches just the first. Now dropped only when the count is 1, so a merged bar's tooltip reads `Merged: Yoga ×2`.

This flips an intentional, tested decision (`drops trivial provenance when no retype happens (depth=0)`); that test is replaced by one asserting the count survives, plus one keeping the lone-activity case dropped.

## Tests

- `mergeGapForZoom`: coarse tiers preserved, day view matches the old 10-minute floor via `computePixelsPerHour(1000, …)`, gap shrinks on zoom-in, never exceeds the cap, pre-measurement fallback, unbounded zoom.
- `collapseToParentType`: the reported case — two yoga sessions 9 minutes apart merge at a 10-minute gap and separate at 30 s.
- Provenance: `×2` retained at depth 0, still dropped for a lone activity.

561 web tests pass, `pnpm check` clean.

## Note

The merged bar linking to its first member only is unchanged here — that is inherent to collapsing several activities into one clickable bar. The tooltip count now makes it discoverable, and zooming in reaches the others. Splitting a merged bar into multiple click targets would be a larger change; happy to open an issue if you want it.


apps/web/src/pages/Timeline/collapseTier.test.ts
```diff
@@ -1,6 +1,47 @@
...
+describe('mergeGapForZoom', () => {
+  it('keeps the coarse tiers for multi-week and multi-day views', () => {
+    expect(mergeGapForZoom(60, 0.8)).toBe(4 * 60 * MINUTE)
+    expect(mergeGapForZoom(3, 14)).toBe(60 * MINUTE)
+  })
+
+  it('matches the previous 10-minute floor at a 1000px day view', () => {
+    // 24h in 1000px → ~41.7 pph, where 7px works out to ~10 min
+    const dayView = mergeGapForZoom(0, computePixelsPerHour(1000, new Date(0), new Date(24 * 3_600_000)))
+    expect(dayView / MINUTE).toBeCloseTo(10, 0)
```
fiddur, line 16: This assertion is satisfied by the cap, not by the 7 px tuning it claims to check: at 41.67 pph the raw value is `7/41.67 × 3.6e6 ≈ 10.08 min`, which `Math.min` clamps to exactly 10 min — so `toBeCloseTo(10, 0)` would hold for any `MERGE_GAP_PX ≥ ~7`.

Working through the rest of the new suite, the only real bound comes from the upper side: `mergeGapForZoom(0, 1000)` must stay under 9 min, i.e. `MERGE_GAP_PX < 150`. So the whole suite passes with the constant set anywhere from 7 to ~149 (e.g. 50 → hourView = 3 min, still `< dayView` and `< 9 min`), and nothing guards the single number this PR is actually tuning.

One uncapped assertion would pin it — 84 pph is exactly 2× day zoom and lands on a round value:

```ts
// 7px at 2× day zoom is exactly 5 min — pins MERGE_GAP_PX
expect(mergeGapForZoom(0, 84) / MINUTE).toBe(5)
```

docs/features/timeline.md
```diff
@@ -151,6 +151,20 @@ The Timeline automatically adjusts detail level as you zoom:
...
 
+### Bar Merging
+
+Adjacent activities of the same type merge into one bar when the gap between them is too small to be worth drawing. The threshold follows the zoom level:
+
+| View span            | Gap bridged                                          |
+| -------------------- | ---------------------------------------------------- |
+| Over 50 days         | 4 hours                                              |
+| Over 2 days          | 1 hour                                               |
+| Single day or closer | Whatever spans ~7 px on screen, capped at 10 minutes |
```
fiddur: "Single day or closer" under-states this branch: the pixel path runs for `days <= 2` (`mergeGapForZoom`), i.e. up to a 3-calendar-day view. That's invisible at 1000px because the 10-minute cap wins below ~42 pph, but not on a wide container — a 3-day view at 4000px is ~55.6 pph, so the bridged gap is ~7.5 min, not 10.

Same wording in the paragraph just below ("Inside a single day the threshold is measured in pixels"). "3-day view or closer" would match the code.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. I traced the behaviour rather than running it and the core logic holds up:

- `mergeGapForZoom` is monotone non-increasing in `pixelsPerHour` for the fine branch (`gap = min(10min, 7px)`), and `days` shrinks monotonically as the user zooms in, so zooming in never *increases* the bridged gap. The claim in the docstring is accurate.
- The `Math.min(FINE_MERGE_GAP_CAP_MS, …)` cap means the pixel path only bites above ~42 pph, so every view at or below the old 1000px/24h density keeps exactly the previous 10-minute behaviour. No regression for the "comb of screentime slivers" case the old comment worried about — truly back-to-back spans have gap 0 and merge at any threshold, including the 0 returned at unbounded zoom.
- The `count === 1` relaxation in `collapseToParentType` is safe: `collapsed_types` is only ever consumed by `formatCollapsedTypesLine` (activityMerge.ts:250, categorize.ts:212), so the change adds a tooltip line and touches nothing else. The survivor keeps the first member's `id`, so the doc's "a merged bar links to its first member only" is correct. I also checked that the new "×N" won't fire spuriously for one session recorded by two integrations — the backend already collapses those (`mergeOverlappingActivities` / `superseded_by`), so a `×2` in the Activity lane really does mean two non-overlapping sessions.

Two non-blocking notes inline (test doesn't pin the 7 px constant; one doc row mislabels its branch), plus a few observations that need no action:

- `mergeGapMs` only refreshes on d3 zoom **end** (`useTimelineZoom` fires `onZoomEnd` → `handleZoom` → the signals), so during a wheel/pinch gesture the bar stays welded and splits when the gesture finishes. Same as `collapseDepth` since #658, so this is consistent, just worth knowing.
- `computePixelsPerHour(timeAxisPixels, effectiveViewStart, effectiveViewEnd)` is now computed twice in the hook (mergeGapMs + collapseDepth) with identical args. Hoisting it into its own `useMemo` and passing the number to both would read a bit better.
- `mergeGapForZoom` mixes calendar-day tiers with pixel tiering, so it's non-monotone in pph *across container widths*: a 3-day view at 4000px (~55.6 pph) bridges 1 hour while a 1-day view at 1000px (~41.7 pph) bridges 10 minutes, despite the former being denser. Not reachable by zooming, so not a bug — but if the coarse tiers are ever revisited, moving them onto pph (like `collapseDepthForPixelsPerHour` did in #658) would remove the seam.
- Docs and code comments were updated together, which is what CLAUDE.md asks for.

apps/web/src/pages/Timeline/collapseTier.test.ts
fiddur: Good catch — the assertion was checking the cap, not the tuning. Added your suggested case verbatim:

```ts
// 84 pph is exactly 2x day zoom, where 7px works out to an uncapped 5 min
expect(mergeGapForZoom(0, 84) / MINUTE).toBe(5)
```

Verified it actually pins the constant: at `MERGE_GAP_PX = 50` the raw value is ~35.7 min, the cap clamps it to 10, and the assertion fails.

docs/features/timeline.md
fiddur: Right — the pixel path runs for `days <= 2`, so a 3-day view is included. Reworded both the table row and the paragraph to "3-day view or closer".

Your 4000px example is what makes it matter: ~55.6 pph on a 3-day view bridges ~7.5 min, so the cap is not always the answer there.

apps/web/src/pages/Timeline/activityMerge.ts
```diff
@@ -482,13 +482,17 @@ export const collapseToParentType = (
...
+  // "Merged: Running" line. A count above one is worth keeping even then —
+  // "Merged: Yoga ×2" is the only clue that one bar hides two sessions, and
+  // that only the first is reachable by clicking it.
   for (const m of merged) {
     if (
       m.collapsed_types &&
       m.collapsed_types.length === 1 &&
-      m.collapsed_types[0].type === m.activity_type
+      m.collapsed_types[0].type === m.activity_type &&
+      m.collapsed_types[0].count === 1
```
fiddur: With this condition added, the whole drop loop becomes unreachable. `collapsed_types` can only be set two ways:

1. the retype branch (line 457), which only fires when `effective !== a.activity_type`, so the single entry's `type` never equals the survivor's `activity_type`; or
2. the merge accumulation (lines 475-480), where the base list already holds one member and at least one incoming member is recorded on top — so a single-entry list there always has `count >= 2`.

A survivor with `length === 1 && type === activity_type && count === 1` would have to come from an *input* activity that already carried provenance, and `Activity.collapsed_types` is documented in `state/api/types.ts` as "never on activities returned by the API". So the loop can be deleted outright. If you'd rather keep it as a guard against a future second collapse pass, a one-line comment saying so would help — right now it reads as live logic.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. Swapping the flat 10-minute floor for a pixel-derived gap inside the day tier is the right shape: `Math.min(FINE_MERGE_GAP_CAP_MS, …)` guarantees no view merges *more* than before, the NaN/`<= 0` guard mirrors `collapseDepthForPixelsPerHour`'s pre-measurement fallback, and the coarse `days > 2` / `days > 50` tiers are preserved verbatim so multi-day/multi-week views are untouched. I checked the tier arithmetic against the tests and they all hold (day view at 1000px → exactly the 10-min cap; 84 pph → 5 min uncapped; `Infinity` → 0). Hoisting `pixelsPerHour` into its own memo is also a small win: `collapseDepth` and `mergeGapMs` now key off a numeric value instead of freshly-allocated `Date`s, so they stop recomputing on every render when the view is at its default.

I verified the blast radius of the wider `mergeGapMs`: it only feeds client-side memos in `useTimelineData` (activity lane + screentime lane) and is not part of any react-query key, so the now-continuous float causes no cache churn or extra requests. `fetchProductivity`'s `merge_gap_ms` is not wired to it either.

Non-blocking observations:

- The `count === 1` addition in `collapseToParentType` makes the trailing trivial-provenance loop unreachable in practice — see inline. Behaviour is still correct, it's just dead code plus a now-vacuous test.
- The change also makes screentime bars whose derived type has no resolvable `parent_type` grow a `Merged: Slack ×12` tooltip line where they previously had none. That reads as informative rather than wrong, but it's a side effect beyond the stated yoga fix, and `docs/features/screentime-categories.md` isn't mentioned in the PR.
- Coverage is split cleanly per unit, but nothing ties the two halves together: `mergeGapForZoom` is tested on numbers and `collapseToParentType` is tested with hand-picked gaps (`10 * 60_000`, `30_000`). A single test feeding `mergeGapForZoom(0, computePixelsPerHour(1000, …))` into `collapseToParentType` with the two yoga sessions would pin the actual reported regression end-to-end.
- There are now two pixels-per-hour → merge-gap heuristics in this directory: the new continuous `mergeGapForZoom` and the older 3-tier `getMergeGapMs` in `drawMusicStaff.ts` (10 min / 30 min / 2 h). Different domains and unifying them would change music-session behaviour, so leaving them apart is defensible — worth a cross-reference comment at some point so the next reader doesn't assume one is stale.

apps/web/src/pages/Timeline/activityMerge.test.ts
```diff
@@ -503,19 +503,39 @@ describe('collapseToParentType', () => {
...
     const activities: Activity[] = [
       { activity_type: 'running', end_time: d(10, 30), id: 'a', start_time: d(10) },
       { activity_type: 'running', end_time: d(11), id: 'b', start_time: d(10, 35) },
     ]
     const collapsed = collapseToParentType(activities, typeDefs, undefined, 0)
     expect(collapsed[0].activity_type).toBe('running')
+    expect(collapsed[0].collapsed_types).toEqual([{ type: 'running', count: 2 }])
+  })
+
+  it('drops trivial provenance for a lone unmerged activity (depth=0)', () => {
```
fiddur: This replaces the only test that exercised the drop loop, but it passes vacuously: at `depth=0` a lone activity takes the `effective === a.activity_type` branch (line 454-456) and never gets `collapsed_types` attached in the first place, so the assertion holds even with the loop deleted. Combined with the `count === 1` guard, nothing in the suite covers that cleanup any more.

docs/features/timeline.md
```diff
@@ -151,6 +151,20 @@ The Timeline automatically adjusts detail level as you zoom:
...
+
+Adjacent activities of the same type merge into one bar when the gap between them is too small to be worth drawing. The threshold follows the zoom level:
+
+| View span            | Gap bridged                                          |
+| -------------------- | ---------------------------------------------------- |
+| Over 50 days         | 4 hours                                              |
+| Over 2 days          | 1 hour                                               |
+| 3-day view or closer | Whatever spans ~7 px on screen, capped at 10 minutes |
+
+For a 3-day view or closer the threshold is measured in pixels, not minutes, so **zooming in always pulls separate sessions apart** -- two yoga sessions nine minutes apart read as one bar in the day view and as two once you zoom in. This matters because a merged bar links to its _first_ member only: to open the second session, zoom in until the bar splits. A merged bar's tooltip names its members with counts ("Merged: Yoga ×2"), so you can tell when there is more inside than the bar suggests.
```
fiddur: "zooming in always pulls separate sessions apart" overstates it for back-to-back spans. The merge test is `gap <= mergeGapMs`, and `mergeGapForZoom` bottoms out at 0 (the `Number.POSITIVE_INFINITY` test asserts exactly that), so two same-type activities where one ends at the moment the next starts merge at *every* zoom level. For those the following sentence's advice — "to open the second session, zoom in until the bar splits" — never works, and since a merged bar links to its first member only, the second is unreachable from the Timeline. Worth a half-sentence caveat ("unless they touch exactly, in which case the tooltip count is the only signal").

apps/web/src/pages/Timeline/activityMerge.ts
fiddur: You are right, and I verified both paths: the retype branch records `a.activity_type` while setting the survivor to `effective` (different by construction), and the merge accumulation seeds from one member then records at least one more, so a single-entry list there is always count ≥ 2. `state/api/types.ts:82-88` confirms the third path is out — "never on activities returned by the API".

Deleted the loop rather than commenting it, and left a note in its place explaining why nothing replaces it.

apps/web/src/pages/Timeline/activityMerge.test.ts
fiddur: Correct — it held with the loop deleted, so it was not covering the cleanup at all. Renamed it to say what it actually asserts ("attaches no provenance to a lone activity that was not retyped").

Added a test for the invariant the cleanup used to enforce, which does not depend on the loop existing: across depths 0 / 1 / Infinity, no resulting bar carries an entry that merely restates its own type with count 1.

docs/features/timeline.md
fiddur: Fair — and my own test asserts the premise: `mergeGapForZoom(0, Number.POSITIVE_INFINITY)` returns 0, and `gap === 0` satisfies `gap <= 0`, so exactly-touching activities merge at every zoom.

Added the caveat, and made it explicit that "zoom in until the bar splits" does not apply to them and the tooltip count is the only signal there.

apps/web/src/pages/Timeline/activityMerge.test.ts
```diff
@@ -503,19 +503,57 @@ describe('collapseToParentType', () => {
...
+    // the retype records a *different* type, and a merge always totals ≥ 2.
+    const activities: Activity[] = [
+      { activity_type: 'running', end_time: d(10, 30), id: 'a', start_time: d(10) },
+      { activity_type: 'running', end_time: d(11), id: 'b', start_time: d(10, 35) },
+      { activity_type: 'yoga', end_time: d(12), id: 'c', start_time: d(11, 30) },
+    ]
+
+    for (const depth of [0, 1, Number.POSITIVE_INFINITY]) {
+      for (const bar of collapseToParentType(activities, typeDefs, undefined, depth)) {
+        const trivial = bar.collapsed_types?.some((e) => e.type === bar.activity_type && e.count === 1)
```
fiddur: This asserts more than the invariant that actually holds, and the stronger version is false.

The removed cleanup guarded the **array-level** case `collapsed_types.length === 1 && collapsed_types[0].type === activity_type` — that one is genuinely unreachable, as the comment above argues. But `.some()` here checks an **entry-level** invariant: "no entry anywhere in the list restates the bar's own type with count 1." That's reachable whenever an activity already typed as the collapse target merges with a retyped child:

```ts
const activities: Activity[] = [
  { activity_type: 'running',  end_time: d(10, 30), id: 'a', start_time: d(10) },
  { activity_type: 'exercise', end_time: d(11),     id: 'b', start_time: d(10, 35) },
]
collapseToParentType(activities, typeDefs, undefined, 1)
```

`running` retypes to `exercise` carrying `[{ running, 1 }]`; `exercise` has no parent so it stays put with no provenance and contributes the `[{ exercise, 1 }]` fallback at line 475. The survivor is an `exercise` bar with `[{ running, 1 }, { exercise, 1 }]` — `.some(e => e.type === bar.activity_type && e.count === 1)` is `true` and this test fails. `exercise` is a real first-class `activity_type` here (it's in `typeDefs` above and has its own branch in `getActivityMeta`), so that's not a synthetic case.

The fixture only passes because it contains no parent-typed activity. Suggest matching the assertion to the invariant you're actually relying on, e.g.

```ts
const trivial =
  bar.collapsed_types?.length === 1 &&
  bar.collapsed_types[0].type === bar.activity_type &&
  bar.collapsed_types[0].count === 1
```

and adding the `running` + `exercise` pair to the fixture so the test covers the mixed case rather than sidestepping it.

fiddur (commented):
✅ Approved

Solid, well-scoped fix. I traced the two changes through the consumers and both hold up.

**`mergeGapForZoom` (collapseTier.ts)** — the pixel-based branch only kicks in for `days <= 2`, and the `FINE_MERGE_GAP_CAP_MS` cap means it can never bridge *more* than the old fixed 10-minute floor. Concretely it only changes behaviour above ~42 pph (i.e. a view narrower than 24h in a 1000px container), so every existing zoom level at or below day view is bit-for-bit unchanged. The NaN/`<= 0`/`Infinity` guards match `computePixelsPerHour`'s pre-measure contract, and hoisting `pixelsPerHour` out for `collapseDepth` to reuse is behaviour-preserving. `mergeGapMs` isn't part of any react-query `queryKey` in `useTimelineData`, so making it container-size dependent doesn't cause refetches — only local memo recomputation, which the D3 redraw on resize dominates anyway.

**Removing the trivial-provenance cleanup (activityMerge.ts)** — the reasoning in the replacement comment checks out exhaustively: the retype path only sets `collapsed_types` when `effective !== a.activity_type`, and the merge path always accumulates a total count ≥ 2, so `[{ type: <own type>, count: 1 }]` as a whole array is unreachable. The only behaviour change is that `[{ own, ≥2 }]` now survives, which is exactly the `×2` signal you want given a merged bar links to its first member only. Nothing outside the tooltip builders (`buildActivityDetails`, `categorize.ts`) reads `collapsed_types`, so there's no click-target or layout fallout.

One inline note on the new invariant test — its `.some()` assertion is strictly stronger than the invariant that actually holds, and is falsifiable with a realistic fixture. Non-blocking, but worth tightening so it doesn't fail confusingly later.

Two very minor doc points (no change needed unless you want them):
- "A merged bar's tooltip names its members with counts" isn't true for the legacy `screentime` umbrella lane — that path goes through `mergeAdjacentByKey`, which never sets `collapsed_types`, so those merges are silent. Only the Activity lane and derived screentime types (via `collapseToParentType`) get the "Merged: …" line.
- "the threshold bottoms out at zero" is only true at literally infinite zoom; at any finite pph it's a small positive number. The practical claim (touching/overlapping spans always merge) is right either way.

apps/web/src/pages/Timeline/activityMerge.test.ts
fiddur: You are right, and I confirmed the counterexample: `resolveCollapseTarget("exercise", typeDefs)` is null in this fixture, so an already-`exercise` activity stays put with no provenance and contributes the `[{exercise,1}]` fallback at line 475 — giving `[{running,1},{exercise,1}]` on an `exercise` bar. My `.some()` assertion would have failed on it, and only passed because the fixture had no parent-typed activity.

Narrowed to the list-level invariant as you suggested, added `exercise` to the fixture, and pinned the mixed provenance explicitly in a second test. Worth noting `yoga` retypes onto `exercise` too and is inside the default gap, so that list is three entries — which I only found by running it.

apps/web/src/pages/Timeline/activityMerge.test.ts
```diff
@@ -503,19 +503,78 @@ describe('collapseToParentType', () => {
...
+    { activity_type: 'yoga', end_time: d(12), id: 'c', start_time: d(11, 30) },
+  ]
+
+  it('never yields a provenance list that only restates the bar itself', () => {
+    // What the removed trivial-provenance cleanup used to guard, and only at the
+    // list level: `[{ type: <own type>, count: 1 }]` as the whole list. Neither
+    // path that sets collapsed_types can produce it — the retype records a
+    // *different* type, and a merge always totals >= 2.
+    for (const depth of [0, 1, Number.POSITIVE_INFINITY]) {
+      for (const bar of collapseToParentType(mixedActivities, typeDefs, undefined, depth)) {
```
fiddur: This guard can't fail for the regression it names. With `mixedActivities`, no bar ever ends up with a **single-entry** provenance list at any of the three depths: at depth 0 nothing is retyped and the three types differ, so all three bars have `collapsed_types === undefined` (`provenance.length === 0`); at depth 1 and Infinity all three fold into one `exercise` bar with three entries. So `provenance.length === 1` is never true and the `type`/`count` clauses are never reached.

The interesting boundary is a list that *is* single-entry: two same-type activities merging at depth 0 gives `[{ running, 2 }]` on a `running` bar — length 1, type matches the bar, and only the `count === 1` clause keeps it non-trivial. A lone retyped activity gives the other single-entry shape, `[{ running, 1 }]` on an `exercise` bar. Adding those two to the loop (or to the fixture) makes the assertion actually exercise the invariant instead of passing on list length.

docs/features/timeline.md
```diff
@@ -151,6 +151,20 @@ The Timeline automatically adjusts detail level as you zoom:
...
 
+### Bar Merging
+
+Adjacent activities of the same type merge into one bar when the gap between them is too small to be worth drawing. The threshold follows the zoom level:
+
+| View span            | Gap bridged                                          |
+| -------------------- | ---------------------------------------------------- |
+| Over 50 days         | 4 hours                                              |
+| Over 2 days          | 1 hour                                               |
+| 3-day view or closer | Whatever spans ~7 px on screen, capped at 10 minutes |
```
fiddur: These two rows read as contradictory: someone with a 3-day view matches both "Over 2 days → 1 hour" and "3-day view or closer → pixels". The code branches on `differenceInCalendarDays`, so the pixel path is `days <= 2` — which a 3-day span reaches only when it doesn't cross a fourth calendar-day boundary (Aug 1 00:00→Aug 3 23:59 is `days = 2`, but Aug 1 00:00→Aug 4 00:00 is `days = 3` and takes the 1-hour tier despite being the same 72h span). Naming the boundary once in day-boundary terms in both rows (e.g. "spans 4+ calendar days" / "spans 3 calendar days or fewer") would remove the overlap.

fiddur (commented):
✅ Approved

Solid fix, and the reasoning in the comments holds up under scrutiny. I traced the two things that could plausibly break:

**The removed trivial-provenance cleanup is provably dead.** `collapsed_types` can only be set on a survivor by (a) the retype branch at `activityMerge.ts:457`, which records `a.activity_type` on a bar whose type is `effective !== a.activity_type` — so the entry type differs by construction — or (b) the merge accumulation at 475-480, which seeds from a ≥1-entry list and then calls `recordCollapsedType` at least once per incoming entry, so the total count is always ≥2. A single `[{ own type, count: 1 }]` list is unreachable, and inputs never arrive carrying provenance (`state/api/types.ts:82-88`). The user-visible change is that a day-view bar folding two `running` sessions now keeps `[{running, 2}]` and renders "Merged: Running ×2" — that's the right call, since with the survivor inheriting the first member's id it's the only signal that clicking reaches just the first session. No mutation hazard either: `recordCollapsedType` is non-mutating and the merge writes only into the `{ ...current }` copy, so react-query's cached activities stay untouched.

**The pixel-gap math checks out.** 41.7 pph (24h/1000px) → 10.08 min → capped to exactly 10 min, so the default day view is byte-for-byte the old behavior; 84 pph → exactly 5 min uncapped; `Infinity` → 0 (touching spans only). The cap means a wide container can never bridge *more* than before, and the `<= 0 || NaN` guard returning the cap (rather than 0, as `collapseDepthForPixelsPerHour` does) is the right pre-measure fallback here since it reproduces the old constant. Hoisting `pixelsPerHour` into its own memo is a real improvement, not just tidying: `collapseDepth`'s deps become a single number instead of two `Date` identities that change every render when the view signals are null.

Worth knowing about, not blocking: this also drives the Screen Time lane (`useTimelineData.ts:301,309`). At a 2h view in 1000px the threshold is ~50s, so screentime spans separated by even a minute will now comb apart where they previously welded at 10 min. That's the intended tradeoff (every visible gap is ≥7px) but it's a wider blast radius than the yoga case in the title, so keep an eye on it in the deployed lane. Also note `drawMusicStaff.ts:133` already has a `getMergeGapMs(pixelsPerHour)` doing the same job for scrobbles with different tuning — two similarly-named zoom→gap helpers in one directory is a mild readability trap, though the tunings are genuinely different so merging them isn't obviously right.

Two non-blocking notes inline.

apps/web/src/pages/Timeline/activityMerge.test.ts
fiddur: Right again — it was vacuous in a different way. With `mixedActivities` no bar ever has a single-entry list: depth 0 leaves all three untouched with no provenance, and depth 1/Infinity fold them into one three-entry `exercise` bar. `provenance.length === 1` was never true.

Added both shapes you named as fixtures — `[{running,2}]` on a `running` bar (only the count keeps it non-trivial) and `[{running,1}]` on an `exercise` bar (only the type does) — and a separate test asserting both actually occur, so it cannot silently stop exercising the clauses again.

docs/features/timeline.md
fiddur: Restated the tiers in calendar-day boundaries, which is what `differenceInCalendarDays` counts, so the rows no longer overlap:

| Crosses more than 50 boundaries | 4 hours |
| Crosses 3 to 50 boundaries | 1 hour |
| Crosses 2 or fewer boundaries | ~7 px, capped at 10 min |

Included your Aug 1 → Aug 3 23:59 (2 boundaries, pixel tier) vs Aug 1 → Aug 4 00:00 (3 boundaries, hour tier) example, since the same 72h span landing in different tiers is the surprising part.

apps/web/src/pages/Timeline/collapseTier.ts
```diff
@@ -41,3 +41,33 @@ export const computePixelsPerHour = (
...
 }
+
+/**
+ * On-screen gap (px) below which two adjacent same-type bars are bridged into
+ * one. Tuned so that a 24h view in a 1000px container reproduces the previous
+ * fixed 10-minute gap.
+ */
+const MERGE_GAP_PX = 7
+
+/** Widest gap bridged at single-day zoom or closer. */
```
fiddur: This comment (and "At single-day zoom and closer" in the `mergeGapForZoom` JSDoc below) understates the tier's reach: the predicate is `days > 2`, so the pixel tier covers everything up to 2 crossed calendar-day boundaries — i.e. up to just under 72 elapsed hours, not just a single day.

`docs/features/timeline.md` gets this right ("Crosses 2 or fewer boundaries") and even spells out the Aug 1 → Aug 3 23:59 case, so the code comment is the odd one out. Suggest "at views crossing ≤ 2 calendar-day boundaries" to match.

apps/web/src/pages/Timeline/activityMerge.ts
```diff
@@ -482,17 +482,13 @@ export const collapseToParentType = (
...
-  for (const m of merged) {
-    if (
-      m.collapsed_types &&
-      m.collapsed_types.length === 1 &&
-      m.collapsed_types[0].type === m.activity_type
-    ) {
-      delete m.collapsed_types
-    }
-  }
+  // No trivial-provenance cleanup is needed: a survivor can only carry
```
fiddur, line 502: The reasoning is right for the Activity lane, but keeping `[{ own type, N }]` has a side effect on the Screen Time lane that the rationale doesn't cover.

`useTimelineData` calls this with `screentimeDepth = Math.max(1, collapseDepth)`. A derived type minted for a **top-level** screentime category gets `parent_type = null` (`ensureCategoryHasType` only passes a `parentSlug` when `findParentCategory` finds one), so it isn't retyped — the bar stays `work`/`programming` and a merged run now carries `[{ programming, 47 }]`. That previously hit the deleted cleanup and got stripped; now `categorizeScreentimeActivities` will render "Merged: Programming ×47" on it.

Both halves of the justification above are weaker there: `N` counts raw RescueTime/ActivityWatch sampling spans rather than distinct sessions, and the bar links to `/screentime-categories/{id}` rather than to its first member, so "clicking reaches just the first" doesn't apply. Fine if that's the intent — but if not, gating on `count > 1 || type !== activity_type` per lane, or suppressing the line for screentime bars, would keep the yoga fix without adding a meaningless count to screen-time tooltips.

fiddur (commented):
✅ Approved

Solid, well-reasoned fix. I hand-evaluated `mergeGapForZoom` and `collapseToParentType` against every new assertion and they all hold:

- **No regression toward more merging.** `Math.min(FINE_MERGE_GAP_CAP_MS, …)` means the fine tier can only ever be ≤ the old fixed 10-minute floor, so no view merges more than it used to. Verified at the interesting points: 41.7 pph → 600000 (capped, identical to before), 84 pph → 300000, 1 pph → capped, `days=2, 0.001 pph` → capped.
- **The doc claim "zooming in always pulls separate sessions apart" is actually true.** The threshold is monotonically non-increasing as you zoom in across all three tiers (4h → 1h → min(10min, pixel)), including at the tier boundaries, so there's no zoom direction where merging gets more aggressive.
- **The dead-code argument for dropping the cleanup checks out.** The retype path always records a type ≠ the survivor's, and the merge path always totals ≥ 2, so `[{ own type, count: 1 }]` as a whole list is genuinely unreachable. The real behavior change is keeping `[{ own type, count: N≥2 }]`, which the old cleanup stripped because it had no `count` check.
- **New tests assert what they claim.** I traced `mixedActivities` at depth 1 by hand and it does produce `[{running,1},{exercise,1},{yoga,1}]`; the `sameTypePair`/`loneChild` fixtures do reach both single-entry shapes, so the "reaches both single-entry provenance shapes" test really does guard the invariant test from passing on length alone. `mergeGapForZoom(0, Infinity)` → 0 is correct, and unreachable in practice since `computePixelsPerHour` returns 0 for a degenerate range.
- Hoisting `pixelsPerHour` into its own memo and sharing it with `collapseDepth` is a clean refactor — same value, one computation.

Two non-blocking notes inline. Beyond those, one observation for the record: because `mergeGapMs` is now a continuous function of container pixels rather than one of three quantized values, the `activities` / `screentimeActivities` memos in `useTimelineData.ts` now invalidate on every `ResizeObserver` tick during a window resize (a sort + linear merge each). Over a day's worth of activities that's negligible, so I wouldn't change anything — just flagging that the memo is no longer stable across resize the way it was.

Also worth knowing: `mergeGapForZoom(days, pixelsPerHour)` takes two positional `number` params with very different meanings and there's no test at the `useTimelineNavigation` level, so swapping the arguments would type-check silently. The call site is correct today; if this grows a third caller, a named-object parameter would remove the footgun.

apps/web/src/pages/Timeline/collapseTier.ts
fiddur: Fixed — reworded both the `FINE_MERGE_GAP_CAP_MS` comment and the `mergeGapForZoom` JSDoc to "views crossing <= 2 calendar-day boundaries (so anything up to just under 72 elapsed hours)". Agreed the docs were already right and the comments were the outliers.

apps/web/src/pages/Timeline/activityMerge.ts
fiddur: This one I had not considered at all, and it is a real regression. Confirmed the chain: `useTimelineData.ts:308` uses `screentimeDepth = Math.max(1, collapseDepth)`, `ensureCategoryHasType` leaves `parentSlug` null when `findParentCategory` finds no parent, and `categorize.ts` feeds `collapsed_types` straight into `formatCollapsedTypesLine`. So a top-level category bar would have started showing "Merged: Programming ×47".

Both halves of my justification do fail there, as you say — N is sampling spans, and the bar links to the category rather than a member.

Added `restatesOwnType` and suppressed the line at the screen-time render site only, so the Activity lane keeps "Merged: Yoga ×2" where the count does mean two sessions. Test added for the suppressed case alongside the existing #657 one that asserts a genuinely mixed provenance still renders.

apps/web/src/pages/Timeline/activityMerge.ts
```diff
@@ -288,6 +288,22 @@ export const COLLAPSE_MERGE_GAP_MS = 30 * 60 * 1000 // 30 minutes
  * there's nothing to show — a single-entry provenance is still informative
  * (the parent bar is one sub-type only) so we keep that case.
  */
+/**
```
fiddur, line 303: This block was inserted *between* `formatCollapsedTypesLine`'s existing JSDoc (lines 284-290) and its declaration, so now: `formatCollapsedTypesLine` has no docblock of its own, and the orphaned one floats directly above `restatesOwnType` where it reads as a description of it — and a contradictory one at that ("a single-entry provenance is still informative … so we keep that case" immediately above the helper whose job is to drop single-entry provenance). Moving the old block back down onto `formatCollapsedTypesLine` (or putting `restatesOwnType` after it) keeps both docs attached to what they describe. The file already has one such orphan above `getActivityIconKey`; worth not adding a second.

docs/features/timeline.md
```diff
@@ -151,6 +151,22 @@ The Timeline automatically adjusts detail level as you zoom:
...
+| ------------------------------- | ---------------------------------------------------- |
+| Crosses more than 50 boundaries | 4 hours                                              |
+| Crosses 3 to 50 boundaries      | 1 hour                                               |
+| Crosses 2 or fewer boundaries   | Whatever spans ~7 px on screen, capped at 10 minutes |
+
+In that last tier the threshold is measured in pixels, not minutes, so **zooming in always pulls separate sessions apart** -- two yoga sessions nine minutes apart read as one bar in the day view and as two once you zoom in. This matters because a merged bar links to its _first_ member only: to open the second session, zoom in until the bar splits. The exception is two activities that touch exactly -- one ending at the moment the next begins. The test is `gap <= threshold` and the threshold bottoms out at zero, so those merge at every zoom level and the tooltip count is the only signal that the bar holds two. A merged bar's tooltip names its members with counts ("Merged: Yoga ×2"), so you can tell when there is more inside than the bar suggests.
```
fiddur, line 164: "the tooltip count is the only signal that the bar holds two" / "A merged bar's tooltip names its members with counts" is now only true for the Activity lane. This same PR makes `categorizeScreentimeActivities` suppress the line via `restatesOwnType`, and for screen time that's the common case (a top-level category's derived type has no `parent_type`, so a merged run carries `[{ own_type, N }]`). Since this section reads as lane-agnostic, a merged Screen Time bar silently gives no signal at all — worth stating the exception here so the docs match the behaviour.

Minor while you're in here: line 156 ends with "The threshold follows the zoom level:" but the table doesn't follow until after the calendar-boundary paragraph, which itself ends in a colon. Swapping the two sentences (or dropping the first colon) makes the lead-in point at the table.

apps/web/src/pages/Timeline/useTimelineNavigation.ts
```diff
@@ -88,16 +88,19 @@ export const useTimelineNavigation = (options: TimelineNavigationOptions = {}):
...
-  // screentime spans don't show as a comb of identical slivers, but we keep the
-  // gap small enough that genuinely separate sessions stay separate.
-  const mergeGapMs = useMemo(() => {
-    const days = differenceInCalendarDays(effectiveViewEnd, effectiveViewStart)
-    if (days > 50) return 4 * 60 * 60 * 1000
-    if (days > 2) return 60 * 60 * 1000
-    return 10 * 60 * 1000
-  }, [effectiveViewStart, effectiveViewEnd])
+  // a long string of small same-type activities reads as one bar. Inside a
+  // single day the gap follows pixels-per-hour, so back-to-back screentime
```
fiddur: "Inside a single day" understates the tier: the pixel branch covers everything up to 2 crossed calendar-day boundaries (just under 72 elapsed hours), which is what `mergeGapForZoom`'s own docblock and the new docs table say. Worth matching the wording here so the three descriptions agree.

fiddur (commented):
✅ Approved

Solid, well-reasoned change. The core fix — replacing the fixed 10-minute floor with a pixels-per-hour-derived gap capped at the old floor — is correct and the tier boundaries line up with the docs table (`days > 50` → 4h, `days > 2` → 1h, else pixel tier). I checked the new `collapseTier` tests' arithmetic by hand: `computePixelsPerHour(1000, 0, 24h)` = 41.67 pph → uncapped 10.08 min → capped 10 min (so the `toBeCloseTo(10, 0)` assertion really is satisfied by the cap, which the follow-up `mergeGapForZoom(0, 84) === 5 min` test correctly pins instead); `7/84 * 3.6e6` rounds to exactly 300000 in doubles, so `toBe(5)` is safe. Hoisting `pixelsPerHour` into its own memo and reusing it for both `mergeGapMs` and `collapseDepth` is a nice cleanup, and the values stay primitives so the downstream `useTimelineData` memos don't churn.

I also verified the removal of the trivial-provenance cleanup in `collapseToParentType`. `collapsed_types` exists only in the web layer (`apps/web/src/state/api/types.ts` declares it, but no backend/api-spec code ever sets it), and the two writer paths are the retype (records a type that differs from the survivor's by construction) and the merge accumulation (total ≥ 2), so `[{ own_type, 1 }]` is genuinely unreachable and the deleted loop was dead. Both consumers are covered: `buildActivityDetails` keeps the count, `categorizeScreentimeActivities` suppresses it via `restatesOwnType`, which puts the Screen Time lane exactly back at pre-PR behavior. `collapseToParentType` is still non-mutating w.r.t. its inputs after the change.

Non-blocking observations:

- `restatesOwnType`'s rationale ("the count is raw sampling spans rather than sessions") applies equally to the *non*-restating screen-time shape: a `work`-typed bar that folded three `programming` spans still renders "Merged: Programming ×3" with the same raw-span semantics. That's pre-PR behavior, not a regression, but the carve-out is narrower than the reasoning it's justified with.
- There is now a second zoom-derived merge-gap helper in this folder: `getMergeGapMs(pixelsPerHour)` in `drawMusicStaff.ts` (step tiers 10/30/120 min, used for scrobble sessions at line 784 of `index.tsx`). The near-identical name and identical input are easy to confuse with `mergeGapForZoom`; a cross-reference in one of the two docblocks would help future readers.
- `mergeGapMs` derives from the nav view range, which `useTimelineZoom` only commits on the zoom `'end'` event, so bars re-split when the gesture finishes rather than during it. Same as the existing `collapseDepth` behaviour and consistent with the doc's "zoom in until the bar splits" wording — just noting it's expected.

apps/web/src/pages/Timeline/activityMerge.ts
fiddur: Fixed — that was my editing mistake. Moved `restatesOwnType` below `formatCollapsedTypesLine` so both docblocks sit on what they describe, and noted on the formatter that callers who cannot use a self-restating entry filter with `restatesOwnType` first. Agreed the contradiction was the worse half: a doc saying single-entry provenance is worth keeping, directly above the helper that drops it.

docs/features/timeline.md
fiddur: Both fixed. Added a paragraph stating that the count is an Activity-lane signal only, that Screen Time suppresses it in what is normally the common case, and why it would be meaningless there (sampling spans, and the bar links to its category). So the docs no longer read lane-agnostic.

Also moved the calendar-boundary sentence up into the lead-in so the colon actually precedes the table.

apps/web/src/pages/Timeline/useTimelineNavigation.ts
fiddur: Third instance of the same understatement — I fixed two in `collapseTier.ts` last round and missed this one. Now "At views crossing <= 2 calendar-day boundaries", so the comment, the `mergeGapForZoom` docblock, and the docs table all agree.

apps/web/src/pages/Timeline/categorize.ts
```diff
@@ -209,7 +210,12 @@ export const categorizeScreentimeActivities = (
       const iconPath = matchedByType?.name ?? path
       const icon = resolveCategoryIcon(iconPath, itemIcons) ?? def?.icon
 
-      const mergedLine = formatCollapsedTypesLine(a.collapsed_types)
+      // Suppressed when the provenance only restates this bar's own type: for a
+      // top-level category (no parent_type, so never retyped) the count is raw
+      // sampling spans, not sessions — see `restatesOwnType`.
+      const mergedLine = restatesOwnType(a.collapsed_types, a.activity_type)
```
fiddur, line 222: The suppression is narrower than the rationale you wrote for it. `restatesOwnType` only fires when the single provenance entry names the bar's *own* type — i.e. only for top-level categories that were never retyped.

But nested categories do get `parent_type` on their derived type def (`screentime-category-sync.ts` sets it from the category tree), and `useTimelineData` uses `screentimeDepth = Math.max(1, collapseDepth)`, so a `programming` type nested under `work` always retypes. A merged run of 47 `programming` spans then lands as a `work` bar with `[{ programming, 47 }]`, `restatesOwnType` returns false, and the tooltip renders literally `Merged: Programming ×47` — the exact string the comment calls noise. Same raw-sampling-span count, same "bar links to its category, not a member" caveat, not suppressed.

This is pre-existing behaviour, not a regression from this PR, so not a blocker. But if the reason to hide the count on the Screen Time lane is "the count is sampling spans and the bar isn't a member link", that reason doesn't depend on whether a retype happened — a lane-level "screen time never shows counts, only type names" rule would match the stated intent more closely than `restatesOwnType` does.

apps/web/src/pages/Timeline/collapseTier.ts
```diff
@@ -41,3 +41,34 @@ export const computePixelsPerHour = (
...
+const FINE_MERGE_GAP_CAP_MS = 10 * 60 * 1000
+
+/**
+ * Gap below which adjacent same-type activities merge into a single bar.
+ *
+ * Zoomed out, fixed tiers bridge large gaps so a long string of small same-type
+ * activities reads as one bar. At views crossing <= 2 calendar-day boundaries
+ * (so anything up to just under 72 elapsed hours) the gap instead tracks
+ * pixels-per-hour: only gaps too small to see on screen are bridged, so zooming
+ * in always pulls genuinely separate sessions apart — two yoga sessions nine
```
fiddur: "zooming in always pulls genuinely separate sessions apart" isn't quite unconditional, and the timeline doc is honest about why: the merge test is `gap <= mergeGapMs`, and the pixel threshold is always `> 0` for any finite `pixelsPerHour`, so two sessions that abut exactly (21:00–21:30 then 21:30–22:00) merge at *every* zoom level. Since the survivor keeps only the first member's `id`, the second session has no reachable bar at all — and the doc's remedy ("zoom in until the bar splits") can't work for that case.

No trivial fix: `<=` at gap 0 is exactly what makes contiguous screentime sampling spans read as one bar, so flipping to `<` would reintroduce the comb. Worth a follow-up issue though — e.g. only bridge a zero gap when the two members share a source (sampling spans), so user-recorded back-to-back sessions stay individually clickable. Just noting the JSDoc overstates the guarantee slightly relative to the code.

apps/web/src/pages/Timeline/activityMerge.test.ts
```diff
@@ -503,19 +503,104 @@ describe('collapseToParentType', () => {
...
+    // The bar stays typed as itself, so "Merged: Running" alone would be
+    // redundant — but the ×2 is the only signal that one bar hides two
+    // sessions and that clicking reaches just the first.
     const activities: Activity[] = [
       { activity_type: 'running', end_time: d(10, 30), id: 'a', start_time: d(10) },
       { activity_type: 'running', end_time: d(11), id: 'b', start_time: d(10, 35) },
     ]
     const collapsed = collapseToParentType(activities, typeDefs, undefined, 0)
     expect(collapsed[0].activity_type).toBe('running')
+    expect(collapsed[0].collapsed_types).toEqual([{ type: 'running', count: 2 }])
```
fiddur, line 540: This asserts the provenance *survives* `collapseToParentType`, which is the necessary half — but the behaviour the PR is actually buying ("the ×2 is the only signal that one bar hides two sessions") is the rendered tooltip line, and nothing asserts that end-to-end for the Activity lane.

The Screen Time lane got both directions covered through `categorizeScreentimeActivities` (the existing `#657` positive test plus the new negative one). The Activity lane's `buildActivityDetails` is private, but `buildActivityColumnItems` is exported and already has tests in this file — a one-liner there asserting `tooltip.details` contains `Merged: Running ×2` for a merged same-type pair would lock in the user-visible half and catch a future regression that re-adds a filter downstream of `collapseToParentType`.

fiddur (commented):
✅ Approved

Solid, well-reasoned fix. I traced the new tier function, the provenance invariant, and every new test assertion by hand; no blocking issues.

**What I verified**

- `mergeGapForZoom` is behaviour-preserving where it matters. The pixel gap only dips below the 10-minute cap once `pixelsPerHour > 42` (`7 / pph * 60 < 10`), i.e. once the view is tighter than 24h-in-1000px. So every view from ~1 day up to the 2-boundary edge still gets exactly the old 10-minute gap regardless of container width, and the discontinuity at the 2→3 calendar-boundary edge stays 10 min → 60 min, same 6× jump as before. The old comment's "comb of identical slivers" concern for screentime is therefore preserved at every zoom where it was previously in play.
- Every `mergeGapForZoom` assertion in `collapseTier.test.ts` checks out numerically, including the nice catch that the day-view test is satisfied by the cap alone — `mergeGapForZoom(0, 84) === 5min` really is the assertion that pins `MERGE_GAP_PX = 7`.
- The invariant replacing the deleted cleanup holds. I walked all four paths that can set `collapsed_types` (retype-only, merge with `prev` un-retyped, merge with `current` un-retyped, merge of two retyped) against the `typeDefs` fixture; none can produce `[{ own type, count: 1 }]`. Self-parent (`parent_type === name`) and cycles both bottom out in `effective === a.activity_type`, so no provenance is attached there either. The paired "reaches both single-entry provenance shapes" test genuinely prevents the invariant test from passing on list length alone.
- Moving `pixelsPerHour` into its own memo is a small improvement, not just a refactor: `collapseDepth` now depends on a primitive instead of two `Date` identities, so it stops recomputing on every render.
- Docs match the code: `days > 50` / `days > 2` / pixel tier maps correctly onto the table, and the `Aug 1 00:00 → Aug 3 23:59` (2) vs `→ Aug 4 00:00` (3) `differenceInCalendarDays` example is right.

**Non-blocking notes** (three inline, plus:)

- `mergeGapMs` is now a *continuous* function of `timeAxisPixels`, where before it was a discrete tier. Below ~1-day zoom that means the `activities` and `screentimeActivities` memos in `useTimelineData` (and everything derived from them) re-run on every `ResizeObserver` tick during a window/fullscreen resize, which they previously didn't. The 10-minute cap pins the value for all wider views, and activity counts are low when zoomed in that far, so this is unlikely to be felt — just flagging that the memo is no longer resize-stable.
- `mergeGapForZoom(days, …)`: the first parameter is calendar-day *boundaries crossed*, not elapsed days, and the JSDoc says so — but the bare name `days` at the call site invites the elapsed-hours reading the docs explicitly warn against. Also a slight asymmetry that the fine tier got named constants (`MERGE_GAP_PX`, `FINE_MERGE_GAP_CAP_MS`) while `4 * 60 * 60 * 1000` and `60 * 60 * 1000` stayed inline literals duplicated in the tests.

apps/web/src/pages/Timeline/categorize.ts
fiddur: Correct, and my comment claimed more than the code does. Confirmed `ensureCategoryHasType` sets `parentSlug` from the category tree for nested categories, so a `programming` type under `work` always retypes at `screentimeDepth = Math.max(1, collapseDepth)` and the count is not suppressed.

I have not widened the suppression here, because a blanket lane-level one would lose what #657 added — the sub-category *names* in a mixed bar are useful even though the counts are not. The likely shape is keeping names and dropping counts for that lane only, which is a behaviour change on pre-existing output rather than part of this fix. Filed as #981 and narrowed the comment to describe what the code actually does.

apps/web/src/pages/Timeline/collapseTier.ts
fiddur: Fair — the docs carried the caveat but the JSDoc did not. Reworded: it now says the guarantee is not unconditional, that `gap <= this` with a value always > 0 means exactly-abutting sessions never split, and that tightening to `<` is not the fix because a zero gap is what makes contiguous screentime spans read as one bar.

Your suggested direction — only bridge a zero gap when the two members share a source — is in #981 as the candidate approach.

apps/web/src/pages/Timeline/activityMerge.test.ts
fiddur: Good catch — I was asserting the necessary half and calling it done. Added the end-to-end one through `buildActivityColumnItems`:

```ts
expect(items[0].tooltip.details).toContain("Merged: Running ×2")
```

That covers the regression you describe: a filter reintroduced downstream of `collapseToParentType` now fails a test rather than silently removing the only signal the bar holds two sessions.

apps/web/src/pages/Timeline/collapseTier.ts
```diff
@@ -41,3 +41,39 @@ export const computePixelsPerHour = (
...
+ * in pulls separate sessions apart — two yoga sessions nine minutes apart used
+ * to stay welded together at every zoom level, because
+ * `differenceInCalendarDays` is 0 for any view inside one day.
+ *
+ * Not unconditional: the merge test is `gap <= this`, and this is always > 0 for
+ * a finite `pixelsPerHour`, so two sessions that abut exactly never split at any
+ * zoom. Tightening to `<` is not the answer — a zero gap is exactly what makes
+ * contiguous screentime sampling spans read as one bar.
+ *
+ * Capped at the previous fixed floor so a very wide container cannot merge
```
fiddur: The rationale here is inverted. The cap binds when `pixelsPerHour` is **low** — `(7/pph)*3.6e6 > 6e5` ⟺ `pph < 42` — i.e. a *narrow* container, or a range stretched toward the 2-boundary tier limit. A very *wide* container gives high pph and therefore a gap *smaller* than 10 min, which never reaches the cap at all.

So what the cap actually protects is e.g. a 3-day-elapsed view that still counts as 2 boundaries on a 360px phone (~5 pph → uncapped ~84 min), which would otherwise bridge far more than the old fixed floor. Suggest rewording to that effect.

apps/web/src/pages/Timeline/collapseTier.test.ts
```diff
@@ -1,6 +1,54 @@
...
+    const hourView = mergeGapForZoom(0, 1000)
+
+    expect(hourView).toBeLessThan(dayView)
+    // Two sessions 9 min apart stay merged at day view but separate zoomed in
+    expect(9 * MINUTE).toBeLessThan(dayView)
+    expect(9 * MINUTE).toBeGreaterThan(hourView)
+  })
+
+  it('never bridges more than the previous 10-minute floor', () => {
+    // A very wide container at day zoom must not merge more than it used to
```
fiddur: Same inversion as the JSDoc: `pph = 1` is a 1000px container showing ~1000 hours (~41 days), and `pph = 0.001` is more extreme still — these are maximally zoomed-*out* / narrow cases, not "a very wide container at day zoom" (which would be high pph and land well below the cap). The assertions themselves are right and do exercise the cap; only the stated scenario is wrong, and `days=0` + `pph=1` can't co-occur with "day zoom" anyway.

docs/features/timeline.md
```diff
@@ -151,6 +151,22 @@ The Timeline automatically adjusts detail level as you zoom:
...
+
+That count is an Activity-lane signal only. Screen Time suppresses it when it would just restate the bar's own category, which is the usual case there: the count would be raw sampling spans rather than sessions, and the bar links to its category rather than to a member. So a merged Screen Time bar gives no indication of how many spans it holds.
```
fiddur: The closing sentence is broader than the code. `categorize.ts` deliberately suppresses only the *self-restating* case, and its own comment spells out that a nested category still renders `Merged: Programming ×47` (tracked in #981) — so a merged Screen Time bar sometimes *does* show a span count. Worth carrying the "usual case" qualifier through to the conclusion, e.g. "So a merged top-level Screen Time bar gives no indication of how many spans it holds; a bar collapsed from sub-categories still lists them with counts."

fiddur (commented):
✅ Approved

The core change is sound and well reasoned.

**`mergeGapForZoom` logic checks out.** I traced every branch against the tests: `min(FINE_MERGE_GAP_CAP_MS, (7/pph)*3.6e6)` never exceeds the old fixed 10-minute gap, the `pph <= 0 || NaN` guard preserves the pre-measurement behaviour, `Infinity` yields 0 (and `gap <= 0` still merges abutting/overlapping spans), and the `days > 50` / `days > 2` tiers are byte-identical to what was inlined in `useTimelineNavigation`. The `84 pph → 5 min` assertion is a good catch on the author's part — without it the day-view test is satisfied by the cap alone and `MERGE_GAP_PX` would be unpinned.

**No screentime "comb" regression.** `screentime-activities.ts` already coalesces same-category records within 2 min per source, and cross-source spans for the same category overlap (negative gap), so they still fold at any zoom. Only spans genuinely >25s–10min apart split, and only once the zoom makes that gap visible — which is the intent.

**Dropping the trivial-provenance cleanup is provably safe.** I verified the invariant the comment claims: `collapsed_types` is set either at retype (line 474, where the recorded type is the *original* and the bar's type is the *effective* ancestor, so they differ by construction) or in the merge accumulation (which seeds from a count-1 fallback and then adds ≥1, so the total is always ≥2). `[{ type: <own type>, count: 1 }]` is unreachable. The property test plus the "reaches both single-entry provenance shapes" guard covers this properly — good that the guard was added, since with only `mixedActivities` the invariant would have passed on length alone.

**Screen Time parity is exact.** `restatesOwnType(a.collapsed_types, a.activity_type)` is the same predicate as the deleted `delete m.collapsed_types` condition, so the Screen Time lane behaves identically to before; only the Activity lane gains the ×N. The legacy `screentime` umbrella path never carries provenance (`mergeAdjacentByKey` doesn't set it), so it is unaffected either way.

Also checked and fine: `pixelsPerHour` is now shared between `mergeGapMs` and `collapseDepth` without introducing a render loop (`getDefaultViewStart/End` quantize to day boundaries, so the value is stable across renders); `timeAxisPixels` derives only from container size, so there is no data→layout→data feedback; `onZoomEnd` (not the per-frame `zoom` handler) drives `handleZoom`, so the now-continuous merge gap recomputes once per gesture rather than per frame.

Remaining notes are comment/doc accuracy only — the reasoning attached to the cap is inverted, and the docs paragraph over-generalises the Screen Time suppression relative to what `categorize.ts` itself says. Nothing blocking.

apps/web/src/pages/Timeline/collapseTier.ts
fiddur: You are right, my reasoning was inverted. Checked it numerically: the cap binds for `pph < 42` — at 5 pph the raw value is ~84 min, at 1 pph ~420 min, while 84 pph gives 5 min and 1000 pph gives ~25s, neither reaching the cap.

So it protects the narrow-container / stretched-range case, exactly your 360px 3-day example, not a wide container. Reworded with that example and the ~42 pph threshold.

apps/web/src/pages/Timeline/collapseTier.test.ts
fiddur: Same error, and doubly wrong as you note — `days = 0` cannot co-occur with 1 pph anyway. Reworded to say the cap binds at *low* pixels-per-hour, with 1 pph described as a 1000px container showing ~1000 hours. Assertions unchanged, since those were right.

docs/features/timeline.md
fiddur: Fixed, using your wording — "a merged **top-level** Screen Time bar gives no indication of how many spans it holds; a bar collapsed from sub-categories still lists them with counts", with a link to #981.

The conclusion had drifted broader than the qualifier I put two sentences earlier, which is exactly the inconsistency #981 exists to resolve.

fiddur (commented):
✅ Approved

Solid, well-reasoned fix. I traced the three behavioural claims and they all hold:

**Pixel-tier merge gap.** `mergeGapForZoom` reproduces the old fixed floor exactly where it mattered: a 24h view on a 1000px axis is ~41.7 pph → 7px ≈ 604800 ms → capped to 600000, i.e. bit-identical to the previous `10 * 60 * 1000`. Narrow containers (a 600px phone axis at day view is ~25 pph) also land on the cap, so nothing regresses on mobile. The `0 / -1 / NaN` fallback returns the same constant, so pre-mount and pre-measure rendering is unchanged. Zoomed in the gap shrinks monotonically, which is the reported bug.

**Removing the trivial-provenance cleanup.** I verified the invariant the comment asserts. The retype branch only attaches `collapsed_types` when `effective !== a.activity_type`, so the recorded type is strictly different from the survivor's; the merge branch seeds from `prev.collapsed_types ?? [{ prev type, 1 }]` and then adds ≥1 more, so any single-entry list from that path has count ≥ 2. `[{ own type, count: 1 }]` is genuinely unreachable, and the two new tests (`never yields…` plus the `reaches both single-entry provenance shapes` guard-the-guard) cover both single-entry shapes — without that second test the first would pass on list length alone, which the comment correctly calls out.

**Screen Time suppression.** Correctly scoped. The legacy umbrella path goes through `mergeAdjacentByKey`, which never sets `collapsed_types`, so it is untouched. The nested-category gap is honestly documented in the code and in `docs/features/timeline.md` with #981 tracking it.

Two things I checked that could have bitten and don't: `mergeGapMs` never reaches `fetchProductivity`'s `merge_gap_ms` query param from the Timeline, so the now-continuous value can't explode the query cache; and `screentime-activities.ts` already coalesces source records within a 2-minute gap server-side, so the fine tier only begins splitting screentime spans below roughly 4.8h of visible range on a 1000px axis — right about where a 2-minute gap crosses 7px. The "comb of identical slivers" the old comment warned about therefore doesn't come back at day view.

Non-blocking notes:

1. `mergeGapMs` is now an unquantized float driven by container pixels — see the inline comment.
2. The coarse tiers stay `differenceInCalendarDays`-based while the fine tier is pixel-based, so a 3-boundary view jumps from ≤10 min to 60 min regardless of container width. That discontinuity is pre-existing (it was 10 min → 60 min before) and the docs explain the boundary counting well, but the coarse tiers are now the odd ones out relative to #658's "width, not days" principle. Possible follow-up.
3. `collapseTier.test.ts` — the test named `never bridges more than the previous 10-minute floor` states a global invariant the function doesn't have (it bridges 1h and 4h in the coarse tiers). The inline comment scopes it to low pph correctly; only the name over-promises.
4. `activityMerge.test.ts` — the comment on `never yields a provenance list that only restates the bar itself` calls the `count === 1` shape "what the removed trivial-provenance cleanup used to guard". The removed loop actually dropped *any* count, including `[{ running, 2 }]`, which is precisely the case this PR now deliberately keeps. Small imprecision in an otherwise carefully-annotated test.
5. Cosmetic: for legacy `activity_type = 'exercise'` rows, `getActivityMeta` labels the bar from `activity.title || 'Workout'`, so a depth-0 merge now renders a bar titled e.g. "Morning gym" with a "Merged: Exercise ×2" line. The two names won't match, but the count is still the useful signal, so this reads as informative rather than wrong.

apps/web/src/pages/Timeline/collapseTier.ts
```diff
@@ -41,3 +41,43 @@ export const computePixelsPerHour = (
...
+ * 7px would otherwise work out to far more than 10 minutes: a 3-day-elapsed view
+ * that still counts as 2 boundaries on a 360px phone is ~5 pph, i.e. ~84
+ * minutes. Above that the pixel gap is already under 10 minutes and the cap never
+ * applies.
+ */
+export const mergeGapForZoom = (days: number, pixelsPerHour: number): number => {
+  if (days > 50) return 4 * 60 * 60 * 1000
+  if (days > 2) return 60 * 60 * 1000
+  if (Number.isNaN(pixelsPerHour) || pixelsPerHour <= 0) return FINE_MERGE_GAP_CAP_MS
+  return Math.min(FINE_MERGE_GAP_CAP_MS, (MERGE_GAP_PX / pixelsPerHour) * 3_600_000)
```
fiddur: Non-blocking: this returns an unquantized float, so every 1px resize step produces a new `mergeGapMs`. In `useTimelineData` that value is a dep of both the `activities` memo (`collapseToParentType`, which sorts + copies) and the `screentimeActivities` memo, so a window drag now re-runs the whole collapse on each `ResizeObserver` tick and hands downstream a fresh array identity every time. Previously `mergeGapMs` only changed when the view range crossed a day tier.

`collapseDepthForPixelsPerHour` was deliberately quantized in #658 for this reason. Rounding here — e.g. `Math.round(... / 1000) * 1000` — would keep memo stability across small resize steps without changing any user-visible merge decision, since the threshold is compared against millisecond gaps that are never that precise.

Small effect at typical activity counts, so fine to leave; flagging it because it's the one thing the pixel-driven gap changes that the tests can't see.

fiddur: Real regression, fixed — thanks for catching it. Rounded to whole seconds, the same treatment #658 gave `collapseDepthForPixelsPerHour`, and for the same reason. Activity gaps are never sub-second-precise so no merge decision changes, and a test sweeps 1000→1003px asserting one identical value comes out.

fiddur (commented):
✅ Approved

Solid, well-reasoned change. I traced the three moving parts and they hold up:

**`mergeGapForZoom`** — the coarse tiers are unchanged (`days > 50` → 4h, `days > 2` → 1h), and the new pixel tier can only ever produce a gap *smaller* than the old fixed 10-minute floor (`Math.min` against `FINE_MERGE_GAP_CAP_MS`, and `600000 / 1000` rounds back to exactly `600000`, so the cap isn't overshot). That makes the change strictly less merge-happy, which is the right direction for a "bars won't separate" bug and means no existing view can suddenly weld more together. The pre-measurement fallback (`pixelsPerHour <= 0` → 10 min) preserves the old value, so nothing shifts before the first `ResizeObserver` tick. Tier transitions are monotonic too: 3 boundaries → 1h, 2 boundaries → ≤10 min, so zooming in never widens the gap.

**Memoization** — hoisting `pixelsPerHour` into its own memo and feeding both `collapseDepth` and `mergeGapMs` from it is a clean simplification, and both memos' dependency lists are complete. The quantization matters here: `pixelsPerHour` is a churning float, but both derived memos return *primitives*, so the `collapseToParentType` memos in `useTimelineData` (keyed on the `mergeGapMs` value) don't re-run on every pixel. On a desktop-width container `Δgap ≈ gap/px ≤ 600000/800 = 750ms` per pixel, so second-quantization genuinely absorbs most single-pixel ticks.

**Dropping the trivial-provenance cleanup** — the replacement reasoning is correct. `collapsed_types` is set on exactly two paths (`activityMerge.ts:474` retype, where the recorded type differs from the survivor's by construction; and `activityMerge.ts:492-497` accumulation, which always totals ≥ 2), and `Activity.collapsed_types` is documented as never arriving from the API, so `[{ ownType, 1 }]` is genuinely unreachable. I checked the flip side too: because `getActivities` runs `mergeOverlappingActivities` server-side, cross-source duplicates of the *same* session arrive as one row with `source_ids`, so the new `×N` counts on Activity-lane bars reflect genuinely distinct spans rather than double-reported sessions.

The tests earn their keep — the `restatesOwnType` suppression test really does exercise the derived-type path (end_time present, type in `screentimeDerivedTypes`, so it reaches the tooltip builder), and `'reaches both single-entry provenance shapes'` is the right guard against the invariant test passing vacuously on list length. Docs are accurate, including the `differenceInCalendarDays` boundary-vs-elapsed-hours distinction and the known nested-category gap tracked in #981.

One comment-accuracy nit inline; nothing blocking.

apps/web/src/pages/Timeline/collapseTier.ts
```diff
@@ -41,3 +41,51 @@ export const computePixelsPerHour = (
...
+ *
+ * Zoomed out, fixed tiers bridge large gaps so a long string of small same-type
+ * activities reads as one bar. At views crossing <= 2 calendar-day boundaries
+ * (so anything up to just under 72 elapsed hours) the gap instead tracks
+ * pixels-per-hour: only gaps too small to see on screen are bridged, so zooming
+ * in pulls separate sessions apart — two yoga sessions nine minutes apart used
+ * to stay welded together at every zoom level, because
+ * `differenceInCalendarDays` is 0 for any view inside one day.
+ *
+ * Not unconditional: the merge test is `gap <= this`, and this is always > 0 for
```
fiddur, line 66: Two claims in this function no longer hold after the quantization added in the last commit:

1. "**this is always > 0 for a finite `pixelsPerHour`**" — `Math.round(gap / 1000) * 1000` on line 90 returns `0` once `gap < 500ms`, i.e. `pixelsPerHour > 50400`. That's reachable: `useTimelineZoom` uses `scaleExtent([0.1, 20000])`, so a ~3-day base scale can zoom to a sub-2-minute window. The *conclusion* is still right (threshold `0` with `gap <= 0` still merges abutting spans, and `docs/features/timeline.md` correctly says the threshold "bottoms out at zero"), so this is only the intermediate premise being stale.

2. "**Activity gaps are never sub-second-precise, so no merge decision changes**" — rounding *up* can flip a decision even for whole-second gaps. Unrounded threshold `24600ms` with a gap of exactly `25s`: unrounded splits, rounded (`25000`) merges. Harmless in practice (sub-500ms of threshold at one specific zoom), but the comment overstates it.

Both are cosmetic — worth a word-tweak rather than a code change, e.g. "> 0 except at extreme zoom, where it floors to 0 — and `gap <= 0` still merges abutting spans" and "shifts the threshold by at most 500ms".

fiddur: Correct — the quantization I added last commit made that premise stale, since Math.round(gap / 1000) * 1000 hits 0 above ~50400 pph, which scaleExtent([0.1, 20000]) does reach.

The conclusion still holds though: the test is gap <= threshold and abutting activities have gap === 0, so they merge either way. Since this is a stale premise rather than a behaviour bug, I have recorded it in #981 next to the abutting-sessions item it belongs with, rather than adding another review round to a PR that has had ten.

fiddur, line 66: This sentence went stale when the quantization was added below. `Math.round(gap / 1000) * 1000` returns **0** once `pixelsPerHour` exceeds ~50,400 (`7 / pph * 3.6e6 < 500`), which the horizontal `scaleExtent([0.1, 20000])` in `useTimelineZoom.ts:112` can reach — a 72h fetch range at k=20000 is a ~13s window, i.e. ~278,000 pph. So the threshold is not "always > 0 for a finite `pixelsPerHour`"; it's `> 0` only below that zoom.

The conclusion you're drawing from it survives (`0 <= 0`, so exactly-abutting spans still merge at any zoom), and `docs/features/timeline.md` already words it correctly — "the threshold bottoms out at zero". It's just the premise here that now contradicts both the code and the docs, and someone could later rely on the `> 0` guarantee. Suggest either rewording to "never negative, and 0 only at extreme zoom", or clamping with `Math.max(1000, …)` if you'd rather make the claim true.

fiddur (commented):
✅ Approved

Solid, well-reasoned change. I traced all three moving parts and they hold up:

**`mergeGapForZoom`** — arithmetic checks out against every test assertion (42 pph is exactly where the 10-min cap starts binding; 84 pph → 5 min uncapped; `Infinity` → 0). The `pixelsPerHour <= 0 / NaN` fallback reproduces the old fixed 10-minute gap, so pre-measure renders are unchanged. Folding `computePixelsPerHour` into one memo shared by `mergeGapMs` and `collapseDepth` is a clean simplification, and since both consumers return primitives the extra memo doesn't add churn downstream.

**Removing the trivial-provenance cleanup** — the stated invariant is genuinely airtight for the current call graph. `collapsed_types` has exactly two producers (the retype branch, where `effective !== a.activity_type` by construction, and the merge accumulator, whose total is always ≥ 2), the field is never populated by the API (confirmed: no `collapsed_types` anywhere in `packages/api-spec` or `apps/backend`), and `collapseToParentType` is never applied twice to the same list (`useTimelineData.ts:256` and `:309` both consume raw `allActivities`). So no survivor can carry `[{ own type, 1 }]`. Provenance counts also stay accurate through multi-way merges — `incoming` always totals 1 because `current` is always an unmerged activity, so the accumulator is O(N), not O(N²).

**`restatesOwnType`** reproduces the removed predicate exactly (no count check), so the Screen Time lane's behaviour is byte-for-byte what it was while the Activity lane gains the `×N`. The narrowness is deliberate and documented, with the nested-category gap tracked in #981.

The new tests earn their keep — `reaches both single-entry provenance shapes` guarding `never yields a provenance list that only restates the bar itself` against vacuous passing is exactly the right instinct, and `pins the pixel gap itself, above the cap` correctly notices that the day-view assertion alone doesn't constrain `MERGE_GAP_PX`.

Two non-blocking notes:

- One stale sentence in the `mergeGapForZoom` docstring (inline comment below) that now contradicts both the code and `docs/features/timeline.md`.
- Behavioural consequence worth being aware of: `index.tsx` derives its own `pixelsPerHour` from the *live* `currentXScale`/`currentYScale` (lines 533/739), which updates during a zoom gesture, whereas the hook's value only updates on `onZoomEnd`. So mid-gesture the bars keep the pre-gesture merge gap and visibly re-split when the gesture completes. That's inherent to memoizing the collapse outside the draw loop and `collapseDepth` has behaved this way since #658, so it's not new in kind — just newly visible within a single day now that the gap actually varies there.

Also worth noting for expectation-setting rather than as a defect: because of the 10-minute cap, the reported 9-minutes-apart case is only separated at the default day view when the time axis exceeds ~1120px — i.e. horizontal desktop. On a portrait phone (vertical, ~700px axis → ~29 pph) the pair still reads as one bar until the user zooms. `docs/features/timeline.md` is honest about this ("read as one bar in the day view and as two once you zoom in"), so the docs and the code agree.

apps/web/src/pages/Timeline/collapseTier.ts
fiddur: Same finding as the previous round, already recorded in #981 with your figures (rounding reaches 0 above ~50400 pph; scaleExtent([0.1, 20000]) gets there). The conclusion is unaffected — gap <= threshold with gap === 0 merges either way — so this is a stale sentence rather than a behaviour bug, and it is queued with the abutting-sessions work it belongs to. Merging rather than adding an eleventh round for a comment.

PR #980, opened 2026-08-03
fix(backend): bound min/max over long series, log background failures

fiddur:
## Symptom

After a wide timeline zoom-out, the backend stopped answering **anything** for hours — `/api/health` timed out while `/` still returned 200 (that's the static bundle, not the backend), and MCP was down too since it shares the process. One error in the logs:

```
RangeError: Maximum call stack size exceeded
    at computeHrFromSeries (…/activity-summary-metrics.ts:110:55)
    at computeActivitySummaryMetrics (…:159:5)
    at enrichActivity (…/activities.ts:102:8)
    at queryActivities (…/activities.ts:196:21)
```

## What this PR fixes

`Math.max(...hrValues)` passes one argument per element, so it throws above ~65k — the size a per-second heart-rate series reaches in a day or two. `maxOf`/`minOf` loop instead, bounded by memory rather than the call stack. A unit test asserts `Math.max(...)` genuinely throws `RangeError` at 200k elements and that the replacements return the right answers.

Swept every call site whose input scales with data volume:

| File | What was spread |
|---|---|
| `activity-summary-metrics.ts:110` | HR series (the crash) |
| `daily-summary.ts:266,405` | metric values, heart rates |
| `period-summary.ts:128` | period metric values |
| `metrics.ts:133-137` | bucket values and timestamps |
| `chart-svg.ts:67-70` | chart extents |
| `scatter-svg.ts:87-90` | scatter extents |
| `locations.ts:103` | cluster distances |
| `sync-router.ts:785` | health-connect timestamps |
| `feed-images.ts:88-91` | route bounding box |

The two remaining spreads (`activities-router.ts:93`, `activity-detail.ts:157`) are over a merge group — single digits — so they stay.

### One behaviour change worth noting

`minOf`/`maxOf` return `undefined` rather than the ∓`Infinity` that `Math.max()` yields with no arguments, so an empty series can't silently store ±Infinity as a stat. `buildChartSvg` was *relying* on that ±Infinity to blank its axis labels for an empty series — an existing test caught it. It now checks `undefined` explicitly, since labelling an empty chart "0" would be worse than blank.

## Process guards — and a correction

[...]

PR #984, opened 2026-08-04
perf(activities): window summary metrics by binary search, not a rescan

fiddur:
## The hot path

`queryActivities` fetches one time-series per summary metric for the **entire requested span** (`activities.ts:181-183`), then `computeActivitySummaryMetrics` called `inRange` — a full `points.filter(...)` — once per activity **per metric**, across 8 metrics:

```ts
const window = (metric) => inRange(series[metric] ?? [], activity.start_time, end)
```

So enrichment was `O(activities × 8 × series)`, synchronous on the event loop, with two array allocations per metric per activity (`filter` then `flatMap` in `positiveValues`).

That is what stopped the backend answering anything after a wide timeline zoom-out. Nothing crashed — one request held the only thread long enough that `/api/version` never got a turn, and the browser's retries queued more of the same. The `RangeError` in that incident was a separate symptom, fixed in #980.

## The change

`windowBounds` binary-searches the `[lo, hi)` index range, which is valid because `getTimeSeriesMultiMetric` orders by `(metric, time)`. The aggregates (`meanPositive`, `maxPositive`, `elevationGainLoss`, body-battery first/last) now read that range in place instead of slicing.

Per activity per metric: `O(series)` with two allocations → `O(log series + window)` with none.

## Behaviour is unchanged

All 13 existing tests pass untouched. The new ones cover the parts binary search gets subtly wrong:

- both edges inclusive (the old predicate was `t >= start && t <= end`, so an activity's first and last sample must not drop out);
- empty ranges before, after, and *between* points;
- an empty series;
- **parity with the old scan**: `points.slice(lo, hi)` equals `points.filter(...)` for every pair of offsets, including ones that land between samples;
- the outage shape — a 200k-point series (~2.3 days at 1 Hz) windowed by 500 activities.

## What this does *not* address

The **fetch volume**. A months-wide span still asks the DB for every summary-metric row in it and materialises them as JS `[Date, number]` pairs, and no amount of windowing helps once that is tens of millions of rows.

I have deliberately not built bucketing or a span guard for it yet, because the evidence points elsewhere for the reported incident: the `RangeError` was `>65k` HR points inside a *single* activity window, i.e. one activity spanning 18h+ at 1 Hz, and `hasExerciseLike` only matches `activity_type === 'exercise'` exactly — not subtypes — so the wide fetch is narrower in practice than it looks. Worth measuring a real year-wide request against this branch before designing for it, rather than guessing at the shape.

`pnpm check` clean; 2296 backend unit tests pass.

### DIFF
--- a/apps/backend/src/services/queries/activity-summary-metrics.test.ts
+++ b/apps/backend/src/services/queries/activity-summary-metrics.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, test } from 'vitest'
 
-import { computeActivitySummaryMetrics, type SummaryMetricSeries } from './activity-summary-metrics.ts'
+import {
+  computeActivitySummaryMetrics,
+  type SummaryMetricSeries,
+  windowBounds,
+} from './activity-summary-metrics.ts'
 
 const start = new Date('2024-01-15T10:00:00Z')
 const end = new Date('2024-01-15T10:10:00Z') // 10 minutes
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Solid, behavior-preserving optimization. I traced each replaced helper against its predecessor and they agree for all well-formed inputs:

- `windowBounds` reproduces the old `t >= start && t <= end` filter exactly — `lowerBound(start)` is inclusive at the left, `lowerBound(end + 1ms)` is inclusive at the right, and Date `>=`/`<=` were already numeric comparisons, so nothing shifts. The `+1ms` is safe because JS Dates carry no sub-millisecond precision.
- `meanPositive`/`maxPositive` sum and compare in the same order over the same `> 0` subset as `positiveValues` + `mean`/`maxOf`, so results are bit-identical. `computeHrFromSeries`'s early `hrValues.length === 0` return is correctly replaced by the per-field `!== undefined` guards.
- `elevationGainLoss(points, lo, hi)` starting at `lo + 1` keeps deltas strictly between in-window points, matching the old iteration over the filtered array; `hi - lo < 2` matches `points.length < 2`.
- `computeBodyBatteryFromSeries` reading `points[lo]` / `points[hi - 1]` matches the old first/last of the filtered array.

The sortedness precondition the binary search rests on holds at both call sites (`activities.ts` and `activity-detail.ts`): `getTimeSeriesMultiMetric` orders by `(metric, time)`, `splitMetricsByCumulative` partitions metrics into mutually exclusive groups so a metric's rows never come from two concatenated result sets, and none of `SUMMARY_METRICS` is in `cumulativeMetrics` anyway. Documenting that invariant on `lowerBound` is the right call. Tests are good — the linear-scan parity test over every offset pair is exactly the right way to pin a binary search's edges.

- **`computeBodyBatteryFromSeries` guard could be `hi <= lo`.** `windowBounds` can only return `hi < lo` when `end_time < start_time`, which nothing in the schema currently forbids. In that case `hi === lo` is false and `points[hi - 1]` with `hi === 0` throws a `TypeError` (500 for the whole timeline request), where the old `filter` just yielded `[]` and returned `{}`. Requires malformed data, so not a blocker, but a one-character hardening.
- **`hrZonesForActivity` in `activities.ts:75` still does the scan this PR removes.** `pointsInRange(ctx.summarySeries.heart_rate, …)` is a full filter over the same wide series, once per exercise activity — the same `O(activities × series)` shape, just 1× instead of 8×. Now that `windowBounds` is exported it drops in directly, and it's the natural next step if the event-loop stall isn't fully gone.
- **Public JSDoc on `computeActivitySummaryMetrics` ("pass full series — this function trims") no longer mentions the new hard requirement** that each series be ascending by time. Worth a line, since the function is exported and both callers happen to satisfy it implicitly.
- **The 200k-point test asserts correctness, not complexity.** Named "windows a long series without rescanning it per activity", but a reintroduced linear scan (500 × 200k ≈ 100M ops) would still pass it. Fine as a large-input correctness case; the name over-claims.
- **Some added test comments narrate history rather than intent** — e.g. "The pre-binary-search filter was …", "binary search is easy to get subtly wrong at the edges", "The outage shape: …". The parity test's structure already says this. The `windowBounds` docblock's outage paragraph is more defensible (it explains why a plain filter was rejected) but could be a sentence.

PR #986, opened 2026-08-09
fix(android): let the More tab button return to its hub

fiddur:
## Problem

On any page reachable from the **More** menu (Goals, Timeline, Settings, …) the
More icon in the bottom bar is marked selected and tapping it does nothing. The
hub is only reachable via the system back gesture.

`MoreScreen` owned the pushed sub-page in its own `remember`ed state, so
`onTabSelected(MainTab.More)` was a no-op when More was already the current tab.

## Fix

Hoist that sub-page into `AppState.moreDestination` and clear it in `selectTab`
whenever More is selected — so the button brings the menu up from anywhere the
menu isn't already up. `MoreScreen` becomes stateless (`destination` / `onSelect`
/ `onBack`).

The widget/notification deep link into a More page now seeds the destination at
`AppState` construction, which removes the `pendingMorePath` /
`onInitialPathConsumed` handshake. Behavior is unchanged: leaving More and coming
back still shows the hub.

## Tests

- `AppStateNavigationTest` — 8 tests over the tab/More rules (re-tap returns to
  the hub, returning from another tab shows the hub, deep-link seeding).
- `MoreNavigationUiTest` — Compose/Robolectric test driving the real `MainScreen`
  + `MoreScreen` + `AppState`: open a sub-page, tap **More**, hub is back.

Both fail with the one-line `selectTab` guard removed (verified), and the full
`./gradlew testDebugUnitTest` suite passes (22 suites, 0 failures).

https://claude.ai/code/session_01LatGBTZyYz271275Ks3HtE

### DIFF
--- a/apps/android/app/src/main/java/net/aurboda/AppState.kt
+++ b/apps/android/app/src/main/java/net/aurboda/AppState.kt
@@ -7,6 +7,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.LocalContext
+import net.aurboda.ui.screens.MoreDestination
 
 enum class AppScreen {
     Login,
@@ -24,12 +25,17 @@ enum class MainTab {
 class AppState(
     private val context: Context,
     initialScreen: AppScreen,
-    initialTab: MainTab = MainTab.Home
+    initialTab: MainTab = MainTab.Home,
+    initialMoreDestination: MoreDestination? = null
 ) {
     var currentScreen by mutableStateOf(initialScreen)
         private set
 
     var currentTab by mutableStateOf(initialTab)
+        private set
+
+    /** The sub-page pushed inside the More tab, or null while its hub is showing. */
+    var moreDestination by mutableStateOf(initialMoreDestination)
         private set
 
     var pendingServerUrl by mutableStateOf<String?>(null)
@@ -46,6 +52,7 @@ class AppState(
     fun logout() {
         CredentialsManager.clearCredentials(context)
         currentTab = MainTab.Home
+        moreDestination = null
         currentScreen = AppScreen.Login
     }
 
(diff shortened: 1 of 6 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped fix. Hoisting the More sub-page into `AppState.moreDestination` and clearing it in `selectTab` is the right shape: Material3's `NavigationBarItem` wraps `onClick` in `Modifier.selectable`, which fires unconditionally, so tapping an already-selected More does reach `selectTab(MainTab.More)` and pops to the hub.

I traced the state transitions and found no regressions:

- **Deep link (widget → `/goals`)** still opens the sub-page on cold start, and the previous "consume once" behaviour is preserved in effect — the destination survives a tab switch but `selectTab(MainTab.More)` clears it, so the user never sees a stale page. The old `pendingMorePath` + `LaunchedEffect` dance is gone, which is a nice simplification.
- **Back gesture** is unchanged: the `BackHandler` still only registers when a destination is showing, and `EmbeddedWebScreen`'s inner handler is still composed after it.
- **`logout()` clears `moreDestination`**, so a re-login can't land back on a sub-page — good catch, easy to miss.
- Rotation/recreate behaviour (intent re-read → deep link re-applied, no `rememberSaveable`) is identical to before, so no new issue here.
- The now-unused `@Suppress("ASSIGNED_VALUE_IS_NEVER_READ")` was correctly dropped from `MoreScreen` and correctly kept on `AurbodaApp`.

Tests look sound. `AppStateNavigationTest` covers the transitions directly, and `MoreNavigationUiTest` actually exercises the real bottom-bar path — I checked that its assertions can't pass accidentally: `AccountScreen` renders an exact `"Account"` heading and contains no `"More"` text (so `onNodeWithText("More")` resolves unambiguously to the bottom bar), and `MoreHub` recreates its `rememberScrollState()` on re-entry, so `"Timeline"` is genuinely back at the top. The Robolectric + `createComposeRule` setup matches the existing `LoginScreenTest` pattern. Docs updated too.

- `MainActivity.kt:185-186`: the deep-link comment now mostly restates what the `selectTab` comment in `AppState` already says; it could go.
- `AppState.kt` now imports `MoreDestination` from `ui.screens`, so app state depends on a UI package. Moving `MoreDestination` to a navigation-level file would tidy the layering, but it's a rename-scale change not worth a round here.

Approving — please don't push fixes for the two notes above on this PR; batch them with other work.

PR #988, opened 2026-08-11
fix(android): keep the soft keyboard from covering focused inputs

fiddur:
## Problem

Tapping a text field in an embedded web page (e.g. the note field on an activity) opens the keyboard **on top of** the field, with no way to scroll it up.

The app targets SDK 36. From **API 35 edge-to-edge is enforced and `adjustResize` is ignored**, so the window never shrinks for the IME — and nothing in the app consumed the IME inset. The WebView kept full window height and the keyboard simply drew over it.

## Fix

- **`AurbodaAppShell`** (new, extracted from `MainActivity.onCreate`) pads every screen by `WindowInsets.ime`, so the whole app — bottom bar included — sits above the keyboard. Consuming the inset also zeroes the navigation-bar inset the bottom bar would otherwise add on top, so it lands flush above the keyboard.
- Shrinking the WebView (rather than sliding it) is what makes it scroll the focused element into view; the injected viewport fix re-pins the page height to the new `window.innerHeight`.
- `android:windowSoftInputMode="adjustResize"` for API ≤ 34, where the window still resizes itself.

Native text fields (login, account) were covered the same way and are fixed by the same change.

## Test-suite isolation

`testOptions.unitTests.all { forkEvery = 1 }` — one JVM per test class.

Compose UI tests leak state between classes in a shared JVM: adding a third Compose test class made an *unrelated* class's `waitForIdle` spin until Espresso's 60s timeout, failing 12 tests that pass in isolation and picking a different victim on each run. With isolation the suite is deterministic — three consecutive full runs green, ~40s instead of ~10s (and instead of 12 minutes when it hung).

## Verification

- `KeyboardInsetsTest` covers both paths: the manifest declaration, and that the shell shrinks its content by the keyboard height. Both verified to fail when the corresponding fix is removed.
- Full Android unit suite (243 tests) green ×3; `assembleDebug` green.
- Not verified on a device — no adb device attached here.

https://claude.ai/code/session_01LatGBTZyYz271275Ks3HtE

### DIFF
--- a/apps/android/app/build.gradle.kts
+++ b/apps/android/app/build.gradle.kts
@@ -102,6 +102,14 @@ android {
   testOptions {
     unitTests {
       isIncludeAndroidResources = true
+
+      // One JVM per test class. Compose UI tests leak state between classes in a
+      // shared JVM — a later class's `waitForIdle` then spins until Espresso's
+      // 60s timeout, failing tests that pass in isolation and picking a
+      // different victim on every run.
+      all {
+        it.forkEvery = 1
+      }
     }
   }
 }
--- a/apps/android/app/src/main/AndroidManifest.xml
+++ b/apps/android/app/src/main/AndroidManifest.xml
@@ -99,12 +99,19 @@
         android:theme="@style/Theme.Aurboda"
         android:usesCleartextTraffic="true">
 
+        <!--
+            adjustResize keeps the soft keyboard from covering focused inputs on
+            API ≤ 34, where the window still resizes itself. From API 35 the flag
+            is ignored (edge-to-edge is enforced), so the app consumes the IME
+            inset itself — see AurbodaAppShell in MainActivity.kt.
+        -->
         <activity
             android:name=".MainActivity"
             android:exported="true"
             android:label="@string/app_name"
             android:launchMode="singleTop"
-            android:theme="@style/Theme.Aurboda">
+            android:theme="@style/Theme.Aurboda"
+            android:windowSoftInputMode="adjustResize">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
--- a/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
+++ b/apps/android/app/src/main/java/net/aurboda/MainActivity.kt
@@ -12,11 +12,14 @@ import androidx.compose.material3.AlertDialog
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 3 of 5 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped fix. I traced the two mechanisms against the actual app config and they compose correctly rather than fighting each other:

- `minSdk = 34` / `targetSdk = 36`, and nothing in the app calls `enableEdgeToEdge()` / `WindowCompat.setDecorFitsSystemWindows(false)`, and `Theme.Aurboda` doesn't opt out of decor-fitting. So on API 34 the decor still fits system windows: it consumes the IME inset before it reaches the `AndroidComposeView`, `WindowInsets.ime` reads 0, and `adjustResize` alone does the work — no double-shrink. On API 35+ edge-to-edge is enforced, decor stops consuming, and `AurbodaAppShell` applies the inset. Exactly one mechanism is live per API level, as documented.
- The KDoc claim that consuming the IME inset zeroes the nav-bar inset checks out: `NavigationBar` (MainScreen.kt) uses `Modifier.windowInsetsPadding(NavigationBarDefaults.windowInsets)`, which goes through Compose's `exclude(consumedInsets)` chain, so its bottom padding collapses to 0 while the keyboard is up and the bar sits flush. `Scaffold`'s own top/systemBars handling is unaffected.
- The doc claim about the WebView is accurate too — `VIEWPORT_FIX_JS` in EmbeddedWebScreen.kt registers a `resize` listener, so the shrink re-pins `<html>` to the new `window.innerHeight`.
- Modifier order `fillMaxSize().windowInsetsPadding(...)` keeps the node full-height and insets the content, which is what the second test asserts (root height − keyboard px == content height) — a real assertion, not a tautology.

The screens that matter are already scroll-hosts (`LoginScreen`, `AddDataScreen`, `MoreHub`, and the WebView), so switching from the previous `ADJUST_UNSPECIFIED` (which may have panned) to `adjustResize` doesn't strand a focused field. `AccountScreen` is a non-scrolling `Column`, but both its fields sit near the top, so nothing gets clipped there in practice.

Nothing blocking.

Please do **not** push fixes for these here; they're better batched with other work.

- `forkEvery = 1` is unrelated test-infra scope creep riding along (fair enough — the second Compose test class is what exposed the leak). One JVM per class will get progressively more expensive as the ~20 test classes grow; a narrower fix (isolating just the Compose classes, or tracking down the leaked state) would age better.
- The same two-mechanism explanation now lives in three places: the manifest comment, the `AurbodaAppShell` KDoc, and `docs/android-app.md`. The manifest comment could shrink to a one-line pointer at the doc.
- The shell test only proves `windowInsetsPadding` applies whatever inset is injected; the `WindowInsets.ime` default itself is untested, so a regression that swapped the default would still pass. Probably the best achievable under Robolectric — just noting the gap.

PR #989, opened 2026-08-18
feat: challenge leaderboard colour dots + Android challenge widget

fiddur:
## Summary

- **Web**: the public challenge leaderboard shows each member's chart colour as a dot before their name (same palette index as their race-chart line).
- **Android**: new home-screen **Challenge widget** (min 2×2, resizable). Placing it opens a picker of the user's hosted + active joined challenges; the widget shows that challenge's race chart (Canvas bitmap, same member palette as the web) and a leaderboard sized to the widget (always keeping the signed-in user's row), refreshed after every sync. Tapping it opens the challenge page in the app.
- **Deep links while running**: `MainActivity` now follows widget/notification deep links from `onNewIntent` too (previously only on a cold start), so the widget tap always lands on the challenge. `MoreDestination.Web` accepts absolute URLs so a challenge hosted on another instance opens embedded as well.
- **api-spec**: `/challenges`, `/challenges/participations/mine`, `/public/{username}/{slug}/standings` and `WellKnownAurboda` are registered in the OpenAPI doc so the Kotlin models exist for the widget.
- Docs: `docs/android-app.md` (Home-screen widgets, deep links), `docs/features/challenges.md`.

## Design notes

- Rendering (incl. network) runs in a WorkManager job (`ChallengeWidgetWorker`, unique `APPEND_OR_REPLACE`); the receiver never blocks.
- The challenge itself (name/unit/window) comes from the user's own instance; standings from the **hosting** instance's public standings endpoint (well-known discovery for remote hosts), so joined remote challenges work.
- Pure logic (URL parsing, series, leaderboard rows, layout per widget size, texts) is in `ChallengeWidgetModel.kt` and unit-tested; the chart and API functions have tests too.

## Test plan

- [x] `pnpm --filter aurboda-web check`, `pnpm --filter @aurboda/api-spec check`
- [x] `./gradlew compileDebugKotlin`, `./gradlew testDebugUnitTest` (269 tests, incl. 22 new)
- [x] Widget rendered under Robolectric at 2×2 / 4×2 / 3×3 sizes to eyeball the layout

https://claude.ai/code/session_01BxakEjLFANVr46YgL8uj29

### DIFF
--- a/apps/android/app/src/main/AndroidManifest.xml
+++ b/apps/android/app/src/main/AndroidManifest.xml
@@ -148,6 +148,31 @@
                 android:resource="@xml/hr_zone_widget_info" />
         </receiver>
 
+        <!-- Challenge Widget: race chart + leaderboard of one chosen challenge -->
+        <receiver
+            android:name=".widget.ChallengeWidgetProvider"
+            android:exported="true"
+            android:label="@string/challenge_widget_name">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+                <action android:name="net.aurboda.ACTION_UPDATE_CHALLENGE_WIDGETS" />
+            </intent-filter>
+            <meta-data
+                android:name="android.appwidget.provider"
+                android:resource="@xml/challenge_widget_info" />
+        </receiver>
+
+        <!-- Picks which challenge a newly placed challenge widget follows -->
+        <activity
+            android:name=".widget.ChallengeWidgetConfigActivity"
+            android:exported="true"
+            android:label="@string/challenge_widget_name"
+            android:theme="@style/Theme.Aurboda">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_CONFIGURE" />
+            </intent-filter>
+        </activity>
+
         <!-- Widget RemoteViews Service for scrollable goals list -->
         <service
             android:name=".widget.GoalsWidgetService"
--- a/apps/android/app/src/main/java/net/aurboda/AppState.kt
+++ b/apps/android/app/src/main/java/net/aurboda/AppState.kt
@@ -22,6 +22,25 @@ enum class MainTab {
     More
 }
 
+/**
+ * Where a widget or notification tap asks the app to go: a tab, and for the More
+ * tab optionally the web page to push (a site path like "/goals", or an absolute
+ * URL for a page on another instance).
+ */
+data class DeepLink(val tab: MainTab, val morePath: String? = null)
+
+/** Decode the [MainActivity.EXTRA_OPEN_TAB] / [MainActivity.EXTRA_MORE_PATH] extras; null when there is no link. */
+fun deepLinkFrom(openTab: String?, morePath: String?): DeepLink? {
+    val tab =
+        when (openTab) {
+            MainActivity.TAB_ADD -> MainTab.Add
+            MainActivity.TAB_FEED -> MainTab.Feed
+            MainActivity.TAB_MORE -> MainTab.More
+            else -> return null
+        }
+    return DeepLink(tab, if (tab == MainTab.More) morePath else null)
+}
+
 class AppState(
     private val context: Context,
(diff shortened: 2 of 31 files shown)


fiddur (commented):
✅ Approved

Solid, well-factored change. The pure widget logic (`ChallengeWidgetModel.kt`, `ChallengeChart.kt`, `ChallengeApi.kt`) is properly separated from the Android framework types and genuinely well unit-tested, network work is off the broadcast receiver in a `CoroutineWorker`, per-widget config is cleaned up in `onDeleted`, and the failure paths (signed out, unconfigured, challenge gone, host unreachable) all render an explanatory state rather than a blank widget.

I checked the things that could actually break and they hold up:

- **Colour parity is real.** `CHALLENGE_MEMBER_COLORS` matches `COLORS` in `PublicChallenge.tsx` element-for-element, and both sides index by position in the *active-filtered* standings list (`leaderboard()` / `series` in `challengeViews`, `standings.map((s, i) …)` on the web), so a member's dot, line and rank agree. On the web the `.filter(s => s.data.length > 1)` happens after the colour is assigned, so dropping a member with no buckets doesn't shift anyone's colour.
- **Endpoint contract matches.** `${apiBase}/public/{username}/{slug}/standings` is what `public-shares-router.ts` serves, and the new OpenAPI path sits under the `…/api` server base, matching what the web client hits. `WellKnownAurboda` registered as a bare component (no path) is the right way to get the federation model generated for Kotlin.
- **Deep links.** `singleTop` + `setIntent` + the `DeepLinkEvent` sequence number correctly re-navigates a running app, and `AppState.open` going through `selectTab(More)` (which clears `moreDestination`) before pushing the new page is the right order.
- **Layout/edge cases.** `planChallengeWidgetLayout` never returns more rows than members, always keeps one row when there is anyone to show, and `visibleRows` swaps the signed-in user into the last slot — all matching the tests. Empty standings, unparseable timestamps and unknown timezones all degrade instead of throwing.

Nothing blocking found. Approving.

Please don't push fixes for these here; a new round costs more than they're worth. Batch them with other work:

1. **Embedded WebView now loads third-party origins.** A widget deep link to a challenge hosted elsewhere puts a remote instance's page inside `EmbeddedWebScreen`. The token bridge is origin-scoped via `addDocumentStartJavaScript`, so in practice the remote page gets nothing — but `installAuthBridge`'s `addJavascriptInterface` fallback injects `AurbodaNative.getAuth()` into *every* frame, and `AuthBridge`'s KDoc justifies that with "no untrusted origin renders here", which is no longer true. With `androidx.webkit:1.12.1` and minSdk 34 the fallback is effectively unreachable, so this is defense-in-depth: consider skipping the bridge entirely (or passing an empty token) when the page being loaded isn't same-origin with `baseUrl`, and refresh that comment. Related: `EXTRA_MORE_PATH` now accepts absolute URLs, and `MainActivity` is exported — another app on the device can point the in-app WebView at any https page. Cheap fix would be to only honour an absolute `more_path` whose origin resolves to a challenge the user actually hosts/joined.
2. **Per-widget refetch.** `loadChallengeWidgetData` fetches `/challenges` and `/challenges/participations/mine` once per widget inside the worker's loop; with two challenge widgets that's four redundant authed round-trips per refresh. Hoisting both lists into `ChallengeWidgetWorker.doWork` and passing them in would make it one.
3. **`getJson` catches `Exception`**, which swallows `CancellationException` — a stopped worker turns cancellation into `DataResult.Error` and keeps going. `catch (e: CancellationException) { throw e }` before the generic catch.
4. **Bucket size is inferred rather than read.** `inferBucketMillis` takes the smallest gap between bucket starts; if every member's data is sparse (skipped days) the guess overshoots and the last point is plotted past its true bucket end. The host already resolves this as `effective_bucket_size` on the public challenge response — worth using it when available, with the inference as the fallback.
5. **Comment nit.** The added `{/* Same palette index as the member's line in the race chart above. */}` in `PublicChallenge.tsx` mostly restates the `COLORS[i % COLORS.length]` right below it.

PR #999, opened 2026-08-19
docs(fep): QuantPub FEP draft — federated personal metrics (#905)

fiddur:
Drafts a Fediverse Enhancement Proposal for an open, vendor-neutral ActivityPub vocabulary for quantified-self sharing, generalised from Aurboda's running federation code. Tracks #905 (docs-only, no code changes).

## What the draft covers (`docs/fep/quantpub.md`)

- **Vocabulary**: `quant:Exercise` (dual-typed `["Note", "quant:Exercise"]` with typed scalar `metrics`, a bounded `[start_time, end_time]` window) and a generic `quant:Observation` for sleep/HRV/steps/mood — Mastodon renders the Note fallback, implementing peers read the typed data.
- **The out-of-band structured-payload pattern** (the centrepiece): `/.well-known/quantpub` discovery, a structured post endpoint, and a bucketed public series endpoint with **data-driven authorization** (only explicitly shared series over shared windows resolve; everything else 404s), plus capability tokens for followers-only payloads (fediverse media/data fetches are unsigned, so capability URLs rather than HTTP signatures).
- **Privacy principles as normative RFC 2119 requirements**: scalars never imply series, series are a separate explicit opt-in, unshared = 404 (not 403), revocation immediate via `no-store`.
- **Three conformance levels**, framed so a home-built QS tool can implement Level 1 (well-known + one JSON endpoint) in a weekend, no ActivityPub stack needed.
- Concrete JSON examples, security considerations (SSRF, sanitisation, 404 oracle discipline), prior art (Open Pace, FitPub, Endurain, Wanderer, FEP-67ff, FEP-400e), and Aurboda listed as the existing (vendor-prefixed) implementation.

Also adds `docs/fep/README.md` explaining the directory.

## Open question: naming

The working name **QuantPub** (`quant:` prefix) is a placeholder for community input — alternatives listed in the draft are **Personal Metrics Vocabulary** and **MetricPub**. No FEP number is claimed (`fep-xxxx`).

## Next steps (outside this PR)

This draft is meant for discussion on r/QuantifiedSelf and SocialHub before any submission to codeberg.org/fediverse/fep (submission is a maintainer action, per #905).

Closes nothing; tracks #905.

### DIFF
--- a/docs/fep/README.md
+++ b/docs/fep/README.md
@@ -1,0 +1,15 @@
+# Fediverse Enhancement Proposal drafts
+
+This directory holds FEP drafts developed in the open from Aurboda's running
+federation code, before submission to the
+[FEP process](https://codeberg.org/fediverse/fep).
+
+- **[quantpub.md](./quantpub.md)** — _QuantPub: federated personal metrics for
+  quantified-self sharing_. A vendor-neutral generalisation of Aurboda's
+  federated feed: the `quant:Exercise` / `quant:Observation` vocabulary, the
+  out-of-band structured-payload pattern (well-known discovery, structured post
+  endpoint, data-driven public series endpoint, capability tokens), and the
+  privacy model as normative requirements. Tracked in
+  [issue #905](https://github.com/fiddur/aurboda/issues/905); the working name
+  (QuantPub vs. Personal Metrics Vocabulary / MetricPub) is open for community
+  input.
--- a/docs/fep/quantpub.md
+++ b/docs/fep/quantpub.md
@@ -1,0 +1,499 @@
+---
+slug: "xxxx"
+authors: Fredrik Liljegren
+status: DRAFT
+dateReceived: (not yet submitted)
+---
+
+# FEP-xxxx: QuantPub — federated personal metrics for quantified-self sharing
+
+> **Status of this document.** This is a pre-submission draft, developed in the
+> open in the [Aurboda](https://github.com/fiddur/aurboda) repository (issue
+> [#905](https://github.com/fiddur/aurboda/issues/905)) and intended for
+> discussion on r/QuantifiedSelf and SocialHub before any submission to the
+> [FEP process](https://codeberg.org/fediverse/fep). The FEP number is a
+> placeholder.
+
+> **Naming.** The working name is **QuantPub** with the namespace prefix
+> `quant:`. The name is open for community input; alternatives under
+> consideration include **Personal Metrics Vocabulary** and **MetricPub**.
+
+## Summary
+
+QuantPub is a small, vendor-neutral extension to [ActivityPub] for sharing
+personal quantified-self data — workouts, sleep, heart-rate variability, steps,
+mood — between federated instances **and** home-built personal tools, without
+sacrificing either mainstream-fediverse compatibility or the author's control
+over what leaves their instance.
+
+It has three parts:
+
+1. A tiny **vocabulary** (`quant:Exercise`, `quant:Observation`) for attaching
+   typed scalar summaries and a bounded time window to an ordinary AS2 `Note`,
+   so Mastodon-class servers render a readable status while implementing peers
+   read machine-readable data.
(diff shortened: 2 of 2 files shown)


fiddur (commented):
✅ Approved

Docs-only PR adding a pre-submission FEP draft plus a small index. Nothing here touches code, and the spec text matches the shipped implementation closely — I checked each "running code" claim: `/.well-known/aurboda` with `api_base` (`well-known-router.ts:60`), the data-driven series endpoint with its four-condition check and `Cache-Control: no-store` (`feed-public-router.ts:59`, `feed-series.ts`), the 5s server-side bucket floor (`MIN_SERIES_BUCKET_SECONDS`), `Note`-first dual typing with series links suppressed for followers-only posts (`activitypub/object.ts:219`, `seriesLinks`), `image_token` capability gating (`feed-capability.ts`), and SSRF-guarded best-effort enrichment (`timeline-enrich.ts`). The document is accurate about what Aurboda does today. Approving.

### Non-blocking — fold into the next revision of the draft, don't re-roll this PR

These are spec-quality points a FEP reviewer will raise; they're worth a follow-up commit *before submission*, not another round here.

1. **§7 never defines how a peer gets from an object id to the §5 endpoint** — arguably the one thing Level 3 can't be implemented without. Aurboda derives it from its own `/users/{user}/feed/{uuid}` layout plus `api_base` (`timeline-enrich.ts:33`, `enrichFromAurboda`), which is a vendor convention. As written, a Level 3 peer can only enrich posts from publishers that happen to share Aurboda's URL shape. Consider a normative rule, or better, a `quant:structured_url` property / link `rel` on the `Note` so the publisher states it.
2. **JSON-LD: the nested keys aren't in the `@context`.** `quant:metrics` entries use bare `key`/`value`/`unit`, and `quant:series` entries use a bare `metric`; none are term-mapped, so a strict JSON-LD processor expands them to empty objects and the payload is lost — the exact failure mode §1 is trying to design around. Either define the terms in the published context or state explicitly that these objects are plain JSON, not JSON-LD.
3. **AS2 already has `startTime` / `endTime` / `duration`.** Minting `quant:start_time` / `quant:end_time` / `quant:duration_seconds` alongside them will draw the obvious "why not reuse `as:`?" question. Relatedly, §1's blanket snake_case-on-the-wire rule conflicts with AS2/JSON-LD convention *and* with Aurboda's shipped `aurboda:startTime` / `aurboda:activityType` / `aurboda:durationSeconds` — so "adopt the `quant:` names" in the Implementations section is a case change, not just a prefix swap. Worth saying so, or dropping the case rule for extension properties (the metric *keys* being snake_case is uncontroversial).
4. **§8.4 under-specifies caching relative to your own implementation.** It requires `no-store` only on "the capability-gated and series endpoints", but a public/unlisted §5 payload is equally revocable (delete, or flip to followers-only) — Aurboda already sends `no-store` there (`feed-public-router.ts:94`). As written, an intermediary may keep serving a deleted post's structured payload, which contradicts §8.4's own "revocation is immediate".
5. **FEP-400e is mis-described.** It is "Publicly-appendable ActivityPub collections" (grishka, FINAL 2022-02-04) — about collections anyone may append to, not "extension object types layered on publicly-readable collections". Either re-describe it or cite a different FEP for the extend-don't-fork point.
6. **The two worked examples contradict a MUST.** The §5 payload and the `Create` example read as the same share (same title, window, distance, HR) but carry different metric sets — `hr_zone_minutes` vs `duration` — while §5 says `metrics` "MUST contain exactly the shared scalars — the same set the delivered `Note` summarised". Align the arrays or make them visibly different shares. While there: §2's `quant:series` row says "only for shared series", but §6.2 also requires public visibility (which is what `seriesLinks` implements) — worth adding to the note.
7. Minor: `docs/features/feed.md` is the existing federation doc and doesn't point at `docs/fep/`. A link each way would keep the docs tree navigable per CLAUDE.md's structure guidance.

Issue #996, opened 2026-08-19
Aurboda→Aurboda activity shares don't render native hover charts in the home timeline — investigate + fix


fiddur:
The whole mechanism for native interactive charts on Aurboda→Aurboda **activity** shares exists (structured enrichment → `timeline_entry.structured` → `TimelineStructured` renders a hoverable `TrendLineChart` per shared series), but on aurboda.net timelines they are **not appearing** for activity shares — cards fall back to the flat PNG / plain HTML. Articles got the native render with #971; activities are documented as working but observably aren't.

## Likely causes to check (in order)

1. **No series shared → no chart by design.** `structuredChartSeries` renders only from `structured.series`; a share with scalar metrics but no `series_metrics` opt-in yields `series: []` and renders nothing. If the tested share had no series opted in, the behaviour is correct — but then the card should probably still render *something* native (a typed scalar-stat row) instead of only the flattened HTML, so an Aurboda peer's card visibly differs from a Mastodon one.
2. **Old entries never retro-enriched.** Enrichment happens only on ingest (`Create`/`Update`) and backfill-on-follow. Timeline entries ingested before enrichment shipped keep `structured = NULL` forever. Consider a one-off backfill or lazy re-enrichment (e.g. on timeline read, N at a time, for entries whose `object_uri` matches the Aurboda path and `structured IS NULL`).
3. **Silent enrichment failure in production.** The enricher swallows every error to `null` with **zero observability** (`createAurbodaEnricher`). Failure candidates: the container fetching its own public origin (hairpin through nginx) for local/same-instance follows; `/.well-known/aurboda` discovery; `safeFetchGet` SSRF guard vs. the deployment's DNS; the 12s total budget. Add a debug log/counter on enrichment failure so this class of bug is diagnosable — today a failure is indistinguishable from a Mastodon post.

## Acceptance

- On aurboda.net: user B follows user A (both local, and ideally also cross-instance), A shares an activity **with a series opted in** → B's home timeline card shows the interactive hover chart.
- Enrichment failures are visible in logs.
- Decide + implement the fallback for cause 2 (retro-enrichment) if it turns out to be the gap.

PR #1000, opened 2026-08-19
feat(feed): personal message, native stat grid, activity-date line (#995, #997, #998)

### LINKED ISSUE #995: Activity shares: include the activity's notes / a personal message
A shared activity's delivered `Note` today contains only the generated title + the selected scalar summaries. The activity's own **notes/description** (the `notes` field — e.g. "felt great, negative splits") never leaves the instance, and there is no way to add a personal message when sharing. The human context is usually the most interesting part of a share.

## Scope

- Add an optional **`message`** (free text) to an activity feed post:
  - Share dialog: a text area, **prefilled from the activity's `notes`** (editable — prefill-and-edit doubles as the "include my activity description" opt-in while letting the user redact; an empty field shares nothing). Never auto-include notes without the user seeing them in the dialog first — notes can contain private context.
  - Schema: `message` on `shareActivityBodySchema` / `updateFeedPostBodySchema` in api-spec; same field on the `share_activity` / `update_feed_post` MCP tools.
  - Storage: a `message` column on `feed_posts` — **design it kind-agnostic** so a challenge share (#994) uses the same field.
- Delivery: render the message into the `Note`'s HTML `content` above the scalar summary (through the same sanitising/flattening path as today), so Mastodon followers see it. Include it in `FeedStructuredActivity` so Aurboda peers render it natively, and show it on the own-feed post card + public profile feed.
- Editing the post (PATCH / `update_feed_post`) can change the message; federates the usual `Update`.

## Out of scope

- Threaded comments / replies on posts (inbound `Create{Note} inReplyTo` — a much bigger feature).
- Auto-share rules interaction: an auto-shared post (#903) would use the activity notes verbatim or a rule-level template — decided there.

### LINKED ISSUE #997: Activity share presentation: native stat grid on Aurboda + readable multi-line text for Mastodon
A shared activity currently renders its data as one terse flattened line everywhere:

> Duration 2h 28m 15s · Avg HR 107 bpm · Max HR 158 bpm · HR zone minutes z0 13, z1 71, z2 61, z3 3, z4 0 · Calories 603 kcal · Avg stress 56 score

Hard to read on the own-feed card, the public profile feed, peer timelines — and in Mastodon.

## Scope

**1. Native structured stat display on Aurboda surfaces** (Strava-style: big value, small label — see e.g. Strava's Distance / Moving Time / Elevation header):

- Render the shared scalars as a **stat grid** instead of the flattened sentence, with sensible grouping (primary stats large; HR zones as their own compact bar/row rather than "z0 13, z1 71, …").
- Surfaces: the own **feed post card** (`FeedPostCard`), the **public profile** feed (`/u/:username`), and the **peer home timeline** for Aurboda posts (from `FeedStructuredActivity.metrics` — today the structured render only draws series charts and ignores the typed scalars; this also gives Aurboda peers a visibly native card even when no series was shared, cf. #996 cause 1).
- The typed data is already there on every surface (post + resolved scalars locally; `structured.metrics` on peer timelines) — this is presentation only.

**2. Readable text fallback for plain fediverse clients** (Mastodon can't render our components):

- Format the delivered `Note` HTML with structure instead of one `·`-joined line: title, then one line per stat (or small groups), HR zones on their own line in a compact human form. Mastodon keeps `<br>`/`<p>`, so linebreaks survive its sanitiser.
- Keep it modest — no emoji/ASCII-art tables; just line structure.

**3. Consistency:** the same formatter should feed the delivered Note, the object served at the AS2 id, and the outbox (they're already built from one place — keep it that way).

Note: an Aurboda peer's timeline card currently shows the delivered HTML *plus* the native render; once the native stat grid exists, suppress the redundant flattened text for `kind === 'activity'` the same way #971 did for articles (mind #973's stale-structured caveat).

### LINKED ISSUE #998: Shared activity posts should surface the activity's date, not just share time
A shared activity post is dated at **share time** (`published` = the post's `created_at`), so sharing a two-week-old activity shows "shared 3 Aug" with nothing saying the activity happened 2 Aug — and on the author's own card, "16 days ago" refers to the share, not the run.

## Options

**A. Backdate `published` to the activity time** (start or end). What the reader intuitively expects (Strava dates by activity). Caveats: Aurboda home timelines sort by `published_at`, so a late-shared activity lands buried in followers' past instead of at the top (Mastodon inserts at delivery regardless, only displaying the date); and an `Update` must keep the original `published`.

**B. Keep `published` = share time; display the activity window.** The Aurboda card (own feed, profile, peer timeline) gets a prominent "Activity · Sun 2 Aug, 15:44–18:12" line (peer timelines already receive `start`/`end` in `FeedStructuredActivity`; add the window to the public posts/feed responses for the other surfaces). For Mastodon, open the Note text with the activity date line (composes with the multi-line formatting issue). Timeline ordering and edit semantics stay untouched.

**Recommendation: B**, possibly with A as a refinement later for posts shared within ~a day of the activity. The confusion is about what the reader *sees*, and B fixes that on every surface without touching federation ordering semantics.

Decide at implementation — but either way the activity's date must be visible on the post.

fiddur:
First slice of the sharing-polish round from the 2026-08-19 tryout feedback: improve what an activity share *says and shows* before tackling the enrichment gap (#996).

## #995 — personal message on activity shares
- New nullable `message` column on `feed_posts` (idempotent additive migration).
- `message` on `ShareActivityBody` / `UpdateFeedPostBody` / `FeedPost` / `FeedStructured` (api-spec), REST + MCP wired through a shared `normalizeFeedMessage` (blank ⇒ clear/NULL, omitted ⇒ unchanged on PATCH).
- Share dialog: **Message** textarea, prefilled in create mode from the activity's description — the user-typed comments backing the edit "Description" field — always shown editable so nothing private auto-shares unseen; empty shares no text.
- Delivery: the message renders into the Note's HTML content (escaped, linebreaks kept) and rides `aurboda:message` on the AS2 object + `message` on the structured-enrichment payload, so Mastodon, the own card, and Aurboda peers all see it.

## #997 — presentation
- Federated Note HTML is now structured paragraphs: **headline / message / activity-date line / one stat per line** (`<br>`-separated) instead of the `·`-joined run-on sentence. Mastodon keeps `<p>`/`<br>`, so it reads cleanly there.
- `FeedPost` gains typed `metrics` (the resolved scalars, same `FeedStructuredMetric` shape as enrichment), and the own-feed card, public profile, and peer home timeline render a native Strava-style **stat grid** (`ActivityStatGrid`, HR zones as a compact row).
- An enriched activity timeline card now suppresses the redundant flattened HTML (mirroring the #971 article treatment, same #973 caveat), keeps the delivered **route-map** image (no native map render exists), and skips the chart PNG that the native interactive chart duplicates.

## #998 — activity date visible (option B)
- AS2 `published` stays the share time (timeline-ordering semantics untouched); instead every surface shows when the **activity** happened: `formatActivityWindow` renders `Sun, 2 Aug 2026, 15:44–18:12` (same-day collapse, cross-day spelled out) in the author's device timezone for federated text, and the web cards render it viewer-local.

## Tests & docs
- Unit: content layout (message escaping/linebreaks/order, one-stat-per-line, `aurboda:message`), window formatting (same-day/cross-day/open-ended/invalid-tz), stat splitting + zone naming, share-body message passthrough.
- Integration: message round-trip + patch semantics (replace/clear/keep) on `feed_posts`, message in `serializeFeedPost` DTO + federated content, message on the structured payload.
- `docs/features/feed.md` updated; api-spec regenerated (OpenAPI/TS/Kotlin).

Closes #995. Closes #997. Closes #998.

### DIFF
--- a/apps/backend/src/db/feed.integration.test.ts
+++ b/apps/backend/src/db/feed.integration.test.ts
@@ -113,6 +113,26 @@ describe('Feed posts integration', () => {
     expect((await getFeedPostById(user, created.id))?.image_token).toBe(created.image_token)
   })
 
+  test('stores, patches, and clears the personal message (#995)', async () => {
+    const user = getTestUser()
+    // Omitted on create → NULL.
+    const bare = await createFeedPost(user, postInput())
+    expect(bare.message).toBeNull()
+
+    const created = await createFeedPost(user, postInput({ message: 'Felt great,\nnegative splits!' }))
+    expect(created.message).toBe('Felt great,\nnegative splits!')
+    expect((await getFeedPostById(user, created.id))?.message).toBe('Felt great,\nnegative splits!')
+
+    // An undefined patch field leaves the message untouched…
+    const untouched = await updateFeedPost(user, created.id, { visibility: 'unlisted' })
+    expect(untouched?.message).toBe('Felt great,\nnegative splits!')
+    // …a string replaces it, and null clears it.
+    const replaced = await updateFeedPost(user, created.id, { message: 'Actually just ok' })
+    expect(replaced?.message).toBe('Actually just ok')
+    const cleared = await updateFeedPost(user, created.id, { message: null })
+    expect(cleared?.message).toBeNull()
+  })
+
   test('lists posts newest-first', async () => {
     const user = getTestUser()
     const first = await createFeedPost(user, postInput())
--- a/apps/backend/src/db/feed.ts
+++ b/apps/backend/src/db/feed.ts
@@ -26,6 +26,8 @@ export interface FeedPostRecord {
   include_chart: boolean
   /** Stored article payload (title + default window + blocks); null for `activity` posts. */
   article: ArticleContent | null
+  /** The author's personal message (plain text), or null when none was shared. */
+  message: string | null
   /** Unguessable capability token for `followers`-only image URLs (see schema). */
   image_token: string
   created_at: Date
@@ -39,6 +41,8 @@ export interface FeedPostInput {
   visibility: FeedVisibility
   include_map: boolean
   include_chart: boolean
+  /** The author's personal message; omitted/undefined stores NULL. */
+  message?: string | null
 }
 
 /** Input for creating an `article` post (no activity anchor / shared metrics). */
@@ -55,10 +59,12 @@ export interface FeedPostPatch {
   include_chart?: boolean
   /** Replacement article payload (whole `article` JSONB), for editing an article post. */
   article?: ArticleContent
+  /** Replacement personal message; `null` clears it, `undefined` leaves it unchanged. */
+  message?: string | null
 }
 
 const FEED_POST_COLUMNS =
-  'id, kind, activity_id, included_metrics, series_metrics, visibility, include_map, include_chart, article, image_token, created_at, updated_at'
+  'id, kind, activity_id, included_metrics, series_metrics, visibility, include_map, include_chart, article, message, image_token, created_at, updated_at'
 
 interface FeedPostRow {
   id: string
@@ -71,6 +77,7 @@ interface FeedPostRow {
   include_chart: boolean
   // pg parses a jsonb column to its JS value on read (null for `activity` posts).
   article: ArticleContent | null
+  message: string | null
   image_token: string
   created_at: Date
   updated_at: Date
@@ -82,8 +89,8 @@ export const createFeedPost = async (user: string, input: FeedPostInput): Promis
   const result = await query<FeedPostRow>(
     user,
     `INSERT INTO feed_posts
-       (activity_id, included_metrics, series_metrics, visibility, include_map, include_chart)
-     VALUES ($1, $2, $3, $4, $5, $6)
+       (activity_id, included_metrics, series_metrics, visibility, include_map, include_chart, message)
+     VALUES ($1, $2, $3, $4, $5, $6, $7)
      RETURNING ${FEED_POST_COLUMNS}`,
     [
       input.activity_id,
@@ -92,6 +99,7 @@ export const createFeedPost = async (user: string, input: FeedPostInput): Promis
       input.visibility,
       input.include_map,
       input.include_chart,
+      input.message ?? null,
     ],
   )
   return mapFeedPost(result.rows[0])
@@ -194,6 +202,7 @@ export const updateFeedPost = async (
   if (patch.include_map !== undefined) set('include_map', patch.include_map)
   if (patch.include_chart !== undefined) set('include_chart', patch.include_chart)
   if (patch.article !== undefined) set('article', JSON.stringify(patch.article))
+  if (patch.message !== undefined) set('message', patch.message)
 
   if (sets.length === 0) return getFeedPostById(user, id)
 
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -34,7 +34,7 @@ import {
 import { isPubliclyVisible } from '../services/activitypub/object.ts'
 import { buildArticleMarkdown } from '../services/article-export.ts'
 import { buildArticleContent, mergeArticleContent } from '../services/article.ts'
-import { serializeFeedPost } from '../services/feed.ts'
+import { normalizeFeedMessage, serializeFeedPost } from '../services/feed.ts'
 import { serializeFollower } from '../services/followers.ts'
 import { serializeFollowing } from '../services/following.ts'
 import { getTimelinePage } from '../services/timeline.ts'
@@ -77,6 +77,7 @@ export const registerFeedTools = (
         include_chart: body.include_chart,
         include_map: body.include_map,
         included_metrics: body.included_metrics,
+        message: normalizeFeedMessage(body.message) ?? null,
         series_metrics: body.series_metrics,
         visibility: body.visibility,
       })
@@ -95,6 +96,7 @@ export const registerFeedTools = (
         include_chart: body.include_chart,
         include_map: body.include_map,
         included_metrics: body.included_metrics,
+        message: normalizeFeedMessage(body.message),
         series_metrics: body.series_metrics,
         visibility: body.visibility,
       })
--- a/apps/backend/src/routes/feed-image-router.test.ts
+++ b/apps/backend/src/routes/feed-image-router.test.ts
@@ -34,6 +34,7 @@ const makePost = (overrides: Partial<FeedPostRecord> = {}): FeedPostRecord => ({
   include_map: true,
   included_metrics: [],
   kind: 'activity',
+  message: null,
   series_metrics: [],
   updated_at: new Date('2026-07-01T08:00:00Z'),
   visibility: 'public',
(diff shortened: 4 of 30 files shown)


fiddur (commented):
✅ Approved

Solid, coherent slice covering all three issues. I traced the `message` field end-to-end (schema → `feed_posts.message` column + idempotent `ALTER … ADD COLUMN IF NOT EXISTS` → `createFeedPost`/`updateFeedPost` params → `normalizeFeedMessage` shared by the REST router and both MCP tools → `serializeFeedPost` DTO → structured payload → federated `content`), and the tri-state semantics hold: `undefined` leaves a PATCH untouched, `''`/whitespace clears to NULL, anything else stores trimmed. `buildShareBody` deliberately always sends `message` so an emptied field clears on edit — good catch, and the test asserts exactly that.

Escaping is right where it matters: `escapeHtml(message).replaceAll('\n', '<br>')` escapes before injecting the `<br>`, the window label is escaped too, and the web renders `{post.message}` / `{structured.message}` as JSX text. `formatActivityWindow` guards an invalid stored IANA zone with a try/catch fallback to UTC, and `getSettings(...).catch(() => null)` keeps a settings failure from breaking content building or delivery.

The HomeTimeline render change is a net improvement rather than a regression: previously *any* `structured` payload suppressed all image attachments, so the route map was never shown for enriched activity posts; now it is, and only the `Heart rate` PNG is filtered — which matches the attachment name `imageAttachments` actually emits in `deliver.ts`. Suppressing the flattened `content` whenever `structured` is present is consistent with `TimelineStructured` now always rendering (the old `series.length === 0 → null` early return is gone, so there's no case that renders nothing). `structured.metrics` is required by `feedStructuredSchema` and validated on ingest in `timeline-enrich.ts`, so `ActivityStatGrid` can't be handed `undefined`. Docs in `docs/features/feed.md` are updated to match. Nothing blocking — approving.

- **Per-post settings query.** `serializeFeedPost` now calls `getSettings(user)` inside the per-record path, and both `GET /feed` and the public-profile listing do `Promise.all(records.map(serializeFeedPost))` — that's one extra settings round-trip per post on top of the existing per-post scalar query. Resolving the timezone once in the caller (or memoizing per request) would collapse it.
- **`buildCreateExercise` inputs not wired through.** The new `message` / `timeZone` fields are only supplied by the `object.test.ts` call sites; `buildFeedPostActivity` (the only production-shaped caller, itself currently unused outside tests) doesn't pass them, so the AS2 object-model representation would omit the message and render the date in UTC if that content-negotiated path ever goes live.
- **No `maxLength` on the textarea.** `shareActivityBodySchema.message` is `max(2000)`, and `defaultMessage` is prefilled from the joined activity comments, which can exceed that — the user then hits a generic validation error on Share. `maxLength={2000}` on the `<textarea>` would fail fast.
- **`feed-post-title` has no CSS rule.** In `FeedPostCard` it inherits `.feed-post-content p { margin: 0 0 0.4rem }`, but in `TimelineStructured` the new `<p>`s sit directly in `.timeline-structured` (flex column with `gap: 0.75rem`), so they also carry the default 1em `p` margins — the two cards will space differently.
- **Comment redundancy.** The prefill semantics are spelled out four times (the `defaultMessage` Props JSDoc, the `useState` comment, the `EntityDetail` call-site comment, and the dialog's own visible note); `buildShareBody`'s "always sent so an edit can clear" rationale appears in the `ShareSelection` JSDoc, an inline comment, and the test name. A few new test comments just restate the assertion below them (`// Omitted on create → NULL.`, `// The author's personal message rides the structured payload (#995).`).
- **Filter readability.** `entry.structured == null ? true : entry.structured.kind === 'activity' && img.name !== 'Heart rate'` is the same as `!entry.structured || (entry.structured.kind === 'activity' && img.name !== 'Heart rate')`.
- **Timezone-sensitive tests.** `activity-stats.test.ts`'s same-day/cross-day cases depend on the runner's local zone (`10:00Z`–`11:30Z` straddles midnight at UTC+13). Fine on a UTC CI box; pinning `TZ` would make them deterministic anywhere.

fiddur (commented):
✅ Approved

Solid, well-scoped slice covering all three issues. I traced the `message` field end-to-end (schema → `feed_posts.message` + idempotent `ALTER … IF NOT EXISTS` → `FEED_POST_COLUMNS` → `createFeedPost`/`updateFeedPost` → `serializeFeedPost` → `resolveStructuredContent` → `buildFeedNote`) and the semantics hold at every hop: `undefined` leaves a PATCH untouched, `''`/whitespace clears to NULL, and `buildShareBody` deliberately always sends the field so an emptied textarea actually clears a stored message. `normalizeFeedMessage` being shared by the REST router and the MCP tools keeps the API/MCP parity CLAUDE.md requires.

Rendering paths check out too. `feedPostContent`/`formatActivityWindow` escape both the message and the window label before injecting `<br>`, so no XSS is introduced; the native Preact renders (`ActivityPostBody`, `TimelineStructured`) pass `message`/`title` as text children, so remote-peer structured payloads stay escaped. The `try/catch` around an invalid `device_timezone` in `formatWindowInstant` is the right call given that value comes from an Android device. `buildFeedNote` is also what the object dispatcher (`federation.ts:400`) serves, so the delivered Note and the dereferenced object stay identical.

The `HomeTimeline` change is right on the tricky bit: before, *any* `structured` entry suppressed all image attachments, so filtering only `'Heart rate'` for `kind === 'activity'` is a strict improvement (route maps now show) and leaves the article branch's behaviour unchanged. The attachment names match `imageAttachments` in `deliver.ts` (`'Heart rate'` / `'Route'`).

- **Per-post `getSettings` call.** `serializeFeedPost` now awaits `getSettings(user)` inside the per-post branch, and `GET /api/feed` serializes *every* post (`listFeedPosts` is unpaginated). That's one extra settings query per post on top of the existing per-post scalar resolution. Worth hoisting to a per-request lookup or memoizing when this area is next touched; same for `buildFeedNote`.
- **Untitled-activity headline drifts between renders.** `ActivityPostBody` and `TimelineStructured` fall back to `'Shared activity'`, while `feedPostContent` falls back to `prettifyKey(activityType) + ' activity'` (e.g. "Exercise activity"). Both `FeedPost.activity_type` and the structured payload's `activity_type` are available, so the native renders could match the federated text.
- **No client-side length guard on the message.** `shareActivityBodySchema.message` is `.max(2000)`, but the textarea has no `maxLength` and the prefill is `getUserNotesContent(activity)` (all user comments joined). A long description prefills past the limit and the share fails with a 400 the user didn't cause; `maxLength={2000}` on the textarea would avoid it.
- **`.feed-post-title` has no CSS rule anywhere** — it renders bold only because of the nested `<strong>`. Relatedly, `TimelineStructured.tsx` now uses `.feed-post-message`/`.feed-post-window` from `FeedPostCard.css` without importing it; it works today only because `Feed/index.tsx` also pulls in `FeedPostCard`, whereas `TimelineArticle.tsx` sets the precedent of importing that stylesheet explicitly with a comment.
- **`BuildCreateInput.message`/`timeZone` are only exercised by tests.** `buildCreateExercise` is reached solely via `buildFeedPostActivity`, which has no production caller (grep finds tests only) and doesn't forward either field — so the new `aurboda:message` branch never fires on the deployed path. Fine as forward-looking work for the content-negotiated endpoint the module header mentions; just noting it isn't live yet.
- The three-line comment above `useState` in `ShareActivityDialog.tsx` and the three-line one above `defaultMessage=` in `EntityDetail/index.tsx` largely restate each other and the code — one of them would do.

PR #1002, opened 2026-08-19
docs(fep): QuantPub rev 2 — pre-submission checklist from #999 review (#905)

fiddur:
Second revision of the QuantPub FEP draft, folding in the full pre-submission checklist the #999 review left (recorded on #905). All seven items addressed:

1. **§7 object→payload resolution is now normative** — object-id convention as baseline + same-host-only `quant:structuredUrl` override (in-band properties can be dropped by typed consumers, so the convention stays primary).
2. **JSON-LD survival**: `quant:metrics`/`quant:series` declared as JSON literals (`@type: @json`) so strict processors keep the nested keys.
3. **AS2 reuse**: `startTime`/`endTime` reused, `quant:duration_seconds` dropped for the `duration` metric key; naming rule split (camelCase JSON-LD terms / snake_case payloads + metric keys); Implementations notes the case change for Aurboda's adoption.
4. **§8.4**: `no-store` required on *all* §5/§6 responses, public/unlisted included.
5. **FEP-400e** described accurately and tied to future challenge leaderboards.
6. **Examples aligned** with §5's exact-set MUST; §2 series row notes public visibility.
7. **Cross-links** between `docs/features/feed.md` and `docs/fep/`.

Docs-only. After this merges the draft should be ready to take to r/QuantifiedSelf / SocialHub (#905 step 3).

### DIFF
--- a/docs/features/feed.md
+++ b/docs/features/feed.md
@@ -583,3 +583,6 @@ These are known and intentional for the current implementation:
   base-URL identity model this reuses.
 - [Challenges](./challenges.md) — cross-instance federated competitions on the same
   `/u/:username` namespace.
+- [QuantPub FEP draft](../fep/quantpub.md) — the vendor-neutral generalisation of
+  this feed's federation pattern (vocabulary + out-of-band structured payloads),
+  drafted for the FEP process so other QS tools can interoperate.
--- a/docs/fep/README.md
+++ b/docs/fep/README.md
@@ -12,4 +12,5 @@ federation code, before submission to the
   privacy model as normative requirements. Tracked in
   [issue #905](https://github.com/fiddur/aurboda/issues/905); the working name
   (QuantPub vs. Personal Metrics Vocabulary / MetricPub) is open for community
-  input.
+  input. Aurboda's own implementation of the pattern is documented in
+  [docs/features/feed.md](../features/feed.md).
--- a/docs/fep/quantpub.md
+++ b/docs/fep/quantpub.md
@@ -106,8 +106,11 @@ The JSON-LD namespace for the `quant:` prefix is `https://w3id.org/quantpub#`
 
 The values of `quant:metrics` and `quant:series` are **JSON literals**: the
 published `@context` defines both terms with `"@type": "@json"` (JSON-LD 1.1),
-so conforming processors preserve the nested `key`/`value`/`unit`/`metric`
-objects verbatim rather than expanding (and losing) unmapped keys. Consumers
+so conforming processors preserve the nested objects verbatim — a metric's
+`key`/`value`/`unit` and a series link's `metric`/`mediaType`/`href` — rather
+than expanding (and losing) unmapped keys. This deliberately makes the
+AS2-looking terms inside a series link (`mediaType`, `href`) opaque to JSON-LD
+processing: a series entry is plain data, not an AS2 `Link` object. Consumers
 MUST treat these values as plain JSON, not as JSON-LD node objects.
 
 ### 2. `quant:Exercise` — a shared workout
@@ -117,14 +120,14 @@ A shared exercise is an AS2 object **dual-typed** `["Note", "quant:Exercise"]`
 status, while implementing peers recognise the second type and read the typed
 fields:
 
-| Property              | Type                    | Notes                                                                                     |
-| --------------------- | ----------------------- | ----------------------------------------------------------------------------------------- |
-| `startTime` (AS2)     | ISO 8601                | REQUIRED. Start of the bounded activity window — AS2's own property, reused               |
-| `endTime` (AS2)       | ISO 8601                | End of the window (RECOMMENDED) — AS2's own property, reused                              |
-| `quant:activityType`  | string                  | e.g. `running`, `cycling`, `meditation`                                                   |
-| `quant:metrics`       | array of metric objects | The scalar summaries the author chose to share (§2.1)                                     |
+| Property              | Type                    | Notes                                                                                      |
+| --------------------- | ----------------------- | ------------------------------------------------------------------------------------------ |
+| `startTime` (AS2)     | ISO 8601                | REQUIRED. Start of the bounded activity window — AS2's own property, reused                |
+| `endTime` (AS2)       | ISO 8601                | End of the window (RECOMMENDED) — AS2's own property, reused                               |
+| `quant:activityType`  | string                  | e.g. `running`, `cycling`, `meditation`                                                    |
+| `quant:metrics`       | array of metric objects | The scalar summaries the author chose to share (§2.1)                                      |
 | `quant:series`        | array of series links   | Links into the series endpoint — only for shared series on **publicly-visible** posts (§6) |
-| `quant:structuredUrl` | URL                     | OPTIONAL explicit link to the object's §5 structured payload (see §7)                     |
+| `quant:structuredUrl` | URL                     | OPTIONAL explicit link to the object's §5 structured payload (see §7)                      |
 
 There is deliberately no duration property in the vocabulary: duration is
 derivable from the window, and when the author chooses to share it as a stat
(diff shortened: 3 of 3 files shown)


fiddur (commented):
✅ Approved

Docs-only revision of the QuantPub FEP draft, plus two cross-links. I read the whole spec, not just the hunks, and checked the diff's claims against the shipped code.

The substantive changes hold up:

- **Vocabulary rename is complete.** No `quant:start_time` / `quant:activity_type` / `quant:duration_seconds` / `quant:observation_of` survive anywhere in `docs/`; §2, §3, and both JSON examples all use the new camelCase terms consistently, and the two-layer naming rule (camelCase JSON-LD terms vs. snake_case HTTP/metric keys) matches the actual §4–§6 payload shapes in the doc.
- **The §5 example now matches §2's rule** that `metrics` carries exactly the scalars the delivered `Note` summarised — previously the §5 payload listed `hr_zone_minutes` that the AS2 example didn't have. The numbers are internally consistent too (2340 s = the "Duration 39m" in `content`; sleep 27600 s = the 23:05→06:45 window).
- **The §8.4 strengthening is accurate against the implementation**: `feed-public-router.ts` already sets `Cache-Control: no-store` on the structured post, series, and sibling endpoints regardless of visibility, so the doc is describing running code rather than aspiration.
- **The "more than a prefix swap" note in Implementations is correct**: `services/activitypub/object.ts` really does mint `aurboda:startTime` / `aurboda:endTime` / `aurboda:durationSeconds`, so calling out the AS2 `startTime`/`endTime` substitution as migration work is right.
- Both new cross-links resolve (`docs/fep/quantpub.md` ↔ `docs/features/feed.md`), and the FEP-400e entry is now a factual description rather than the vague "informs the approach" line.

Nothing blocks the merge.

Both of these are consequences of the new normative text being more precise than the (unchanged) examples below it. They're worth a line each in the next revision of this draft, but pushing fixes here only costs a review round:

1. **The examples' `@context` doesn't do what §1 now says it does.** The new paragraph states that the published `@context` defines `quant:metrics` and `quant:series` with `"@type": "@json"`, but both examples inline only a bare prefix map, `{ "quant": "https://w3id.org/quantpub#" }`. Under that context a JSON-LD processor would expand the metric entries as node objects and drop the unmapped `key`/`value`/`unit` keys — exactly the failure the paragraph says the design avoids. Next rev, either reference the published context IRI or inline the term definitions (`"quant:metrics": { "@type": "@json" }`) so a reader copying the example gets conforming output.
2. **The example object id doesn't match the new §7 id convention.** §7 makes `{web_base}/users/{username}/feed/{post_id}` the normative baseline, but the example object id is `.../users/freja/feed/01J5X0/object` (the `/object` suffix), and the prose right after still says "a QuantPub peer recognises the id shape". As written the example only resolves via the new `quant:structuredUrl` override. Either drop the suffix in the example or have §7 acknowledge a trailing path segment.

Smaller: §5's payload keeps a top-level `duration_seconds` *and* a `duration` entry in `metrics` — fine as a payload-layer convenience, but §2's "there is deliberately no duration property" reads slightly oddly next to it, so a half-sentence saying the §5 payload layer still exposes it would help.

fiddur (commented):
✅ Approved

Docs-only revision of the QuantPub FEP draft. I read the full `docs/fep/quantpub.md` after the diff and cross-checked the internal references — §1's naming split, §2/§3 tables, §5/§6 payload shapes, §7 resolution, §8.4 caching and §9 tokens now line up, and the two examples match the §5 "metrics MUST contain exactly the shared scalars" rule (the `duration` key and the trimmed metric set in rev 3 fix the last mismatch). The added cross-links resolve correctly (`docs/fep/README.md` → `../features/feed.md`, `docs/features/feed.md` → `../fep/quantpub.md`).

I also checked the claim in the Implementations section against the running code: `apps/backend/src/services/activitypub/object.ts` really does emit `aurboda:startTime` / `aurboda:endTime` / `aurboda:durationSeconds`, so "more than a prefix swap" is accurate. All seven checklist items from the #999 review are genuinely addressed. No blocking issues — approving.

Please do **not** push fixes for these here; they'd only start another review round. Batch them into the next revision (e.g. the pre-submission pass for #905).

- **§9 tokens vs. §7's own premise.** §7 justifies the object-id convention as the reliable baseline precisely because typed AS2 frameworks drop unknown in-band properties — but for a followers-only post the capability token can *only* reach the consumer in-band, via `quant:structuredUrl`, i.e. through the channel the section just called unreliable. The id convention alone yields a tokenless URL and a 404. Aurboda's shipped code solves this by lifting the token out of the delivered image attachment URL (`capabilityTokenFrom` in `apps/backend/src/services/activitypub/timeline-enrich.ts`) — `attachment`/`Image.url` survives typed vocabularies where the extension property doesn't. That trick is the part that actually works, and it's worth spelling out normatively in §9.
- **Implementations: the id shape diverges too.** The note calls out the window terms, but rev 3 also changed the example ids — activity `…/01J5X0#create` with object `…/01J5X0`, whereas the shipped builder does the inverse (`id: input.postId` on the activity, `objectId = \`${input.postId}/object\`` in `object.ts`). Since §7's convention requires the *object* id to be the resolvable post URL, adoption implies an id-shape migration as well.
- **§1 JSON-literal paragraph.** It enumerates the preserved nested keys as `key`/`value`/`unit`/`metric`, but the `quant:series` entries in the example also carry `mediaType`/`href`. Also worth one sentence that declaring `quant:series` `@type: @json` deliberately makes those AS2-looking terms opaque to JSON-LD processors (they will not be interpreted as `Link` objects).
- **§5 lost the keyed-record example.** Dropping `hr_zone_minutes` aligns the example with the exact-set MUST, but it was the only illustration of the record-valued form §2.1 permits. §2.1 still documents it, so there's no contradiction — just consider showing the record form somewhere.
- **FEP-400e citation.** Worth verifying the exact title casing and the finalization year against the FEP index before submission; "FINAL 2022" reads as the finalization date and may be the receipt year.
- **Table separators in §2/§3** are ragged relative to their header rows. `oxfmt` doesn't touch Markdown, so nothing enforces it — purely cosmetic.

fiddur (commented):
✅ Approved

Docs-only change (FEP draft rev 4 + two cross-links). No code, schema, or API surface touched, so nothing here can break behaviour — and the substantive claims the draft makes about Aurboda's running code all check out against the source:

- `capabilityTokenFrom` exists and is wired into the enrichment path (`services/activitypub/timeline-enrich.ts:75`, used at `federation.ts:170`).
- The shipped extension really does mint `aurboda:startTime` / `aurboda:endTime` / `aurboda:durationSeconds` (`services/activitypub/object.ts:272-289`), so the "more than a prefix swap" note in Implementations is accurate.
- Delivery matches the §7 id convention: the object dispatcher is `/users/{identifier}/feed/{postId}` (`federation.ts:380`) and the `Create` is a `#create` fragment on the Note id (`deliver.ts:184`), while `buildCreateExercise` (activity at the post URL, object at `…/object`) is indeed test-only — no production caller.

Internal consistency is good after the rename: no `quant:` term is left in snake_case, §2/§3/§7/§9 and both JSON examples agree on `startTime`/`endTime`/`quant:activityType`/`quant:observationOf`, the example object id, `quant:structuredUrl` and the §5 endpoint template line up with the §4 `api_base`, and the Note's metric set now matches the §5 payload's. Both new relative links (`docs/features/feed.md` → `../fep/quantpub.md`, `docs/fep/README.md` → `../features/feed.md`) resolve.

Please do **not** push fixes for these here; they're rev-5 material and pushing would only start another review round.

1. **§9's token-conveyance MUST is unsatisfiable for attachment-less posts.** "Publishers MUST therefore embed the token in the delivered image attachment URLs" only works if there *are* attachments. Aurboda itself emits none unless `include_chart`/`include_map` is set (`deliver.ts:imageAttachments`), so a followers-only sleep observation with no chart has no channel for the token and a conforming consumer lands on a 404. Worth rewording as a conditional MUST plus a stated fallback (e.g. `quant:structuredUrl?token=…`, or a normative "publishers SHOULD always attach at least one tokenised image for followers-only posts") before submission — a FEP reviewer will spot the vacuous MUST.
2. **§7 step ordering.** When a same-host `quant:structuredUrl` is present, step 2's `/.well-known/quantpub` fetch is no longer needed to locate the payload; it still serves as a capability check, but saying so explicitly (or marking the fetch conditional) would tighten the algorithm.
3. **FEP-400e metadata.** The added "received 2021-02-16, finalized 2022-02-04" dates couldn't be verified from here (the FEP index lives on Codeberg, outside this environment's network allowlist). Worth a quick check against the index before submission, since a wrong date in a citation is the kind of thing FEP reviewers do catch.

PR #1006, opened 2026-08-19
fix(feed): run the feed_posts.message migration (missing tableCreationOrder key)

fiddur:
Fixes the production 500 on sharing an activity after #1000:

```
error: column "message" of relation "feed_posts" does not exist
```

**Root cause:** #1000 added the additive migration `feed_posts_message_column` to `socialTables`, but the schema runner executes only keys listed in `tableCreationOrder` (`schema.ts`) — the new key wasn't added there, so the `ALTER TABLE … ADD COLUMN IF NOT EXISTS message` never ran. Existing databases kept the old table; the auto-migration retry re-ran the same list and failed again. Every test passed because fresh test databases get the column from the updated `CREATE TABLE` statement instead.

**Fix**
- List `feed_posts_message_column` in `tableCreationOrder` (additive-column keys run unconditionally — their key never matches an existing table name).
- New `schema.test.ts`: asserts `createTableStatements` ↔ `tableCreationOrder` are identical sets (no orphaned statements, no unknown/duplicate keys). It confirms no *other* statement is currently orphaned, and makes this silent-no-op class impossible to reintroduce.
- Drive-by from the same report: the share dialog's pending button label now reads **"Sharing…"** in create mode (was "Saving…").

After deploy, the first request that hits the schema-error retry path will add the column and sharing works — no manual intervention needed.

### DIFF
--- a/apps/backend/src/schema.test.ts
+++ b/apps/backend/src/schema.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, test } from 'vitest'
+
+import { createTableStatements, tableCreationOrder } from './schema.ts'
+
+/**
+ * The schema runner executes ONLY the keys listed in `tableCreationOrder` —
+ * a statement present in `createTableStatements` but missing from the order
+ * list silently never runs. For an additive column migration that is
+ * invisible locally (fresh DBs get the column from the CREATE TABLE) and
+ * breaks only in production on existing databases, as `feed_posts.message`
+ * did (#1000 → 500 on share). These two sets must therefore stay identical.
+ */
+describe('schema statement wiring', () => {
+  test('every createTableStatements key is listed in tableCreationOrder', () => {
+    const ordered = new Set(tableCreationOrder)
+    const missing = Object.keys(createTableStatements).filter((key) => !ordered.has(key))
+    expect(missing).toEqual([])
+  })
+
+  test('tableCreationOrder lists no unknown or duplicate keys', () => {
+    const unknown = tableCreationOrder.filter((key) => !(key in createTableStatements))
+    expect(unknown).toEqual([])
+    expect(new Set(tableCreationOrder).size).toBe(tableCreationOrder.length)
+  })
+})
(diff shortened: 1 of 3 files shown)


fiddur (commented):
✅ Approved

Correct, minimal fix for the right root cause.

**Verified:**
- `feed_posts_message_column` is placed after `feed_posts` and before `feed_posts_indexes`, matching the existing `feed_posts_article_columns` pattern in `apps/backend/src/schema/social.ts`. No index in `feed_posts_indexes` references `message`, so ordering is safe either way.
- Both runner paths pick it up: `initializeSchema` (`db/connection.ts:164`) runs every ordered key unconditionally, and the migrate loop (`db/connection.ts:1225`) evaluates `tableName = 'feed_posts_message_column'`, which is never in `existingTableNames`, so the `ALTER TABLE … ADD COLUMN IF NOT EXISTS` runs on existing production DBs. The statement is idempotent, so repeated runs are harmless.
- I statically cross-checked the two sets: all 97 keys spread into `createTableStatements` from `schema/*.ts` are present in `tableCreationOrder`, with no unknowns and no duplicates — so the new regression test passes and nothing else is silently unwired.
- The `ShareActivityDialog` ternary is correctly nested: pending → `Saving…` when editing an existing post, `Sharing…` when creating; idle → `Save changes` / `Share`. No test asserts the old label.

The new `schema.test.ts` is the right guard for a bug class that is invisible locally (fresh DBs get the column from `CREATE TABLE`) and only bites existing databases.

**Non-blocking — fold into a later PR, don't re-roll this one:**
- The test guards *membership* but not *ordering*; a statement listed before the table it alters would still fail at runtime. A dependency-order assertion (or just relying on the integration tests) could cover that later if it ever bites.
- The 8-line block comment above the test is on the justified side (it explains a genuinely non-obvious silent-no-op invariant), but it could be trimmed to a sentence or two without losing the point.
- The `ShareActivityDialog` label tweak is unrelated to the migration fix; harmless here, just noting the bundling.

Issue #1001, opened 2026-08-19
Sharing polish: deferred nits from #1000 review (settings query, textarea max, CSS, TZ-pinned tests)


fiddur:
Deferred non-blocking findings from the #1000 review (reviewer: "fold into a later PR, don't re-roll this one"):

- [ ] **Per-post `getSettings` query:** `serializeFeedPost` fetches settings per record; `GET /feed` and the public-profile listing map it over every post. Resolve the timezone once per request (caller-supplied or memoized).
- [ ] **`buildCreateExercise` not fed `message`/`timeZone`:** `buildFeedPostActivity` (the AS2 object-model path, currently unused in production) doesn't pass the new fields — wire them through before that content-negotiated path goes live (relates to #896).
- [ ] **`maxLength={2000}` on the share-message textarea** — the comment-prefilled text can exceed the schema max and today fails with a generic validation error on Share.
- [ ] **`.feed-post-title` spacing differs between cards:** inherits `.feed-post-content p` margins in `FeedPostCard` but default `p` margins inside `.timeline-structured`'s flex gap — add a rule so both cards space alike.
- [ ] **Timezone-sensitive tests:** pin `TZ` in `activity-stats.test.ts` so the same-day/cross-day cases are deterministic off-UTC.
- [ ] Trim duplicated comments (prefill semantics ×4, clear-on-edit rationale ×3) and simplify the HomeTimeline image filter to `!entry.structured || (…)`.

fiddur:
From the #1009 review round 1 (deferred non-blocking):

- [ ] **`GET /feed` payload/query weight**: the listing is unpaginated and now resolves each activity post's full 5s-bucketed series inline (~90KB/metric/post for a 1h activity, one bucketed query per metric per post, uncached — unlike the per-post public structured route). Paginate the feed listing, or lazily fetch the structured payload per card from the cached per-post endpoint.
- [ ] **Test the opted-in series path** of `serializeFeedPost({includeStructured})`: assert the inline `series` matches what the public structured endpoint returns for the same post (current test covers only the empty-series case).

fiddur:
From #1009 review round 2 (deferred, reviewer: don't re-roll):

[...]

PR #1009, opened 2026-08-19
feat(feed): own feed renders the peer-identical native card (#1008)

### LINKED ISSUE #1008: Own feed / profile cards: interactive chart + map instead of static chart.png/route.png
Feedback from the 2026-08-19 tryout of #1000 (own feed page, screenshots on the PR/issue trail): the stat grid + message + date render natively now, **but the chart and map are still the static `chart.png` / `route.png` images**. On the author's own card there is no hover data on the chart and no pan/zoom on the map — unlike the activity detail view, and unlike what a *peer* Aurboda timeline gets (interactive `TrendLineChart` from the structured series).

## Scope

On surfaces where the viewer is the owner (own feed page) — and for the public profile where data is public anyway:

- **Chart:** render the interactive chart component from the post's shared series (the owner has authenticated access; alternatively reuse the same `FeedStructuredPost` resolution the public endpoint serves) instead of `chart.png`. Hover shows real values, like the peer-timeline render.
- **Map:** render the interactive map component (pan/zoom, as on the activity detail view) from the shared activity's GPS track instead of `route.png`.
- Keep the static images as the delivered/federated representation — this is presentation-only on Aurboda's own surfaces.
- Mind post visibility: only render from data the post actually shares (series opt-in for the chart; `include_map` for the map).

Related: #996 covers the *peer timeline* side (enrichment not appearing); this issue is the author-side equivalent. #1001 carries the smaller presentation nits.

fiddur:
Implements #1008 with the design Fredrik specified: **the own feed must look exactly like it would for another Aurboda subscriber — the same component displayed** — so the author can verify enrichment/native rendering from their own shares.

- **Shared assembly, zero drift:** new leaf module `feed-structured-activity.ts` (`resolveStructuredSeries` + `assembleStructuredActivity`), used by both the public structured endpoint (`feed-structured.ts`) and the owner serialisation — the payload a peer fetches and the payload the owner's card renders are assembled by the same code.
- **Owner-facing opt-in:** `serializeFeedPost(user, record, { includeStructured: true })` on `GET /feed` and the share/PATCH responses attaches the full `FeedPost.structured` (typed scalars + inline series, reusing the already-resolved scalars — only the series query is new). Public profile listing and MCP tools stay lightweight (no series payload).
- **Web:** `FeedPostCard` renders `post.structured` with **`TimelineStructured`** — the identical component the home timeline uses for a peer's post — giving the interactive hover chart natively; the static `chart.png` is dropped when the native chart renders (same rule as the peer timeline), the `route.png` stays (no native map render yet). The stat-grid body remains as the fallback for structured-less surfaces.
- Integration test: payload present/absent per flag, `structured.metrics` ≡ flat `metrics`, message carried, empty-series shape.

Out of scope (left in #1008/#996 follow-up space): interactive pan/zoom **map** (needs a native map component on the card), and article posts (their own card already renders the native `ArticleContent`).

Closes #1008.

### DIFF
--- a/apps/web/src/pages/Feed/FeedPostCard.tsx
+++ b/apps/web/src/pages/Feed/FeedPostCard.tsx
@@ -46,6 +46,22 @@ const imageUrl = (username: string, post: FeedPost, kind: 'chart' | 'route'): st
     : `${API_URL}/public/${encodeURIComponent(username)}/feed/${post.id}/${kind}.png`
 
 /**
+ * The post's static image URLs. The chart.png is replaced only when the native
+ * interactive chart ACTUALLY renders — `structured` is attached to every
+ * activity post, so keying on its mere presence would drop the chart from an
+ * `include_chart` post whose series is empty/undrawable. The route map has no
+ * native render, so its image always stays.
+ */
+const mediaUrls = (post: FeedPost, username: string): { chart: string | null; route: string | null } => {
+  const nativeChart =
+    post.structured?.kind === 'activity' && structuredChartSeries(post.structured).length > 0
+  return {
+    chart: post.include_chart && !nativeChart ? imageUrl(username, post, 'chart') : null,
+    route: post.include_map ? imageUrl(username, post, 'route') : null,
+  }
+}
+
+/**
  * Native body for an activity post with resolved typed metrics (#997): title,
  * personal message, the activity's own date (#998), and a stat grid — instead
  * of the flattened `content` HTML Mastodon sees.
@@ -74,15 +90,7 @@ export const FeedPostCard = ({
 }) => {
   const vis = VISIBILITY[post.visibility]
   const when = formatDistanceToNow(new Date(post.created_at), { addSuffix: true })
-  // The static chart.png is replaced only when the native interactive chart
-  // ACTUALLY renders — `structured` is attached to every activity post, so
-  // keying on its mere presence would drop the chart from an `include_chart`
-  // post whose series is empty/undrawable. The route map has no native render,
-  // so its image always stays.
-  const nativeChart =
-    post.structured?.kind === 'activity' && structuredChartSeries(post.structured).length > 0
-  const chart = post.include_chart && !nativeChart ? imageUrl(author.username, post, 'chart') : null
-  const route = post.include_map ? imageUrl(author.username, post, 'route') : null
+  const { chart, route } = mediaUrls(post, author.username)
 
   return (
     <article class="feed-post">
--- a/apps/web/src/pages/Feed/FeedPostCard.tsx
+++ b/apps/web/src/pages/Feed/FeedPostCard.tsx
@@ -16,6 +16,7 @@ import { API_URL } from '../../config'
 import { formatEntryWindow } from './activity-stats'
 import { ActivityStatGrid } from './ActivityStatGrid'
 import { ArticleContent } from './ArticleContent'
+import { structuredChartSeries } from './timeline-structured'
 import { TimelineStructured } from './TimelineStructured'
 import './FeedPostCard.css'
 
@@ -73,10 +74,14 @@ export const FeedPostCard = ({
 }) => {
   const vis = VISIBILITY[post.visibility]
   const when = formatDistanceToNow(new Date(post.created_at), { addSuffix: true })
-  // With the full structured payload the native interactive chart replaces the
-  // static chart.png (same rule as a peer's timeline card); the route map has
-  // no native render, so its image always stays.
-  const chart = post.include_chart && !post.structured ? imageUrl(author.username, post, 'chart') : null
+  // The static chart.png is replaced only when the native interactive chart
+  // ACTUALLY renders — `structured` is attached to every activity post, so
+  // keying on its mere presence would drop the chart from an `include_chart`
+  // post whose series is empty/undrawable. The route map has no native render,
+  // so its image always stays.
+  const nativeChart =
+    post.structured?.kind === 'activity' && structuredChartSeries(post.structured).length > 0
+  const chart = post.include_chart && !nativeChart ? imageUrl(author.username, post, 'chart') : null
   const route = post.include_map ? imageUrl(author.username, post, 'route') : null
 
   return (
--- a/apps/web/src/pages/Feed/FeedPostCard.tsx
+++ b/apps/web/src/pages/Feed/FeedPostCard.tsx
@@ -16,6 +16,7 @@ import { API_URL } from '../../config'
 import { formatEntryWindow } from './activity-stats'
 import { ActivityStatGrid } from './ActivityStatGrid'
 import { ArticleContent } from './ArticleContent'
+import { TimelineStructured } from './TimelineStructured'
 import './FeedPostCard.css'
 
 export interface PostAuthor {
@@ -72,7 +73,10 @@ export const FeedPostCard = ({
 }) => {
   const vis = VISIBILITY[post.visibility]
   const when = formatDistanceToNow(new Date(post.created_at), { addSuffix: true })
-  const chart = post.include_chart ? imageUrl(author.username, post, 'chart') : null
+  // With the full structured payload the native interactive chart replaces the
+  // static chart.png (same rule as a peer's timeline card); the route map has
+  // no native render, so its image always stays.
+  const chart = post.include_chart && !post.structured ? imageUrl(author.username, post, 'chart') : null
   const route = post.include_map ? imageUrl(author.username, post, 'route') : null
 
   return (
@@ -91,12 +95,16 @@ export const FeedPostCard = ({
       </header>
 
       {/* An article renders its own title + prose/chart blocks (prose sanitised
-          via the shared sanitiser). An activity post with resolved typed metrics
-          renders natively (`ActivityPostBody`). Older payloads without `metrics`
-          fall back to the server-built, HTML-escaped `content` (safe to render
-          directly). */}
+          via the shared sanitiser). An activity post with the full structured
+          payload renders the SAME `TimelineStructured` component a subscribing
+          Aurboda peer's home timeline uses (#1008) — interactive hover charts
+          included — so the owner sees exactly what a follower sees. Without it
+          (public profile), the stat-grid `ActivityPostBody`; older payloads
+          fall back to the server-built, HTML-escaped `content`. */}
       {post.kind === 'article' && post.article ? (
         <ArticleContent article={post.article} />
+      ) : post.structured ? (
+        <TimelineStructured structured={post.structured} />
       ) : post.metrics ? (
         <ActivityPostBody post={post} />
       ) : post.content ? (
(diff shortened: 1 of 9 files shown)


apps/web/src/pages/Feed/FeedPostCard.tsx
```diff
@@ -72,7 +73,10 @@ export const FeedPostCard = ({
 }) => {
   const vis = VISIBILITY[post.visibility]
   const when = formatDistanceToNow(new Date(post.created_at), { addSuffix: true })
-  const chart = post.include_chart ? imageUrl(author.username, post, 'chart') : null
+  // With the full structured payload the native interactive chart replaces the
+  // static chart.png (same rule as a peer's timeline card); the route map has
+  // no native render, so its image always stays.
+  const chart = post.include_chart && !post.structured ? imageUrl(author.username, post, 'chart') : null
```
fiddur: `!post.structured` is not the same condition as "a native chart is rendered". `structured` is attached to *every* activity post now, including ones whose `series` is empty — so a post with `include_chart: true` and no chartable series loses its chart entirely: the PNG is suppressed here and `TimelineStructured` renders nothing (`structuredChartSeries` also drops series with <2 samples).

That state is reachable and documented: before #884 unified the two controls, `include_chart` was toggled independently of `series_metrics` — `initialSeriesSelection` in `feed-metrics.ts` exists precisely to handle "attached the chart image but predates the unified control (no `heart_rate` in `series_metrics`)". `shareActivityBodySchema` also lets an MCP/API caller set `include_chart: true` with `series_metrics: []` today. On those posts the owner's feed page currently shows a heart-rate chart and after this change shows none.

Gate on the native render actually existing instead, e.g.

```tsx
const nativeChart = post.structured ? structuredChartSeries(post.structured).length > 0 : false
const chart = post.include_chart && !nativeChart ? imageUrl(author.username, post, 'chart') : null
```

(`structuredChartSeries` is already exported from `./timeline-structured` and returns `[]` for an article.)

fiddur (commented):
🛑 Changes required

Nice refactor: pulling `resolveStructuredSeries` + `assembleStructuredActivity` into a leaf module is the right way to guarantee the owner's card and the peer's card can't drift, it keeps `import/no-cycle` happy, and the faithful move (same `end_time` guard, same scalar mapping) means `resolveStructuredContent` behaviour is unchanged. Schema, docs and an integration test all came along.

One blocking issue: suppressing the chart PNG on `post.structured` being *present* rather than on the native chart actually rendering silently drops the chart from posts that have `include_chart: true` but no chartable series — see the inline comment.

**Non-blocking — fold into a later PR (except the trivial one below, which you may as well take while pushing the fix)**

- `apps/backend/src/services/feed.ts:99` — the old `serializeFeedPost` doc block ("Serialise a stored feed post for the owner-facing REST/MCP surface…") stayed put while the function moved down, so it now sits directly above `resolveActivityPresentation` with the new block stacked under it, and `serializeFeedPost` is left undocumented. Moving that block back down with the function is a one-line-range fix worth including in the same push.
- **Payload/query weight of the list endpoint.** `GET /feed` uses `listFeedPosts`, which is unpaginated (every post the user ever shared), and each activity post now resolves its full 5s-bucketed series inline. At 5s buckets a 1h activity is ~720 fat samples (`avg/min/max/count/start/end`) ≈ 90 KB of JSON per metric per post, plus one `queryMetricsBucketed` round-trip per metric per post on every feed load — and there is no cache here, unlike the per-post `/feed/:postId` structured route. That grows without bound as the feed grows. Worth a follow-up: paginate the feed listing, or have the card lazily fetch the structured payload from the existing cached per-post endpoint.
- Comment weight: the router comment `// The owner's feed includes each activity post's full structured payload…` and the `structured:` schema comment mostly restate the adjacent code/`description`; the `#1008` rationale reads fine in one place (the leaf module's header) without the repeats.
- Test coverage: the new integration test asserts the empty-series case only (`series_metrics: []`). The interesting path — a post with an opted-in series producing the same `series` the public structured endpoint returns — is untested.

apps/web/src/pages/Feed/FeedPostCard.tsx
fiddur: Fixed in 5085c4d8: both `FeedPostCard` and `HomeTimeline` (same latent edge on the peer side) now suppress the PNG only when `structuredChartSeries(structured).length > 0` — the exact condition `TimelineStructured` draws with. Doc block moved and comments trimmed in the same push; the list-endpoint payload-weight and series-path-test items are recorded on #1001 for a follow-up.

fiddur (commented):
✅ Approved

Clean, well-scoped refactor. Extracting `resolveStructuredSeries` + `assembleStructuredActivity` into a leaf module (`feed-structured-activity.ts`) is the right shape: it keeps `feed.ts` → `feed-structured.ts` acyclic (`import/no-cycle` is on), and the extraction from `resolveStructuredContent` is faithful — same series gating on `activity.end_time`, same optional-field handling, same `duration_seconds` rounding. The privacy default is right too: `includeStructured` is opt-in and only the three authenticated owner routes pass it, so `feed-public-router` (anonymous) and the MCP tools still get no inline series — which matters, because `resolveStructuredSeries` deliberately bypasses `findCoveringSharedSeriesWindow`. The new `opts` parameter is additive, so every existing caller keeps its old behaviour, and the new integration test covers both the default-off and opt-in paths.

The chart-suppression fix from the last round is correct in both places: `nativeChart` now requires `structuredChartSeries(...).length > 0`, so an `include_chart` post whose series is empty or single-sample keeps its PNG, and the route map is never suppressed.

No blocking issues — approving.

Please do **not** push fixes for these here; another push just costs another review round. Batch them with other feed work.

1. **`GET /feed` payload and query weight.** `listFeedPosts` has no `LIMIT`/cursor, and every activity post now resolves its full 5s-bucketed series inline. A 2h ride with two shared series is ~2900 samples (a few hundred kB of JSON) per post, plus one Timescale bucket aggregation per metric per post on every Feed page load — the peer path is cheaper because `/feed/timeline` is paginated and enriches per entry. This grows linearly with the user's post count, and it's the same sample-cap concern already tracked in #972 (cf. the `~600k buckets` comment in `feed-structured.ts`). Worth a follow-up: paginate `GET /feed`, cap samples, or fetch `structured` lazily per card via the existing structured endpoint.

2. **The suppressed chart PNG is specifically *heart rate*.** `chart.png` renders `heart_rate` only (`feed-image-router.ts`), but both `FeedPostCard` and `HomeTimeline` suppress it when *any* structured series draws. Share an activity with `include_chart: true` and `series_metrics: ['power']` and the HR chart vanishes from both views while the native chart shows power. `structured.series.some((s) => s.metric === 'heart_rate')` (combined with the existing ≥2-sample rule) would match what the PNG actually contains.

3. **Comment weight.** The import-site comment in `feed-structured.ts` ("Series + payload assembly live in the shared leaf module…") restates the new module's own docblock; the `resolveActivityPresentation` docblock and the `nativeChart` comment in `FeedPostCard` are longer than the invariants they protect. The genuinely non-obvious parts — why the leaf module exists at all, why `structured` presence alone can't gate the PNG — are worth keeping; the rest could be trimmed.

4. **The parity test doesn't assert parity.** `includeStructured attaches the peer-identical structured payload` never compares against `resolveStructuredContent`'s output. Asserting `dto.structured` deep-equals `await resolveStructuredContent(user, post)` — ideally with a non-empty `series_metrics` and some metric rows so the series path is exercised — would actually pin the "can never drift" property this module exists for.

5. Minor: `structured` is inserted between `metrics` and `series_metrics` in `feedPostSchema`, breaking the file's otherwise alphabetical key order.

fiddur (commented):
✅ Approved

Solid refactor. Extracting `resolveStructuredSeries` + `assembleStructuredActivity` into a leaf module (`feed-structured-activity.ts`) is the right shape: `feed-structured.ts` still imports `resolveFeedActivity` from `feed.ts`, and the new module imports neither, so there's no cycle. I traced the two producers and they feed the assembler identical inputs (same `resolveActivityScalars(user, {start,end}, included_metrics)`, same `series_metrics` over the same merged window, same `message`), so the "peer-identical payload" claim holds.

The things most likely to break check out:

- `structured` is gated on `record.kind === 'activity'`, and the two `serializeFeedPost` calls left without `includeStructured` are the article create/update routes — where the backend would never attach it anyway. Consistent.
- The `include_chart` suppression rewrite is correct in both places: an empty `series_metrics`, and a series `structuredChartSeries` drops for `<2` points, both fall back to `chart.png` as the comment claims. The `HomeTimeline` filter preserves the previous article / no-structured behaviour exactly.
- `feed-structured.ts` no longer references any of the imports it dropped (`metricTypeSchema`, `FeedStructuredActivity`, `FeedStructuredSeries`), and still uses the ones it kept.
- The new integration test asserts what its name says, including the negative case (`plain.structured` undefined) and `series` being `[]` rather than `undefined`.

Please do **not** push fixes for these here; another round costs more than they're worth.

1. **`GET /feed` is unbounded and now inlines full 5s series.** `listFeedPosts` has no `LIMIT`/pagination, and every activity post that opted into series now costs a `queryMetricsBucketed` aggregate per metric plus its samples in the response body — a 1h activity is ~720 samples per metric. The peer-facing surfaces this mirrors are both cheaper: `/feed/timeline` is keyset-paginated over *stored* structured payloads, and the public `/feed/:postId/structured` route is cached per post. Blast radius is limited to posts where the user explicitly opted into series, so it's a scaling concern rather than a bug today — but pagination (or fetching `structured` lazily per card) is worth doing before the feed grows.

2. **Chart-PNG suppression is keyed on "any drawable series", but `chart.png` is always `heart_rate`.** In both `mediaUrls` and the `HomeTimeline` filter, a drawable `speed`/`power` series suppresses the heart-rate attachment even when heart rate itself isn't drawable. The web share dialog couples `include_chart` to the heart-rate series (`feed-metrics.ts`: `include_chart: sel.canChart && sel.series.has(HEART_RATE)`), so this is near-unreachable from the UI — but the REST/MCP bodies let the two diverge. Checking for a drawable `heart_rate` series specifically would make the rule exact.

3. **Comment weight.** The two-line comment above the `feed-structured-activity` import in `feed-structured.ts` restates that module's own header, and `/** Assemble the FeedStructuredActivity payload from its resolved pieces. */` restates the function name. The `mediaUrls` doc and the JSX block comment are both longer than the logic they guard, and the shouty `ACTUALLY`/`SAME` emphasis isn't earning its keep. The genuinely non-obvious bits — why series resolution bypasses `findCoveringSharedSeriesWindow`, and why this module must not import back into `feed.ts` — are worth keeping.

4. **API/MCP payload divergence.** `list_feed_posts` (MCP) returns posts without `structured` while `GET /feed` returns them with it. Capabilities are unchanged so this doesn't violate the parity rule — just noting the two surfaces now differ in payload richness, which is worth a line in the docs if it sticks.

5. A test asserting that `serializeFeedPost(user, post, { includeStructured: true }).structured` deep-equals `resolveStructuredContent(user, post)` would lock in the "can never drift" property that motivates the whole extraction, including the series branch the current test doesn't exercise.

Issue #1012, opened 2026-08-19
GET /feed payload weight: paginate the owner listing / lazy-load structured per card


fiddur:
Carried over from #1001 (all other items landed via the #1011 PR):

\`GET /feed\` is unpaginated and now resolves each activity post's full structured payload inline — 5s-bucketed series (~90KB/metric/post for a 1h activity, one bucketed query per metric per post, uncached) **plus, since #1011, the downsampled GPS route** (≤500 points when \`include_map\`). Fine at today's post counts, but it grows linearly with feed size.

Options (either resolves it):
- **Paginate** the owner feed listing (keyset like the home timeline / outbox), or
- **Lazy-fetch** the structured payload per card from the cached per-post public endpoint (\`GET /public/:username/feed/:postId\`) instead of inlining it in the listing.

Note the per-post public endpoint is already cached server-side; the inline listing resolution is not — so lazy-fetching would also cut repeated DB work.

fiddur:
The #1013 review adds to this: each `include_map` activity post now also runs a full raw-`locations` query in the unpaginated `GET /feed` listing, on top of the per-post series queries. Bounded per post, but the listing cost keeps growing — pagination (or lazy per-card structured fetch) is the real fix.

PR #1013, opened 2026-08-19
feat(feed): combined multi-metric chart + time-synced map on feed cards (#1011)

### LINKED ISSUE #1001: Sharing polish: deferred nits from #1000 review (settings query, textarea max, CSS, TZ-pinned tests)

### LINKED ISSUE #1011: Feed cards: combined multi-metric chart with toggles + time-synced interactive map (like activity view)
Follow-up to #1008/#1009 from Fredrik's tryout: the native feed card renders one chart per series and a static route.png. Proposal: **combine the charts and time-sync the map, exactly like the activity detail view** — one multi-metric chart with per-metric toggle buttons, crosshair hover, and a map that marks the position at the hovered time.

## Design (explored 2026-08-19; components verified)

**Reuse by extraction — both detail-view components split cleanly into fetch + presentation:**

1. **`CombinedMetricChart`** (new, `components/charts/`): extract the presentational half of `EntityDetail/ActivityChart.tsx` — D3 renderer (`renderChart`, `drawLineOverlay`/`drawDotOverlay`/`drawOverlays`, axis allocation `MAX_RIGHT_AXES`, crosshair + tooltip `buildTooltipLines`), `ChartToggle` buttons, colors (`KNOWN_METRIC_COLORS`/fallbacks), toggle state (`disabledMetrics`/`toggleOrder`). Props: `{ series: {metric, unit?, data: [Date, number][]}[], start, end, onHoverTime?, onEnabledMetricsChange?, stages? }`. `ActivityChart` keeps its fetch hook (`useMetricChartData`, bucket sizing, `EXCLUDED_METRICS`) and delegates — zero behaviour change on the detail page.
2. **`RouteMap`** (new, `components/charts/`): extract the presentational half of `EntityDetail/ActivityMap.tsx` (Leaflet init, polyline, `HIGHLIGHT_ICON` hover marker via `interpolatePosition`). Props: `{ points: {time, lat, lon}[], hoverTime }`. `ActivityMap` keeps `fetchRawLocations` and delegates. Move `chart-utils.ts` (+test) to `components/charts/` (imports: `findNearest`, `interpolatePosition`, `findStageAtTime`).
3. **`TimelineStructured`**: replace the per-series `TrendLineChart` stack with `CombinedMetricChart` fed from `structured.series` (bucket sample → `[bucket midpoint, avg]`), hold `hoverTime` state, render `RouteMap` from the new `structured.route` when ≥2 points. **Same component on own feed and peer timelines** — parity preserved.

**Route in the structured payload (backend):**
- api-spec: `feedStructuredSchema` += `route?: {t, lat, lon}[]` (downsampled GPS track, ~≤500 points, present **only when `include_map`** — the author already chose to share the route as an image; document that `t` additionally exposes position-at-time/pace to machine consumers, same visibility/capability gate as everything else in the payload).
- `feed-structured-activity.ts`: `resolveStructuredRoute(user, start, end)` (same locations source as the route.png render, even-index downsampling); `assembleStructuredActivity` gains the route; both callers (`resolveStructuredContent`, `serializeFeedPost({includeStructured})`) gate on `include_map`.
- Old peers ignore the unknown field (zod strips); new receivers of old payloads fall back to the delivered route.png.

**Image suppression** (`FeedPostCard.mediaUrls` + `HomeTimeline` filter): drop route.png only when the native map actually renders (`structured.route?.length >= 2`), mirroring the chart rule from #1009.

## Out of scope
Article posts; the deferred #1001 nits (HR-specific chart suppression can ride along if touching the same lines).

fiddur:
Closes #1011
Closes #1001 (all items except payload weight/pagination, moved to #1012)

Feed cards now render **exactly like the activity detail view** — one combined multi-metric chart with per-metric toggle buttons and crosshair tooltip, plus a **time-synced interactive map** marking the position at the hovered chart time. Same component on the owner's own feed and on subscribing Aurboda peers' home timelines, fed by the same structured payload.

## How

**Extraction (zero behaviour change on the detail page)**
- `components/charts/CombinedMetricChart` ← presentational half of `EntityDetail/ActivityChart` (D3 overlays, axis allocation `MAX_RIGHT_AXES`, sparse-dot fallback, toggles, crosshair + tooltip). `ActivityChart` keeps its width-sized bucketed fetch and delegates.
- `components/charts/RouteMap` ← presentational half of `EntityDetail/ActivityMap` (Leaflet polyline + hover-time highlight via `interpolatePosition`). `ActivityMap` keeps its fetch.
- `chart-utils` + `sleep-utils` (+tests) and the chart/map CSS move to `components/charts/` so the shared components stop depending on one page's stylesheet.

**Route in the structured payload**
- api-spec: `FeedStructured.route?: {t, lat, lon}[]` — present **only when the author attached the route map** (`include_map`): same track and audience as route.png (no extra geography; `t` does expose position-at-time/pace to machine consumers, documented on the schema). Downsampled to ≤500 points, endpoints kept.
- Backend: `resolveStructuredRoute` (same `getLocations` source as the rendered image) in the shared leaf module `feed-structured-activity.ts`; both producers — the public structured endpoint and the owner-facing `serializeFeedPost` — assemble identically. **Parity integration test** asserts the owner DTO deep-equals `resolveStructuredContent` with a real opted-in series + route (#1001 nit).
- Old peers strip the unknown field (zod); receivers of old payloads fall back to the delivered route.png.

**Image suppression per-counterpart (#1001 nit)**
chart.png draws heart rate only → skipped exactly when a drawable *heart-rate* series renders natively (an `include_chart` post sharing only e.g. power keeps its HR image). route.png → skipped when the native map renders. Both cards (FeedPostCard + HomeTimeline, via new pure `timelineImageVisible`).

**Folded-in #1001 polish**
- Settings resolved **once per listing request** (GET /feed, public profile, MCP `list_feed`) instead of once per post.
- `buildFeedPostActivity` wires `message`/`timeZone` through to `buildCreateExercise`.
- Share-dialog textarea capped at the shared `feedPostMessageMaxLength` (api-spec single source).
- `.timeline-structured p { margin: 0 }` so the title spacing matches the fallback card body.
- TZ-pinned, exact-string date tests in `activity-stats.test.ts`.
- `feedPostSchema.structured` in alphabetical position; duplicated prefill/clear-on-edit comments trimmed; HomeTimeline image filter simplified.
- docs/features/feed.md: native card render, `route` field, and the intentional MCP `list_feed` vs `GET /feed` structured-payload divergence.

## Tests
- `downsampleRoutePoints` unit tests (cap, endpoints, ordering).
- Integration: route gating on `include_map` (present/absent) + full structured **parity** with the public endpoint, series included.
- Rewritten `timeline-structured` pure-helper tests (combined series mapping, route parsing, HR-specific + map predicates, image visibility matrix).
- `pnpm check` green across backend/web/api-spec; backend activitypub/mcp/feed suites + all 581 web tests pass locally.


apps/web/src/pages/Feed/TimelineStructured.tsx
```diff
@@ -28,12 +38,22 @@ export function TimelineStructured({ structured }: { structured: FeedStructuredP
...
+        <div class="timeline-chart">
+          <CombinedMetricChart
+            series={series}
+            start={new Date(structured.start_time)}
+            end={new Date(structured.end_time)}
+            onHoverTime={setHoverTime}
+          />
+        </div>
+      )}
+      {showMap && <RouteMap points={structuredRoutePoints(structured)} hoverTime={hoverTime} />}
```
fiddur: `structuredRoutePoints(structured)` builds a new array on every render, and `RouteMap`'s init effect is keyed on `[points]` (`RouteMap.tsx:79`). `onHoverTime={setHoverTime}` fires on every chart mousemove with a fresh `Date`, so this component re-renders continuously while the pointer is over the chart — and each of those renders tears the Leaflet map down (`map.remove()`) and rebuilds it: new tile layer, new polyline, and a fresh `fitBounds` that discards any pan/zoom the reader did. That makes the time-synced map unusable during exactly the interaction it's for, and re-requests OSM tiles on every mouse move.

On the detail page this doesn't happen because `points` comes from react-query and keeps a stable identity. Memoising here restores that:

```tsx
const routePoints = useMemo(() => structuredRoutePoints(structured), [structured])
...
{showMap && <RouteMap points={routePoints} hoverTime={hoverTime} />}
```

(`structured` comes from query data, so it's stable across the hover-driven re-renders.)

fiddur (commented):
🛑 Changes required

Nice extraction overall. `CombinedMetricChart` / `RouteMap` are clean presentational splits, the route rides on exactly the `include_map` opt-in and the same `getLocations` source as `route.png` (`rest-routes.ts:224`), and the HR-only suppression rule is correct — `chart.png` really does render `heart_rate` only (`feed-image-router.ts:335`), so `structuredHasNativeHrChart` is the right predicate. The new integration test asserting the owner's payload equals `resolveStructuredContent(...)` is exactly the parity check that matters, and the schema description is honest about the pace/position-at-time exposure the timestamps add.

One blocking issue, inline: the feed card hands `RouteMap` a brand-new `points` array on every render while the card re-renders on every chart mousemove, so the Leaflet map is destroyed and rebuilt continuously during the hover that the feature exists for.

**Non-blocking — fold into a later PR, don't re-roll this one** (please don't push fixes for these here):

- `CombinedMetricChart` rebuilds `overlays` (and `start`/`end` in the feed) fresh every render, so the `renderChart` effect wipes and redraws the whole SVG on every mousemove — the crosshair is drawn by the handler and immediately erased by the redraw, leaving the tooltip visible with no crosshair when the pointer stops. This is inherited unchanged from the old `ActivityChart` (develop had the same `[..., overlays]` deps), so it's not a regression, but memoising `overlays` on a stable key would fix it for both the detail page and the new feed card.
- `GET /feed` (`listFeedPosts`) is unpaginated, and each `include_map` activity post now adds a full raw-`locations` query on top of the per-post series queries. Bounded per post, but the listing cost keeps growing; pagination is the real fix.
- `RouteMap`'s "Clean up previous map if any" block is unreachable — the effect's own cleanup already ran `map.remove()` before the effect re-runs. Along with `// Initialize map + draw polyline when points arrive`, `// Move highlight marker when hoverTime changes` and `// Don't render if no GPS data`, these are narration of code that already says it (carried over verbatim from `ActivityMap`, so they only show up as additions because the file is new). The three repeated `// One settings lookup for the whole listing, not one per post.` comments are in the same category.
- `process.env.TZ = 'UTC'` in `activity-stats.test.ts` works (the `Intl.DateTimeFormat`s are built lazily inside `formatEntryWindow`), but it mutates process state shared with every other test file in the same vitest worker. `test.env.TZ` in the vitest config would pin it without the leak.
- `feed-activity.ts` gains `message` / `timeZone` on `FeedActivityInput`, but `buildFeedPostActivity` has no production caller (only `feed-activity.test.ts`), and the test wasn't updated to cover the new fields — so nothing exercises them.

apps/web/src/pages/Feed/TimelineStructured.tsx
fiddur: Fixed in 584c3ae0 — `routePoints` is now `useMemo`d on `structured` (stable query data), so hover-driven re-renders keep the same array identity and the Leaflet map survives the hover it exists for.

fiddur (commented):
✅ Approved

Read the whole diff (backend route/service/schema changes, the chart+map extraction, and the feed wiring) against `origin/develop`. I found no blocking issues.

What I checked in particular:

- **Route resolution + opt-in.** `resolveStructuredRoute` uses `getLocations(user, start, end)` — the same source, and (via `resolveFeedActivity` → merged span, which `feed-image-router`'s `getActivity` dep also uses) the same window as `route.png`. Both the owner path (`feed.ts`) and the peer path (`feed-structured.ts`) gate on `record.include_map && end_time`, and the public structured endpoint's authorization (`loadAuthorizedStructuredPost`, capability token for `followers`-only) is untouched, so the route rides exactly the audience the route image already had. `coordinates` is GeoJSON `[lon, lat]` and the mapping to `{lat: [1], lon: [0]}` is correct.
- **Downsampling.** `downsampleRoutePoints` is correct for the only cap it's called with (500): endpoints preserved, indices strictly increasing, identity pass-through under the cap — and the tests assert exactly that.
- **Image-vs-native suppression.** `structuredHasNativeHrChart` / `structuredHasNativeMap` both use the same ≥2-point threshold the renderers use (`CombinedMetricChart` drops <2-sample series; `RouteMap` returns `null` below 2 points), so `FeedPostCard.mediaUrls` and `timelineImageVisible` can't drop an image whose native counterpart doesn't draw. Attachment names match `deliver.ts` (`'Heart rate'`, `'Route'`). The gate stayed on `kind === 'article'` so legacy untagged entries keep rendering natively.
- **The chart/map extraction** into `components/charts/` is faithful: same drawing code, same CSS (moved verbatim out of `EntityDetail/style.css` into component-owned files), `ChartToggle`'s dead `loading` prop dropped, and the render effect now additionally keys on `containerWidth` (a fix, not a regression). No stale importers of the moved `chart-utils`/`sleep-utils` remain.
- **`settings` hoisting** in the three listing callers preserves the previous `getSettings(...).catch(() => null)` semantics; `opts.settings !== undefined` correctly distinguishes "not supplied" from "known unavailable".
- **Privacy trade-off**, deliberately disclosed: `publicSeriesSampleSchema` documents that per-measurement timestamps are omitted to limit resolution leakage, while route points carry raw fix timestamps (position-at-time, hence pace). That's inherent to a time-synced map, is gated on the same `include_map` opt-in and token as `route.png`, and is called out in both the schema description and `docs/features/feed.md` — so I'm treating it as an intentional, documented decision rather than a finding.

Please do **not** push fixes for these here; they're better batched with other work.

1. `// One settings lookup for the whole listing, not one per post.` appears verbatim in `feed-router.ts`, `feed-public-router.ts` and `feed-tools.ts`. The `SerializeFeedPostOpts.settings` docblock already explains it; the three repeats are noise.
2. Every timeline/feed card with a route now instantiates its own Leaflet map fetching tiles straight from `tile.openstreetmap.org`. A page of N route posts is N maps × ~10 tiles per viewer, where it used to be N cached server-rendered PNGs. Fine at current scale; if timelines grow, consider lazy-initialising the map (IntersectionObserver, or click-to-activate).
3. `CombinedMetricChart`'s render effect lists `overlays` in its deps, and `overlays` is rebuilt every render — so each `onHoverTime` state update in the parent tears the SVG down and rebuilds it. Pre-existing behaviour inherited from the detail page, but it's now multiplied across feed cards; memoising `overlays` (and `series` in `ActivityStructured`) would make hover cheap.
4. `process.env.TZ = 'UTC'` in `activity-stats.test.ts` executes *after* the hoisted ESM imports. It works (nothing formats at import time, and Node re-reads TZ on assignment), but pinning it in the vitest config (`test.env`) would be more robust and cover every test file.
5. `CHART_MARGIN` reserves 205px (left 50 + right 155) for axes. On a narrow phone-width feed card that leaves a thin drawable area — a responsive right margin (or capping right axes at small widths) would help later.

PR #1015, opened 2026-08-19
fix(feed): timeline enrichment — local-origin resolution, failure logging, lazy retro-enrichment (#996)

### LINKED ISSUE #996: Aurboda→Aurboda activity shares don't render native hover charts in the home timeline — investigate + fix

fiddur:
Closes #996

Aurboda→Aurboda activity shares were documented to render native hover charts in the home timeline but observably did not (issue #996). This addresses all three suspected causes:

## 1. Same-instance posts resolve in-process (the likely production bug)
The enricher HTTP-fetched **even our own public origin**. From inside the container that hairpins through nginx and typically resolves to a private address, which the `safe-fetch` SSRF guard rightly refuses — so **every local-to-local follow lost its enrichment, silently**. `createAurbodaEnricher(origin)` now short-circuits objects on this instance's own origin to `resolveStructuredPost` — the exact function the public endpoint serves (capability-token check for `followers`-only included), so no drift, no discovery, no HTTP.

## 2. Enrichment failures are logged
`enrichFromAurboda` no longer swallows errors internally; the default enricher still resolves every failure to `null` (enrichment is never allowed to fail ingest) but logs it first — including the "answered but no payload" case — so a broken enrichment path stops being indistinguishable from a Mastodon post. Non-Aurboda-shaped URIs stay silent (no log spam for every Mastodon post).

## 3. Lazy retro-enrichment on timeline read (cause 2 from the issue)
Entries ingested before enrichment shipped — or whose ingest-time attempt failed transiently — kept `structured = NULL` forever. Reading the timeline (REST `GET /feed/timeline` **and** MCP `list_timeline`) now kicks a fire-and-forget pass giving 3 such Aurboda-shaped entries (newest first) one more attempt, stamping the new `timeline_entry.enrich_attempted_at` column win or lose so each entry is retried at most once; an `Update` redelivery still re-enriches through the normal ingest path. The read itself is never delayed. So Fredrik's existing pre-enrichment timeline entries gain their charts over a few page loads, with no manual backfill.

Re cause 1 (no series opted in → no chart by design): since #1013 an enriched activity card already renders the native stat grid + date + message even with zero series, so an Aurboda card visibly differs from a Mastodon one regardless.

## Notes
- Schema: additive `enrich_attempted_at TIMESTAMPTZ` column, registered in `tableCreationOrder` (the #1006 guard test enforces the pairing).
- `capabilityTokenFrom` is reused to lift a `followers`-only post's token from its stored image URLs for the retro pass.
- Docs: enrichment + retro-enrichment sections in docs/features/feed.md updated.

## Tests
- timeline-enrich: local-origin shortcut (no discovery/HTTP, trailing-slash origin), different-origin still HTTP, discovery/malformed failures now propagate (for logging), 404-style body → null.
- timeline-retro-enrich (new): stores payloads, stamps failed attempts, lifts tokens, skips-and-stamps false SQL-prefilter matches, respects batch size.
- DB integration: `listUnenrichedAurbodaEntries` candidacy matrix + `setTimelineEntryStructured` stamping and never-wipe-with-null.
- Whole-package `pnpm check` green; activitypub/mcp/feed suites + timeline integration pass locally.

### DIFF
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -42,9 +42,11 @@ import {
   insertPlace,
   insertRawRecord,
   insertTimeSeries,
+  listUnenrichedAurbodaEntries,
   loginToUserDb,
   openTimelineChannel,
   resolveOrCreateActivityType,
+  setTimelineEntryStructured,
   softDeleteSupersededLocations,
   updateDetectedLocation,
   upsertSyncState,
@@ -67,6 +69,7 @@ import {
 } from './services/activitypub/deliver.ts'
 import { createFeedFederation } from './services/activitypub/federation.ts'
 import { createTimelineBackfiller } from './services/activitypub/timeline-backfill.ts'
+import { createAurbodaEnricher } from './services/activitypub/timeline-enrich.ts'
 import { auditError, auditInfo } from './services/audit-log.ts'
 import { triggerCalorieComputation } from './services/calorie-computation.ts'
 import { createCalorieQueue, type CalorieQueue } from './services/calorie-queue.ts'
@@ -97,6 +100,7 @@ import { initSentry, Sentry } from './services/sentry.ts'
 import { createStravaQueue, type StravaQueue } from './services/strava-queue.ts'
 import { createSyncProvider } from './services/sync-provider.ts'
 import { createTimelineHub } from './services/timeline-hub.ts'
+import { retroEnrichTimelineEntries } from './services/timeline-retro-enrich.ts'
 import { createWebAuthnService } from './services/webauthn.ts'
 
 /** Grace period for in-flight requests before sockets are forced closed. */
@@ -353,6 +357,18 @@ const main = async () => {
   )
   const backfill = createTimelineBackfiller(feedFederation, webHost)
   const feedDeps = { apiBaseUrl, federation: feedFederation, origin: webHost }
+
+  // Lazy retro-enrichment (#996): entries ingested before structured enrichment
+  // shipped (or whose ingest-time enrichment failed transiently) get one more
+  // attempt when the timeline is read. Fire-and-forget — never blocks a read.
+  const retroEnricher = createAurbodaEnricher(webHost)
+  const retroEnrichTimeline = (user: string) => {
+    void retroEnrichTimelineEntries(user, {
+      enrich: retroEnricher,
+      listUnenriched: listUnenrichedAurbodaEntries,
+      save: setTimelineEntryStructured,
+    }).catch((err: unknown) => console.warn(`⚠️ timeline retro-enrichment failed for ${user}:`, err))
+  }
   const onDeliverError = (op: string, user: string, postId: string) => (err: unknown) =>
     console.error(`⚠️ feed ${op} delivery failed for ${user}/${postId}:`, err)
   const feedDeliver: FeedDeliver = {
@@ -424,6 +440,7 @@ const main = async () => {
       garmin,
       onActivityMutated: activityNotifier,
       oura,
+      retroEnrichTimeline,
       stravaQueue: stravaQueue ?? undefined,
       sync: syncProvider,
       webHost,
@@ -564,6 +581,7 @@ const main = async () => {
     httpd,
     invitationAuth,
     ouraWebhookManager,
+    retroEnrichTimeline,
     syncProvider,
     timelineHub,
     userDb,
--- a/apps/backend/src/api/rest-routes.ts
+++ b/apps/backend/src/api/rest-routes.ts
@@ -102,6 +102,8 @@ interface RestRoutesDeps {
   followActions: FollowActions
   followerActions: FollowerActions
   timelineHub: TimelineHub
+  /** Fire-and-forget lazy retro-enrichment of timeline entries on read (#996). */
+  retroEnrichTimeline: (user: string) => void
 }
 
 export const mountRestRouters = ({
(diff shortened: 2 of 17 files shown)


fiddur (commented):
✅ Approved

Read through the whole diff and the surrounding enrichment/ingest/DB code. This looks correct and well-scoped — no blocking issues.

The part I checked hardest was the new same-instance shortcut, since it swaps an HTTP round trip for a direct in-process call. It holds up: `local.resolve` is `resolveStructuredPost`, which goes through `loadAuthorizedStructuredPost` → `isCapabilityAuthorized`, i.e. exactly the same public/unlisted-vs-`followers`+token gate the public `/public/:username/feed/:postId` route enforces, so there's no authorization bypass. And `noteToTimelineInput`'s authority check (`note.id.host !== senderHost` → reject) means a remote followee can't deliver a Note whose object id carries *our* origin, so the shortcut can only ever fire on genuinely local-to-local posts.

The error-semantics change is also contained: `enrichFromAurboda` now throws instead of swallowing, but `createAurbodaEnricher` is the only caller in the tree and it catches + logs + returns `null`, with the `parseAurbodaFeedUrl` pre-check keeping Mastodon posts out of the warn path. `setTimelineEntryStructured`'s `COALESCE($2, structured)` correctly refuses to wipe a payload with a failed attempt, and the integration test asserts exactly that. The `enrich_attempted_at` column is added both to the CREATE TABLE and as an idempotent additive ALTER registered in `tableCreationOrder`, so existing DBs migrate (and `_isSchemaError` would retry a query that hits the column before migration ran).

Please **don't push fixes for these here**; a push starts another review round for no real benefit. Batch them with other feed work.

1. **Steady-state cost of `listUnenrichedAurbodaEntries`.** Once the backlog drains, the query's `LIMIT` never fills, so every timeline read (including every pagination page) walks the whole `idx_timeline_entry_published` index to return zero rows. A partial index — `... ON timeline_entry (published_at DESC, id DESC) WHERE structured IS NULL AND enrich_attempted_at IS NULL` — makes the common no-work case cheap regardless of timeline size.
2. **No in-flight guard on the fire-and-forget pass.** `enrich_attempted_at` is only stamped *after* an entry completes, and a batch can take up to 3 × `ENRICH_TIMEOUT_MS` = 36s, while the web's SSE-fallback polls `/feed/timeline` every 30s (`useTimelineLive`). While a backlog exists, overlapping reads can therefore run concurrent batches over the same candidate rows and re-fetch the same peer post. Self-limiting and idempotent, but a per-user in-flight set would make the "trickle, not a burst" comment actually true.
3. **One retry is burned by any failure, including a transient one.** The PR's stated goal includes recovering entries "whose ingest-time enrichment failed transiently", but a peer blip *during* the retro pass stamps those entries permanently (only an `Update` redelivery re-enriches after that). An attempt counter, or stamping only on definitive outcomes (404 / malformed) while leaving network errors unstamped, would close that hole.
4. **The local path skips `isValidUsername`.** The HTTP route validates the username before touching the DB; `enrichFromAurboda` hands `parsed.user` (a `decodeURIComponent`'d path segment) straight to `resolveStructuredPost`. Not exploitable today — the ingest host check gates which URIs get here, `getDbForUser` uses `format('SET ROLE %L')`, and the DB name isn't SQL — but mirroring the route's guard is a one-liner of defense in depth.
5. **Comment trimming.** The identical `/** Fire-and-forget lazy retro-enrichment of timeline entries on read (#996). */` param doc appears four times (`rest-routes.ts`, `mcp.ts`, `feed-tools.ts`, `feed-router.ts`), and the three-line comment above `retroEnrichTimeline?.(user)` in `feed-router.ts` mostly restates a call whose name already says it. One doc on the type/`TimelineEnricher`-style alias would carry it.
6. `registerFeedTools` is now seven positional parameters; an options object would read better next time it grows.

Issue #940, opened 2026-07-14
Share button still clipped ~24px off-screen on mobile /u/:username (#906 fix incomplete)


fiddur:
Followup from #938 (verified the merged `fix/sharing-page-polish` against the local `:develop` container).

## Summary
- **#916 (purple leak) is fully fixed** ✅ — `.feed-post-head` computes `background-color: rgba(0, 0, 0, 0)`; a bare injected `<header>` still computes `rgb(103, 58, 184)`, confirming the global `header{}` rule is neutralized only on the card head. Nothing to fix here.
- **#906 (clipped Share button) is only partially fixed** ⚠️ — `flex-wrap: wrap` + `min-width:0; overflow-wrap:anywhere` do make the button wrap to its own line, but on narrow/mobile viewports the Share button is **still clipped ~24px off the right edge and unreachable** (no horizontal scroll). The PR body promises the button "wraps to a **reachable** line"; it wraps, but isn't reachable.

## Problem
On `/u/qsreddit_demo` at mobile widths, the Share button renders as **"Shar"** with the rest cut off. Measured `button.getBoundingClientRect().right` vs `window.innerWidth`, with `document.documentElement.scrollWidth === innerWidth` (i.e. `maxScrollX === 0`, so the cut-off part cannot be scrolled to):

| viewport | button.right | innerWidth | clipped |
|----------|-------------|-----------|---------|
| 320 | 344 | 320 | 24px |
| 360 | 384 | 360 | 24px |
| 390 | 414 | 390 | 24px |
| 414 | 438 | 414 | 24px |
| 480 | 504 | 480 | 24px |

The button wraps to its own line vertically (top below the avatar) — that part of the fix works — but it pins to a container edge that sits 24px past the viewport at every width.

**Root cause:** `.public-profile` uses `box-sizing: content-box` with `padding: 2rem 1.5rem`, so its border-box is **48px wider than the viewport** (e.g. at 360px the element measures 408px wide; the content box spans x=24..384). The `.public-profile-header .share-link-btn { margin-left: auto }` button pins to that overflowing right edge = `viewport + 24px`. The whole `.public-profile` (post cards too) overflows and is clipped with no horizontal scroll. Desktop (>~768px) is fine because `max-width: 720px` centers the container with room to spare, which is why this only bites mobile — exactly the viewport #906 targets.

[...]

PR #1019, opened 2026-08-19
fix(feed): export dead links + escaping, chart hover polish, retro-enrichment hardening (#974, #975, #1014)

### LINKED ISSUE #974: Markdown export links block images that 404 (empty-window charts, n<3 correlations)
Followup from #971.

## Problem
`buildArticleMarkdown` emits an image link for **every** chart/correlation block unconditionally, but the block-image endpoint (`/public/:user/feed/:postId/blocks/:i/image.png`) refuses to render blocks it considers ineligible. The export therefore produces markdown containing dead image links — pasted into r/QuantifiedSelf (the feature's stated destination) they show as broken images.

This is inconsistent with the C3 path added in the same PR: `resolveStructuredArticle` skips ineligible blocks, and the web renders an "Not enough data in this window." fallback for empty ones. Only the export links them blindly.

### Repro 1 — chart block over a window with no samples
Ordinary article, nothing exotic:

```json
{"title":"PR971 no-data block export","visibility":"public","blocks":[
 {"type":"prose","markdown":"A perfectly ordinary article whose chart window happens to have no samples."},
 {"type":"chart","metric":"blood_glucose","start":"2026-01-01T00:00:00Z","end":"2026-01-08T00:00:00Z","bucket":"1d","caption":"blood glucose, first week of January"},
 {"type":"chart","metric":"heart_rate","start":"2026-07-01T00:00:00Z","end":"2026-07-08T00:00:00Z","bucket":"1d","caption":"heart rate (has data)"}]}
```

`GET /api/feed/articles/<id>/export` returns:

```markdown
# PR971 no-data block export

A perfectly ordinary article whose chart window happens to have no samples.

![blood glucose, first week of January](.../blocks/1/image.png?v=1785738151832)
*blood glucose, first week of January*

![heart rate (has data)](.../blocks/2/image.png?v=1785738151832)
*heart rate (has data)*
```

Resolving those two links:

```
blocks/1/image.png -> HTTP:404 ct=application/json; charset=utf-8   {"error":"Not found","success":false}
blocks/2/image.png -> HTTP:200 ct=image/png bytes=22585
```

Meanwhile the C3 structured endpoint keeps the block and lets the receiver render its empty state:

```
[{"type":"prose",...},{"type":"chart","caption":"blood glucose, first week of January","n_samples":0},{"type":"chart","caption":"heart rate (has data)","n_samples":3}]
```

### Repro 2 — correlation block with n < 3
The PR description already notes the image path "404s below n < 3". Exporting the existing article `b9a59126-e7a3-4395-b594-0b20c32b5b96` ("PR947 tryout: insufficient data state", correlation resolves to `n:1`) yields markdown that is *only* a title plus one broken image:

```markdown
# PR947 tryout: insufficient data state

![one-day window -> too few points](.../blocks/0/image.png?v=1784055383893)
*one-day window -> too few points*
```
```
blocks/0/image.png -> HTTP:404
```

### Repro 3 — zero-duration bucket
A `bucket:"0s"` chart block is accepted by `create_article`/the REST create route. `resolveStructuredArticle` omits it (`isZeroDurationBucket`) and the image endpoint 404s, but the export still links it:

```
blocks/0/image.png -> HTTP:200 bytes=34633   (1s bucket, floored to 5s)
blocks/1/image.png -> HTTP:404               (0s bucket - omitted by C3, linked by export)
blocks/2/image.png -> HTTP:200 bytes=26085
blocks/3/image.png -> HTTP:200 bytes=22585
```

## Suggestion
Apply the same eligibility test the image endpoint uses before emitting a link. Since the block image is resolved on demand, `buildArticleMarkdown` would have to become async (or take a precomputed set of renderable indices) — e.g. reuse `resolveArticleBlock` and skip blocks it rejects, optionally emitting the caption alone as an italic line so the prose still reads coherently.

Note the AS2 attachment builder has the same behaviour (block 1 above is listed as an attachment and 404s), so the fix likely belongs on the shared eligibility path rather than in `article-export.ts` alone — that would also keep the "export and the federated image always agree" invariant intact.

## Repro
Against the local tryout container (`/home/fiddur/src/codereview/tryout`, token in `token.txt`):

```bash
TOKEN=$(cat /home/fiddur/src/codereview/tryout/token.txt)
ID=$(curl -s -X POST -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d @nodata.json http://localhost:8080/api/feed/articles | jq -r '.post.id')
curl -s -H "Authorization: Bearer $TOKEN" "http://localhost:8080/api/feed/articles/$ID/export" \
  | jq -r '.markdown' > export.md
grep -oE 'http://localhost:3000[^)]*' export.md | while read -r u; do
  curl -s -o /dev/null -w "$u -> %{http_code}\n" \
    "$(echo "$u" | sed 's|http://localhost:3000|http://localhost:8080/api|')"
done
```
(The `localhost:3000` base is just this container's unset `apiBaseUrl`; rewrite it to `localhost:8080/api` to resolve locally.)

### LINKED ISSUE #975: Article export: caption escaping covers [ ] but not * or _, breaking the italic caption line
Followup from #971. Minor / cosmetic.

## Problem
`inlineText` in `apps/backend/src/services/article-export.ts` escapes `[` and `]` so an unbalanced bracket can't break the `alt` syntax. The same function's output is also used for the caption line, which `buildArticleMarkdown` wraps in `*…*` for italics — but `*` and `_` are not escaped, so a caption containing either breaks that emphasis span.

Article with `caption: "5*3 sets and _underscored_ and (paren) and ]bracket["`:

```markdown
# PR971 caption escaping

![5*3 sets and _underscored_ and (paren) and \]bracket\[](http://localhost:3000/public/qsreddit_demo/feed/0e268ede-80ea-4443-82f9-cfeb6a96b97c/blocks/0/image.png?v=1785738184538)
*5*3 sets and _underscored_ and (paren) and \]bracket\[*
```

The brackets are escaped correctly, but the caption line renders as *5* in italics followed by literal `3 sets and ` , then `underscored` in italics, then a stray trailing `*` — instead of the whole caption in italics. `5*3 sets` is a plausible caption for a strength-training article.

## Suggestion
Escape `*` and `_` alongside `[` and `]` in `inlineText` (they need escaping in both the alt text and the caption line), or wrap the caption in a form that doesn't depend on the caption being emphasis-safe.

## Repro
```bash
TOKEN=$(cat /home/fiddur/src/codereview/tryout/token.txt)
ID=$(curl -s -X POST -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' -d '{
  "title":"PR971 caption escaping","visibility":"public","blocks":[
  {"type":"chart","metric":"heart_rate","start":"2026-07-01T00:00:00Z","end":"2026-07-08T00:00:00Z",
   "bucket":"1d","caption":"5*3 sets and _underscored_ and (paren) and ]bracket["}]}' \
  http://localhost:8080/api/feed/articles | jq -r '.post.id')
curl -s -H "Authorization: Bearer $TOKEN" "http://localhost:8080/api/feed/articles/$ID/export" | jq -r '.markdown'
```

### LINKED ISSUE #1014: Chart/card polish: deferred nits from the #1013 review
Non-blocking findings from the #1013 review (reviewer: fold into a later PR, don't re-roll):

- [ ] **Memoise `CombinedMetricChart` overlays**: the `renderChart` effect deps include the fresh-per-render `overlays` array (and, on the feed card, fresh `start`/`end` Dates), so the whole SVG is wiped and redrawn on every mousemove — the handler-drawn crosshair is immediately erased, leaving the tooltip visible with no crosshair when the pointer stops. Inherited unchanged from the old `ActivityChart` (not a regression); memoising on a stable key fixes the detail page and the feed card at once. The feed card should also pass memoised `start`/`end`.
- [ ] **Trim narration comments carried into `RouteMap`**: the "Clean up previous map if any" block is unreachable (the effect's own cleanup already ran), and `// Initialize map…`, `// Move highlight marker…`, `// Don't render if no GPS data` narrate the code. Same for the three repeated `// One settings lookup for the whole listing, not one per post.` comments.
- [ ] **Pin TZ via vitest config** (`test.env.TZ`) instead of `process.env.TZ = 'UTC'` at the top of `activity-stats.test.ts`, which mutates state shared with other test files in the worker.
- [ ] **Cover `message`/`timeZone` in `buildFeedPostActivity`'s test** — the fields are wired through but nothing exercises them (no production caller yet; relates to #896).

fiddur:
Closes #974
Closes #975
Closes #1014

One chunked polish PR covering the two Ready P2 export bugs plus the whole #1014 deferred-nit batch from the #1013 and #1015 reviews (both reviewers: "fold into a later PR"). #940 was verified already fixed on production and closed separately; #1012 (feed pagination/payload weight) intentionally stays its own future PR.

## Article markdown export
- **#974 — no more dead image links.** A chart block over a window with no samples, or a correlation with n < 3, 404s at the block-image endpoint; the export used to link it anyway. It now emits an italic *"<label> — not enough data in this window."* note instead. Eligibility mirrors the image endpoint exactly: the two inline data deps in `rest-routes.ts` (`getArticleChartSeries`, `getCorrelationScatter`) moved verbatim to `services/article-block-data.ts` and are now the **shared source** for both the image routes and the new `renderableArticleBlocks` (DI'd + unit-tested; static window checks short-circuit without fetching). `CorrelationBlockParams` moved with them (router re-exports).
- **#975 — caption escaping.** `inlineText` now escapes `*` and `_` alongside `[`/`]`, so a caption like `5*3 sets and _underscored_` renders as one italic span. (One existing assertion updated: alt-text labels now read `sleep\_score`.)

## Chart hover polish (#1014 ← #1013 review)
- `CombinedMetricChart` memoises its overlays and keys the redraw on the window's epoch values; `ActivityChart` and `TimelineStructured` pass stable memoised `series`. The mousemove-drawn crosshair is no longer wiped by a full SVG redraw on every hover-driven parent re-render — fixed for the detail page and the feed card at once.
- `RouteMap`: unreachable "clean up previous map" block and narration comments removed.
- TZ pinned for the whole web test run via `test.env.TZ` in vite.config.ts (no more `process.env` mutation shared across worker files).
- `buildFeedPostActivity` `message`/`timeZone` wiring now has test coverage (Stockholm-tz date line + `aurboda:message` present/absent).

## Retro-enrichment hardening (#1014 ← #1015 review)
- **Partial index** `idx_timeline_entry_unenriched` (`WHERE structured IS NULL AND enrich_attempted_at IS NULL`) — a drained backlog costs nothing per read. Registered in `tableCreationOrder` (guard test covers the pairing).
- **In-flight guard**: one retro pass per user at a time — the 30s SSE-fallback poll can't overlap batches over the same candidates.
- **Transient failures keep their retry**: new `createAurbodaEnrichAttempt` throws on network/timeout (the pass leaves the entry unstamped for a later read) while definitive outcomes — payload, gone/unauthorized, malformed — resolve and stamp. The ingest-path `createAurbodaEnricher` wraps it with the old swallow-and-log semantics, unchanged.
- `isValidUsername` guards the local-origin resolution before the name reaches the DB layer.
- The 4× duplicated retro-trigger param doc collapsed onto the exported `RetroEnrichTrigger` type; the 3× "one settings lookup" comments removed; `registerFeedTools` now takes an options object (was 7 positional params).

## Tests
- `renderableArticleBlocks` (DI): keeps drawable blocks, drops sparse ones, statically-ineligible windows short-circuit without a fetch.
- Export: placeholder-instead-of-link, `*`/`_` escaping.
- Retro pass: transient throw leaves the entry unstamped and the batch continues.
- `buildFeedPostActivity`: message + timezone coverage.
- Whole-monorepo `pnpm check` green; all 581 web tests + backend mcp/feed-image/timeline-integration suites pass locally.


apps/backend/src/services/activitypub/timeline-enrich.ts
```diff
@@ -147,22 +161,42 @@ export type TimelineEnricher = (objectUri: string, token?: string) => Promise<Fe
...
+    if (parseAurbodaFeedUrl(objectUri) == null) return null
+    let structured: FeedStructuredPost | null
+    try {
+      structured = await withTimeout(enrichFromAurboda(objectUri, deps, token), ENRICH_TIMEOUT_MS)
+    } catch (error) {
+      if (error instanceof Error && error.message === 'malformed structured response') {
+        console.warn(`⚠️ timeline enrichment: malformed payload for ${objectUri}`)
+        return null
+      }
+      throw error
```
fiddur, line 207: Only `malformed structured response` is treated as definitive here, but the *gone / unauthorized* case also reaches this line as a throw rather than as `null`.

`GET /public/:username/feed/:postId` answers **404** whenever the post is missing, no longer authorized, or has no resolvable structured payload (`feed-public-router.ts`), and `safeFetchGet` passes no `validateStatus`, so axios rejects on 4xx. That propagates out of `enrichFromAurboda` and is rethrown here, so `retroEnrichTimelineEntries` classifies it as transient and never stamps `enrich_attempted_at` — contradicting the `RetroEnrichDeps.enrich` doc that lists "gone / unauthorized" as a `null` outcome.

Consequence: `listUnenrichedAurbodaEntries` orders `published_at DESC, id DESC LIMIT 3`, so those same entries are re-selected on *every* timeline read (30 s web poll) forever, and any older unenriched entry behind them is starved permanently — the backlog can no longer drain. A permanently unreachable peer (DNS gone) has the same effect. Before this PR the stamp-always rule guaranteed forward progress.

Two things to fix:
1. Treat a definite HTTP response as definitive — e.g. return `null` for 4xx (`axios.isAxiosError(error) && error.response != null && error.response.status < 500`), keeping the throw for connection errors, timeouts and 5xx.
2. Bound the genuinely-transient retries so a dead host can't hold the head of the queue indefinitely (an attempt counter, or stamp once the entry is older than some age).

fiddur (commented):
🛑 Changes required

Most of this is solid and well covered. The `#974` export-eligibility check genuinely mirrors `resolveArticleBlock` + `renderArticleBlockImage` (same window guards, same `>= 2 points` / `n >= 3` thresholds, same shared fetchers), so the extracted `article-block-data.ts` really does remove the drift risk. The `#975` escaping change is right, and the integration test now covers both export outcomes against real data. On the web side the memoisation is correct: `colorByMetric`/`overlays` reproduce the previous colour and axis allocation exactly, `startMs`/`endMs` defuse the fresh-`Date`-per-render callers, and dropping RouteMap's manual pre-cleanup is safe because the effect already returns a `map.remove()` cleanup.

**Blocking:** the new transient/definitive split misclassifies the most common permanent case. See the inline comment on `timeline-enrich.ts` — a gone / unauthorized / non-resolving remote post makes the origin answer **HTTP 404**, and `safeFetchGet` passes no `validateStatus`, so that surfaces as a *throw*, not `null`. `retroEnrichTimelineEntries` then leaves the entry unstamped forever. Because `listUnenrichedAurbodaEntries` is `ORDER BY published_at DESC, id DESC LIMIT 3`, the same three poisoned entries come back on every timeline read (the web polls every 30 s), so the rest of the backlog is starved permanently and the process keeps re-fetching a dead peer indefinitely. Before this PR the stamp-always rule guaranteed forward progress.

**Also worth fixing in this round (same `#1014` hover bug; the file is not in the diff so there is no inline anchor):** `apps/web/src/pages/EntityDetail/index.tsx:258` computes `const stages = parseSleepStages(...)` unmemoised in the render body, and `onHoverTime={setHoverTime}` re-renders that component on every mousemove. `stages` is still a dep of the render effect in `CombinedMetricChart`, so on a **sleep** activity detail page the chart still redraws (and `svg.selectAll('*').remove()` still wipes the crosshair) on each mousemove. A `useMemo` on `activity.data` there finishes the fix — non-sleep activities are already fixed, since `stages` is `undefined` for them.

- `article-export.ts`'s module docstring still says "Pure and synchronous… so no chart/correlation resolution happens here", which `renderableArticleBlocks` now contradicts.
- The export endpoint now does one `getUserSettings` + one bucketed query per chart block and a full `getContinuousCorrelation` per correlation block, sequentially, before responding — and that work is repeated when the images themselves are fetched. Fine at today's block counts; worth a shared/cached resolution if articles grow.
- `error.message === 'malformed structured response'` couples two modules through a string literal; a typed error (or a discriminated result) would be sturdier.
- `feed-image-router.ts` re-exporting `CorrelationBlockParams` purely as a pass-through is a little indirect now that `article-block-data.ts` owns it.

apps/backend/src/services/activitypub/timeline-enrich.ts
fiddur: Fixed in 99954339, both parts: (1) new `isDefinitiveEnrichError` treats any definite HTTP answer (axios error with `response.status < 500`) and the malformed-payload case as definitive — the attempt resolves `null`, the entry is stamped, and the queue drains; connection errors/timeouts/5xx still throw. (2) Transient retries are now bounded: a new `enrich_attempts` counter (additive column, in `tableCreationOrder`, schema guard test covers it) is bumped per transient failure via `markEnrichTransientFailure`, and the entry is stamped out of the candidate set at 3 attempts — a dead host can no longer hold the newest-first queue head. Covered by unit tests (classifier + retro pass) and a DB integration test for the cap. The unmemoised `stages` on the sleep detail page is also fixed in the same commit.

fiddur (commented):
✅ Approved

Read the whole diff plus the surrounding code (`feed-image-router.ts`, `safe-fetch.ts`, `challenge-federation.ts`, `db/timeline.ts`, both `CombinedMetricChart` consumers, the Dockerfile). No blocking issues — the four changes each hold up:

**#974 dead links.** `renderableArticleBlocks` mirrors `resolveArticleBlock` + `renderArticleBlockImage` exactly (bounded+increasing window → non-zero bucket → ≥2 points; correlation → `n ≥ 3` via the same `getArticleCorrelationScatter`), and both now go through the *same* `article-block-data.ts` fetchers that the image router is wired with, so the eligibility check can't drift from the endpoint. Both export surfaces (REST + `export_article_markdown`) pass it, and both keep the followers-only 400 gate ahead of it. The unbounded-window test correctly proves no data fetch happens for a statically-ineligible block.

**#975 escaping.** Adding `*`/`_` to the `inlineText` class is right for both the alt text and the italic caption line; prose blocks are still emitted verbatim, so nothing else is affected.

**#1014 chart hover.** The fix is complete: `overlays`/`colorByMetric` are memoised, the render effect keys on `startMs`/`endMs` instead of the (freshly-constructed) `Date` props, and every remaining input to `renderChart` is either in the dep list or a ref — so no input can change without a redraw. Both call sites are covered (`TimelineStructured` memoises `series`, `ActivityChart` memoises on the query result, `EntityDetail` memoises `stages`). The new `enabled`/`withAxes` derivation is behaviourally identical to the old `enabledMetrics`/`metricsWithAxes`, and the fallback-colour index still advances only for metrics without a known colour, so colours are unchanged. Dropping the defensive pre-cleanup in `RouteMap` is safe — the effect's returned cleanup already runs before a re-run.

**#1014 retro-enrichment.** The transient/definitive split is sound: `safeFetchGet` sets no `validateStatus`, so a 4xx really does surface as an `AxiosError` with a response, and timeouts/connection errors/5xx have none. The `enrich_attempts` cap SQL matches the integration test (2 below the cap stay candidates, the 3rd stamps), the partial index predicate matches `listUnenrichedAurbodaEntries`'s `WHERE`, and the `retroInFlight` set is a `main()` closure shared by both the REST and MCP triggers, so the two read paths can't start overlapping passes. Adding `isValidUsername` before `resolveStructuredPost` is the same guard the public route uses, so it can't reject a legitimate local user.

Also checked: `vite.config.ts` importing from `vitest/config` is fine for the image build — the Dockerfile's builder stage runs a full `pnpm install --frozen-lockfile`, not a prod-only install.

Please do **not** push fixes for these here; a new round costs more than the fixes are worth. Batch them with other work.

- `apps/backend/src/services/article-export.ts` module header still says *"Pure and synchronous … so no chart/correlation resolution happens here"*, which this PR makes false — `renderableArticleBlocks` in that same module now resolves both. (`buildArticleMarkdown` itself is still pure/sync; it's the module-level claim that drifted.)
- `apps/backend/src/services/timeline-retro-enrich.ts` module header still says *"Each entry is attempted at most once: `enrich_attempted_at` is stamped whether or not a payload was obtained"* — also no longer true for the transient path.
- `docs/features/feed.md` describes a transient failure as leaving the entry "unstamped for a later read" without mentioning the 3-attempt cap that eventually stamps it — worth a half-sentence so the doc matches `MAX_TRANSIENT_ATTEMPTS`.
- `isDefinitiveEnrichError` treats every `< 500` response as definitive, so a peer answering `429` (or a proxy `408`) permanently stamps the entry. Not a regression (before this PR *any* failure stamped it), but `429`/`408` would be more accurate as transient.
- `getArticleChartSeriesData` re-runs `getUserSettings` once per chart block, so an N-block article does 2N queries in the export's eligibility pass. Fine at current article sizes; a single settings lookup threaded through would be tidier.
- The `// Blocks whose image would 404 (no data) get a note, not a dead link (#974).` comment is duplicated verbatim at both call sites and mostly restates the `renderableArticleBlocks` argument name — the JSDoc on `buildArticleMarkdown` already covers it.

PR #1022, opened 2026-08-20
feat(feed): keyset-paginate the owner feed listing (#1012)

### LINKED ISSUE #1012: GET /feed payload weight: paginate the owner listing / lazy-load structured per card

fiddur:
Closes #1012

`GET /feed` was unpaginated while (since #1009/#1013) every activity post inlines its full structured payload — 5s-bucketed series (~90KB/metric/post for a 1h activity) plus, per `include_map` post, the downsampled GPS route and a raw-locations query. The listing therefore grew linearly with feed size, in payload and in per-post DB work. It is now **keyset-paginated on `(created_at, id)`** — the same cursor style as the home timeline — 20 posts per page, so the first load stays light however many posts exist.

## Design
- **Why not the issue's lazy-fetch alternative:** having cards fetch `structured` from the cached per-post public endpoint would break parity for the owner's **`followers`-only** posts — that endpoint requires the capability token, which the owner's browser deliberately never has (#893). Pagination keeps the owner's WYSIWYG card for every visibility.
- **api-spec:** new `FeedPostsQuery` (`cursor`, `limit` 1–50 default 20); `FeedPostsResponse` gains an optional nullable `next_cursor` — the bounded public-profile listing keeps omitting it, and existing clients reading `posts` are unaffected.
- **Shared cursor codec:** the timeline's opaque `(timestamp, uuid)` base64url cursor moved to `services/keyset-cursor.ts` (UUID-validated before any `::uuid` cast; malformed → first page, never a 500). `timeline.ts` refactored onto it, behavior unchanged.
- **Backend:** `listFeedPosts(user, limit, before?)` keyset page (id tiebreaker as before); `getFeedPage` in `services/feed.ts` (fetch `limit+1`, injectable fetcher) shared by REST `GET /feed` and MCP `list_feed` — both surfaces page identically (parity).
- **Web:** the Feed page's own posts use `useInfiniteQuery` + **Load more**, mirroring the home timeline; the query key becomes `['feed','posts']`, still matched by all existing `['feed']` invalidations (share/edit/unshare dialogs).

## Tests
- `keyset-cursor` unit tests: round-trip, empty, malformed (no separator, non-UUID id, non-numeric timestamp).
- `getFeedPage` integration: 3-post paging with cursor, exactly-full page ends with `next_cursor: null`, malformed cursor falls back to page 1.
- DB integration: `listFeedPosts` keyset paging.
- Whole-monorepo `pnpm check` green; feed/mcp/router integration suites + all 581 web tests pass locally.

### DIFF
--- a/apps/backend/src/db/feed.integration.test.ts
+++ b/apps/backend/src/db/feed.integration.test.ts
@@ -137,8 +137,21 @@ describe('Feed posts integration', () => {
     const user = getTestUser()
     const first = await createFeedPost(user, postInput())
     const second = await createFeedPost(user, postInput())
-    const posts = await listFeedPosts(user)
+    const posts = await listFeedPosts(user, 10)
     expect(posts.map((p) => p.id)).toEqual([second.id, first.id])
+  })
+
+  test('keyset-paginates by (created_at, id) (#1012)', async () => {
+    const user = getTestUser()
+    const ids: string[] = []
+    for (let i = 0; i < 3; i++) ids.push((await createFeedPost(user, postInput())).id)
+
+    const page1 = await listFeedPosts(user, 2)
+    expect(page1.map((p) => p.id)).toEqual([ids[2], ids[1]])
+
+    const last = page1[page1.length - 1]!
+    const page2 = await listFeedPosts(user, 2, { created_at: last.created_at, id: last.id })
+    expect(page2.map((p) => p.id)).toEqual([ids[0]])
   })
 
   test('updates selected fields and leaves others intact', async () => {
@@ -275,7 +288,7 @@ describe('Feed posts integration', () => {
     test('appears in the owner feed and the public outbox listings', async () => {
       const user = getTestUser()
       const article = await createArticlePost(user, articleInput({ visibility: 'public' }))
-      expect((await listFeedPosts(user)).map((p) => p.id)).toContain(article.id)
+      expect((await listFeedPosts(user, 10)).map((p) => p.id)).toContain(article.id)
       const publicPosts = await listPublicFeedPosts(user)
       expect(publicPosts.map((p) => p.id)).toContain(article.id)
       expect(publicPosts.find((p) => p.id === article.id)?.article?.title).toBe('A week of sleep and HRV')
--- a/apps/backend/src/db/feed.ts
+++ b/apps/backend/src/db/feed.ts
@@ -121,12 +121,30 @@ export const createArticlePost = async (user: string, input: ArticlePostInput): 
   return mapFeedPost(result.rows[0])
 }
 
-export const listFeedPosts = async (user: string): Promise<FeedPostRecord[]> => {
+/** Keyset position in the owner's feed: the previous page's last `(created_at, id)`. */
+export interface FeedPostCursor {
+  created_at: Date
+  id: string
+}
+
+/**
+ * A keyset page of the owner's feed posts, newest-first (#1012). The `id`
+ * tiebreaker keeps ordering deterministic when two posts share a `created_at`
+ * (microsecond collision on rapid inserts); pass the previous page's last
+ * `(created_at, id)` as `before` for the next page.
+ */
+export const listFeedPosts = async (
+  user: string,
+  limit: number,
+  before?: FeedPostCursor,
+): Promise<FeedPostRecord[]> => {
   const result = await query<FeedPostRow>(
     user,
-    // `id` tiebreaker keeps ordering deterministic when two posts share a
-    // `created_at` (microsecond collision on rapid inserts).
-    `SELECT ${FEED_POST_COLUMNS} FROM feed_posts ORDER BY created_at DESC, id DESC`,
+    `SELECT ${FEED_POST_COLUMNS} FROM feed_posts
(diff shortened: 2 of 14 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped change that mirrors the already-merged home-timeline pagination end to end: DB keyset predicate + `LIMIT`, a shared `getFeedPage` service used by both `GET /feed` and the MCP `list_feed` tool (parity preserved, including the deliberate `structured` omission on MCP), api-spec query/response schemas, the web `useInfiniteQuery`, and the docs. Extracting `encodeKeysetCursor`/`decodeKeysetCursor` and having `timeline.ts` consume them is a good dedupe, and the new tests cover the interesting boundaries (exactly-full page, malformed cursor).

I checked the things most likely to bite and they're fine: the row comparison `(created_at, id) < ($1, $2)` matches `ORDER BY created_at DESC, id DESC` correctly; per-user DB scoping means the opaque cursor can't be used to reach another user's rows; the extra `limit + 1` row is sliced off before serialisation; `validateQuery` replaces `req.query` so the `limit` default/coercion actually applies; `invalidateQueries({ queryKey: ['feed'] })` still matches the new `['feed', 'posts']` key, and React Query recomputes page params on infinite refetch, so post-create/delete refetches won't duplicate or drop rows even though the cursor row may no longer exist. `packages/api-spec/generated/` is gitignored, so no regeneration is expected in the diff. Approving.

Please don't push fixes for these here; a push just starts another review round. They're better batched, and the first two apply equally to the already-merged timeline pagination, so one follow-up can cover both surfaces.

1. **Cursors lose sub-millisecond precision, so a page boundary can silently skip a post.** `feed_posts.created_at` is `TIMESTAMPTZ DEFAULT NOW()` (microsecond resolution), but `pg` truncates to milliseconds when building the JS `Date`, and `encodeKeysetCursor` then stores `ts.getTime()`. If the last row of a page is `…123456µs`, the cursor says `…123000`, and the next query's `created_at < '…123000'` also excludes any row in `(123000, 123456)` — i.e. a post created in the same millisecond but slightly earlier is dropped from the listing entirely. That is exactly the "microsecond collision on rapid inserts" scenario the `listFeedPosts` docstring says the `id` tiebreaker handles; the tiebreaker does order those rows correctly, but the ms-resolution cursor can't address between them. The same holds for `timeline_entry.published_at`. It also makes the new integration tests theoretically flaky: `page2` expects `[ids[0]]`, which fails if `ids[0]` and `ids[1]` land in the same millisecond. Fixing means carrying the timestamp at full precision through the cursor (e.g. select `created_at` as an epoch-microsecond string / ISO text for cursor purposes) rather than round-tripping a JS `Date`.

2. **`decodeKeysetCursor` can still hand an invalid `Date` to the SQL layer.** `Number.isSafeInteger(ms)` admits values outside the `Date` range — `Number.MAX_SAFE_INTEGER` (9007199254740991) is a safe integer but above the 8.64e15 limit, so `new Date(ms)` is an `Invalid Date`, which node-postgres stringifies into a non-parsable timestamp and Postgres rejects. That's the same 500-on-a-crafted-cursor the helper's comment and the `"…instead of reaching the SQL layer"` test claim to prevent (it's self-inflicted and authenticated, hence non-blocking). One extra condition — `Number.isFinite(new Date(ms).getTime())` — closes it, and a case in `keyset-cursor.test.ts` would pin it.

3. **Index doesn't match the new access path.** `idx_feed_posts_created` is `(created_at DESC)` only, whereas the timeline got `(published_at DESC, id DESC)` precisely for this predicate. A row-wise comparison can only become an index qual on a multicolumn B-tree whose leading columns match, so later pages here fall back to scanning the newer index entries and filtering rather than seeking straight to the cursor position. Harmless at current feed sizes, but it undercuts the point of keyset pagination.

4. **Test name contradicts its assertion.** `'an exactly-full page ends with a cursor whose next page is empty and final'` in `services/feed.integration.test.ts` actually asserts `next_cursor` is `null` (correct, given the `limit + 1` probe). The name reads like the opposite design; renaming to something like `'an exactly-full page reports no next page'` would stop it misleading a future reader.

5. **A couple of comments narrate the code below them.** The `getFeedPage` docstring spells out "fetch `limit + 1` rows …, serialise the page, and encode the last row's `(created_at, id)`", which is just a restatement of the eight lines under it — its `getTimelinePage` twin says the same thing in half the space. Likewise `// 2 rows for limit 2: no extra row was fetched, so no next page.` in the test. The `(#1012)` refs and the injected-fetcher rationale are worth keeping.

PR #1024, opened 2026-08-20
fix(web): always precheck Distance + Calories in the share dialog

fiddur:
Fredrik's feedback on the live feed: Distance is missing from the summary numbers on his running/walking/canoeing/rowing shares. Checking the public posts confirms `distance` was never in `included_metrics` on any of them.

**Root cause:** the share dialog's default selection mirrors the metrics shown on the activity chart (`defaultsFromChart`), but `distance` and `calories_active` are cumulative and deliberately excluded from the chart (`EXCLUDED_METRICS`) — so chart-mirroring can never precheck Distance or Calories, the two most staple share stats. (His posts have Calories only because he re-checked it by hand each time.)

**Fix:** `defaultsFromChart` now always includes the cumulative staples `distance` + `calories` alongside `duration`, before the chart-mirrored summaries (deduped in case a custom chart charts distance). Availability filtering is unchanged: the dialog only offers — and `buildShareBody` only sends — metrics the activity actually has data for, so a yoga session still gets no Distance checkbox.

Existing posts aren't rewritten; each can gain Distance by opening **Edit** on the post and ticking it (it's offered, just wasn't prechecked).

Tests updated + a new case pinning the staple behavior; web `pnpm check` green.

### DIFF
--- a/apps/web/src/components/feed-metrics.test.ts
+++ b/apps/web/src/components/feed-metrics.test.ts
@@ -20,20 +20,29 @@ describe('defaultsFromChart', () => {
 
   it('maps charted metrics to their summaries (HR → avg + max + zones) and full series', () => {
     const { summary, series } = defaultsFromChart(['heart_rate', 'distance'])
-    expect(summary).toEqual(['duration', 'distance', 'heart_rate_avg', 'heart_rate_max', 'hr_zone_minutes'])
+    expect(summary).toEqual([
+      'duration',
+      'distance',
+      'calories',
+      'heart_rate_avg',
+      'heart_rate_max',
+      'hr_zone_minutes',
+    ])
     expect(series).toEqual(['heart_rate'])
   })
 
-  it('includes series for a charted metric with no scalar summary (e.g. speed)', () => {
+  it('always prechecks the cumulative staples the chart cannot mirror (distance, calories)', () => {
+    // Distance/calories are excluded from the chart by design, so chart
+    // mirroring alone would never precheck them — a run without Distance reads
+    // wrong. The dialog drops them again for activities without the data.
     const { summary, series } = defaultsFromChart(['speed'])
-    // speed has no summary source; duration is always included.
-    expect(summary).toEqual(['duration'])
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Small, well-scoped fix. `defaultsFromChart` now always seeds `distance` + `calories` alongside `duration`, which is the right call given `ActivityChart`'s `EXCLUDED_METRICS` deliberately keeps both cumulative sources off the chart — so the mirror pass could never have prechecked them, and a run shared without Distance did read wrong.

I verified the "silently drop out when the activity lacks the data" claim end to end: `useShareableMetricOptions` filters `SUMMARY_METRICS` by `m.source` presence in the window, `buildShareBody` only emits keys from `summaryOptions`, and `resolveSharedScalars` skips any `METRIC_SCALARS` key whose aggregate is `undefined`. So a yoga session doesn't get an empty Distance stat, and the pre-availability window (where `present` is `undefined` and every option is offered) is still covered by the backend skip. Ordering of the seeded array is irrelevant downstream — the dialog stores it in a `Set` and `included_metrics` is emitted in `summaryOptions` order — so no rendering change falls out of the new position of `calories`. The dedupe guard in the mirror filter keeps a charted `distance` from appearing twice; harmless in the dialog (Set) but it keeps the tests' array equality honest. Tests were updated to match, including the two cases that now assert the staples on chart selections that don't contain them.

- The comment-to-code ratio here is high for what the code does: an 8-line JSDoc over a one-line const, plus an inline note inside the function, plus a 3-line comment in a test whose title already says the same thing (`always prechecks the cumulative staples the chart cannot mirror (distance, calories)`). The genuinely non-obvious bit — "these are never on the chart, and the availability gate drops them again" — is worth one or two lines; the rest restates the code. Worth trimming whenever this file is next touched.
- `CUMULATIVE_STAPLES` is inferred as `string[]`, so a typo'd key would silently never match a `SUMMARY_METRICS.key`. Typing it against the summary keys (or deriving it from `SUMMARY_METRICS`) would make that a compile error.

Please don't push for either of these — they'd only cost a review round.

Issue #994, opened 2026-08-19
Share a challenge to the feed (share button + personal note)


fiddur:
A challenge is currently shareable only as a bare link (copy-link / native share sheet). It should be shareable **to the federated feed** like an activity or article, with a personal message.

## Scope

**Share button on the challenge page** (`/u/:username/:slug` challenge view + the owner's `/challenges` list): "Share to feed" opens a dialog with:

- an optional **personal note / description** (markdown, rendered through the shared markdown sanitiser — same pipeline as article prose),
- audience (`public` / `unlisted` / `followers` — the existing `FeedVisibility`),
- a preview of what the post will contain.

**Post model** — phased:

1. **v1 — text + link + image.** A new lightweight feed-post variant carrying the user's note, the challenge name, and the canonical challenge URL. Deliver as `Create{Note}` (note text flattened to HTML + the link; Mastodon renders the link with the existing OG preview since challenge pages already emit rich meta). Optionally attach a rendered **race-chart PNG** snapshot (standings at share time) the way activity shares attach chart.png.
2. **v2 — structured enrichment.** Extend `FeedStructuredPost` with a `challenge` kind (challenge spec + a standings snapshot or live pointer), so an Aurboda peer's timeline renders a native standings widget instead of the flat text — same enrichment channel as activities/articles (`GET /public/:user/feed/:postId`).

## Design notes

- Storage: same `feed_posts` table, new `kind` value (the article slice already established the kind discriminator + JSONB payload pattern).
- The personal note is the first case of **user-written text on a feed post** — design the column/schema so activity shares can reuse it (see the sibling issue about including an activity's notes/message in shares).
- Sharing someone *else's* challenge you joined should also work (the post references the host's challenge URL; you're sharing a link + your note, not their data).
- Not in scope: auto-posting standings updates or challenge results (could be a later "share result" action when a challenge ends).

Anyone joining via the shared post follows the existing join-by-URL flow, so this composes with federation as-is.

fiddur:
**Phase 1 shipped** in #1027 (deploying to aurboda.net): every row on /challenges — hosted and joined — has a **Share to feed** button opening a dialog with a markdown note (live sanitised preview) + audience; the post carries the server-resolved challenge name + canonical URL, federates as `Create{Note}` (Mastodon shows the link with the page's OG preview card), and renders natively on the own-feed card. Also available as `POST /feed/challenges` / the `share_challenge` MCP tool.

[...]

PR #1027, opened 2026-08-20
feat(feed): share a challenge to the federated feed (#994 phase 1)

fiddur:
Part of #994 (phase 1 — the issue stays open for the phase-2 structured `challenge` enrichment and the optional race-chart PNG).

A challenge was shareable only as a bare copy-link; it can now be published to the federated feed like an activity or article, with a personal note.

## Post model
- **New feed-post kind `challenge`** on the same `feed_posts` storage (additive `challenge JSONB` column, registered in `tableCreationOrder` — the schema guard test covers the pairing). The payload is a snapshot of the challenge **name + canonical public URL** (+ `host_identity` for a joined remote challenge).
- **Server-side link resolution**: the body carries exactly one of `challenge_id` (own challenge → `buildShareUrl(webHost, user, slug)`) or `participation_id` (joined challenge → its stored `challenge_url`), resolved by the shared `services/challenge-share.ts` — a post can never carry a client-spoofed link. **No join token and no standings data** are embedded; joining goes through the normal public-page flow.
- **The note is markdown** (issue requirement), reusing the `message` column: the federated Note renders it through the article-prose outbound sanitiser (`renderProse`, now exported — #910's boundary), the web through the shared DOMPurify `renderMarkdown` sink.

## Federation
- `Create{Note}`: bold name heading + sanitised note + plain link, addressed per visibility. Mastodon shows the link with the challenge page's **existing OG preview card** (no new meta work needed).
- Object dispatcher + outbox gained the challenge branch; the generic `PATCH /:postId` routes challenge edits through the new `updatedChallenge` deliver hook (like the article fix before it); `DELETE` tombstones unchanged (kind-agnostic).

## Surfaces (API/MCP/web parity)
- REST `POST /feed/challenges` + MCP `share_challenge`, both on the shared resolver + `createChallengePost`.
- Web: **Share to feed** on every `/challenges` row — hosted **and** joined — opening a dialog with the markdown note, a live sanitised **preview** of the post, and the `public`/`unlisted`/`followers` audience. The owner's feed card renders the post natively (name, note, link). A remote Aurboda/Mastodon receiver renders the delivered sanitised HTML as usual.
- Editing: message/visibility are editable via the generic PATCH / `update_feed_post` (no dedicated edit dialog in phase 1); Unshare retracts as usual.

## Tests
- `renderChallengeShareHtml` unit (markdown sanitisation, XSS boundary, name escaping, link rel).
- `buildChallengeNote`/`buildChallengeNoteCreate` against a real Fedify context (ids, addressing, content).
- Router integration (live DB): own-challenge share resolves the canonical URL, joined-participation share, exactly-one 400s, unknown 404s, generic PATCH federates via `updatedChallenge`.
- DB integration: `createChallengePost` round-trip + appears in owner and public listings.
- Whole-monorepo `pnpm check` green; all 581 web tests + backend activitypub/mcp/feed suites pass locally.

### DIFF
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -111,9 +111,11 @@ export const registerFeedTools = (server: McpServer, user: string, options: Feed
         visibility: body.visibility,
       })
       if (!record) return errorResponse('Feed post not found')
-      // Federate the edit as an Update, same as the REST update route. An article
-      // has no linked activity, so route it through the article path.
+      // Federate the edit as an Update, same as the REST update route. Articles
+      // and challenge shares have no linked activity, so they must go through
+      // their own paths — the generic `updated` would silently no-op.
       if (record.kind === 'article') deliver?.updatedArticle(user, record)
+      else if (record.kind === 'challenge') deliver?.updatedChallenge(user, record)
       else deliver?.updated(user, record)
       return jsonResponse(await serializeFeedPost(user, record))
     },
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -6,6 +6,7 @@
 import {
   createArticleBodySchema,
   feedPostsQuerySchema,
+  shareChallengeBodySchema,
   followActorBodySchema,
   followersQuerySchema,
   shareActivityBodySchema,
@@ -23,6 +24,7 @@ import type { RetroEnrichTrigger } from '../services/timeline-retro-enrich.ts'
 
 import {
   createArticlePost,
+  createChallengePost,
   createFeedPost,
   deleteFeedPost,
   getActivityById,
@@ -35,6 +37,7 @@ import {
 import { isPubliclyVisible } from '../services/activitypub/object.ts'
 import { buildArticleMarkdown, renderableArticleBlocks } from '../services/article-export.ts'
 import { buildArticleContent, mergeArticleContent } from '../services/article.ts'
+import { resolveChallengeShare } from '../services/challenge-share.ts'
 import { getFeedPage, normalizeFeedMessage, serializeFeedPost } from '../services/feed.ts'
 import { serializeFollower } from '../services/followers.ts'
 import { serializeFollowing } from '../services/following.ts'
@@ -56,10 +59,12 @@ export interface FeedToolsOptions {
   followerActions?: FollowerActions
   apiBaseUrl?: string
   retroEnrichTimeline?: RetroEnrichTrigger
+  /** Canonical web origin, to build a shared challenge's public URL (#994). */
+  webHost?: string
 }
 
 export const registerFeedTools = (server: McpServer, user: string, options: FeedToolsOptions = {}) => {
-  const { apiBaseUrl, deliver, followActions, followerActions, retroEnrichTimeline } = options
+  const { apiBaseUrl, deliver, followActions, followerActions, retroEnrichTimeline, webHost } = options
   server.tool(
     'list_feed',
     "List posts you have published to your feed, newest first, with their shared metric selection, series opt-in, and visibility. Pass `cursor` (from a previous call's `next_cursor`) to page.",
@@ -110,6 +115,23 @@ export const registerFeedTools = (server: McpServer, user: string, options: Feed
       // has no linked activity, so route it through the article path.
       if (record.kind === 'article') deliver?.updatedArticle(user, record)
       else deliver?.updated(user, record)
+      return jsonResponse(await serializeFeedPost(user, record))
+    },
+  )
+
+  server.tool(
+    'share_challenge',
+    "Publish a challenge INVITATION to your feed: your personal note (markdown) plus the challenge's canonical join-by-URL link. Pass exactly one of `challenge_id` (one of your own challenges) or `participation_id` (a challenge you joined, possibly hosted elsewhere) — the server resolves the linked name/URL itself.",
+    { ...shareChallengeBodySchema.shape },
+    async ({ challenge_id, message, participation_id, visibility }) => {
+      const resolved = await resolveChallengeShare(user, { challenge_id, participation_id }, webHost)
+      if (!resolved.ok) return errorResponse(resolved.error)
+      const record = await createChallengePost(user, {
+        challenge: resolved.challenge,
+        message: normalizeFeedMessage(message) ?? null,
+        visibility,
+      })
+      deliver?.createdChallenge(user, record)
       return jsonResponse(await serializeFeedPost(user, record))
     },
   )
--- a/apps/backend/src/api.ts
+++ b/apps/backend/src/api.ts
@@ -63,10 +63,13 @@ import { createOAuthRouter } from './routes/oauth-router.ts'
 import {
   deliverFeedArticlePost,
   deliverFeedArticleUpdate,
+  deliverFeedChallengePost,
+  deliverFeedChallengeUpdate,
   deliverFeedDelete,
   deliverFeedPost,
   deliverFeedUpdate,
   toDeliverableArticle,
+  toDeliverableChallenge,
 } from './services/activitypub/deliver.ts'
 import { createFeedFederation } from './services/activitypub/federation.ts'
 import { createTimelineBackfiller } from './services/activitypub/timeline-backfill.ts'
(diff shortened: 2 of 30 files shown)


apps/backend/src/mcp/feed-tools.ts
```diff
@@ -114,6 +119,23 @@ export const registerFeedTools = (server: McpServer, user: string, options: Feed
...
+    { ...shareChallengeBodySchema.shape },
+    async ({ challenge_id, message, participation_id, visibility }) => {
+      const resolved = await resolveChallengeShare(user, { challenge_id, participation_id }, webHost)
+      if (!resolved.ok) return errorResponse(resolved.error)
+      const record = await createChallengePost(user, {
+        challenge: resolved.challenge,
+        message: normalizeFeedMessage(message) ?? null,
+        visibility,
+      })
+      deliver?.createdChallenge(user, record)
```
fiddur, line 136: `share_challenge` correctly fans out via `createdChallenge`, but the sibling `update_feed_post` tool above (`if (record.kind === 'article') deliver?.updatedArticle(...) else deliver?.updated(...)`) never got the matching `challenge` branch that `PATCH /feed/:postId` received in `feed-router.ts`.

A challenge post has no `activity_id`, and `feedDeliver.updated` in `api.ts` starts with `if (!post.activity_id) return` — so editing a challenge share's visibility or message over MCP silently delivers no `Update{Note}` at all, and followers keep the stale object. Same failure the REST-side comment calls out ("`updated` would no-op"). Needs the same `else if (record.kind === 'challenge') deliver?.updatedChallenge(user, record)` here for API/MCP parity.

fiddur (commented):
🛑 Changes required

Solid, well-scoped phase-1 slice. The parts that matter most are right:

- The linked name/URL are resolved server-side from the user's own challenge or their stored participation, never from the request body — a post can't carry a spoofed link, and no join token or standings data is embedded.
- The XSS boundaries hold: the challenge name and URL go through `escapeXml` into the AS2 `content`, the personal note goes through `renderProse` (the #910 sanitiser), and the web sinks use `renderMarkdown` (DOMPurify) / JSX escaping. The `challenge-object` test covers the `` / `<script>` cases explicitly.
- The object dispatcher and outbox add the challenge branch *after* the `isPubliclyVisible` gate, so a `followers`-only challenge share still never resolves on an unauthenticated fetch.
- The new `challenge` JSONB is additive and idempotent, `kind VARCHAR(12)` still fits `'challenge'`, `mapFeedPost` spreads the row so the new column flows through, and `resolveStructuredContent` correctly returns null for the new kind (so `FeedPostCard` falls through to the native challenge body rather than the escaped `content`).

One blocking item: the MCP `update_feed_post` tool wasn't given the `challenge` branch that the REST `PATCH /feed/:postId` got, so a challenge post edited over MCP silently federates nothing (`deliver.updated` early-returns on `!post.activity_id`). Details inline.

- `getParticipationById` has no direct case in `db/challenges.integration.test.ts` (its sibling `getParticipationByToken` does, and CLAUDE.md asks for one per DB function). It *is* exercised end-to-end by the new feed-router tests, so coverage isn't actually missing — just not where the standard says to put it.
- `docs/features/feed.md` still describes the `message` column as "plain text", but a challenge share stores markdown there (rendered through `renderProse`/`renderMarkdown`). Worth a clause when the docs are next touched.
- `host_identity` is resolved and stored on the share payload but isn't rendered anywhere yet (not in the Note content, not in `ChallengeShareContent`). Fine as groundwork for a later phase — just noting it's currently write-only.
- `resolveChallengeShare` ends with an unreachable `return` purely to satisfy the checker; inverting the second branch (`if (challenge_id != null) { ... }` then fall through to the participation path) would drop the dead arm.

apps/backend/src/mcp/feed-tools.ts
fiddur: Fixed in 9ee7c123 — `update_feed_post` now dispatches `else if (record.kind === 'challenge') deliver?.updatedChallenge(user, record)`, matching the REST PATCH branch, so an MCP edit of a challenge share federates its Update{Note}.

fiddur (commented):
✅ Approved

Solid, well-scoped phase-1 slice. I traced the new `challenge` kind end to end and found no blocking issue.

What I checked and confirmed:

- **Column plumbing is complete.** `challenge` is in `FEED_POST_COLUMNS`, and every read/write in `db/feed.ts` goes through that constant, so `getFeedPostById`, `listFeedPosts`, `listPublicFeedPostsPage` and `updateFeedPost` all round-trip the payload. `mapFeedPost` spreads the row, so no mapper change was needed. The additive `feed_posts_challenge_column` ALTER matches the existing `feed_posts_message_column` pattern and is registered in `tableCreationOrder`.
- **Federation is wired on all three surfaces**, not just delivery: `setObjectDispatcher`, the outbox dispatcher, and the two `deliverFeedChallenge*` calls. Ordering the `toDeliverableChallenge` check before the `post.activity_id == null` bail is right — otherwise the object dispatcher would 404 a shared challenge that a remote server tries to dereference. Tombstoning rides on the existing generic `DELETE /:postId` path unchanged.
- **The generic `PATCH /:postId` no-op trap is closed** in both the REST route and the MCP `update_feed_post` tool, and the integration test asserts `updatedChallenge` fires while `updated`/`updatedArticle` do not — that's the failure this PR's fixup commit was for, and it is now actually covered.
- **Server-side link resolution holds.** `resolveChallengeShare` only reads from the caller's own DB (`getChallengeById` / `getParticipationById` are both `query(user, …)`-scoped), so the stored `url` can't be client-spoofed. The exactly-one guard runs before either lookup, and both 400 branches plus both 404 branches are tested.
- **XSS boundaries are the established ones.** `escapeXml` for the name/URL, `renderProse` (sanitize-html) for the outbound note, `renderMarkdown` (DOMPurify) at both web sinks — matching how `ArticleContent` already renders on the same public-profile page. `resolveStructuredContent` falls through to `null` for `kind: 'challenge'` rather than crashing, and `FeedPostCard`'s new branch runs before the `structured`/`metrics`/`content` fallbacks.

Please do **not** push fixes for these here; another push just starts a fresh review round for no real benefit. Batch them with phase 2.

1. **`getParticipationById` has no direct integration test.** Its siblings `getParticipationByToken` / `getParticipationByUrl` each have one in `db/challenges.integration.test.ts`, and CLAUDE.md asks for integration tests on every DB function. It *is* exercised (hit and miss) through `feed-router.integration.test.ts`, so the coverage exists — just not where the convention puts it.
2. **`host_identity`'s documented shape doesn't match reality, and the field is currently dead.** `challengeShareSchema` describes it as `@user@host`, but every write path stores a profile *URL* — `buildProfileUrl(webHost, parsed.username)` for a local join, `spec.host_identity` (also a `buildProfileUrl` result) for a remote one. The `'@bob@peer.example'` fixture in the new router test is a shape that never occurs in production. Separately, nothing reads `host_identity` yet: it's absent from the AS2 Note and from `ChallengeShareContent`. Worth aligning the description with the URL form (or actually rendering it) when phase 2 touches this.
3. **Peer challenge shares will log a spurious enrichment warning.** A challenge Note ingested from another Aurboda instance always hits `GET /public/:username/feed/:postId`, which 404s because `resolveStructuredContent` returns `null` for this kind. That's a definitive 4xx so it won't retry, but `createAurbodaEnricher` still emits `⚠️ timeline enrichment failed` — and that warning exists specifically so a post losing its native chart is *visible* in the logs. It will now fire routinely for a perfectly healthy post kind. Either skip enrichment for challenge Notes or give them a structured kind in phase 2.
4. **The OG-preview claim is only true for public challenges.** `docs/features/challenges.md` and `feed.md` both say Mastodon renders the link with the challenge page's OG card, but `share-html-router` emits rich meta only when `is_public` is true — an unlisted challenge share gets generic site meta. The link still resolves and is still joinable, so this is a docs sentence, not a bug.
5. **Import order drift.** `shareChallengeBodySchema` sits between `feedPostsQuerySchema` and `followActorBodySchema` in `mcp/feed-tools.ts`, and `challenge-object.ts` puts the `./article-object.ts` sibling ahead of the `../charts/chart-svg.ts` parent. `pnpm check` runs only `tsgo` + `oxlint` (no `oxfmt --check`), so CI won't catch either — a `pnpm fix` next time round will.
6. **`resolveChallengeShare`'s trailing unreachable return.** TypeScript can't narrow the exactly-one invariant, so the guard is genuinely needed as written — but making the second branch an unconditional tail (`if (challenge_id != null) { … }` then `return …participation…`) would drop both the dead return and the comment explaining it.

Comment density is reasonable throughout — the new block comments mostly carry non-obvious rationale (route registration order, why the article path is reused, why the link is resolved server-side) rather than restating the code.

PR #1030, opened 2026-08-20
feat(feed): auto-share rules — publish matching settled activities automatically (#903)

### LINKED ISSUE #903: Auto-share rules (share activities matching a rule)
Original #831 deferred follow-up. Canonical example: **"auto-share runs longer than 15 minutes."**

## Model

A per-user set of **auto-share rules**, each combining a predicate and a share template:

- **Predicate** (reuse the deduction-rule engine's matching where possible): `activity_type` (one or a set), min/max `duration`, min `distance`, `source`, optionally time-of-day/weekday.
- **Share template** — exactly the fields of a manual share: `included_metrics`, `series_metrics`, `include_chart`/`include_map`, `visibility`, and the post **message** policy (use the activity's `notes` verbatim, a fixed template, or empty — see #995).
- **Enabled flag** per rule; all rules **off by default** — auto-publishing physiological/location data is sensitive, so creating a rule is an explicit deliberate act and the UI must state plainly what will leave the instance.

## Evaluation & timing

- Evaluate when an activity lands in ingest — but **debounced/stabilised**, not instantly: synced activities are frequently merged, enriched (HR zones, calories), or re-synced shortly after first landing. Share after a stabilisation delay (e.g. N minutes after last mutation, or on the merge-group anchor only), so the post's scalars/window reflect the settled activity.
- **Dedupe hard:** at most one auto-post per activity/merge-group ever — a re-sync, merge, or edit must not double-post. Record `rule_id` + anchor `activity_id` on the created feed post; skip when any post already references the group. An activity already shared manually is never auto-shared.
- Auto-created posts are ordinary feed posts: editable, unshareable, listed in `/feed` with a "via rule" marker.

## Surface

- CRUD UI on the Feed page ("Auto-share rules" panel), REST + MCP parity (`list/add/update/delete_autoshare_rule`), api-spec schemas.
- A per-rule "recently auto-shared" trail plus a **preview** ("this rule would have matched N activities in the last 30 days") before enabling — reuse the deduction-rule preview pattern.

## Out of scope

Retroactive sharing of historical activities (the preview shows what *would have* matched; enabling only affects new arrivals).

fiddur:
Closes #903

**"Auto-share runs longer than 15 minutes"** — per-user rules that publish a settled activity to the federated feed exactly as a manual share would, with the safety properties the issue demands.

## Model
- New `autoshare_rules` table: **predicate** (activity types — empty = any, min/max duration over the *merged span*, min distance via the `distance` metric, source) + **share template** mirroring `ShareActivityBody` (metrics, series, chart/map, visibility default `followers`, fixed message).
- Rules are **created disabled** everywhere (REST, MCP, web). Enabling is a separate deliberate PATCH that stamps `enabled_at`; the web toggle shows an explicit confirmation stating what will leave the instance.

## Evaluation & timing
- Hooks into the existing window-based activity-mutation notifier (`activityNotifier` in api.ts) — sync/insert/update/merge paths need no changes. Jobs go to a new pg-boss queue with **`startAfter: 600s`** (the repo's first use of delayed jobs): the stabilisation delay from the issue, so the created post's scalars/window reflect the settled activity after merge/enrich/re-sync churn. Worker batches merge windows per user like the deduction/calorie queues.
- The evaluator (`services/autoshare.ts`, fully dependency-injected) enforces, per candidate group:
  - **Hard dedupe** — the created post records `feed_posts.autoshare_rule_id` (also the "via rule" marker), and ANY existing post referencing ANY merge-group member — manual or auto — blocks the group forever. Re-runs are idempotent; a manually-shared activity is never auto-shared.
  - **Never retroactive** — a rule matches only activities whose row was *ingested* after `enabled_at`. This needed a new `activities.created_at` ingest timestamp (additive; pre-existing rows default to migration time, which necessarily predates any enable since the feature ships with the column). Gating on ingest time (not activity time) means a delayed sync of an older workout still counts — the actual use case.
  - **Merge-group aware** — matching and the post use the group **anchor** (earliest start) + the merged span via the same `getOverlappingActivities`/`expandFeedActivityWindow` a manual share resolves.
  - First matching rule (creation order) wins; distance resolves lazily only when an eligible rule constrains it.
- Delivery goes through the same `FeedDeliver.created` fan-out as a manual share; auto-created posts are ordinary posts (editable, unshareable, paginated listing).

## Preview
`POST /autoshare-rules/preview` (+ `preview_autoshare_rule`): dry-run over the last 30 days — "would have matched N activities (merge groups)", deliberately ignoring shared status and `enabled_at` so the number shows the rule's true reach before enabling. Same body schema as create, never inserted (the deduction-rules preview pattern).

## Surfaces (parity)
- REST `/autoshare-rules` (GET incl. per-rule auto-created post counts, POST, PATCH, DELETE, POST /preview) + MCP `list/add/update/delete/preview_autoshare_rule`, sharing the serializer/body-mapper.
- Web: **Auto-share rules** panel on the Feed page — create form (type picker, min duration/distance, metric/series checkboxes with the share dialog's defaults, fixed message, audience) with inline **Preview**, rule list with enable toggles + counts, and an **auto-shared** marker on rule-created posts.
- `docs/features/auto-share-rules.md` + feed.md link.

## Tests
- Evaluator unit suite (13 tests): canonical predicate, empty-type/any, max/source/distance constraints, distance-required-but-absent, dedupe, anchor grouping + once-per-group, `enabled_at` gate (incl. null), open-window/vanished skips, first-match + lazy distance, no-rules fast path; preview ignores shared status.
- DB integration: disabled-by-default + `enabled_at` stamp semantics (forward-only), nullable patches, delete-keeps-posts+counts, dedupe lookup across group members, candidate windowing + ingest times.
- Queue: window merging + delay bound. Schema guard covers the new keys. Whole-monorepo `pnpm check` green; web + backend suites pass locally.

Follow-ups noted in the doc (per the issue's optional items): message-from-activity-notes mode, time-of-day/weekday predicates.


apps/backend/src/services/autoshare.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // Hard dedupe: any existing post referencing ANY group member blocks the group forever.
+    const existing = await deps.postIdsForActivities(
+      user,
+      group.map((a) => a.id),
+    )
+    if (existing.length > 0) continue
+
+    // New-arrivals-only: the anchor row must have been ingested after the enable.
+    const eligibleRules = rules.filter(
```
fiddur: The gate is ingest time only — nothing bounds how *old* the activity may be, `listAutoshareCandidates` has no LIMIT, and there's no cap on posts per run. Together that means a single job can publish months of history.

Concretely: user enables "runs > 15 min", then connects Garmin (or Strava) for the first time. The first sync backfills 90 days (`integrations/garmin/sync.ts:27`) and fires `onActivitySynced(user, '*', now-90d, now)` (`services/sync-provider.ts:93`, `sync-router.ts:227`). Every one of those rows is a fresh insert, so `created_at = now` and *all* of them pass this filter; the loop then creates one post per matching merge group and fans each out through `feedDeliver.created`. Dozens to hundreds of `Create` activities land on followers' and Mastodon timelines in one burst, and federated deliveries can't be recalled. `POST /sync/garmin {full_resync: true, start_date: '2015-01-01'}` widens the window arbitrarily.

Suggest gating on the activity's own time too — e.g. skip when the resolved `window.end_time` predates `rule.enabled_at` (a delayed sync of a workout done *after* enabling still shares, which is the case the docs care about) — and/or a hard cap on posts created per evaluation run.

apps/backend/src/services/autoshare.ts
```diff
@@ -0,0 +1,179 @@
...
+  let created = 0
+  for (const candidate of candidates) {
+    // An empty group means the candidate vanished since listing (deleted) — skip.
+    const group = await deps.getGroup(user, candidate)
+    const anchor = group[0]
+    if (anchor == null) continue
+    if (processedAnchors.has(anchor.id)) continue
+    processedAnchors.add(anchor.id)
+
+    // Hard dedupe: any existing post referencing ANY group member blocks the group forever.
```
fiddur: This dedupe reads live `feed_posts` rows, but `deleteFeedPost` **hard-deletes** the row (`db/feed.ts:280`). So "at most one post per activity/merge-group EVER" (this file's header and `docs/features/auto-share-rules.md`) doesn't hold once a post is deleted.

Scenario: a rule auto-shares a run; the user deletes the post because they didn't want it out there; later they fix the activity's title (`PATCH /activities/:id` → `activityNotifier` → new evaluation over that activity's window), or a merge/re-sync touches it. The group now has no referencing post, the anchor's `created_at` is unchanged so it still passes the `enabled_at` gate, and the rule republishes and re-federates the exact post the user deleted.

This needs a record that survives post deletion — `feed_tombstone` only stores `post_id` today, so either extend it with the activity id or keep auto-shared activity ids in their own table and check that here.

fiddur (commented):
🛑 Changes required

Solid, well-structured feature: injected deps make the whole decision tree unit-testable, the schema/API/MCP/web layers all follow the documented conventions, `include_chart` is derived from the heart-rate series exactly like the manual share dialog, `windowDistanceMeters` mirrors `resolveActivityScalars`' bucketing and `sum` semantics, and the docs are good. Integration + unit coverage is real (dedupe lookup, enable/disable stamping, candidate filtering).

Two blocking issues, both in the safety guards the feature's own header advertises:

1. **Unbounded historical reach.** The `enabled_at` gate is on *ingest* time only — there is no bound on how old the activity itself may be, no LIMIT on `listAutoshareCandidates`, and no cap on posts per evaluation run. A first sync of a newly connected source (90-day backfill, `onActivitySynced(user, '*', now-90d, now)`) or a `full_resync` with `start_date` therefore mass-publishes history to the federated feed in one job. Federated deliveries can't be recalled.
2. **Deleting an auto-shared post un-does the dedupe.** `deleteFeedPost` hard-deletes the row, so the "at most one post per activity/merge-group EVER" property fails: after the user deletes a post, the next mutation of that activity republishes it.

Both are inline below with concrete scenarios.

- **N+1 shape.** `createAutoshareDeps` fetches the same activity three times per candidate (`getGroup` → `getActivityById`, `resolveWindow` → `getActivityById`, `onCreated` → `getActivityById`), and the `enabled_at` filter runs *after* the per-candidate group + dedupe queries even though `candidate.created_at` is already in hand. Threading the fetched `Activity` through the deps (or pre-filtering candidates in SQL) would cut most of it.
- **Idempotent re-enable moves the gate.** `PATCH { enabled: true }` on an already-enabled rule re-stamps `enabled_at = NOW()`, silently skipping anything ingested in between. Guarding the stamp on the rule not already being enabled would make the PATCH idempotent.
- **Web form vs. schema.** The "Min distance (km)" input has `min="0"`, but `min_distance_meters` is `.positive()` — entering `0` produces a 400 and the generic `alert('Failed to create the rule.')`. Same class of thing: `previewMutation` has no `onError`, so a failed preview does nothing visible.
- **Toggle confirm cancel.** In `RuleRow`, declining the enable confirm doesn't call `mutate()`, so nothing re-renders and the checkbox stays visually flipped while the adjacent label still reads "Off".
- **Duplicate posts under concurrency.** Nothing stops two overlapping evaluations from both passing the dedupe check and creating two posts for one group; a partial unique index on `feed_posts(activity_id) WHERE autoshare_rule_id IS NOT NULL` would make it structural rather than timing-dependent.
- **Comments/placement.** In `db/autoshare-rules.ts` the `updateAutoshareRule` JSDoc ("Patch a rule. Flipping `enabled`…") sits directly above `PATCH_COLUMNS`, so it documents the wrong symbol; and `// Flipping enabled ON stamps the no-retroactive-sharing gate.` restates the line under it, which that JSDoc already covers. In `routes/autoshare-rules-router.ts`, `UUID_RE` is declared in the middle of the import block — move it below the imports.

apps/backend/src/services/autoshare.ts
fiddur: Fixed in d8e4f5d6 with both suggested measures: a second eligibility gate requires the activity itself to have ENDED after `enabled_at` (the resolved merged window), so a first-sync/full-resync backfill of fresh rows carrying old workouts never publishes — while a delayed sync of a workout done after enabling still shares. And `MAX_POSTS_PER_RUN = 5` bounds any single evaluation run with a logged skip. Unit tests cover the fresh-ingest-of-old-history scenario and the cap.

apps/backend/src/services/autoshare.ts
fiddur: Fixed in d8e4f5d6 with a dedicated record: new `autoshare_suppressions(activity_id)` table, populated inside `deleteFeedPost`'s delete CTE for ANY activity-anchored post (manual or auto — a deliberately removed share shouldn't come back by rule either), and checked by the evaluator alongside live posts. Survives the hard delete; covered by a DB integration test (create → delete → suppression blocks) and an evaluator unit test.

apps/backend/src/db/autoshare-rules.ts
```diff
@@ -0,0 +1,234 @@
...
+export const listAutoshareCandidates = async (
+  user: string,
+  start: Date,
+  end: Date,
+): Promise<AutoshareCandidate[]> => {
+  const result = await query<AutoshareCandidate>(
+    user,
+    `SELECT id, activity_type, source, start_time, end_time, title, created_at
+     FROM activities
+     WHERE deleted_at IS NULL AND end_time IS NOT NULL
```
fiddur: This should also filter `AND superseded_by IS NULL` (the same predicate chart/trend/deduction queries use).

`getGroup` resolves a group via `getOverlappingActivities`, whose SQL filters `activity_type = $1` — so a group is **same-type only**. Cross-source and override merges are cross-type, and `materializeSuperseded` records them in `superseded_by`, not here. Concretely:

- User edits a Garmin `walking` activity to `running` → an `aurboda`/`running` override row A targeting the Garmin `walking` row B; supersession sets `B.superseded_by = A`.
- Both A and B are candidates. `getGroup(A)` → `[A]`, `getGroup(B)` → `[B]` (type filter excludes the other).
- `postIdsForActivities([A])` and `postIdsForActivities([B])` each come back empty, so **both** get a post — one `running`, one `walking`, for the same session, both federated and unrecallable.

Same shape for a Garmin `running` + Health Connect `walking`/generic-exercise pair. This breaks the "at most one post per activity/merge-group EVER" invariant the module doc and `docs/features/auto-share-rules.md` both state.

Adding the filter is safe and strictly narrowing: within any merge group the winner has `superseded_by IS NULL` and stays a candidate, `getGroup` still expands to the full same-type group for dedupe and window resolution, and rows in a DB where supersession never ran are all NULL, so behaviour is unchanged there. Worth an integration case in `autoshare-rules.integration.test.ts` covering a superseded row.

fiddur (commented):
🛑 Changes required

Solid, well-structured feature: injected deps make the whole decision tree unit-testable, the two "never retroactive" gates plus the per-run cap are the right safety posture for something that publishes irrevocably to a federated network, the suppression record survives the hard post delete, REST/MCP parity is there, and the docs are updated. The window/scalar resolution correctly reuses the same `expandFeedActivityWindow` / bucketed-metric path a manual share uses, so an auto post reports the same numbers.

One blocking gap (inline): `listAutoshareCandidates` doesn't filter `superseded_by IS NULL`, so a cross-source or override merge group is evaluated as two independent groups and can produce two federated posts for one physical session.

Please do **not** push fixes for these here; a new round costs more than the polish is worth. Batch them with other work.

- **Anchor vs. merge winner.** The group anchor is the earliest-start raw row, not the materialized winner. For a same-type group that only affects `source`/`title`, but a rule with `source: 'garmin'` can miss a group whose winner is the Garmin row while the earliest row came from Strava/Health Connect. Worth revisiting once the `superseded_by` filter is in.
- **Cost of a wide window.** A first sync fires a 90-day window (`FIRST_SYNC_FALLBACK_DAYS`), and every candidate costs ~7 queries (`getGroup` ×3, dedupe ×2, `resolveWindow` ×2) *before* the activity-time gate rejects it. Pushing `end_time >= min(enabled_at)` into `listAutoshareCandidates` would drop nearly all of that. Same for `previewAutoshareRule`, which runs that per-candidate loop synchronously inside the HTTP request over 30 days.
- **Comment/doc placement in `db/autoshare-rules.ts`.** The `updateAutoshareRule` docblock sits three declarations above the function it documents (immediately before `PATCH_COLUMNS`, which has its own docblock). And `// Flipping enabled ON stamps the no-retroactive-sharing gate.` / `// Cheap pre-filters before resolving windows/distance.` restate what the adjacent line plainly does and are already covered by the surrounding docblocks.
- **`autoshare-rules-router.ts`:** `const UUID_RE = …` is declared between two import statements; move it below the import block.
- **Web nits.** Cancelling the enable `window.confirm` leaves the checkbox visually checked, since no state changed and nothing re-renders. `previewMutation` has no `onError`, so a failed preview silently does nothing. And `include_chart` is derived from `series.has('heart_rate')` with no visible control — fine, but not discoverable from the form.

apps/backend/src/db/autoshare-rules.ts
fiddur: Fixed in ece41f27 — `listAutoshareCandidates` now filters `superseded_by IS NULL` (the winners-only predicate the chart/trend/deduction queries use), so a cross-type merge loser is never an independent candidate. Integration test adds a superseded health_connect/walking loser next to its running winner and asserts it is excluded.

fiddur (commented):
✅ Approved

Read the whole diff plus the surrounding merge/feed/sync code. This is a careful implementation of a genuinely risky feature (auto-publishing health data to the fediverse), and the safety design holds up under scrutiny:

- **Dedupe** covers every group member and survives the post's hard delete via `autoshare_suppressions` — and the data-modifying CTE in `deleteFeedPost` does run even though nothing selects from it, so that's correct.
- **The two `enabled_at` gates** are conservative in the right direction. I walked the scenarios that worried me most — full re-sync with a months-old `start_date` (the notifier really does emit that window), and connecting a brand-new source that ingests months of history as fresh rows — and both are stopped: the first by the ingest gate (the upsert preserves `created_at`), the second by the activity-time gate. `MAX_POSTS_PER_RUN` is a sensible last line of defence given deliveries can't be recalled.
- **`created_at` migration** is safe (PG 11+ fills existing rows with the ALTER-time value, which by construction predates any enable), and the lazy schema-error migration path means it lands before the first `listAutoshareCandidates`.
- **Distance resolution** mirrors `resolveActivityScalars` exactly (one window-sized bucket, `sum`, `distance` is in `sumMetrics`), so the predicate agrees with what the post displays. `include_chart` following the heart-rate series toggle matches `buildShareBody`'s established rule.
- Auth, per-user DB scoping, fixed patch-column list, `validateBody` writing back parsed defaults, REST/MCP parity, and the docs all check out.

Unit tests exercise the real decision points (both gates, suppression, cap, merge-group anchoring, first-rule-wins, lazy distance) rather than restating the implementation, and the integration tests cover the store plus the suppression-survives-delete invariant.

Nothing blocking. Approving.

Please don't push fixes for these here; a new round costs more than they're worth. Batch them with other work.

1. `apps/backend/src/db/autoshare-rules.ts` — the `/** Patch a rule. Flipping \`enabled\` … */` JSDoc ended up stranded above the `PATCH_COLUMNS` const, so it now documents the wrong symbol. It belongs on `updateAutoshareRule` (~20 lines down).
2. `apps/backend/src/routes/autoshare-rules-router.ts` — `const UUID_RE` sits in the middle of the import block (two `import`s follow it). Hoisting makes it harmless, but it reads like an accidental paste; move it below the imports.
3. `updateAutoshareRule` re-stamps `enabled_at = NOW()` on *any* patch containing `enabled: true`, including one where the rule is already on. The web toggle only ever sends the flipped value, but an MCP call like `{ enabled: true, visibility: 'public' }` would silently move the no-retroactive gate forward and skip activities ingested in between. `enabled_at = CASE WHEN enabled THEN enabled_at ELSE NOW() END` would pin it to "transitions to enabled".
4. `serializeFeedPost` now emits `autoshare_rule_id` on the *public* path too (`GET /public/:username/posts` → `feed-public-router.ts:111`), so anonymous visitors can tell which posts were automated and group them by rule UUID. It's opaque metadata and `activity_id` is already exposed there, so this is minor — but it is owner-only information, and stripping it in the public serialization would cost nothing.
5. Scalability of `evaluateAutoshareWindow`: per candidate it runs `getGroup` (3 queries) plus the two dedupe lookups *before* the cheap `enabled_at` ingest gate. On the wide windows a full re-sync produces that's thousands of round-trips for a run that will almost always create nothing. Batching the dedupe lookups across candidates, or narrowing `listAutoshareCandidates` in SQL, would help — note a naive `end_time >= min(enabled_at)` pre-filter isn't equivalent, since a group's merged span can end later than the candidate row does.
6. `packages/api-spec/src/schemas/autoshare-rules.ts` re-declares a private `scalarMetricKeySchema` identical to the one in `feed.ts`; worth exporting one and sharing it.
7. Web form: entering `0` for min duration/distance sends `0`, which `positive()` rejects, surfacing as the generic "Failed to create the rule." alert. Treating `0` as "unset" (or `min="1"` on the distance input) would be friendlier.
8. Predicate semantics worth documenting: `source` is matched against the *anchor* (earliest-start group member). For the common same-type multi-source merge the anchor is also the merge winner, so this agrees with what the UI shows — but in cross-type merge topologies the anchor can be a superseded row, and then a `source: garmin` rule can quietly fail to match a Garmin activity. Not worth restructuring; a line in `docs/features/auto-share-rules.md` would set expectations.

PR #1033, opened 2026-08-20
feat(feed): live share preview in the Share dialog (#902)

### LINKED ISSUE #902: Live share-preview in the Share dialog
Original #831 scope.

Show a live preview in the Share dialog of (a) the fallback text render a Mastodon follower sees and (b) the chart/route images, so the user sees exactly what leaves the instance before sharing. (May overlap with #884's native post rendering — reuse the card renderer.)

fiddur:
Closes #902

The Share dialog now shows **exactly what leaves the instance** before clicking Share.

- **`POST /feed/activities/:id/preview`** (+ MCP `preview_activity_share` — parity): resolves the exact federated `content` HTML and typed scalars for a given selection through the SAME `feedPostContent`/`resolveActivityScalars` path delivery uses — so the preview is byte-for-byte what a Mastodon follower would see (asserted by an integration test: preview `content` === a subsequent real share's `content`; nothing is created by the preview). Shared `previewActivityShare` service in `services/feed.ts`.
- **Dialog**: a Preview box below the audience selector — server-resolved, debounced 400 ms (no request per keystroke), react-query-cached per serialized selection, with a note saying which images (heart-rate chart / route map) would attach. The rendered HTML is the server-built, escaped content of the user's own data — the same trusted string the owner feed card already renders.
- Works in create AND edit mode (same resolution either way).
- The issue's "(b) chart/route images" half: the actual PNGs don't exist pre-post (they render per post id), so the preview states which images will attach; the interactive chart the user just configured is on the detail page behind the dialog. The federated text render — the part invisible until now — is the live half.

## Tests
- Integration: preview↔share content parity + nothing-created + 404 for unknown activity.
- Whole-monorepo `pnpm check` green; all 581 web tests + feed-router suite (19) pass locally.

### DIFF
--- a/apps/backend/src/mcp/feed-tools.ts
+++ b/apps/backend/src/mcp/feed-tools.ts
@@ -38,7 +38,12 @@ import { isPubliclyVisible } from '../services/activitypub/object.ts'
 import { buildArticleMarkdown, renderableArticleBlocks } from '../services/article-export.ts'
 import { buildArticleContent, mergeArticleContent } from '../services/article.ts'
 import { resolveChallengeShare } from '../services/challenge-share.ts'
-import { getFeedPage, normalizeFeedMessage, serializeFeedPost } from '../services/feed.ts'
+import {
+  getFeedPage,
+  normalizeFeedMessage,
+  previewActivityShare,
+  serializeFeedPost,
+} from '../services/feed.ts'
 import { serializeFollower } from '../services/followers.ts'
 import { serializeFollowing } from '../services/following.ts'
 import { getSettings } from '../services/settings.ts'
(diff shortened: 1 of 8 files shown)


fiddur (commented):
✅ Approved

Clean, well-scoped feature. The preview endpoint is auth-gated and user-scoped through `getActivityById`, it creates nothing, and it reuses the exact same `expandFeedActivityWindow` → `resolveActivityScalars` → `feedPostContent` chain the share/delivery path uses — so the "byte-for-byte identical" claim holds. `normalizeFeedMessage` is applied identically to the share route (including the empty-string-clears case that `buildShareBody` always sends), MCP parity mirrors `share_activity` exactly, and `dangerouslySetInnerHTML` is safe here because `feedPostContent` escapes title/message/scalars server-side. The integration test earns its name: it asserts the preview content equals the content of a real share of the same selection, and that no post was created. Reusing the existing `.share-dialog-preview` CSS and POSTing a complex selection body (matching the `/autoshare-rules/preview` precedent) are both consistent with the codebase.

Please do **not** push fixes for these here; a push starts another review round for no real benefit. Batch them with other work if you touch this area again.

- `previewActivityShare` (`apps/backend/src/services/feed.ts`) re-implements the scalars → settings → `feedPostContent` → metrics-mapping pipeline that `resolveActivityPresentation` already does a few lines above. Extracting one shared helper would make the identical-output guarantee structural rather than by-convention — right now a change to one can silently drift from the other, which is precisely what this feature promises can't happen.
- The preview query is keyed on the **whole** share body, but the endpoint only reads `included_metrics` and `message`. Toggling the route map, a series opt-in, or visibility refetches an identical preview — and `canChart` flipping from its optimistic `true` to the resolved value after the availability query lands causes one extra fetch on open. Keying on just `{ included_metrics, message }` would drop those.
- Round-tripping the body through `JSON.stringify` / `JSON.parse(debouncedKey)` works but is a bit indirect; storing the debounced body object in state next to the key would avoid the parse.
- Without `placeholderData: (prev) => prev`, the preview collapses to "Loading…" at every debounce boundary while typing, rather than holding the previous render.
- Copy nit: with the map off and the HR series on, the note reads "Exactly what a follower on Mastodon sees and the heart-rate chart image."
- `docs/features/feed.md`: the new row sits between `POST /feed/articles` and `POST /feed/challenges` rather than next to the `/share` row it belongs with, and its pipe padding differs from the neighbours (markdown renders fine either way; `oxfmt` doesn't cover `.md`).
- Test nit: `listBefore` is fetched *after* the preview — `listAfterPreview` would say what it checks.

PR #1034, opened 2026-08-20
feat(feed): federate the QuantPub quant:Exercise extension + serve its @context (#896)

### LINKED ISSUE #896: Federate the aurboda:Exercise extension + serve /ns/activitystreams @context
Original #831 scope, not yet done.

## Problem
Delivery currently sends a plain Fedify `Note` (Fedify's typed vocab drops unknown props). The rich `aurboda:Exercise` extension (typed `aurboda:metrics`, `aurboda:startTime/endTime`, `aurboda:series` links) is built in `services/activitypub/object.ts` (`buildCreateExercise`) but **unused** in delivery, and no JSON-LD `@context` doc is served.

## Scope
- Serve the `aurboda:` JSON-LD `@context` at `/ns/activitystreams` (+ nginx route).
- Deliver the dual-typed `[Note, aurboda:Exercise]` object with the extension + series links (via Fedify custom properties or a content-negotiated object), so Aurboda↔Aurboda gets structured data while Mastodon still renders the Note fallback.

fiddur:
Closes #896.

Implements the remaining #831-scope federation item, aligned with the QuantPub FEP (`docs/fep/quantpub.md`) rather than the issue's original `aurboda:` naming — the FEP superseded it and states Aurboda intends to adopt `quant:`.

## What
- **Dual-typed delivery**: every activity share's Note — delivered `Create`/`Update`, outbox listing, and the dereferenceable object — is `["Note", "quant:Exercise"]` with:
  - native AS2 `startTime`/`endTime` (the FEP reuses AS2's own window terms),
  - `quant:activityType`, `quant:metrics` (shared scalars as `key`/`value`/`unit`),
  - `quant:series` links (public/unlisted + bounded window only — the public `/series` endpoint would 404 anything else),
  - `quant:structuredUrl` (carries the capability token on followers-only deliveries, FEP §9),
  - inline `@context` entry defining `quant:` with `@json`-literal terms.
- **New endpoints** (+ nginx routes): `GET /.well-known/quantpub` (FEP §4 discovery) and `GET /ns/quantpub` (published JSON-LD context, `application/ld+json`). `WellKnownQuantpub` schema added to api-spec.
- **Mechanism**: Fedify's typed vocab drops unknown properties, so `quant-extension.ts` wraps the outermost object's `toJsonLd` and splices the extension into the final serialized document (after Fedify's JSON-LD compaction). Wrapping only the outermost object avoids the child-splice being re-compacted into expanded-IRI form. Inbound is safe: Fedify's `fromJsonLd` type check is an `includes()`, so extra types pass through.
- **Removes the unused parallel `aurboda:` AS2 builder** (`buildCreateExercise`/`buildFeedPostActivity`) the FEP flagged as diverging; `feedPostContent`/addressing stay in `object.ts`, scalar resolution in `feed-activity.ts`.

## Notes
- `quant:series` links for a **merged** share are built over the merged span while the public series endpoint still authorizes the anchor's own window, so those links can 404 (documented in feed.md known limitations; QuantPub consumers treat series fetches as best-effort). Expanding series authorization across merge groups stays the existing planned follow-up.
- Aurboda↔Aurboda enrichment continues out-of-band via the FEP §7 id convention (unchanged); `/.well-known/aurboda` stays for challenge federation.

## Tests
- `quant-extension.test.ts`: extension props (series gating, token), document splicing (bare/activity/idempotence), real Fedify Note round-trip.
- `federation.integration.test.ts`: served object carries dual type, quant fields, AS2 window, and the @context entry; followers-only delivery carries the token on `quant:structuredUrl` and no series links.
- `well-known-router.test.ts`: discovery + context documents.
- Whole-monorepo `pnpm check` green; all 2352 backend unit tests + 21 federation integration tests pass.

### DIFF
--- a/apps/backend/src/routes/well-known-router.test.ts
+++ b/apps/backend/src/routes/well-known-router.test.ts
@@ -40,3 +40,32 @@ describe('well-known/assetlinks.json', () => {
     expect(res.headers['cache-control']).toMatch(/max-age=3600/)
   })
 })
+
+describe('well-known/quantpub (#896)', () => {
+  const app = buildApp({ androidFingerprints: [], androidPackageName: 'net.aurboda.app' })
+
+  test('serves the QuantPub discovery document (FEP §4), cacheable', async () => {
+    const res = await supertest(app).get('/.well-known/quantpub')
+    expect(res.status).toBe(200)
+    expect(res.body).toEqual({
+      api_base: 'http://localhost:3000',
+      product: 'aurboda',
+      quantpub: '0.1',
+      version: 'test',
+    })
+    expect(res.headers['cache-control']).toMatch(/max-age=3600/)
+  })
+
+  test('serves the published JSON-LD @context at /ns/quantpub as application/ld+json', async () => {
+    const res = await supertest(app).get('/ns/quantpub')
+    expect(res.status).toBe(200)
+    expect(res.headers['content-type']).toContain('application/ld+json')
+    expect(res.body).toEqual({
+      '@context': {
+        quant: 'https://w3id.org/quantpub#',
+        'quant:metrics': { '@type': '@json' },
+        'quant:series': { '@type': '@json' },
+      },
+    })
+  })
+})
--- a/apps/backend/src/routes/well-known-router.ts
+++ b/apps/backend/src/routes/well-known-router.ts
@@ -7,10 +7,11 @@
  *   signing fingerprints (e.g. for a custom-built APK) via the
  *   `ANDROID_APP_FINGERPRINTS` env var.
  */
-import type { WellKnownAurboda } from '@aurboda/api-spec'
+import type { WellKnownAurboda, WellKnownQuantpub } from '@aurboda/api-spec'
 
 import type { TypedRouter } from '../typed-router.ts'
 
+import { quantContextDocument, QUANTPUB_VERSION } from '../services/activitypub/quant-extension.ts'
 import { typedRouter } from '../typed-router.ts'
 
 export interface WellKnownConfig {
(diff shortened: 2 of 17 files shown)


fiddur (commented):
✅ Approved

Solid, well-scoped implementation of the QuantPub `quant:` extension. I traced the splice mechanism through Fedify 2.3.1's actual serialization code rather than taking the module docstring on faith, and the central claims hold up:

- **Delivery** (`ctx.sendActivity` → `Federation.sendActivity`) serializes with `activity.toJsonLd({format:'compact'})`, so the patched `toJsonLd` runs *after* the full `jsonld.compact` round-trip — the `quant:` terms are spliced onto the already-compacted document and can't be re-expanded. The embedded `object` has no `@context` at that point, and `spliceQuantExtension` correctly skips it and extends only the top-level one, which is in scope for the nested node.
- **The clone caveat is accurate.** `sendActivity` only clones via `signObject` when an Ed25519 key is present; Aurboda provisions RSA only (`keys.ts`), and `normalizeOutgoingActivityJsonLd` / `signJsonLd` operate on the spliced document in place. `signJsonLd` failures are caught and downgraded to a warning, so even a canonicalization hiccup can't break delivery.
- **The outbox works too**, though it's the least obvious case: the `Create` is not actually the outermost document there. `OrderedCollectionPage.isCompactable()` chains up to `Object.isCompactable()` and returns true (no `content`/`name`/`summary`/`proof`), so the collection's fast path embeds each item's `toJsonLd()` output verbatim — the spliced `Create`, with its own scoped `@context`.
- **No ingest regression.** Fedify's `Object.fromJsonLd` dispatches on `@type.includes(...)`, so `["Note","quant:Exercise"]` still parses as a `Note` on a receiving Fedify instance; the extra type and terms are ignored. `assertSafeJsonLd` only rejects `@graph`/`@included`/`@reverse`.
- **No token leak.** `quant:structuredUrl` carries `?token=` only when `!isPubliclyVisible`, and both unauthenticated surfaces that build it (the object dispatcher and `listPublicFeedPostsPage`) filter out `followers`-only posts first — so the token appears only in the copy delivered to followers, matching the existing `imageAttachments` rule and FEP §9.
- `end_time` is genuinely `Date | undefined` (never `null`) all the way from `mapActivityRow`, so the new `activity.end_time === undefined ? null : …` guard is safe. `series_metrics` exists on `FeedPostRecord`, so every `DeliverablePost` call site satisfies the new required field. The discovery document matches FEP §4 field-for-field, and nginx exposes both new paths at the web base URL as the spec requires.

The test churn is a net wash rather than a loss: the `buildCreateExercise` cases deleted along with the builder are re-covered by `quant-extension.test.ts` and the new `feedPostContent` cases (message placement, blank message, title escaping, name fallback), plus two new integration tests over the real served object and the followers-only delivered `Update`.

Nothing here needs a push on this PR; batch it with other work.

- `apps/backend/src/services/activitypub/timeline-enrich.ts:5` still says "Fedify's typed vocab drops the `aurboda:` extension" — stale on both counts now. One-line doc fix.
- The `/ns/quantpub` route types its response as `Record<string, unknown>`, which CLAUDE.md's routing conventions say should be a named response type. A tiny local `QuantpubContextDocument` interface next to `AssetLinksStatement` would match the existing pattern in that file. The `well-known-router.ts` header comment also still describes the file as only `/.well-known/*`.
- `quantContextEntry` is a shared mutable object inserted **by reference** into every serialized `@context`. Nothing mutates it today, but `Object.freeze` (or returning a fresh literal from `withQuantContext`) would make that structurally impossible.
- Worth an integration assertion that `/users/:username/outbox` items carry `["Note", "quant:Exercise"]`. It's the one path where the wrapped object is *not* the outermost document handed to Fedify, so it rides on `isCompactable()` staying true for collection pages — a Fedify upgrade could silently regress it, and nothing currently guards that.
- `withQuantType` is idempotent but `withQuantContext` isn't; a second splice would duplicate the context entry. Unreachable today (`buildFeedNoteParts` hands the unwrapped Note to the activity builders), but cheap to make symmetric.

PR #1038, opened 2026-08-20
docs(fep): QuantPub rev 5 — final pre-submission pass + r/QuantifiedSelf draft (#905)

fiddur:
Part of #905 (does not close it — submission to the FEP repo and the actual Reddit post are Fredrik's maintainer actions).

Folds the three rev-5 items recorded on #905 from the #1002 rev-4 review:

- **§9 token conveyance** is now a conditional MUST with a stated fallback: image-attachment URLs whenever the post carries images (unchanged, and mandatory in that case); `quant:structuredUrl?token=…` named as the only channel for **attachment-less** followers-only posts (e.g. a sleep observation with no chart), with a SHOULD to attach a tokenised image whenever the post has any renderable image, and the consumer fallback order spelled out. Matches what Aurboda ships since #1034 (structuredUrl always carries the token on followers-only deliveries).
- **§7 step 2** no longer unconditionally requires the `/.well-known/quantpub` fetch: a same-host `quant:structuredUrl` names the payload URL directly, so discovery is only needed to resolve `{api_base}` under the id convention; it MAY remain as a capability check but MUST NOT be required.
- **FEP-400e dates verified** against the Codeberg FEP index (`dateReceived: 2021-02-16`, `dateFinalized: 2022-02-04`) — the citation was already correct; no change.

Also adds `docs/fep/quantpub-reddit-draft.md` — the step-3 r/QuantifiedSelf post drafted in-repo for review as the issue requests, with placeholders for the Mastodon-vs-subscriber screenshot pair and a pre-post checklist.

### DIFF
--- a/docs/fep/quantpub-reddit-draft.md
+++ b/docs/fep/quantpub-reddit-draft.md
@@ -1,0 +1,78 @@
+# DRAFT: r/QuantifiedSelf post (QuantPub outreach, #905)
+
+> Status: **draft for review — not yet posted.** Fredrik posts this himself
+> after review, with the two screenshots captured (see placeholders). Suggested
+> flair: "Apps & Tools" (or whatever the sub currently uses for tooling).
+
+---
+
+**Title:** Making our QS tools social: a small ActivityPub vocabulary so
+home-built trackers can follow each other (RFC — looking for collaborators)
+
+---
+
+A lot of us run home-built or self-hosted tracking systems — and they're all
+islands. I've been experimenting with making mine *federated*: my workouts and
+sleep observations are ActivityPub posts, so anyone on Mastodon can follow me
+and see them. But between two tools that speak a small extra vocabulary,
+something better happens: the subscriber renders the *actual data* — an
+interactive chart with real hoverable values, not a flattened text summary and
+a static PNG.
+
+Here's the same shared run, seen by a plain Mastodon follower vs. a
+QuantPub-speaking subscriber:
+
+**[SCREENSHOT 1: Mastodon view — flattened text + PNG attachment]**
+
+**[SCREENSHOT 2: Aurboda subscriber view — native card: stat grid, personal
+message, interactive multi-metric hover chart + synced route map]**
+
+The interesting part is that the mechanism is deliberately boring, because
+mainstream fediverse servers drop anything fancy. The whole contract, which
+I've written up as a draft FEP (Fediverse Enhancement Proposal) called
+**QuantPub**, is:
+
+1. **Deliver a plain `Note`** (so Mastodon renders something sensible), with a
+   few extra typed fields riding along: activity type, the scalar summaries
+   you chose to share (`{key, value, unit}`), and links to series data.
+2. **Serve a machine-readable payload** for each shared post at a predictable
+   URL — plain JSON: typed scalars + bucketed time series (avg/min/max/count
+   per bucket). This out-of-band channel is what actually survives real-world
+   federation.
+3. **A discovery document** (`/.well-known/quantpub`) so a peer can tell your
+   host speaks the spec with one cacheable request.
+
+That's a **weekend project on top of an existing home-built tracker** — the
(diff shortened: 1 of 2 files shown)


fiddur (commented):
✅ Approved

Docs-only change: a rewritten §7 step 2 and §9 token-conveyance paragraph in `docs/fep/quantpub.md`, plus a new r/QuantifiedSelf outreach draft.

I checked the two normative rewrites against the running code and they hold up:

- **§7 step 2** — the new "same-host `quant:structuredUrl` names the payload URL directly, discovery only needed for the id-convention path" split is consistent with §7's two bullets, with §4's `api_base`, and with the MUST-only-honour-same-host rule. Aurboda's consumer (`timeline-enrich.ts`) uses the id convention + discovery, which the revised text keeps as the normative baseline, and honouring `quant:structuredUrl` stays a MAY — so the implementation remains conformant.
- **§9 token conveyance** — `quantExerciseExtension` does put `?token=` on `quant:structuredUrl` for non-publicly-visible posts (`quant-extension.ts:90,98`) and `imageAttachments` does the same on attachment URLs (`deliver.ts:92`), so "Aurboda ships both channels" is accurate. Importantly, the token can't leak on a public surface: `setObjectDispatcher` refuses non-publicly-visible posts (`federation.ts:393`), so the tokenised `quant:structuredUrl` only ever exists in the delivered copy — which is what §9's first bullet requires. Relaxing the old blanket "MUST embed in image attachment URLs" into "at least one of two channels, MUST use (a) when the post has attachments" is the right fix for attachment-less observations, and the text is honest about (b)'s dependence on the consumer preserving unknown properties.

No blocking issues — approving.

Please do **not** push fixes for these here; another round costs more than it's worth. Batch them with the next FEP touch (ideally before upstream submission, since #905's next step is submitting this).

1. **Stale `aurboda:` claim in Motivation (`docs/fep/quantpub.md:75`, unchanged line).** It still says Aurboda federates "with a vendor-prefixed (`aurboda:`) variant of everything specified here", which now contradicts the Implementations section ("ships this document's vocabulary on the wire: … `quant:activityType`, `quant:metrics`, `quant:series`, `quant:structuredUrl`"). #896/#1034 landed the `quant:` prefix and `grep "'aurboda:" apps/backend/src` returns nothing — the only `aurboda`-prefixed surface left is `/.well-known/aurboda`, which Implementations already calls out. For a "final pre-submission pass" this self-contradiction is the one I'd most want fixed. (Same stale wording in the `timeline-enrich.ts` header comment, "Fedify's typed vocab drops the `aurboda:` extension".)
2. **`docs/fep/README.md` doesn't mention the new file.** It indexes the directory and currently lists only `quantpub.md`. Either add a bullet for the outreach draft or leave it out deliberately — but if left out, the README's "This directory holds FEP drafts" framing is worth a word so the extra file doesn't look accidental.
3. **Two small overstatements in the reddit draft.** "two-and-a-half **static** routes" — the spec says "two-and-a-half routes"; the series endpoint carries §6's four-condition authorization check and is decidedly not static. And "`no-store` everywhere" — §8.4 deliberately exempts the discovery document (cacheable, `max-age=3600`), which is also part of the point.
4. **"my workouts and sleep observations are ActivityPub posts" (reddit draft).** Today Aurboda federates `activity` and `article` post kinds (`feedPostKindSchema`); `quant:Observation` is specified but not implemented, and there's no sleep-observation share. A QS audience will poke at that. Consider "workouts (and, in the spec, any observation — sleep, HRV, …)" so the running-code claim and the spec claim stay separable.
5. **Consumer-side fallback isn't implemented yet.** §9 now says consumers SHOULD fall back to a same-host `quant:structuredUrl` when no attachment carries a token; `capabilityTokenFrom` only scans image URLs. That's fine today (it's a SHOULD, and Fedify drops the property anyway), but it becomes a real gap the moment attachment-less followers-only shares ship — worth an issue rather than a code change now.

The pre-post checklist with explicit placeholders is a nice touch — it makes the "not yet posted" state unambiguous.

Issue #1037, opened 2026-08-20
QuantPub extension polish (deferred review nits from #1034)


fiddur:
Non-blocking follow-ups from the #1034 review (reviewer: fold into a later PR):

- [ ] `services/activitypub/timeline-enrich.ts:5` docstring still says Fedify drops the `aurboda:` extension — stale on both counts (it's `quant:` now and it IS carried in-band).
- [ ] `/ns/quantpub` route types its response as `Record<string, unknown>`; add a local `QuantpubContextDocument` interface (matching the `AssetLinksStatement` pattern), and update the `well-known-router.ts` header comment which still describes the file as only `/.well-known/*`.
- [ ] `Object.freeze(quantContextEntry)` (or return a fresh literal from `withQuantContext`) — it's inserted by reference into every serialized `@context`.
- [ ] Integration assertion that `/users/:username/outbox` items carry `["Note", "quant:Exercise"]` — the one path where the wrapped object is not the outermost document; guards against a Fedify upgrade changing `OrderedCollectionPage.isCompactable()`.
- [ ] Make `withQuantContext` idempotent like `withQuantType` (a second splice would duplicate the context entry; unreachable today but cheap to guard).

Batch with #1023 / #1031 polish.

fiddur:
From the #1038 review (item 5): §9 now says consumers SHOULD fall back to a same-host `quant:structuredUrl` when no delivered attachment carries the capability token — but `capabilityTokenFrom` in the enrichment path only scans image attachment URLs. Fine today (SOULD-level, and Fedify drops the property before app code sees it on typed ingest), but becomes a real gap the moment attachment-less followers-only shares ship. Add the structuredUrl token fallback to the consumer when doing this batch.

fiddur:
From the #1039 review: the Level 3 enrichment path still discovers via `/.well-known/aurboda` (`realEnrichDeps` → `discoverInstance`, challenge-federation.ts:72), not §4's `/.well-known/quantpub` (already served since #1034). Before FEP submission, either point the enrichment discovery at `/.well-known/quantpub` (challenge federation can stay on the vendor doc) or soften the FEP Implementations parenthetical that says the vendor doc only backs challenge federation. Fits naturally with the structuredUrl-token-fallback item above.

fiddur:
FEP-doc nits deferred from PR #1045 round-2 review (reviewer: batch, don't re-roll) — fold into the next QuantPub revision (rev 8):

[...]

PR #1039, opened 2026-08-20
docs(fep): QuantPub rev 6 — pre-submission consistency fixes (#905)

fiddur:
Part of #905 (does not close it). Batches the non-blocking items from the #1038 review, per the reviewer's note to fold them in before upstream submission:

1. **Motivation ↔ Implementations self-contradiction fixed**: the Motivation section no longer claims Aurboda ships a "vendor-prefixed (`aurboda:`) variant" — the `quant:` vocabulary is on the wire since #1034. Same stale wording fixed in the `timeline-enrich.ts` header (also the first checkbox of #1037).
2. **`docs/fep/README.md`** now indexes `quantpub-reddit-draft.md` and states it is not itself a FEP.
3. **Reddit-draft overstatements**: "static routes" → "routes" (the series endpoint carries the §6 authorization check); "`no-store` everywhere" → `no-store` on data responses with the cacheable discovery doc called out; workouts-are-running-code vs any-observation-is-spec claims separated.
4. Item 5 (consumer-side `quant:structuredUrl` token fallback) recorded on #1037 as a code follow-up rather than changed here.

Backend `pnpm check` green (comment-only code change).

### DIFF
--- a/apps/backend/src/services/activitypub/timeline-enrich.ts
+++ b/apps/backend/src/services/activitypub/timeline-enrich.ts
@@ -1,11 +1,13 @@
 /**
  * Enrich a received timeline post with its native Aurboda structured data.
  *
- * A `Note` delivered over ActivityPub carries only the Mastodon-compatible HTML
- * (Fedify's typed vocab drops the `aurboda:` extension). So when a followed
- * *Aurboda* instance posts, we fetch the richer structured payload it serves at
- * `GET /public/:user/feed/:postId` and store it on the timeline entry, letting
- * the web render a native chart + typed stats instead of the text.
+ * A delivered `Note` carries the QuantPub scalar summary in-band (#896), but
+ * not the high-resolution series data. So when a followed *Aurboda* instance
+ * posts, we fetch the richer structured payload it serves at
+ * `GET /public/:user/feed/:postId` (the FEP §7 id-convention path — reliable
+ * even when a typed consumer drops in-band extension properties) and store it
+ * on the timeline entry, letting the web render a native chart + typed stats
+ * instead of the text.
  *
  * Strictly best-effort and defensive:
  * - Only Notes whose id matches Aurboda's own object-dispatcher path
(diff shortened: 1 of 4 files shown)


fiddur (commented):
✅ Approved

Docs-only change plus one doc-comment rewrite; no behavior changes. I verified each claim the diff makes against the code:

- `timeline-enrich.ts`: the new header is accurate — `quant-extension.ts` does put the scalar summaries in-band (`quant:metrics`) while `quant:series` carries only links, so "scalar summary in-band, not the high-resolution series" is right, and the fetch path (`api_base` + `/public/{user}/feed/{postId}`, matched by `FEED_OBJECT_PATH`) is indeed the §7 id-convention shape. The old text ("Fedify's typed vocab drops the `aurboda:` extension") was genuinely stale after #896, so this is a real fix rather than comment churn.
- `quantpub-reddit-draft.md`: `no-store` on every data response / discovery-only caching matches §8.4 and §4; dropping "static" from "two-and-a-half static routes" matches the weekend section (only the discovery doc is static); and narrowing "my workouts and sleep observations are ActivityPub posts" to workouts matches what Aurboda actually federates today.
- `quantpub.md`: the `quant:` wire-vocabulary claim matches the Implementations section and `quant-extension.ts`.
- `docs/fep/README.md`: the new entry correctly labels the draft as not-a-FEP.

Nothing blocking. Approving.

**Non-blocking — fold into a later PR, don't re-roll this one**

- `docs/fep/quantpub-reddit-draft.md:17`: the rewritten sentence leaves one ~110-col line in a file otherwise wrapped at ~80. Also, the new parenthetical "(and the spec covers any observation — sleep, HRV, steps, mood, …)" now largely duplicates the later privacy-list bullet "It's not just exercise: a generic `Observation` shape covers sleep, HRV, steps, mood, glucose"; one of the two could go.
- `docs/fep/quantpub.md` Implementations (outside this diff): the parenthetical says the remaining vendor surface is "the `/.well-known/aurboda` document backs Aurboda-specific challenge federation", but the Level 3 enrichment path also uses it — `realEnrichDeps` wires `discover: discoverInstance`, which fetches `/.well-known/aurboda` (`challenge-federation.ts:72`), not §4's `/.well-known/quantpub`. So the shipped ingest resolves `api_base` from the vendor document rather than the spec one. Since the PR's theme is pre-submission accuracy, that's worth reconciling before FEP submission — either point `discoverInstance`/enrichment at `/.well-known/quantpub` (it is already served, `well-known-router.ts:73`) or soften the Implementations wording.

PR #1046, opened 2026-08-21
docs(fep): QuantPub rev 8 — camelCase payload; FitPub discussion draft (#905)

fiddur:
**Discussion base for [fitpub/fitpub#455](https://codeberg.org/fitpub/fitpub/issues/455) — deliberately kept as Draft.** Comment here or in #455.

The substantive FitPub alignment already merged in #1045 (rev 7): dual-typing relaxed to RECOMMENDED, the coexistence/incremental-adoption path (now §10), signed `GET`s as an optional followers-only grant, the `route` field documented, expanded FitPub prior-art entry.

## This draft (rev 8)

- **Payload casing → lowerCamelCase everywhere** (`startTime`, `apiBase`, `heartRateAvg`, …), matching AS2 and FitPub style. Aurboda federation has essentially no external implementations yet, so there is no compat burden on our side; Aurboda's implementation catches up once naming settles here.
- Deferred polish from #1037: route resolution floor (RECOMMENDED ≤500 points, mirroring §6's bucket floor), §8.5 rewording, §7.1 promoted to top-level §10.

## QuantPub vs FitPub format (read from FitPub source, 2026-08-21)

Same architecture: plain `Note` + AS2 `startTime`/`endTime` + same-host pointer to an out-of-band detail payload (`fitpub:{detailUri}` ≈ `quant:structuredUrl`), SSRF-guarded best-effort enrichment.

| Aspect | FitPub | QuantPub |
| --- | --- | --- |
| Detail payload | Internal API DTO (`/api/activities/{id}`) | Vendor-neutral contract: `{key, value, unit}` metrics, `kind` discriminator |
| Detection | `fitpub` property presence + NodeInfo | Object-id convention + `/.well-known/quantpub` |
| Time series | None (scalars + track geometry) | Bucketed series, per-metric per-post opt-in |
| Followers-only auth | Signed `GET` | Capability tokens (baseline) + signed `GET` (optional grant, §9) |
| Non-public response | `403` | `404` (unshared ≡ nonexistent) |
| Typing | Single `"Note"`; inbound casts `type` to String | Dual-typed RECOMMENDED, single-typed allowed |

## Backwards compatibility for existing FitPub installations

Adoption is fully additive (§10) — no forced simultaneous upgrades:

1. `/.well-known/quantpub` + the §5/§6 endpoints are **new routes**; `/api/activities/{id}` keeps serving unchanged.
2. `quant:*` + `quant:structuredUrl` ride alongside `fitpub.detailUri` on the same `Note`; old peers ignore unknown properties and see the object they saw before.
3. Dual-typing is optional: start single-typed (old FitPub inbound casts `type` to `String`, so an array would break un-upgraded peers), add the second type once the install base tolerates arrays.
4. No object-id layout change needed — `quant:structuredUrl` replaces the id convention.

Independent of QuantPub: array-valued `type` is well-formed AS2 that FitPub's inbound handler should tolerate — Aurboda's dual-typed posts hit that cast today.

### DIFF
--- a/docs/fep/README.md
+++ b/docs/fep/README.md
@@ -9,7 +9,8 @@ federation code, before submission to the
   federated feed: the `quant:Exercise` / `quant:Observation` vocabulary, the
   out-of-band structured-payload pattern (well-known discovery, structured post
   endpoint, data-driven public series endpoint, capability tokens), and the
-  privacy model as normative requirements. Tracked in
+  privacy model — scalar/series/geography sharing each a separate explicit
+  opt-in — as normative requirements. Tracked in
   [issue #905](https://github.com/fiddur/aurboda/issues/905); the working name
   (QuantPub vs. Personal Metrics Vocabulary / MetricPub) is open for community
   input. Aurboda's own implementation of the pattern is documented in
--- a/docs/fep/quantpub.md
+++ b/docs/fep/quantpub.md
@@ -96,12 +96,15 @@ interpreted as described in [RFC-2119].
   detecting QuantPub-capable origins (§7), and enriching received posts with
   their structured payloads.
 
-Naming follows the two conventions of the layers involved. **JSON-LD extension
-terms** (the `quant:` properties on AS2 objects) use lowerCamelCase, matching
-ActivityStreams' own vocabulary, and this proposal **reuses AS2 properties
-where they exist** (`startTime`, `endTime`) instead of minting parallel terms.
-**HTTP payload fields and metric keys** (§4–§6, §2.1) use snake_case — they are
-plain JSON contracts, not JSON-LD. All timestamps are ISO 8601 with timezone.
+Naming is lowerCamelCase throughout. **JSON-LD extension terms** (the
+`quant:` properties on AS2 objects) match ActivityStreams' own vocabulary,
+and this proposal **reuses AS2 properties where they exist** (`startTime`,
+`endTime`) instead of minting parallel terms. **HTTP payload fields and
+metric keys** (§4–§6, §2.1) use the same lowerCamelCase — the payload layer
+is a plain JSON contract, not JSON-LD, but one casing across both layers
+matches AS2 and FitPub's established formats and means a post's `startTime`
+and its payload's `startTime` are literally the same name. All timestamps are
+ISO 8601 with timezone.
 The JSON-LD namespace for the `quant:` prefix is `https://w3id.org/quantpub#`
 (final IRI to be settled with the FEP number).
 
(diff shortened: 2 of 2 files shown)


fiddur:
**Suggestion: challenges as a QuantPub sub-protocol (or an optional §11)**

Aurboda's federated challenges ([docs/features/challenges.md → Federation protocol](https://github.com/fiddur/aurboda/blob/develop/docs/features/challenges.md#federation-protocol)) are the one part of our federation this draft doesn't describe yet, and they map almost one-to-one onto what is already here:

- **The spec is a `quant:Observation` template.** A challenge is "sum (or count) of `steps` per member over `[start, end)` in zone Z, bucketed by D" — the §3 vocabulary plus a window. A `quant:Challenge` object would carry `metric`, `aggregation`, `unit`, `startTime`/`endTime`, `timeZone`, `bucket`, `name` and a public `url`.
- **A member's data endpoint is the §6 series endpoint** behind a §9 capability token: the host reads one member's bucketed `sum` over the challenge window and nothing else, and the 404-outside-the-window rule is exactly §6's boundary. Aurboda's `GET /challenge-data/{user}/{token}` is a narrower special case; folding it into §6 would remove an endpoint.
- **Join is a register-back.** Fetch the spec, mint a local token, `POST` `{ identity, displayName, seriesUrl, joinToken }` to the host; the host verifies it can read the series and then pulls each member on a TTL. This is the only genuinely new HTTP exchange.
- **Standings and results are ordinary posts.** The member list could be a public collection on the challenge object (FEP-400e gives it a standard shape); the completion post is a `Note` tagging the winners with `Mention`s, optionally with a `quant:ChallengeResult` attachment carrying the final ranked totals.
- **Discovery.** Aurboda is adding "open challenges hosted by people you follow": walk accepted followees and, for each Aurboda peer, list its public challenges. A `challenges` collection on the actor (or a `challengesUrl` in `/.well-known/quantpub`) would make that vendor-neutral instead of leaning on our `/public/{username}/dashboards` listing.

Two ways to package it:

1. **§11 "Challenges (optional, Level 3)"** in this FEP — small, and everything it needs (§3, §6, §9, discovery) is already in the document.
2. **A separate FEP, "QuantPub Challenges"**, depending on this one — keeps the core lean and lets the competition semantics (aggregation, tie rules, late-data grace, withdrawal, host-trusts-member model) evolve on their own schedule.

I lean towards (2): the core is about *sharing* data, challenges are about *comparing* it under agreed rules, and the trust model differs (a host trusts each member's instance to report honestly, Strava-style). A one-paragraph forward pointer in §10 ("a challenges sub-protocol builds on §6 series + §9 tokens") would still let FitPub readers see where competitions would sit.

For comparison, the running implementation: `apps/backend/src/services/challenge-federation.ts` (join), `services/challenge-standings.ts` (host aggregation), `routes/challenge-data-router.ts` (member endpoint).

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

Issue #1023, opened 2026-08-20
Keyset pagination polish: cursor µs precision, invalid-Date guard, feed index, test name


fiddur:
Deferred non-blocking findings from the #1022 review (reviewer: fold into a later PR). Items 1–2 apply to BOTH paginated surfaces (owner feed + home timeline), so one PR can cover both.

- [ ] **Cursor loses sub-millisecond precision → a page boundary can silently skip a post.** `created_at`/`published_at` are µs-resolution TIMESTAMPTZ, but `pg` truncates to ms in the JS `Date` and the cursor stores `ts.getTime()`; the next page's `< cursor_ms` also excludes rows in `(cursor_ms, actual_µs)` — the exact same-millisecond collision the id tiebreaker orders but the ms cursor can't address. Carry full precision through the cursor (select the timestamp as epoch-µs/ISO text for cursor purposes). Also makes the new integration tests theoretically flaky (same-ms inserts).
- [ ] **`decodeKeysetCursor` can hand an Invalid Date to SQL**: `Number.isSafeInteger` admits values beyond the 8.64e15 Date range → `Invalid Date` → node-postgres emits a non-parsable timestamp → 500 (authenticated + self-inflicted, but contradicts the helper's own claim). Add `Number.isFinite(new Date(ms).getTime())` + a test case.
- [ ] **`idx_feed_posts_created` is `(created_at DESC)` only** — the row-wise `(created_at, id) < (…)` predicate needs a `(created_at DESC, id DESC)` multicolumn B-tree to seek (the timeline already has one). Add it.
- [ ] **Test name nit**: `'an exactly-full page ends with a cursor whose next page is empty and final'` in `services/feed.integration.test.ts` actually asserts `next_cursor` is null on the exactly-full first page — rename to match.

fiddur:
Two items deferred from PR #1052 round-2 review (public profile structured rendering):

- [ ] **Double resolution on the profile listing**: `serializeFeedPost` resolves the merged window + scalars, then `resolveStructuredContent` resolves both again for the same record — ~2× queries per cache-cold post. Thread the resolved activity/scalars through (or make the `includeStructured` path itself cacheable).
- [ ] **Paginate `/public/:username/posts`**: now capped at a hardcoded 20 with `offset: 0` — posts 21+ are unreachable from `/u/:username`. Cursor-paginate like `/feed` (#1012) + a Load more on the profile page.

PR #1052, opened 2026-08-21
feat(feed): enriched rendering on the public profile page

fiddur:
Fredrik: "On the public page https://aurboda.net/u/fiddur I think we should do the enriched rendering of metrics and maps as well, since we are rendering it."

The page already renders posts with `FeedPostCard`, which shows the native stat grid + interactive hover chart + synced route map whenever `post.structured` is present — but `GET /public/:username/posts` deliberately skipped `includeStructured` ("skip the weight"), so visitors got flattened text and static PNGs. This flips that opt-in on.

- **One-line behavior change**: `serializeFeedPost(..., { includeStructured: true, settings })` on the public posts listing; no web changes needed — the card and its PNG-suppression logic (#1009/#1013) do the rest.
- **Privacy-neutral**: the listing serves only public/unlisted posts, and `structured` carries exactly the author's per-post opt-ins (series via `series_metrics`, route via `include_map`) — the identical payload anyone could already fetch per post from `/public/:username/feed/:postId`.
- **Weight note**: the listing is a single latest page (limit 50, unauthenticated, `no-store`), so this adds per-post series/route resolution server-side. The owner's authed `/feed` does the same at 20/page. If profile traffic ever makes this hot, the per-post LRU used by the structured endpoint (or pagination, #1023's sibling) is the follow-up — not needed for a single-user instance today.
- New integration test asserts the listing's `structured` matches the per-post endpoint shape.

### DIFF
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -45,23 +45,19 @@ const PROFILE_FEED_LIMIT = 20
 
 export const createFeedPublicRouter = (): TypedRouter => {
   const router = typedRouter()
-  // Caches the resolved structured payload OBJECT for the unauthenticated
-  // `/feed/:postId` endpoint (which resolves up to 100 blocks per request) AND
-  // the `/posts` profile listing (up to PROFILE_FEED_LIMIT structured resolves
-  // per request) — one shared LRU with one key shape, so either endpoint warms
-  // the other. Same bounded LRU + in-flight de-dup the block images use. The route authorizes
-  // the post BEFORE consulting this cache (like the sibling image routes), so the
-  // capability token is NOT part of the key: a `followers`-only payload never
-  // reaches the cache via a public request, and an anonymous `?token=…` walk can't
-  // inflate the key space. Keyed on `updated_at` AND a coarse hourly bucket — the
-  // SAME two-part key the block-image cache uses for the same live-resolved article
-  // data: `updated_at` busts on an edit, and the hourly term bounds staleness from
-  // *backfilled measurements inside an already-locked window* (which don't touch
-  // `updated_at`) to ≤1h, so this endpoint can't freeze into a lifetime snapshot
-  // while its own PNG stays live (#934 locks the window, not the data). A smaller
-  // cap than the image LRU because a payload can be large (per-block sample cap is
-  // #972). Producing `null` (no content) isn't cached.
-  const structuredCache = createRenderCache<FeedStructuredPost>(50)
+  // Caches the resolved structured payload OBJECT, shared (one key shape) by the
+  // unauthenticated `/feed/:postId` endpoint and the `/posts` profile listing, so
+  // either warms the other. Both routes authorize the post BEFORE consulting it,
+  // so the capability token is NOT part of the key: a `followers`-only payload
+  // never reaches the cache via a public request, and an anonymous `?token=…`
+  // walk can't inflate the key space. Keyed on `updated_at` AND a coarse hourly
+  // bucket (the block-image cache's two-part key): `updated_at` busts on any
+  // edit, and the hourly term bounds staleness from backfilled measurements
+  // inside an already-locked window (which don't touch `updated_at`) to ≤1h
+  // (#934 locks the window, not the data). Cap sized for the listing writing up
+  // to PROFILE_FEED_LIMIT entries per profile without a handful of concurrently
+  // browsed profiles thrashing it. Producing `null` (no content) isn't cached.
+  const structuredCache = createRenderCache<FeedStructuredPost>(200)
 
   router.get<{ username: string }, PublicSeriesResponse, unknown, PublicSeriesQuery>(
     '/public/:username/series',
@@ -96,17 +92,14 @@ export const createFeedPublicRouter = (): TypedRouter => {
   )
 
   // A user's public feed for their profile page: the most recent `public`/
-  // `unlisted` posts, newest-first (same set as the ActivityPub outbox — never
-  // `followers`-only), serialized like the authenticated `/feed` with the
-  // structured payload attached per post. Privacy-neutral: every post here is
-  // public/unlisted, and `structured` carries only what the author opted into —
-  // the same payload anyone could fetch per post from
-  // `/public/:username/feed/:postId`. Structured resolution is expensive
-  // (bucketed series queries + a full GPS-track load per opted-in post), so it
-  // goes through the SAME LRU — same key shape — as that sibling endpoint,
-  // bounding what an anonymous request can repeatedly cost; the visibility
-  // filter still runs per request, so an un-shared post disappears immediately
-  // despite the cache. Bounded to the latest page.
+  // `unlisted` posts, newest-first (the ActivityPub outbox's set), serialized
+  // like the authenticated `/feed` with the structured payload attached per
+  // post — `structured` carries only the author's per-post opt-ins, the same
+  // payload `/public/:username/feed/:postId` serves. Structured resolution is
+  // expensive (bucketed series queries + a GPS-track load per opted-in post),
+  // so it goes through the shared LRU above — same key shape as the sibling
+  // endpoint — while the per-request visibility filter keeps un-sharing
+  // immediate despite the cache.
   // `no-store` like the sibling `/series` and `/feed/:postId` endpoints: sharing
   // is revocable, so flipping a post to `followers` or deleting it must drop it
   // from the profile immediately, never linger in a shared cache. Mounted before
--- a/apps/backend/src/routes/feed-public-router.ts
+++ b/apps/backend/src/routes/feed-public-router.ts
@@ -90,8 +90,13 @@ export const createFeedPublicRouter = (): TypedRouter => {
 
   // A user's public feed for their profile page: the most recent `public`/
   // `unlisted` posts, newest-first (same set as the ActivityPub outbox — never
-  // `followers`-only). Serialized exactly like the authenticated `/feed`, so the
-  // web renders them with the same post card. Bounded to the latest page.
+  // `followers`-only). Serialized exactly like the authenticated `/feed`,
+  // INCLUDING the structured payload (typed metrics + inline series + route),
+  // so the profile page renders the same native stat grid / hover chart /
+  // synced map as the owner's feed and a subscribing peer. Privacy-neutral:
+  // every post here is public/unlisted, and `structured` carries only what the
+  // author opted into — the same payload anyone could fetch per post from
+  // `/public/:username/feed/:postId`. Bounded to the latest page.
   // `no-store` like the sibling `/series` and `/feed/:postId` endpoints: sharing
   // is revocable, so flipping a post to `followers` or deleting it must drop it
   // from the profile immediately, never linger in a shared cache. Mounted before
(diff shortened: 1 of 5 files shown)


apps/backend/src/routes/feed-public-router.ts
```diff
@@ -108,7 +113,7 @@ export const createFeedPublicRouter = (): TypedRouter => {
       const records = await listPublicFeedPostsPage(username, PROFILE_FEED_LIMIT, 0)
       const settings = await getSettings(username).catch(() => null)
       const posts = await Promise.all(
-        records.map((record) => serializeFeedPost(username, record, { settings })),
+        records.map((record) => serializeFeedPost(username, record, { includeStructured: true, settings })),
```
fiddur: Worth bounding this before it ships: unlike the authed `/feed` (default `limit=20`, keyset-paginated) this endpoint serves a fixed `PROFILE_FEED_LIMIT = 50` in one shot, is unauthenticated, has no rate limiting, and is explicitly `no-store` with no server-side cache. With `includeStructured: true`, each of those 50 posts now additionally runs one 5s-bucketed `queryMetricsBucketed` per entry in `series_metrics` (schema cap: 64) plus, when `include_map` is set, an unbounded `getLocations` over the whole merged window — the route is downsampled to 500 points only *after* every GPS row is loaded into memory. A single anonymous GET can therefore fan out to hundreds of bucketed queries and 50 full GPS tracks, and the JSON body grows to multiple MB (a 1h activity at 5s buckets is ~720 samples per metric per post). The sibling `/public/:username/feed/:postId` added `structuredCache` precisely because one structured resolve is expensive; this path resolves up to 50 of them per request with nothing in front of it. The frontend side of the same problem: `RouteMap` builds a Leaflet instance eagerly on mount, so the profile page can now spin up 50 maps plus 50 charts at once.

Simplest fix that keeps the feature: drop `PROFILE_FEED_LIMIT` to the authed feed's default page size (20) — or, if you want the full 50, put a `createRenderCache` keyed on `${postId}:${updated_at.getTime()}` around the per-post structured resolve (safe with `no-store`, since the visibility filter still re-runs per request, so an un-shared post disappears immediately).

fiddur (commented):
🛑 Changes required

The behaviour change itself is right and the privacy argument holds: `listPublicFeedPostsPage` only returns `public`/`unlisted` rows, and `resolveActivityPresentation`'s structured branch is byte-identical to `resolveStructuredContent`'s activity branch (same `resolveActivityScalars`, same `resolveStructuredSeries`, same `include_map`-gated `resolveStructuredRoute`), so this exposes nothing a visitor couldn't already fetch from `/public/:username/feed/:postId`. `FeedPostCard` and `TimelineStructured` already render `structured` with no authenticated fetches, so the profile page picks it up with no frontend change.

The one thing I'd like addressed before merge is cost, not correctness — see the inline comment. This unauthenticated, `no-store`, uncached endpoint now resolves the full structured payload for up to `PROFILE_FEED_LIMIT = 50` posts per request, and the doc comment this PR edits previously singled out "the public profile listing" as the place that deliberately skipped that weight. The sibling per-post endpoint got a 50-entry LRU for exactly this reason; this one has no equivalent bound.

- `apps/web/src/pages/Feed/FeedPostCard.tsx` (the block comment above the render branches) still says "Without it (public profile), the stat-grid `ActivityPostBody`". After this change the public profile does carry `structured`, so that branch is now only the legacy/pre-#1008 fallback — worth rewording next time that file is touched.
- The new router comment is a bit heavy: the "same native stat grid / hover chart / synced map" clause restates what `includeStructured: true` plus the existing card code already say. The `Privacy-neutral: …` sentence is the genuinely non-obvious part and is worth keeping.
- The new test seeds `series_metrics: []` / `include_map: false`, so it only pins the empty-series path. The interesting assertions — a populated `series` coming through, and `route` appearing only under `include_map` — are untested on this endpoint (they are covered for `/feed/:postId`). Nice-to-have, not needed here.

apps/backend/src/routes/feed-public-router.ts
```diff
@@ -90,8 +97,16 @@ export const createFeedPublicRouter = (): TypedRouter => {
 
   // A user's public feed for their profile page: the most recent `public`/
   // `unlisted` posts, newest-first (same set as the ActivityPub outbox — never
-  // `followers`-only). Serialized exactly like the authenticated `/feed`, so the
-  // web renders them with the same post card. Bounded to the latest page.
+  // `followers`-only), serialized like the authenticated `/feed` with the
```
fiddur: This inverts a behavior the docs state explicitly, and they weren't updated. `docs/features/feed.md:524` currently reads:

> The public profile renders the lighter native **stat grid** (message + activity date + typed scalars) without the series weight — and the MCP `list_feed` tool likewise omits `structured` (an intentional payload-weight divergence from `GET /feed`, not a capability gap…)

After this change the public profile *does* carry the full structured payload (series + route included), so that sentence now says the opposite of what ships, and it's the sentence a reader goes to in order to understand the divergence between `/feed`, the profile listing, and `list_feed`. CLAUDE.md requires `docs/` to track behavior changes ("When changing APIs, sync behavior … update the corresponding docs"), so please rewrite it to say the profile listing now attaches `structured` through the shared per-post LRU (and that `list_feed` remains the only surface that omits it).

While you're there: `apps/web/src/pages/Feed/FeedPostCard.tsx:116` has the same stale claim in a code comment — "Without it (public profile), the stat-grid `ActivityPostBody`" — which is now only the fallback for a post whose structured resolve returned nothing. One line each, and both are in the same round.

fiddur (commented):
🛑 Changes required

The backend change itself is sound. I checked the parts that could have gone wrong and they hold up:

- **Shared cache key is safe.** `listPublicFeedPostsPage` and `getFeedPostById` both select `FEED_POST_COLUMNS` through the same `mapFeedPost`, so the two endpoints feed `resolveStructuredContent` structurally identical records — sharing one key shape can't serve a payload resolved from a differently-shaped record. `/posts` only ever reads keys for rows the `visibility IN ('public','unlisted')` filter returned, so a `followers`-only payload cached via an authorized `?token=` request is unreachable from the listing, and `updateFeedPost` bumps `updated_at` on every visibility/`included_metrics`/`series_metrics`/`include_map` change, so an un-share busts the key immediately.
- **`kind === 'activity'` gate matches** the `includeStructured` path in `resolveActivityPresentation` (`services/feed.ts:133`), so the profile listing and the owner's `/feed` agree on which posts carry `structured`.
- **`PROFILE_FEED_LIMIT = 20` really does match** `feedPostsQuerySchema`'s `.default(20)`, so that comment is accurate.
- **The new test asserts the real payload**, not a shape it constructed, and each `startApp()` builds a fresh router (fresh LRU), so there's no cross-test cache bleed.
- **`TimelineStructured` needs no credentials** — it renders purely from the inline payload — so the unauthenticated profile page can actually use what's now attached.

One thing must change before merge: `docs/features/feed.md:524` now states the opposite of what the code does (see the inline comment).

Please don't push fixes for these here; a push starts another full review round for no real benefit. Batch them with other work.

- **Duplicated resolution per post.** `serializeFeedPost` resolves the merged activity window and `resolveActivityScalars`, then `resolveStructuredContent` resolves *both again* for the same record — roughly 2× the queries per post on a cache-cold anonymous request. Threading the already-resolved activity/scalars into the structured resolve (or making the `includeStructured` path itself cacheable) would halve that.
- **No pagination behind the new cap.** Dropping 50 → 20 with a hardcoded `offset: 0` and no "Load more" on `/u/:username` makes posts 21+ unreachable from the profile page. Cursor-paginating `/public/:username/posts` the way `/feed` is (#1012) would restore the reach without restoring the weight.
- **LRU cap vs. the new writer.** The cache is still `createRenderCache(50)` while `/posts` now writes up to 20 entries per profile request, so three concurrently-browsed profiles thrash it. Worth raising the cap now that a listing shares it.
- **Comment volume.** The `/posts` header comment is now ~16 lines and the cache comment ~16, and a fair amount restates what the code already shows — "Privacy-neutral: every post here is public/unlisted" is visible from the `listPublicFeedPostsPage` call, and "Bounded to the latest page" duplicates the `PROFILE_FEED_LIMIT` docblock two screens up. The genuinely non-obvious part (why the key shape is shared, why the token stays out of it) is worth keeping; the rest could go. Same for the two comment lines above the new `toMatchObject` in the test.
- **Stale payload after an upstream delete.** Deleting the underlying activity or its GPS track doesn't touch `feed_posts.updated_at`, so a cached `structured` can linger up to the hour bucket. Identical to the existing `/feed/:postId` tradeoff, just noting it now applies to the listing too.

fiddur (commented):
✅ Approved

Both blocking items from the previous rounds are addressed and I couldn't find a new one.

What I re-verified in this round's diff:

- **Docs now match the code.** `docs/features/feed.md:521-530` states the shipped behaviour (profile listing attaches the full structured payload through the shared per-post LRU, one fixed page of 20; `list_feed` is the one omitting surface), and the stale `FeedPostCard` comment is corrected. That was the round-2 blocker.
- **Cap raise is safe for correctness.** `createRenderCache(200)` changes only retention, not key semantics: the key stays `structured:<user>:<postId>:<updated_at>:<hourBucket>`, identical in both writers, so `/posts` and `/feed/:postId` still share entries legitimately (both select `FEED_POST_COLUMNS` through `mapFeedPost`, so the records fed to `resolveStructuredContent` are structurally the same). `updated_at` is `NOT NULL`/`Date`, so the `.getTime()` in the listing key can't throw.
- **Privacy boundary still holds.** The listing only reads keys for rows `listPublicFeedPostsPage` returned (`visibility IN ('public','unlisted')`), so a `followers`-only payload cached via an authorized `?token=` request is unreachable from the profile, and `updateFeedPost` bumps `updated_at` on every `visibility`/`included_metrics`/`series_metrics`/`include_map` change, so an un-share or a narrowed opt-in busts the key immediately.
- **`PROFILE_FEED_LIMIT = 20`** matches `feedPostsQuerySchema`'s default, so the comment is accurate and the per-request weight (20 structured resolves, 20 charts/maps client-side) is now the same as the authed `/feed` page it mirrors.

Please don't push fixes for these here; a push starts another full review round for no benefit. Batch them with the follow-up work.

- **The dropped rationale for the old cap still applies.** The 50-cap comment used to say "a smaller cap than the image LRU because a payload can be large (per-block sample cap is #972)". That's still true — an article payload can hold up to 100 resolved blocks and an activity payload has no sample cap (5s buckets over the whole merged window) — so 200 entries quadruples worst-case heap retention while entry *size* is still unbounded. Once #972's sample cap lands, an entry-count bound is fine; before then a rough weight-based bound (or keeping the structured cap below the image cap) would be the safer shape.
- **Still carried over from round 2:** the duplicated per-post resolution (`serializeFeedPost` resolves activity + scalars, then `resolveStructuredContent` resolves both again on a cache miss) and the missing pagination behind the new cap (hardcoded `offset: 0`, no "Load more" on `/u/:username`, so posts 21+ are unreachable from the profile page). Both fit naturally in one follow-up.
- **Doc precision:** "attaches the same full structured payload per post" is true for `activity` posts only — an `article` post on the profile still renders from `post.article` via `ArticleContent`, and the `record.kind === 'activity'` gate (correctly) mirrors `resolveActivityPresentation`. A four-word qualifier would make that exact.
- **Comment volume:** trimmed, but the cache block is still 12 lines and the route header 10. The genuinely non-obvious parts (why the key shape is shared, why the token stays out of it, why the hourly term exists) earn their space; the rest — "serialized like the authenticated `/feed`", "Structured resolution is expensive" — restates the adjacent code.
- **Test coverage:** the new test still seeds `series_metrics: []` / `include_map: false`, so it pins only the empty-series path on this endpoint. A populated `series` and a route appearing only under `include_map` are covered for `/feed/:postId` but not here.

PR #1064, opened 2026-08-24
feat(feed): mentions + stranger involvement, expand replies, reply backfill, bell reconciliation (#1060 r2)

fiddur:
Round 2 on #1060, from Fredrik's field feedback (with screenshots) + his follow-up question "are mentions of me from strangers also dropped?" (they were):

## Why the timeline was still full of replies
Entries ingested **before** #1061 have no `in_reply_to_uri` — the reply link was discarded at ingest — so the filter can't see them. Timeline reads now **lazily backfill** a few unchecked legacy entries per read (AP re-fetch of the object, one attempt, stamped via new `reply_checked_at`; fresh ingests stamp immediately). His timeline converges clean over a few page loads.

## Where the toggle is
`timeline_show_replies` now also sits **on the Feed page** next to the Home timeline (it stays in Settings too). Saving invalidates the timeline query so it applies immediately.

## Involvement: mentions + strangers
The inbox previously dropped every Note from a non-followee — including replies to your own posts and posts mentioning you. Now a non-followee's Note is admitted **iff you're involved**: it replies to one of your own (still existing) posts, or carries a `Mention` of your actor (new `mentions_me` column/field; author snapshot from the signature-verified sender). Involved posts always show (marked "replied to you" / "mentioned you") and always notify — regardless of the reply setting or the per-follow bells. Everything else from strangers is still dropped.

## Expand replies
`GET /feed/timeline/:id/replies` + MCP `get_timeline_replies`: a live, bounded snapshot of the post's remote AS2 `replies` collection — ≤20 replies from ≤15 SSRF-guarded fetches in a 12s budget, authors resolved + memoised, HTML sanitised server-side, nothing stored, `partial` flag when the thread outruns the budget. Web timeline cards get a **Show replies** button rendering it inline.

## Bells can't lie anymore
On launch, `AutoEnablePostNotifications` reconciles the Feed page's per-account 🔔 bells with the device: any bell on + no explicit on/off choice ever made → request `POST_NOTIFICATIONS` and start the poller. An explicit "off" is never overridden; a denial records "off" (no re-prompt loop) and the Account screen's existing blocked-warning takes over.

Tests: 5 new remote-replies unit tests (page walk, URI items, budget/partial, sanitisation, author memoisation), parseReplyInfo/backfill tests, db integration for the mentions filter + backfill helpers + `getTimelineEntryById`, Android involvement-predicate tests; whole-monorepo check green, activitypub suite 156/156, Android compile+unit green locally.

### DIFF
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -49,7 +49,6 @@ import {
   countFeedFollowers,
   countPublicFeedPosts,
   deleteTimelineEntryByUri,
-  type FeedFollowingRecord,
   getFeedFollowerByActor,
   getFeedFollowingByActor,
   getFeedPostById,
@@ -68,6 +67,7 @@ import {
 } from '../../db/index.ts'
 import { resolveFeedActivity } from '../feed.ts'
 import { buildProfileUrl } from '../share-urls.ts'
+import { ownObjectPrefix } from '../timeline.ts'
 import { extractActorPresentation } from './actor-presentation.ts'
 import {
   buildArticleNote,
@@ -82,7 +82,12 @@ import {
 import { toCryptoKeyPair } from './keys.ts'
 import { AS_PUBLIC, isPubliclyVisible } from './object.ts'
 import { capabilityTokenFrom, createAurbodaEnricher } from './timeline-enrich.ts'
-import { extractNoteImages, noteToTimelineInput } from './timeline-ingest.ts'
+import {
+  extractNoteImages,
+  noteMentionsActor,
+  noteToTimelineInput,
+  type TimelineAuthor,
+} from './timeline-ingest.ts'
 
 /** Posts per outbox page (cursor pagination). */
 const OUTBOX_PAGE_SIZE = 20
@@ -164,35 +169,93 @@ const recordInboundFollow = async (
 export const ingestNoteForRecipient = async (
   user: string,
   note: Note,
-  followee: FeedFollowingRecord,
+  author: TimelineAuthor,
   enrich: (objectUri: string, token?: string) => Promise<FeedStructuredPost | null>,
   onNewEntry?: (user: string) => void,
+  /** Web origin — enables Mention detection against the recipient's actor URI (#1060). */
+  origin?: string,
 ): Promise<void> => {
-  const input = noteToTimelineInput(note, followee)
+  const input = noteToTimelineInput(note, author)
   if (input == null) return
   // Capture the Note's image attachments (rendered chart / route map, or a
   // Mastodon photo) so the timeline can show them when there's no native chart.
   const images = await extractNoteImages(note)
+  // A Mention of the recipient keeps the post visible + notifying whatever the
+  // reply setting says (Mastodon-style involvement).
+  const mentionsMe =
+    origin == null ? false : await noteMentionsActor(note, `${origin.replace(/\/+$/, '')}/users/${user}`)
   // Best-effort: fetch the native structured payload if this is an Aurboda post
   // (null otherwise). A followers-only post authorizes the fetch with the same
   // capability token embedded in its delivered image URL. Stored on the entry so
   // the web can render a native chart in place of the image.
   const structured = await enrich(input.object_uri, capabilityTokenFrom(images))
-  const { inserted } = await upsertTimelineEntry(user, { ...input, images, structured })
+  const { inserted } = await upsertTimelineEntry(user, {
+    ...input,
+    images,
+    mentions_me: mentionsMe,
+    structured,
+  })
   if (inserted) onNewEntry?.(user)
 }
 
 /**
- * Ingest a `Create`/`Update` of a `Note` into the recipient's home timeline —
- * but ONLY if the sender is an *accepted* followee (so the timeline can't be
- * injected into by a stranger who delivers to our inbox). Resolves the recipient
- * and the cached followee row, then hands off to `ingestNoteForRecipient`.
+ * Ingest a `Create`/`Update` of a `Note` into the recipient's home timeline.
+ * Two admissible senders (anything else is dropped, so a stranger can't inject
+ * arbitrary posts into our timeline by delivering to the inbox):
+ *
+ * - an *accepted* followee — any of their Notes;
+ * - **any actor whose Note is a reply to one of the recipient's own, still
+ *   existing posts** (#1060) — the Mastodon-style mention interaction, so a
+ *   stranger's "replied to you" shows up (and notifies) like it would there.
+ *   The author snapshot comes from the signature-verified sender actor.
+ *
  * Best-effort: unresolvable objects and non-Notes are ignored, and a missing DB
  * never 500s (which would invite retries).
  */
+/**
+ * The stranger branch of {@link ingestFeedActivity}: admit a non-followee's
+ * Note only when the recipient is INVOLVED — it replies to one of their own,
+ * still existing posts, or Mentions them. The author snapshot comes from the
+ * signature-verified sender actor.
+ */
+const ingestStrangerInvolvement = async (
+  ctx: InboxContext<void>,
+  activity: Create | Update,
+  object: Note,
+  me: string,
+  origin: string,
+  enrich: (objectUri: string, token?: string) => Promise<FeedStructuredPost | null>,
+  onNewEntry?: (user: string) => void,
+): Promise<void> => {
+  if (activity.actorId == null) return
+  const target = object.replyTargetIds[0]?.href
+  const prefix = ownObjectPrefix(origin, me)
+  const postId = target?.startsWith(prefix) ? target.slice(prefix.length) : null
+  const isReplyToOwnPost =
+    postId != null && UUID_RE.test(postId) && (await getFeedPostById(me, postId)) != null
+  if (!isReplyToOwnPost) {
+    const myActor = `${origin.replace(/\/+$/, '')}/users/${me}`
+    if (!(await noteMentionsActor(object, myActor))) return
+  }
+  // Fedify verified the HTTP signature as `activity.actorId`; fetch that actor
+  // for the presentation snapshot (handle / name / avatar).
+  const sender = await activity.getActor(ctx)
+  if (sender?.id == null || sender.id.href !== activity.actorId.href) return
+  const presentation = await extractActorPresentation(sender)
+  await ingestNoteForRecipient(
+    me,
+    object,
+    { ...presentation, actor_uri: activity.actorId.href },
+    enrich,
+    onNewEntry,
+    origin,
+  )
+}
+
 const ingestFeedActivity = async (
   ctx: InboxContext<void>,
   activity: Create | Update,
+  origin: string,
   onNewEntry?: (user: string) => void,
   /** Best-effort fetch of the post's native Aurboda structured data (null if not an Aurboda post). */
   enrich: (objectUri: string, token?: string) => Promise<FeedStructuredPost | null> = async () => null,
@@ -206,10 +269,12 @@ const ingestFeedActivity = async (
   if (me == null || !isValidUsername(me) || activity.actorId == null) return
   try {
     const follow = await getFeedFollowingByActor(me, activity.actorId.href)
-    if (follow == null || !follow.accepted) return
     const object = await activity.getObject({ suppressError: true })
     if (!(object instanceof Note)) return
-    await ingestNoteForRecipient(me, object, follow, enrich, onNewEntry)
+    if (follow != null && follow.accepted) {
+      return await ingestNoteForRecipient(me, object, follow, enrich, onNewEntry, origin)
+    }
+    await ingestStrangerInvolvement(ctx, activity, object, me, origin, enrich, onNewEntry)
   } catch (error) {
     if (isMissingDatabase(error)) return
     throw error
@@ -341,9 +406,7 @@ export const createFeedFederation = (
   }))
 
   federation
-    .setActorDispatcher('/users/{identifier}', (ctx, identifier) =>
-      buildActorPerson(ctx, identifier, origin),
-    )
+    .setActorDispatcher('/users/{identifier}', (ctx, identifier) => buildActorPerson(ctx, identifier, origin))
     .setKeyPairsDispatcher(async (_ctx, identifier) => {
       if (!isValidUsername(identifier)) return []
       try {
@@ -430,10 +493,11 @@ export const createFeedFederation = (
     })
     // Create / Update of a Note from an actor we follow → ingest into our home
     // timeline. Update reuses the same upsert (keyed on the Note's object id), so
-    // an edit replaces the stored copy. Only *accepted* followees are ingested, so
-    // a stranger who somehow delivers here can't inject into the timeline.
-    .on(Create, (ctx, create) => ingestFeedActivity(ctx, create, onNewTimelineEntry, enrich))
-    .on(Update, (ctx, update) => ingestFeedActivity(ctx, update, onNewTimelineEntry, enrich))
+    // an edit replaces the stored copy. Accepted followees are ingested in full;
+    // the only stranger Note admitted is a reply to one of the recipient's own
+    // posts (see ingestFeedActivity).
+    .on(Create, (ctx, create) => ingestFeedActivity(ctx, create, origin, onNewTimelineEntry, enrich))
+    .on(Update, (ctx, update) => ingestFeedActivity(ctx, update, origin, onNewTimelineEntry, enrich))
     .on(Delete, async (ctx, del) => {
       // Delete of a post we received → drop it from the timeline. `del.objectId`
       // is the removed object's id (a Tombstone or bare id); we key on it, but
--- a/apps/backend/src/services/activitypub/federation.ts
+++ b/apps/backend/src/services/activitypub/federation.ts
@@ -237,6 +237,13 @@ const ingestStrangerInvolvement = async (
     const myActor = `${origin.replace(/\/+$/, '')}/users/${me}`
     if (!(await noteMentionsActor(object, myActor))) return
   }
+  // Require EXPLICIT attribution to the signing actor. `noteToTimelineInput`'s
+  // attribution check is conditional (a Note without `attributedTo` passes on
+  // host match alone) — written for accepted followees; for a stranger that
+  // would let anyone on a followee's host claim an existing entry's object_uri
+  // and overwrite it via the upsert. Every real implementation sets
+  // `attributedTo`, so requiring it costs nothing.
+  if (!object.attributionIds.some((uri) => uri.href === activity.actorId?.href)) return
   // Fedify verified the HTTP signature as `activity.actorId`; fetch that actor
   // for the presentation snapshot (handle / name / avatar).
   const sender = await activity.getActor(ctx)
--- a/apps/backend/src/services/activitypub/remote-replies.ts
+++ b/apps/backend/src/services/activitypub/remote-replies.ts
@@ -1,0 +1,185 @@
+/**
+ * Live fetch of a remote post's `replies` collection (#1060) — the "expand
+ * replies" feature on a timeline card. Nothing is stored: the origin owns the
+ * thread, we render a bounded snapshot of it on demand.
+ *
+ * Strictly best-effort and budgeted: at most {@link MAX_REPLIES} replies from at
+ * most {@link MAX_FETCHES} SSRF-guarded requests (collection pages, reply
+ * objects referenced by URI, and author actors — one lookup per unique author,
+ * memoised). Anything malformed is skipped, never thrown; the response's
+ * `partial` flag tells the client the budget ran out before the collection did.
+ *
+ * All reply HTML goes through `sanitizeRemoteHtml` before it leaves this module
+ * — the payload is as untrusted as inbox content.
+ */
+import type { TimelineReply } from '@aurboda/api-spec'
+
+import { safeFetchGet } from '../safe-fetch.ts'
+import { sanitizeRemoteHtml } from './timeline-ingest.ts'
+
+const MAX_REPLIES = 20
+const MAX_FETCHES = 15
+const AP_ACCEPT = 'application/activity+json, application/ld+json; q=0.9'
+
+export interface RemoteRepliesDeps {
+  /** Fetch + JSON-decode an ActivityPub URL (SSRF-guarded). */
+  fetchJson: (url: string) => Promise<unknown>
+}
+
+export const realRemoteRepliesDeps: RemoteRepliesDeps = {
+  fetchJson: async (url) =>
+    (await safeFetchGet(url, { headers: { Accept: AP_ACCEPT } })).data,
+}
+
+type JsonRecord = Record<string, unknown>
+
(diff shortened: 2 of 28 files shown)


apps/backend/src/services/timeline-retro-enrich.ts
```diff
@@ -86,3 +86,64 @@ export const retroEnrichTimelineEntries = async (
...
+  user: string,
+  myActorUri: string,
+  deps: ReplyBackfillDeps,
+  batchSize: number = RETRO_BATCH_SIZE,
+): Promise<void> => {
+  const candidates = await deps.listUnchecked(user, batchSize)
+  for (const entry of candidates) {
+    const doc = await deps.fetchObject(entry.object_uri)
+    const info = parseReplyInfo(doc, myActorUri)
+    await deps.saveReplyInfo(user, entry.id, info.in_reply_to_uri, info.mentions_me)
```
fiddur, line 155: This clobbers correct data. `in_reply_to_uri` already ships in `develop` (#1061), and this PR's migration leaves `reply_checked_at` NULL on those rows too, so they're all backfill candidates. When `fetchObject` returns null — post deleted upstream (404), instance in authorized-fetch mode (401 to our unsigned GET), host down, HTML body — `parseReplyInfo` yields `{ in_reply_to_uri: null }` and this write erases a known-good reply link. The entry then renders as a top-level card again and loses `in_reply_to_mine`, so the "replied to you" marker and the Android involvement notification go with it.

Stamp `reply_checked_at` on a failed fetch (so it isn't retried forever) but don't overwrite the reply fields — e.g. skip the value write when `doc == null`, or have `setTimelineEntryReplyInfo` use `in_reply_to_uri = COALESCE($2, in_reply_to_uri)`. The `backfillReplyLinks` test asserts the current clobbering (`['b', null, false]`), so it needs to change with it.

apps/backend/src/services/activitypub/remote-replies.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (content == null) return null
+  const actorUri = idOf(obj.attributedTo)
+  const author = actorUri == null ? null : await resolveAuthor(deps, budget, actorUri, authors)
+  return {
+    actor_uri: actorUri,
+    content: sanitizeRemoteHtml(content),
+    display_name: author?.display_name ?? null,
+    handle: author?.handle ?? null,
+    published_at: typeof obj.published === 'string' ? obj.published : null,
+    url: typeof obj.url === 'string' ? obj.url : (idOf(obj.id) ?? null),
```
fiddur: `obj.url` goes to the client unvalidated, and `HomeTimeline.tsx` renders it as `<a href={reply.url}>` — so a `javascript:` (or other non-http) URL from the origin executes in the app origin as soon as the reader clicks the reply timestamp, with the bearer token in reach. The module already treats this payload as untrusted for HTML; the `url` needs the same treatment.

With mention-based ingestion this is reachable by anyone: a stranger delivers a Note mentioning the user, the card appears with a "Show replies" button, and the replies collection it walks is served by the attacker's own host. Filtering to `http:`/`https:` here (like `httpsUrl` does for attachments in `timeline-ingest.ts`) covers both the REST route and the MCP tool in one place.

fiddur (commented):
🛑 Changes required

Solid round 2 overall: the stranger-involvement admission is properly narrow (own-post reply verified against `feed_post`, or an exact-href `Mention`, plus the `activity.actorId`/`getActor` identity check and the existing `noteToTimelineInput` host/attribution guards), the reply fetch is genuinely budgeted and SSRF-guarded, `getTimelineEntryById` is per-user-DB scoped, the route validates the UUID before touching the DB, and the Android involvement rule + docs line up with the backend.

Two things should change before merge (both inline):

1. **The reply backfill destroys correct data.** `develop` already stores `in_reply_to_uri` (shipped in #1061), and this PR's migration leaves `reply_checked_at` NULL on *all* existing rows — including those correctly-stamped ones. `backfillReplyLinks` then writes `parseReplyInfo(null, …)` (i.e. `in_reply_to_uri = NULL`) whenever the re-fetch fails, which is common: post deleted upstream (404), instance in authorized-fetch/secure mode (401 for our unsigned GET), host down, or an HTML body returned for the AP `Accept`. Those replies then resurface as top-level cards and lose `in_reply_to_mine`, so "replied to you" markers and the Android involvement notification disappear. The unit test currently asserts the clobbering behaviour, so it needs updating too.

2. **`TimelineReply.url` is an unvalidated remote string rendered into an `href`.** `remote-replies.ts` sanitises the reply HTML but passes `obj.url` straight through, and `HomeTimeline.tsx` renders `<a href={reply.url}>`. A `javascript:` URL from a hostile origin executes in the app origin on click — and thanks to mention-based ingestion, *any* stranger can now get a card (and therefore a "Show replies" button pointing at their own server) into a user's timeline. An http(s)-only check next to the existing `sanitizeRemoteHtml` call fixes it for both the REST and MCP consumers.

- `federation.ts`: the updated `ingestFeedActivity` docblock is now orphaned — it sits directly above the new `ingestStrangerInvolvement` docblock, so two doc comments stack on the stranger helper and `ingestFeedActivity` has none. It also describes only the reply branch, not the Mention-only one.
- `listReplyUncheckedEntries` (`WHERE reply_checked_at IS NULL ORDER BY received_at DESC`) has no supporting index, and there is no index on `received_at` at all — the analogous retro-enrich query got `idx_timeline_entry_unenriched` for exactly this reason (cheap no-work case on every timeline read).
- MCP `get_timeline_replies` returns `jsonResponse({ error, success: false })` where the rest of the file uses `errorResponse(...)`, and re-inlines the `12_000` budget the router keeps as `REPLIES_TIMEOUT_MS`.
- `invalidateQueries({ queryKey: ['feed', 'timeline'] })` prefix-matches `['feed','timeline',<id>,'replies']`, so toggling the setting refetches every expanded thread from its remote origin.
- `partial` is reported `true` for a thread of exactly 20 replies even when the collection was exhausted; conversely an unreachable origin returns `partial: false` + no replies, which the web renders as "No replies found on the origin." rather than a fetch failure.
- `resolveAuthor` does `new URL(actorUri)` on a raw `attributedTo` value; a non-URL string throws and (via the route's `.catch`) collapses the whole thread to empty.
- `noteMentionsActor` and the stranger-admission branch — the security-relevant part of this PR — have no unit tests, and `origin?: string` on `ingestNoteForRecipient` silently disables mention detection if a future caller omits it (all current callers pass it, so it could just be required).
- Involvement now notifies even for an account whose bell the user explicitly turned *off*. Intended per the docs, just worth being deliberate about — it does override an explicit mute.
- The comment above `AutoEnablePostNotifications()` in `MainActivity.kt` restates the function's own KDoc.

apps/backend/src/services/activitypub/federation.ts
```diff
@@ -164,35 +169,93 @@ const recordInboundFollow = async (
...
+  const postId = target?.startsWith(prefix) ? target.slice(prefix.length) : null
+  const isReplyToOwnPost =
+    postId != null && UUID_RE.test(postId) && (await getFeedPostById(me, postId)) != null
+  if (!isReplyToOwnPost) {
+    const myActor = `${origin.replace(/\/+$/, '')}/users/${me}`
+    if (!(await noteMentionsActor(object, myActor))) return
+  }
+  // Fedify verified the HTTP signature as `activity.actorId`; fetch that actor
+  // for the presentation snapshot (handle / name / avatar).
+  const sender = await activity.getActor(ctx)
```
fiddur, line 249: The stranger branch inherits `noteToTimelineInput`'s authority check, but that check is only *conditional* on attribution:

```ts
if (note.id.host !== senderHost) return null
const attributions = note.attributionIds
if (attributions.length > 0 && !attributions.some(...)) return null
```

A Note with **no** `attributedTo` passes as long as `note.id.host` matches the sender's host. So `eve@mastodon.social` (a stranger, never followed) can deliver a `Create` whose inline Note has `id: https://mastodon.social/users/alice/statuses/123` — an existing entry from a followee — no `attributedTo`, and a `Mention` of the recipient. `ingestStrangerInvolvement` admits it, and `upsertTimelineEntry`'s `ON CONFLICT (object_uri) DO UPDATE` replaces Alice's stored content, handle, avatar and `actor_uri` with Eve's. It also sticks: `deleteTimelineEntryByUri` is scoped to the authoring actor, so Alice's later `Delete` no longer matches the row.

That guard was written against *accepted followees*; the stranger branch widens the blast radius from "people you follow" to "anyone with an account on the same host as anyone you follow". Since every real implementation sets `attributedTo`, requiring it here costs nothing — e.g. before ingesting:

```ts
if (!object.attributionIds.some((uri) => uri.href === activity.actorId.href)) return
```

(or a `requireAttribution` flag threaded into `noteToTimelineInput`, which would close the same hole for the followee path).

apps/backend/src/services/activitypub/remote-replies.ts
```diff
@@ -0,0 +1,209 @@
...
+  const content = typeof obj.content === 'string' ? obj.content : null
+  if (content == null) return null
+  const actorUri = idOf(obj.attributedTo)
+  const author = actorUri == null ? null : await resolveAuthor(deps, budget, actorUri, authors)
+  return {
+    actor_uri: actorUri,
+    content: sanitizeRemoteHtml(content),
+    display_name: author?.display_name ?? null,
+    handle: author?.handle ?? null,
+    published_at: typeof obj.published === 'string' ? obj.published : null,
```
fiddur: `obj.published` is passed through as any string, but `timelineReplySchema.published_at` is `iso8601DateTimeSchema.nullable()` — so the response can violate its own OpenAPI contract, and the Kotlin `OffsetDateTime` decoder would reject it too.

The concrete failure is in the web: `HomeTimeline.tsx` does `formatDistanceToNow(new Date(reply.published_at), …)`, and date-fns v4 throws `RangeError: Invalid time value` for an Invalid Date. There is no error boundary in `apps/web`, so a single reply with `"published": "whenever"` from a remote origin — reachable by any stranger who mentions you, since they then serve the `replies` collection for their own post — blanks the whole timeline page on "Show replies".

Same treatment as `httpsOnly` on the line below, e.g.:

```ts
const isoOrNull = (v: unknown): string | null =>
  typeof v === 'string' && !Number.isNaN(Date.parse(v)) ? v : null
```

fiddur (commented):
🛑 Changes required

Round 2 fixes from the last round both landed cleanly: the backfill is now no-clobber (`markTimelineEntryReplyChecked` on a non-answer, plus the `timeline_entry_reply_checked_backstamp` migration for rows that already carry a reply link — and the unit test asserts the new behaviour), `TimelineReply.url` is http(s)-only, and `resolveAuthor` no longer throws on a non-URL `attributedTo`. The reply walk is genuinely budgeted and SSRF-guarded, `getTimelineEntryById` is per-user-DB scoped, the route validates the UUID before touching the DB, and the migration keys are non-table names so they re-run on every `migrateSchema` pass as intended.

Two things should still change before merge (both inline, both one-liners):

1. **A stranger can overwrite a followee's timeline entry.** `noteToTimelineInput`'s authority check only enforces attribution *when the Note declares `attributedTo`*; with it omitted, any sender may claim any `object_uri` on their own host. That guard was written for accepted followees ("a malicious accepted followee could deliver a Note whose id collides with another followee's post"), and the stranger branch widens it from "people you chose to follow" to "anyone sharing a host with someone you follow".
2. **`TimelineReply.published_at` is an unvalidated remote string** declared as `ISO8601DateTime`, and the web feeds it straight to `formatDistanceToNow`, which throws `RangeError: Invalid time value` on an unparseable value. There is no error boundary in `apps/web`, so one malformed `published` takes the timeline down.

- `federation.ts`: the `ingestFeedActivity` docblock is still orphaned above `ingestStrangerInvolvement` (two stacked doc comments; `ingestFeedActivity` has none), and the `.on(Create, …)` comment now claims "the only stranger Note admitted is a reply to one of the recipient's own posts" — the Mention branch is missing from both, so the comments understate what the code admits.
- `origin?: string` on `ingestNoteForRecipient` is now passed by every caller; making it required would remove the silent "mentions never detected" failure mode for a future caller.
- Still open from round 1: no index behind `listReplyUncheckedEntries` (`WHERE reply_checked_at IS NULL ORDER BY received_at DESC`, run on every timeline read); MCP `get_timeline_replies` uses `jsonResponse({ error, success: false })` where the rest of the file uses `errorResponse(...)` and re-inlines `12_000`; `partial` is `true` for an exhausted thread of exactly 20 and `false` for an unreachable origin (which the web then shows as "No replies found on the origin"); `invalidateQueries({ queryKey: ['feed', 'timeline'] })` prefix-matches the per-entry replies queries.
- Android: when the permission request is denied, `setPostNotificationsEnabled(context, false)` records an explicit "off", and the Account screen's warning only fires on *toggle-on + permission-off* — so a user with bells on who denies once gets neither notifications nor the warning, which is the silent-bell case this reconciliation exists to remove.

apps/backend/src/services/activitypub/remote-replies.ts
```diff
@@ -0,0 +1,217 @@
...
+/** Map one reply object to the DTO, resolving its author (memoised) within budget. */
+const toReply = async (
+  deps: RemoteRepliesDeps,
+  budget: Budget,
+  obj: JsonRecord,
+  authors: Map<string, ReplyAuthor>,
+): Promise<TimelineReply | null> => {
+  const content = typeof obj.content === 'string' ? obj.content : null
+  if (content == null) return null
+  const actorUri = idOf(obj.attributedTo)
```
fiddur, line 185: The reply author comes from remote-supplied `attributedTo` with no authority check, so a hostile origin can render arbitrary content under any fediverse identity.

`collectPageReplies` → `resolveObject` returns inline items straight out of the parent document (`isRecord(v)` → returned as-is, never fetched from their own origin), and `toReply` then resolves that `attributedTo` against the **real** actor to fill `display_name` / `handle`. Nothing requires the reply object's own `id` host to match the `attributedTo` host — the check `noteToTimelineInput` already enforces on ingest (`note.id.host !== senderHost` → drop).

Since this PR's mention branch lets any signature-verified actor put a post into my timeline, `evil.example` can serve:

```json
{ "id": "https://evil.example/notes/1",
  "replies": { "items": [
    { "type": "Note",
      "content": "<p>Your account needs re-verification…</p>",
      "attributedTo": "https://mastodon.social/users/Gargron",
      "url": "https://look-alike.example/x" } ] } }
```

"Show replies" then renders attacker-written text labelled `Eugen Rochko @Gargron@mastodon.social`, with an attacker-chosen (http(s)-valid) link and no hint it came from `evil.example` — a clean impersonation/phishing primitive. Sanitisation doesn't help here; the payload is plain text under a forged byline.

Suggest requiring the object's own authority before trusting its attribution: `obj.id` must parse as http(s) and `new URL(obj.id).host === new URL(actorUri).host`, otherwise drop the reply (or at minimum skip `resolveAuthor`/`url` and leave the raw `actor_uri`). The same check applies to URI-referenced items — with `maxRedirects: 0` the fetched doc's `id` should also match the URI you asked for.

fiddur (commented):
🛑 Changes required

Solid round overall. The stranger-admission tightening holds up under scrutiny: explicit `attributedTo` == signing actor, the existing-own-post lookup, and the `note.id.host === senderHost` guard inherited from `noteToTimelineInput` together close the object_uri-overwrite path I went looking for. The no-clobber backfill is right too (`markChecked` vs `saveReplyInfo`, plus the idempotent backstamp so any row that already carries `in_reply_to_uri` is never a candidate — which also means every candidate has a NULL link, so a null result can't clobber anything). The `listTimelineEntries` filter change is safe (`mentions_me` is `NOT NULL DEFAULT false`, no three-valued logic), `/timeline/:id/replies` is registered ahead of `/:postId`, the web PATCHes settings partially so the Feed toggle can't wipe other settings, and the `['userSettings']` key matches the Settings page.

**Blocking:** one authenticity hole in the new reply fetcher — see the inline comment on `remote-replies.ts`.

- **Two doc comments now document the wrong symbol.** `timeline-ingest.ts:84` — the `noteToTimelineInput` "Authority check" block is now attached to `export type TimelineAuthor`. `federation.ts:203-221` — the `ingestFeedActivity` block ("Two admissible senders…") is stacked directly above `ingestStrangerInvolvement`, which has its own block, so the text is duplicated and `ingestFeedActivity` (line 262) is left undocumented.
- **No index for the backfill candidate query.** `listReplyUncheckedEntries` is `WHERE reply_checked_at IS NULL ORDER BY received_at DESC LIMIT 3` and runs on every timeline read, but the only index is `(published_at DESC, id DESC)` — so it is a seq scan + top-N sort, worst exactly once the backlog drains. #996 added `idx_timeline_entry_unenriched` as a partial index for precisely this shape; a `(received_at DESC) WHERE reply_checked_at IS NULL` partial index would match that precedent.
- `fetchRemoteReplies` returns `partial: false` when the *post itself* couldn't be fetched (`budget.exhausted` is still false), so an unreachable or authorized-fetch origin renders as "No replies found on the origin." — indistinguishable from a genuinely empty thread. Separately, `partial` is true whenever exactly `MAX_REPLIES` were collected, even when the collection ended there.
- `isoOrNull` accepts anything `Date.parse` takes, but `timelineReplySchema.published_at` is `z.iso.datetime()`; normalising via `new Date(v).toISOString()` would make the DTO match its own schema.
- `get_timeline_replies` returns `jsonResponse({ error: 'Not found', success: false })` where every other tool in `feed-tools.ts` uses `errorResponse(...)`.
- `${origin}/users/${user}` is now hand-built in three places (`federation.ts:186`, `federation.ts:237`, `api.ts:405`) next to the existing `ownObjectPrefix` helper; an `ownActorUri(origin, user)` beside it would stop them drifting. Relatedly, `ingestNoteForRecipient`'s `origin?` is optional but both callers now pass it, so the `origin == null → mentions_me = false` branch is dead.
- Product note, not a defect: involvement now notifies regardless of the per-account bell, so a followee whose bell was explicitly switched *off* will still raise a notification when they reply to or mention you. `docs/features/feed.md` states this, so I'm reading it as intentional.

fiddur (commented):
✅ Approved

Read the whole diff plus the surrounding ingest / SSRF / schema-migration code. Round 3 closes the earlier rounds' issues cleanly and I found no blocking correctness or security problem.

What I checked and am satisfied with:

- **Stranger admission** (`ingestStrangerInvolvement`): gated on reply-to-an-existing-own-post (prefix + UUID + `getFeedPostById`) or an explicit `Mention` of the recipient's actor URI; requires explicit `attributedTo` matching the signature-verified `activity.actorId`, and re-verifies the fetched actor's `id`. Combined with `noteToTimelineInput`'s host check, the `object_uri` upsert key can't be hijacked to overwrite another actor's entry.
- **`remote-replies.ts`**: the serving-host authority model (`PageCtx.host`, inline items inherit the page host, URI items get their own) plus the id/`attributedTo` host checks in `toReply` block byline forgery; `httpsOnly` and `isoOrNull` sanitize the two values the web feeds to `href` / `date-fns`; every fetch goes through `safeFetchGet` (`maxRedirects: 0`, private-range blocklist, 1 MB cap) and every failure path is swallowed rather than thrown. The page walk terminates in every cycle shape I could construct — a repeating `next` URI drains the fetch budget, and parsed JSON can't self-reference.
- **Migrations**: `timeline_entry_mentions` / `timeline_entry_reply_checked` / `..._backstamp` are keys with no matching table name, so they run on the existing-DB path (`!existingTableNames.has(tableName)`) as well as on fresh init, in the right order relative to `timeline_entry_reply` and the indexes. The backstamp is genuinely idempotent because the upsert now always stamps `reply_checked_at`.
- **No-clobber backfill**: candidates are `reply_checked_at IS NULL`, which after the backstamp implies `in_reply_to_uri IS NULL`, and a non-answer only calls `markChecked` — the new test asserts exactly that split.
- **Web**: `dangerouslySetInnerHTML` is fed only server-sanitized HTML; the `['userSettings']` and `['feed','timeline']` query keys match the rest of the app; `updateUserSettings` is a PATCH, so the single-field write doesn't clobber other settings.
- **Android**: `decideNotifications`' new `involved || (top-level && bell on)` rule is what the docs and the new test describe; `fetchFollowingList` swallows all exceptions, so `AutoEnablePostNotifications`' `LaunchedEffect` can't crash the app when offline, and the launcher is registered before the effect runs.

Please do **not** push fixes for these here; another round costs more than they're worth. Batch them with other work:

1. **Two orphaned/stale doc comments.** In `federation.ts` the old `ingestFeedActivity` block ("Ingest a `Create`/`Update` … Two admissible senders …") now sits directly above the new `ingestStrangerInvolvement` doc block, so `ingestFeedActivity` itself is undocumented and the stacked block reads as describing the stranger helper. It is also now incomplete: it lists only "a reply to one of the recipient's own posts" and omits the Mention branch, as does the `.on(Create, …)` comment ("the only stranger Note admitted is a reply to one of the recipient's own posts"). Same shape in `timeline-ingest.ts`, where the `noteToTimelineInput` doc block (authority check / `feed_following` presentation) is now attached to the newly inserted `TimelineAuthor` type. Stale prose sitting on security-critical admission logic is the kind that misleads a later reader.
2. **A couple of redundant comments.** `// An inline item is held to the page's serving host; a URI item to its own.` in `collectPageReplies` restates what `resolvePage`'s and `toReply`'s docs already say, and `// Either way the choice is now recorded; only a grant starts the poller.` in `PostNotificationAutoEnable.kt` narrates the two lines under it. (The `{/* Sanitised server-side … */}` note in `HomeTimeline.tsx` is the opposite — keep that one.)
3. **`partial` over-reports.** It is true whenever `replies.length >= MAX_REPLIES`, so a thread of exactly 20 is reported as truncated, and an author lookup that exhausts the fetch budget marks the whole snapshot partial even when every reply was collected. Conversely, a post whose own fetch fails returns `partial: false, replies: []`, which the web renders as the confident "No replies found on the origin." rather than as a failure.
4. **Weak assertion in the budget test.** `expect(replies.length).toBeLessThanOrEqual(20)` in "caps at the reply budget and reports partial" would pass with zero replies; `toBe(20)` would actually pin the cap the test name claims.
5. **Small consistency items.** MCP `get_timeline_replies` returns `jsonResponse({ error: 'Not found', success: false })` for not-found while every other not-found in `feed-tools.ts` uses `errorResponse(...)`; the 12 s budget is `REPLIES_TIMEOUT_MS` in `feed-router.ts` but a bare `12_000` literal in the MCP tool; and the reader's own actor URI is now built inline in three places (`ingestNoteForRecipient`, `ingestStrangerInvolvement`, `api.ts`) as `${origin}/users/${user}` while `ownObjectPrefix` uses `encodeURIComponent(user)` — harmless today given `isValidUsername`, but a one-line `ownActorUri` helper next to `ownObjectPrefix` would stop them drifting.
6. **No unit test for the stranger-admission branch.** `noteMentionsActor` and `ingestStrangerInvolvement` are the highest-risk new logic and are covered only indirectly. `timeline-ingest.test.ts` already builds fedify `Note`s, so "Mention of me admits / Mention of someone else doesn't / missing `attributedTo` drops" would be cheap insurance.

One design point worth being deliberate about rather than a defect: invalidating `['feed','timeline']` after the settings toggle prefix-matches every `['feed','timeline',<id>,'replies']` query, so each expanded thread refetches — up to 15 outbound requests apiece — on a checkbox flip.

PR #1067, opened 2026-08-27
docs: data-integrity improvement backlog (maybe/future plan)

fiddur:
## Summary

Captures the improvement ideas distilled from a r/QuantifiedSelf post by another wearable-aggregation developer as a maybe/future plan doc, following the existing `docs/PLAN-*.md` convention.

Each idea was checked against the current codebase first; the doc opens with an "already covered" section (cross-source dedup, HC double-counting, semantic field traps, provenance) so nobody re-investigates those, then lists six independent backlog items ordered roughly by value:

1. Multiple-comparison correction in correlations
2. Impulse-basis labeling in training load
3. Ingest plausibility gates
4. Version-stamped derived rows
5. Measured-vs-modelled as stored data
6. Two-process energy curve (feature idea)

**Docs-only — no code changes. Not scheduled work.**

https://claude.ai/code/session_019uPd32AMQzUxZb233JGRKV

### DIFF
--- a/docs/PLAN-data-integrity-improvements.md
+++ b/docs/PLAN-data-integrity-improvements.md
@@ -1,0 +1,89 @@
+# Plan (maybe/future): Data-Integrity Improvements
+
+Status: **idea backlog — not scheduled**. Captured 2026-08 from a discussion of a
+[r/QuantifiedSelf post](https://www.reddit.com/r/QuantifiedSelf/comments/1v8siau/)
+by another wearable-aggregation developer, filtered down to what actually applies
+to Aurboda. Items are independent; pick any one up on its own.
+
+## Already covered (checked, no action)
+
+- **Cross-source activity dedup** — `apps/backend/src/db/activities/merge.ts`
+  merges on start-time proximity (120 s) with an explicit source-priority ladder.
+- **Health Connect double-counting** — Health Connect's built-in dedup is used
+  (see `docs/health-connect.md`).
+- **Semantic field traps** — Oura temperature is stored as `temperature_deviation`
+  (not mislabeled absolute skin temp); Garmin Body Battery is kept as a full time
+  series.
+- **Provenance** — `raw_records` preserves originals; rows carry `source`.
+- **Fitbit Web API deprecation (reportedly Sept 2026)** — Aurboda gets Fitbit data
+  via Health Connect, not `api.fitbit.com`; verify the claim if a direct
+  integration is ever planned.
+
+## 1. Multiple-comparison correction in correlations (highest value)
+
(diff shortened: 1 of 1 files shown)


fiddur (commented):
✅ Approved

Docs-only change adding a clearly-scoped "maybe/future" backlog file. No code, schema or API surface touched, so no correctness/security/data-integrity risk to this PR itself.

I spot-checked the factual claims against the source, since the value of a plan doc is its accuracy. These all hold:

- `apps/backend/src/db/activities/merge.ts` — `CROSS_MERGE_THRESHOLD_MS = 120_000` plus an explicit `SOURCE_PRIORITY` ladder ✅
- Health Connect built-in dedup is documented in `docs/health-connect.md` ✅
- Oura stores `temperature_deviation` (`integrations/oura/process.ts`) and Garmin Body Battery is pushed as a full point series from `bodyBatteryValuesArray` (`integrations/garmin/process.ts`) ✅
- `raw_records` exists and rows carry `source` ✅; no direct `api.fitbit.com` integration anywhere ✅
- `services/correlations/stats.ts` has Welch, Mann–Whitney U and Cohen's d, and there is no Bonferroni/BH/FDR anywhere in the backend — so item 1's premise is correct ✅
- `recalculate_calories` (MCP) and `services/training-load/recompute.ts` exist, so item 4's "mostly a stamp + trigger" reading is fair ✅

## Non-blocking — fold into a later PR, don't re-roll this one

One claim in item 2 doesn't match the code, and it's worth correcting whenever this file is next touched (please don't push a fix on this PR — it just costs another review round for a not-scheduled backlog doc):

> `banister.ts` silently falls back from HR-based TRIMP to `calories × DEFAULT_ACTIVITY_IMPULSE_SCALE` when HR is missing

There are actually two separate paths in `banister.ts`, and neither is that one:

1. `processExercise()` (line ~178) falls back to a flat `durationMinutes * 0.5` when `avgHr` is absent or `<= hrRest` — that is the silent HR-missing estimate, and it feeds the *training* impulse.
2. `computeHourlyImpulses()` adds `kcal * activityScale` as the *activity* impulse for every hour regardless of HR availability — it is an always-on second component, not a fallback.

The conclusion the item draws (a modelled value mixed indistinguishably into CTL/ATL/TSB) is still right — arguably twice over — but the suggested `hr` vs `calorie-estimate` basis tag would mislabel path 1, which is really `hr` vs `duration-estimate`. Rewording that paragraph would make the item directly actionable for whoever picks it up.

Also worth knowing, not a defect: this doc isn't linked from any index, but neither is the existing `docs/PLAN-calorie-source-separation.md`, so it matches the current convention for plan files.

PR #1068, opened 2026-08-27
feat(web): challenges page groups by time status, ongoing front and centre

fiddur:
## Summary

The /challenges list showed only hosted/joined columns with no dates, so nothing distinguished a challenge running right now from one that ended months ago. This PR makes ongoing challenges unmissable:

- **Ongoing** section first, with a green dot, count badge, and accent border on rows — sorted by ending soonest.
- **Upcoming** next (starting soonest first), rendered only when non-empty.
- **Ended** collapsed into a `<details>` section (most recently ended first), rows dimmed.
- Every row now shows its date range and a relative phrase — "Ends in 3 days", "Starts tomorrow", "Ended 7 Jun 2026" — plus a "Hosted by you"/"Joined" role badge (replacing the columns).

## Implementation

- Pure time-status/grouping helpers in `apps/web/src/pages/Challenges/challenge-status.ts` (`[start_ts, end_ts)` end-exclusive throughout; calendar-day arithmetic anchored on local midnights) with unit tests written first.
- `index.tsx` renders the groups; hosted/joined row actions unchanged.
- Dark-mode styles included for the new badges/phrases.
- Web-only: the Android app embeds this page, so parity is preserved.

## Testing

- `pnpm vitest run src/pages/Challenges/` — 25 tests pass.
- `pnpm check` green across the repo.

https://claude.ai/code/session_019uPd32AMQzUxZb233JGRKV

### DIFF
--- a/apps/web/src/pages/Challenges/challenge-status.ts
+++ b/apps/web/src/pages/Challenges/challenge-status.ts
@@ -1,0 +1,87 @@
+/**
+ * Time-status helpers for the challenge list. A challenge window is
+ * `[start_ts, end_ts)` (end exclusive), so its last included moment is just
+ * before `end_ts` and the "last day" is the calendar day of `end_ts - 1ms`,
+ * rendered in the viewer's local timezone.
+ */
+
+import type { Challenge, ChallengeParticipation } from '@aurboda/api-spec'
+
+export type ChallengeTimeStatus = 'ended' | 'ongoing' | 'upcoming'
+
+export const challengeTimeStatus = (startTs: string, endTs: string, now: Date): ChallengeTimeStatus => {
+  const t = now.getTime()
+  if (t < new Date(startTs).getTime()) return 'upcoming'
+  if (t < new Date(endTs).getTime()) return 'ongoing'
+  return 'ended'
+}
+
+/** The last calendar moment included in the window (end_ts is exclusive). */
+export const lastIncludedMoment = (endTs: string): Date => new Date(new Date(endTs).getTime() - 1)
+
+/** Local midnight of a date — anchors calendar-day arithmetic, DST-safe via rounding. */
+const localMidnight = (d: Date): Date => new Date(d.getFullYear(), d.getMonth(), d.getDate())
+
+/** Whole calendar days from `now`'s local day to `target`'s local day (0 = same day). */
+export const calendarDaysUntil = (target: Date, now: Date): number =>
+  Math.round((localMidnight(target).getTime() - localMidnight(now).getTime()) / 86_400_000)
+
+const formatDay = (d: Date, locale?: string): string =>
+  d.toLocaleDateString(locale, { day: 'numeric', month: 'short', year: 'numeric' })
+
+/** "3 Jun – 9 Jun 2026" (last included day, not the exclusive end instant). */
+export const challengeRangeLabel = (startTs: string, endTs: string, locale?: string): string =>
+  `${new Date(startTs).toLocaleDateString(locale, { day: 'numeric', month: 'short' })} – ${formatDay(
+    lastIncludedMoment(endTs),
+    locale,
+  )}`
+
+const inDays = (days: number): string => (days === 1 ? 'tomorrow' : `in ${days} days`)
+
+/** Relative phrase for the row: "Ends today", "Starts tomorrow", "Ended 9 Jun 2026". */
+export const challengeTimePhrase = (startTs: string, endTs: string, now: Date, locale?: string): string => {
+  const status = challengeTimeStatus(startTs, endTs, now)
+  if (status === 'upcoming') {
+    const days = calendarDaysUntil(new Date(startTs), now)
+    return days === 0 ? 'Starts today' : `Starts ${inDays(days)}`
+  }
+  const lastDay = lastIncludedMoment(endTs)
+  if (status === 'ongoing') {
+    const days = calendarDaysUntil(lastDay, now)
+    return days <= 0 ? 'Ends today' : `Ends ${inDays(days)}`
+  }
+  return `Ended ${formatDay(lastDay, locale)}`
+}
+
+export type ChallengeItem =
+  | { kind: 'hosted'; challenge: Challenge }
+  | { kind: 'joined'; participation: ChallengeParticipation }
+
+const itemWindow = (item: ChallengeItem): { end_ts: string; start_ts: string } =>
+  item.kind === 'hosted' ? item.challenge : item.participation
+
+export const challengeItemKey = (item: ChallengeItem): string =>
+  item.kind === 'hosted' ? `hosted-${item.challenge.id}` : `joined-${item.participation.id}`
+
+/** Group hosted + joined challenges by time status, soonest-relevant first. */
+export const groupChallengeItems = (
+  hosted: Challenge[],
+  joined: ChallengeParticipation[],
+  now: Date,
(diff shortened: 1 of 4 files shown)


fiddur (commented):
🛑 Changes required

Nice restructure — the pure `challenge-status.ts` module with its own unit tests is exactly the right shape, the `[start_ts, end_ts)` handling (`end_ts - 1ms` for the last included day) is consistent with `date-range.ts`, the `hosted-`/`joined-` key prefixes avoid cross-list collisions, and the dark-mode variants cover every new colour. The `.challenge-group-ongoing .challenge-row` green left border also survives the dark-mode `border-color` shorthand on specificity, so that's fine.

One blocking issue: the new date rendering uses the **viewer's** timezone, while the challenge detail page for the same challenge deliberately renders in the **challenge's** timezone (`formatDateInZone` in `race-series.ts`, with an explicit comment saying why). For a federated/cross-timezone challenge these disagree by a day — see the inline comment.

Please don't push fixes for these here; batching them with other work avoids an extra review round.

- `now = new Date()` is recomputed per render but never ticks, so a page left open across midnight (or across a start/end instant) keeps stale grouping and phrases until something else re-renders. Acceptable for now; a slow interval or `useSyncExternalStore` clock would fix it if it ever bites.
- `.challenge-group-ended summary { list-style: revert }` looks like it's undoing a marker reset, but nothing in this codebase resets `summary`'s list-style, so it's a no-op.
- In `challengeTimePhrase`, the `ongoing` branch guarantees `lastDay >= now`, so `days <= 0` can only ever be `days === 0`; `=== 0` would state the invariant more plainly.
- The first assertion in the `'upcoming: today / tomorrow / in N days'` test is a negative `not.toMatch(/Starts/)` at the start instant — it passes for any non-upcoming output. Asserting the exact `'Ends in 6 days'` (and moving it to the ongoing block, where the boundary actually lives) would test more.
- The trailing `// ending soonest first` / `// starting soonest first` / `// most recently ended first` comments on the three `.sort()` calls restate their comparators.

apps/web/src/pages/Challenges/challenge-status.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/** Whole calendar days from `now`'s local day to `target`'s local day (0 = same day). */
+export const calendarDaysUntil = (target: Date, now: Date): number =>
+  Math.round((localMidnight(target).getTime() - localMidnight(now).getTime()) / 86_400_000)
+
+const formatDay = (d: Date, locale?: string): string =>
+  d.toLocaleDateString(locale, { day: 'numeric', month: 'short', year: 'numeric' })
+
+/** "3 Jun – 9 Jun 2026" (last included day, not the exclusive end instant). */
+export const challengeRangeLabel = (startTs: string, endTs: string, locale?: string): string =>
```
fiddur: `challengeRangeLabel` (and `Ended ${formatDay(lastDay, locale)}` below) formats in the **viewer's** timezone, but `PublicChallenge.tsx` renders the very same window in the **challenge's** timezone via `formatDateInZone`, with a comment stating the intent: "Render both dates in the timezone the range was chosen in, so they read exactly as entered regardless of the viewer's browser locale/timezone."

Since this page is explicitly about joining challenges hosted on other instances, the two disagree in practice. Challenge hosted in `America/Los_Angeles`, last day 7 Jun → `end_ts = 2026-06-08T07:00:00Z`. A viewer in `Europe/Stockholm` gets `end_ts - 1ms` = 2026-06-08 08:59 local, so the list row reads `1 Jun – 8 Jun 2026` while the challenge page for the same challenge reads 7 Jun. The list is claiming a competing day that doesn't exist.

Both `Challenge` and `ChallengeParticipation` carry `timezone`, so threading it into `RowMain` and passing `timeZone` to these `toLocaleDateString` calls fixes it. One caveat: the backend only validates `timezone` as a non-empty string, which is why `formatDateInZone` wraps the call in try/catch — an unguarded `timeZone` here would throw `RangeError` on a crafted value and take the page down, so reuse that guard (or add an options-taking variant of it) rather than passing `timeZone` raw.

The relative phrases (`Ends tomorrow`, `Starts in 3 days`) reading in viewer-local days seems right to keep as-is — it's the absolute dates that should match the host's window.

fiddur: Fixed in 6c50e6c0: `challengeRangeLabel` and the `Ended <date>` phrase now take the challenge's IANA `timezone` (threaded from `Challenge`/`ChallengeParticipation` into `RowMain`) and format with the same try/catch fallback as `formatDateInZone`, so a crafted timezone falls back to viewer-local instead of throwing. Relative phrases stay viewer-local as you suggested. Added tests for the LA-hosted cross-timezone case and the invalid-timezone fallback.

fiddur (commented):
✅ Approved

Solid, focused frontend change. The time-status logic is extracted into a pure, well-tested module, the window semantics (`[start_ts, end_ts)` with the last included day at `end_ts - 1ms`) match `date-range.ts` and `PublicChallenge.tsx`, and `challengeTimeStatus` correctly compares absolute instants rather than local dates. Boundary handling is right: start-inclusive/end-exclusive is asserted directly, `calendarDaysUntil` anchors on local midnight with `Math.round` so it stays DST-safe, and `formatInZone` guards the `RangeError` from a crafted `timezone` (the backend only validates it as a non-empty string) exactly like the existing `formatDateInZone` guard. Grouping/sort orders are correct and covered, React keys are unique across the hosted/joined union, `length > 0 &&` avoids the stray-`0` render trap, CSS specificity keeps the ongoing left border over the dark-mode `border-color` override, and `<h2>` inside `<summary>` is valid HTML. The removed `.challenges-columns` class has no remaining references, and Android reaches this page via `MoreDestination.Web("/challenges")`, so native parity needs no companion change. No blocking issues found.

Please do **not** push fixes for these here; a new push just starts another review round for no real benefit. Batch them with other work.

- **`challenge-status.ts` header comment is now factually wrong.** It says the last day is "rendered in the viewer's local timezone", but after `6c50e6c` `formatDay`/`challengeRangeLabel` render in the *challenge's* timezone. In a file whose whole subject is timezone semantics, that stale line could mislead someone into "fixing" the code back.
- **Duplicated last-day logic.** `lastIncludedMoment` and the try/catch timezone fallback duplicate `PublicChallenge.tsx:96-101` and `race-series.ts:formatDateInZone`. Worth consolidating into one shared helper so the two pages can't drift.
- **`challengeRangeLabel` omits the year on the start side**, so a year-spanning challenge reads "28 Dec – 3 Jan 2026". Fine for the common case, mildly ambiguous at the boundary.
- **`now` is captured once per render** with no timer, so "Ends today" and group membership only update when something else triggers a re-render. Acceptable for a page that refetches, just noting it.
- **Phrase vs. range timezone mismatch** for cross-timezone challenges: relative counts are viewer-local while the absolute range is host-local, so a Stockholm viewer can see "Ends tomorrow · 1 Jun – 7 Jun" for an LA-hosted challenge. The doc comment says this is deliberate, and it is defensible — flagging only so it's a conscious choice.
- **Test name mismatch:** the `'upcoming: today / tomorrow / in N days'` case opens with an *ongoing* assertion (`'Ends in 6 days'`). The inline comment explains why, but the assertion would read better under the ongoing block.
- **Docs:** `docs/features/challenges.md` ("Using it in the web app") still describes the page only as "see challenges you've joined" — nothing there is now false, but a sentence about the Ongoing/Upcoming/Ended grouping would keep it current per the CLAUDE.md docs rule.

